PR #18, opened 2025-05-12
feat: Added Config Overrides

AmitsinghTanwar007:
## Description
I have made a function named config from metadata inside the utils.rs of grpc_server which mutate the original config details if x-config-override contains some json.
And this function is used inside the flows specified under Payments.

## Motivation and Context
The change is made So that If a connector required some specific configuration details for a specific request or time period. then without changing the whole config of the server we can provide the details related to the config inside the request header.

## How did you test it?
I tested it using cargo test test_config_override

### DIFF
--- a/backend/grpc-server/src/server/payments.rs
+++ b/backend/grpc-server/src/server/payments.rs
@@ -1,6 +1,6 @@
 use crate::{
     configs::Config,
-    utils::{auth_from_metadata, connector_from_metadata},
+    utils::{auth_from_metadata, config_from_metadata, connector_from_metadata},
 };
 use connector_integration::types::ConnectorData;
 use domain_types::types::generate_payment_void_response;
@@ -120,7 +120,7 @@ impl PaymentService for Payments {
         request: tonic::Request<PaymentsAuthorizeRequest>,
     ) -> Result<tonic::Response<PaymentsAuthorizeResponse>, tonic::Status> {
         info!("PAYMENT_AUTHORIZE_FLOW: initiated");
-
+        let config = config_from_metadata(request.metadata(), self.config.clone())?;
         let connector = connector_from_metadata(request.metadata())?;
         let connector_auth_details = auth_from_metadata(request.metadata())?;
         let payload = request.into_inner();
@@ -138,18 +138,16 @@ impl PaymentService for Payments {
         > = connector_data.connector.get_connector_integration_v2();
 
         // Create common request data
-        let mut payment_flow_data = match PaymentFlowData::foreign_try_from((
-            payload.clone(),
-            self.config.connectors.clone(),
-        )) {
-            Ok(data) => data,
-            Err(e) => {
-                return Err(tonic::Status::invalid_argument(format!(
-                    "Invalid request data: {}",
-                    e
-                )))
-            }
-        };
+        let mut payment_flow_data =
+            match PaymentFlowData::foreign_try_from((payload.clone(), config.connectors.clone())) {
+                Ok(data) => data,
+                Err(e) => {
+                    return Err(tonic::Status::invalid_argument(format!(
+                        "Invalid request data: {}",
+                        e
+                    )))
+                }
+            };
 
         let should_do_order_create = connector_data.connector.should_do_order_create();
 
@@ -191,7 +189,7 @@ impl PaymentService for Payments {
 
         // Execute connector processing
         let response = match external_services::service::execute_connector_processing_step(
-            &self.config.proxy,
+            &config.proxy,
             connector_integration,
             router_data,
         )
--- a/backend/grpc-server/src/server/payments.rs
+++ b/backend/grpc-server/src/server/payments.rs
@@ -224,7 +224,7 @@ impl PaymentService for Payments {
         request: tonic::Request<PaymentsSyncRequest>,
     ) -> Result<tonic::Response<PaymentsSyncResponse>, tonic::Status> {
         info!("PAYMENT_SYNC_FLOW: initiated");
-
+        let config = config_from_metadata(request.metadata(), self.config.clone())?;
         let connector = connector_from_metadata(request.metadata())?;
         let connector_auth_details = auth_from_metadata(request.metadata())?;
         let payload = request.into_inner();
@@ -311,7 +311,7 @@ impl PaymentService for Payments {
         request: tonic::Request<RefundsSyncRequest>,
     ) -> Result<tonic::Response<RefundsSyncResponse>, tonic::Status> {
         info!("REFUND_SYNC_FLOW: initiated");
-
+        let config = config_from_metadata(request.metadata(), self.config.clone())?;
         let connector = connector_from_metadata(request.metadata())?;
         let connector_auth_details = auth_from_metadata(request.metadata())?;
         let payload = request.into_inner();
@@ -368,6 +368,7 @@ impl PaymentService for Payments {
         request: tonic::Request<PaymentsVoidRequest>,
     ) -> Result<tonic::Response<PaymentsVoidResponse>, tonic::Status> {
         info!("PAYMENT_CANCEL_FLOW: initiated");
+        let config = config_from_metadata(request.metadata(), self.config.clone())?;
         let connector = connector_from_metadata(request.metadata())?;
         let connector_auth_details = auth_from_metadata(request.metadata())?;
         let payload = request.into_inner();
@@ -421,6 +422,7 @@ impl PaymentService for Payments {
         &self,
         request: tonic::Request<IncomingWebhookRequest>,
     ) -> Result<tonic::Response<IncomingWebhookResponse>, tonic::Status> {
+        let config = config_from_metadata(request.metadata(), self.config.clone())?;
         let connector = connector_from_metadata(request.metadata())?;
         let connector_auth_details = auth_from_metadata(request.metadata())?;
         let payload = request.into_inner();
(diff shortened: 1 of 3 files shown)


NishantJoshi00:
what's the usecase for this?

backend/grpc-server/src/utils.rs
```diff
@@ -94,6 +99,153 @@ pub fn auth_from_metadata(
...
+                        }
+                    }
+                }
+                // Add other connectors as needed
+                _ => {
+                    tracing::warn!("Unknown connector in config override: {}", connector_name);
+                }
+            }
+        }
+    }
```
NishantJoshi00: not scalable enough

backend/grpc-server/src/utils.rs
```diff
@@ -94,6 +99,153 @@ pub fn auth_from_metadata(
     }
 }
 
+pub fn config_from_metadata(
```
NishantJoshi00: this could be done in a better way, figure out a way to create a future proof solution, which will not require change even if something were to change in the configurations

knutties: Should we call this `config_from_request`  instead of `config_from_metadata`?

backend/domain_types/src/types.rs
```diff
@@ -34,22 +34,22 @@ use serde::Serialize;
 use std::borrow::Cow;
 use std::{collections::HashMap, str::FromStr};
 use utoipa::ToSchema;
-#[derive(Clone, serde::Deserialize, Debug)]
+#[derive(Clone, serde::Deserialize, Debug, serde::Serialize)]
```
knutties: Let us just declare serde::Deserialize and serde::Serialize in the use statement above and use the non-qualified references here.

backend/grpc-server/src/configs.rs
```diff
@@ -1,10 +1,9 @@
-use std::path::PathBuf;
-
-use domain_types::types::{Connectors, Proxy};
-
 use crate::{consts, error::ConfigurationError, logger::config::Log};
+use domain_types::types::{Connectors, Proxy};
+use std::path::PathBuf;
+// use serde::{de::{self, Deserializer, MapAccess, Visitor}, Deserialize};
```
knutties: Let us just declare serde::Deserialize and serde::Serialize in the use statement here and use the non-qualified references later.

backend/grpc-server/src/utils.rs
AmitsinghTanwar007: if it seems correct then i can change but the config_from_metadata is getting the config from metadata and then working on it.

backend/grpc-server/src/utils.rs
```diff
@@ -140,7 +233,8 @@ macro_rules! implement_connector_operation {
             request: tonic::Request<$request_type>,
         ) -> Result<tonic::Response<$response_type>, tonic::Status> {
             tracing::info!(concat!($log_prefix, "_FLOW: initiated"));
-
+            let config = $crate::utils::config_from_metadata(request.metadata(), self.config.clone())
```
knutties: @AmitsinghTanwar007 - wondering aloud here - should the resolved config (header + toml merged final config) be injected from a middleware itself instead of actually doing it in the operation macro.

cc: @jarnura

backend/grpc-server/src/config_overrides.rs
```diff
@@ -0,0 +1,78 @@
...
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    #[allow(clippy::expect_used)]
```
AmitsinghTanwar007: I am working on this If anybody has a more correct way to remove this .expect from this function please let me know

jarnura: do the update only when the values are there.

backend/grpc-server/src/config_overrides.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn call(&mut self, mut req: Request<Body>) -> Self::Future {
+        let default_config = Config::new().expect("Failed to load default config");
+        // Extract x-config-override header
+        let config_override = req
+            .headers()
+            .get("x-config-override")
+            .and_then(|h| h.to_str().ok())
+            .unwrap_or("")
```
NishantJoshi00: don't have this as empty, it's fine to be `None`

backend/grpc-server/src/server/disputes.rs
```diff
@@ -257,6 +262,14 @@ impl DisputeService for Disputes {
             .cloned()
             .unwrap_or_else(|| "unknown_service".to_string());
         grpc_logging_wrapper(request, &service_name, |request| async {
+            let config = match request.extensions().get::<Config>() {
+                Some(config) => config.clone(),
+                None => {
```
NishantJoshi00: same here

NishantJoshi00: Don't write match statement in the open, you are repeating this piece of code everywhere. Follow DRY

backend/grpc-server/src/server/payments.rs
```diff
@@ -68,12 +75,10 @@ pub struct Payments {
...
-        connector_data: ConnectorData,
-        payment_flow_data: &mut PaymentFlowData,
-        connector_auth_details: ConnectorAuthType,
-        payload: &PaymentServiceAuthorizeRequest,
-        connector_name: &str,
-        service_name: &str,
+        config: &Config,                         // Infra state
+        data: HandleOrderData,                   // L2 mirco state
+        payment_flow_data: &mut PaymentFlowData, // L2 macro state
+        payload: PaymentServiceAuthorizeRequest, // Api types
```
NishantJoshi00: what is the change, and why is this required in the PR?

backend/grpc-server/src/utils.rs
```diff
@@ -140,6 +142,99 @@ pub fn auth_from_metadata(
     }
 }
 
+pub fn config_from_metadata(
```
NishantJoshi00: add unit tests for this

backend/grpc-server/src/utils.rs
```diff
@@ -140,6 +142,99 @@ pub fn auth_from_metadata(
...
+                error_identifier: 400,
+                error_message: format!("Cannot deserialize merged config: {e}"),
+                error_object: None,
+            }))
+        });
+    }
+    Ok(config)
+}
+
+fn merge_configs(override_val: &Value, base_val: &Value) -> Value {
```
NishantJoshi00: and this

backend/grpc-server/tests/test_config_override.rs
```diff
@@ -0,0 +1,117 @@
...
+                }
+            },
+            "proxy": {
+                "idle_pool_connection_timeout": 30,
+            },
+        });
+
+        request.metadata_mut().insert(
+            "x-config-override",
+            override_config.to_string().parse().unwrap(),
```
NishantJoshi00: avoid using unwraps, use expects' instead

backend/grpc-server/src/server/payments.rs
AmitsinghTanwar007: there was a clippy error of many arguments so i asked arun raj that can i add a struct to make some attributes common so he said yes.

backend/grpc-server/src/config_overrides.rs
NishantJoshi00: The output of the function is a future, which holds a result, instead of expect do map_err

backend/grpc-server/src/server/disputes.rs
```diff
@@ -113,11 +115,9 @@ impl DisputeService for Disputes {
             let dispute_data = SubmitEvidenceData::foreign_try_from(payload.clone())
                 .map_err(|e| e.into_grpc_status())?;
 
-            let dispute_flow_data = DisputeFlowData::foreign_try_from((
-                payload.clone(),
-                self.config.connectors.clone(),
-            ))
-            .map_err(|e| e.into_grpc_status())?;
+            let dispute_flow_data =
```
NishantJoshi00: what happened here? can you run `cargo +nightly fmt`?

NishantJoshi00 (changes requested):
paste screenshots of successful tests

backend/grpc-server/src/server/disputes.rs
AmitsinghTanwar007: yes i used both the commands cargo +nightly fmt as well as cargo fmt.

backend/grpc-server/src/utils.rs
```diff
@@ -140,6 +142,86 @@ pub fn auth_from_metadata(
     }
 }
 
+pub fn config_from_metadata(
```
NishantJoshi00: I am curious about the cost of this operation. Can you benchmark this and add the details for this? Not high priority but we should know what we are dealing with, as this would matter significantly during scale. If an integration specifically depends on changing the metadata

NishantJoshi00 (commented):
Also, can you add a real world test case? A screenshot for adyen with a config change or something else that simulates this for a specific flow

NishantJoshi00:
there are some merge conflicts

jarnura (approved):

su-shivanshmathur (approved):

PR #27, opened 2025-05-20
feat(connector): Added cards flow and unit tests for Authorizedotnet

swetasharma03:
## Description

Added authorizedotnet connector with the following flows
- Authorize
- Capture
- Psync 
- Void
- Refund
- RSync

### Additional Changes
- [ ] This PR modifies the API contract
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

using `cargo test`
```
export AUTHORIZENET_API_KEY=""
export AUTHORIZENET_KEY1=""
```
image

PR #39, opened 2025-06-06
feat(core): added resposne preprocessing in macros

chaitak-gorai:
## Description

Before parse_struct some connector have a preprocessing of raw response (like xml to json conversion in elavon, UTF-8 encoding parsing in authorizedotnet)

Added a optional bool key to pass `preprocess_response` in macro_rules! macro_connector_implementation for each flow.
If true `preprocess_response_bytes` function will be called before parse struct.
`preprocess_response_bytes` need to be implemented inside `create_all_prerequisites` in connector code.

 

### Additional Changes

## How did you test it?

Used in #27

### DIFF
--- a/backend/connector-integration/src/connectors/macros.rs
+++ b/backend/connector-integration/src/connectors/macros.rs
@@ -179,7 +179,38 @@ macro_rules! expand_fn_get_request_body {
 pub(crate) use expand_fn_get_request_body;
 
 macro_rules! expand_fn_handle_response {
-    ($connector: ty, $flow: ident, $resource_common_data: ty, $request: ty, $response: ty) => {
+    // When preprocess_response is true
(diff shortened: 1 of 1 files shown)


backend/connector-integration/src/connectors/macros.rs
```diff
@@ -179,7 +179,38 @@ macro_rules! expand_fn_get_request_body {
 pub(crate) use expand_fn_get_request_body;
 
 macro_rules! expand_fn_handle_response {
-    ($connector: ty, $flow: ident, $resource_common_data: ty, $request: ty, $response: ty) => {
+    // When preprocess_response is true
+    ($connector: ty, $flow: ident, $resource_common_data: ty, $request: ty, $response: ty, true) => {
```
ImSagnik007, line 183: can we have the boolean as a tagged argument?

ImSagnik007 (approved):

PR #51, opened 2025-06-20
ci: add GitHub Actions workflow for GHCR docker publishing

kanikac199:
### Additional Changes


.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,111 @@
+name: Docker Build and Push Check
+
+on:
+  push:
+    branches:
+      - main
```
SanchithHegde: Do you need the image to be pushed on every commit to main? If not, we can remove this trigger.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+            os: ubuntu-latest
+          - platform: linux/arm64
+            tag: linux-arm64
+            os: ubuntu-24.04-arm
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.ref }}
```
SanchithHegde: This value / variable would be available only during the `workflow_dispatch` event, this would not be available for the `push` event.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+          context: .
+          file: Dockerfile
+          platforms: ${{ matrix.platform }}
+          tags: ghcr.io/${{ github.repository }}:${{ env.SHORT_SHA }}-${{ matrix.tag }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          push: true
+        env:
+          CARGO_INCREMENTAL: ${{ github.event_name == 'pull_request' && '1' || '0' }}
```
SanchithHegde: This pipeline is not at all triggered by the `pull_request` event, so we won't need to test for the `pull_request` event?

.github/workflows/docker-to-ghcr-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+        run: |
+          COMMIT_SHA=$(git rev-parse HEAD)
+          SHORT_SHA=${COMMIT_SHA::7}
+          REF_INPUT="${{ github.event.inputs.ref }}"
+          
+          # Use ref input if provided, otherwise use commit hash
+          if [ -n "$REF_INPUT" ]; then
+            # Clean the ref input to make it suitable for Docker tag (replace / with -)
+            REF_CLEAN=$(echo "$REF_INPUT" | sed 's/\//-/g')
+            VERSION="0.1.0-nightly.$(date +'%Y%m%d')-${REF_CLEAN}"
```
knutties: Let us add hour/minute to allow for multiple builds in a day.

.github/workflows/build-and-deploy.yml
gopikrishna000: removed it

.github/workflows/build-and-deploy.yml
gopikrishna000: now using 
ref: ${{ github.event.inputs.ref || github.ref}}

.github/workflows/build-and-deploy.yml
gopikrishna000: changed it to 
CARGO_INCREMENTAL: 0

SanchithHegde:
Can we close this PR in favor of #66?

PR #54, opened 2025-06-23
Removing hyperswitch dependency

utkarsh-611:
### Additional Changes


backend/cards/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+error-stack = "0.4.1"
+serde = { version = "1.0.189", features = ["derive"] }
+thiserror = "1.0.69"
+time = "0.3.41"
+regex = "1.11.1"
+
+# First party crates
+hyperswitch_masking = { git = "https://github.com/juspay/hyperswitch", tag = "v1.111.4", package = "masking" }
```
knutties, line 15: should we just call this `masking` ?

backend/cards/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+common_utils = { version = "0.1.0", path = "../common_utils" }
```
knutties: you can skip the version here I guess.  Just the path should suffice.

backend/common_utils/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "common_utils"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+common_enums = { version = "*", path = "../common_enums" }
```
knutties: we can skip the version if we are depending on relative path.

backend/connector-integration/Cargo.toml
```diff
@@ -6,12 +6,10 @@ edition = "2021"
 [dependencies]
 domain_types ={ version = "0.1.0", path = "../domain_types" }
 base64 = "0.21"
-hyperswitch_domain_models = { git = "https://github.com/juspay/hyperswitch", tag = "v1.111.4" }
-hyperswitch_common_utils = { git = "https://github.com/juspay/hyperswitch", tag = "v1.111.4", package = "common_utils" }
-hyperswitch_common_enums = { git = "https://github.com/juspay/hyperswitch", tag = "v1.111.4", package = "common_enums" }
-hyperswitch_api_models = { git = "https://github.com/juspay/hyperswitch", tag = "v1.111.4", package = "api_models" }
+common_enums = { version = "*", path = "../common_enums" }
```
knutties: we can remove the version attribute for all the relative dependencies (lines 9, 10, 11, 23)

backend/connector-integration/src/utils/xml_utils.rs
```diff
@@ -1,5 +1,5 @@
 use bytes::Bytes;
-use hyperswitch_interfaces::errors as hs_errors;
+use interfaces::errors as hs_errors;
```
knutties: would be good to keep this as `errors` itself.  Better to not call stuff in here as hs as much as possible.

backend/grpc-server/Cargo.toml
```diff
@@ -40,6 +40,10 @@ grpc-api-types = { version = "0.1.0", path = "../grpc-api-types" }
 connector-integration = { version = "0.1.0", path = "../connector-integration" }
 external-services = { version = "0.1.0", path = "../external-services" }
 domain_types = { version = "0.1.0", path = "../domain_types" }
+interfaces = { version = "0.1.0", path = "../interfaces" }
```
knutties: lines 43, 44, 45, 46 - let us omit versions.

backend/interfaces/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "interfaces"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+common_utils = { version = "0.1.0", path = "../common_utils" }
```
knutties: let us omit versions in lines 8, 9, 10.

backend/domain_types/Cargo.toml
```diff
@@ -5,18 +5,28 @@ edition = "2021"
 
 [dependencies]
 grpc-api-types = { version = "0.1.0", path = "../grpc-api-types" }
+common_enums = { version = "*", path = "../common_enums" }
```
knutties: lines 8, 9, 10 - remove version attribute.

backend/external-services/Cargo.toml
```diff
@@ -33,4 +28,10 @@ async-trait = "0.1"
 tracing = { version = "0.1.40" }
 tracing-appender = { version = "0.2.3" }
 tracing-subscriber = { version = "0.3.18", default-features = true, features = ["env-filter", "json", "registry"] }
-tracing-attributes = "0.1.27"
\ No newline at end of file
+tracing-attributes = "0.1.27"
+
+domain_types ={ version = "*", path = "../domain_types"}
```
knutties: same comment as before.  let us skip `version` attribute.

Cargo.toml
```diff
@@ -5,7 +5,10 @@ members = [
   "backend/connector-integration",
   "backend/external-services",
   "backend/grpc-server", 
-  "sdk/rust-grpc-client"
+  "sdk/rust-grpc-client",
+  "backend/common_enums",
+  "backend/common_utils",
+  "backend/interfaces",
```
knutties, line 11: `domain_types` is not needed here as member?

jarnura (approved):

backend/cards/Cargo.toml
utkarsh-611: we will create a masking crate, which will contain all the functions and types related to masking like Email, Card Number, Account Number etc. So, to avoid conflicting names, we can call this as hyperswitch_masking only

SanchithHegde: Why would you need a separate masking crate for this repo? Why duplicate all the effort?

PR #57, opened 2025-06-25
feat: adding integrity framework support

utkarsh-611:
### Additional Changes


backend/external-services/src/service.rs
```diff
@@ -177,14 +181,28 @@ where
         }
         None => Ok(router_data),
     };
+
+    let result_with_integrity_check = match result {
+        Ok(data) => {
+            data.request
+                .check_integrity(&data.request.clone(), None)
+                .map_err(|err| ConnectorError::IntegrityCheckFailed {
```
utkarsh-611, line 189: Here I am failing currently, but what should be the behaviour here, here fallbacks are writtens in euler. For response verification in euler we are throwing INTERNAL_ERROR
Screenshot 2025-06-25 at 5 59 54 PM

jarnura (approved):

backend/domain_types/src/router_request_types.rs
```diff
@@ -65,3 +66,52 @@ pub struct ConnectorCustomerData {
     pub payment_method_data: Option<PaymentMethodData>,
     // pub split_payments: Option<SplitPaymentsRequest>,
 }
+
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct AuthoriseIntegrityObject {
+    /// Authorise amount
+    pub amount: MinorUnit,
+    /// Authorise currency
+    pub currency: common_enums::Currency,
```
knutties: Just `Currency` suffices.  You have already added the `use` statement.

jarnura (approved):

knutties:
@utkarsh-611 @jarnura - how is the integrity object populated in the domain type?  It is not coming in the proto payload input - so wondering what will be used to populate the domain type.  Or is this PR in preparation to update the proto in a later iteration?

utkarsh-611:
> @utkarsh-611 @jarnura - how is the integrity object populated in the domain type? It is not coming in the proto payload input - so wondering what will be used to populate the domain type. Or is this PR in preparation to update the proto in a later iteration?

@knutties when converting from api_type (proto) to domain_types we will keep this value as None. This integrity_object will be populated by the connector integration (L3) which then will be picked by the framework to do the integrity check

jarnura (dismissed):

jarnura (dismissed):

PR #58, opened 2025-06-25
fix: proto fixes (Add Implementations for RefundService and DisputeService)

Shailendra-03:
### Additional Changes

Unit tests execution: 
Screenshot 2025-06-26 at 3 31 06 PM

### DIFF
--- a/backend/grpc-server/src/app.rs
+++ b/backend/grpc-server/src/app.rs
@@ -8,7 +8,7 @@ use grpc_api_types::{
         payment_service_server, refund_service_handler, refund_service_server,
     },
 };
-use std::{future::Future, net};
+use std::{future::Future, net, sync::Arc};
 use tokio::{
     signal::unix::{signal, SignalKind},
     sync::oneshot,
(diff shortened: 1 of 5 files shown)


chaitak-gorai:
Bro can you verify by running tests in `elavon_payment_flows_test` if all the tests are running fine.

backend/grpc-server/src/app.rs
```diff
@@ -96,7 +101,13 @@ impl Service {
     pub async fn new(config: configs::Config) -> Self {
         Self {
             health_check_service: crate::server::health_check::HealthCheck,
-            payments_service: crate::server::payments::Payments { config },
+            payments_service: crate::server::payments::Payments {
+                config: config.clone(),
```
knutties: Do we need to clone here?

@NishantJoshi00 @SanchithHegde @jarnura  - what is the typical Rust approach here - this is the toml config - we possibly want to have only copy?

backend/grpc-server/src/app.rs
```diff
@@ -96,7 +101,13 @@ impl Service {
...
+            },
+            refunds_service: crate::server::refunds::Refunds {
+                config: config.clone(),
```
knutties: same as above.

backend/grpc-server/src/app.rs
jarnura: If we going to read the data only then clone not required, we can have Arc and we can just clone the Arc and not the object

jarnura (changes requested):

backend/grpc-server/src/app.rs
Shailendra-03: Added this change

jarnura (approved):

PR #68, opened 2025-07-01
feat: adding source verification framework

utkarsh-611:
### Additional Changes

### DIFF
--- a/backend/external-services/src/service.rs
+++ b/backend/external-services/src/service.rs
@@ -131,12 +131,11 @@ where
                 .with_label_values(&[&method.to_string(), service_name, connector_name])
                 .observe(external_service_elapsed);
             tracing::info!(?response, "response from connector");
-            
+
             match response {
                 Ok(body) => {
                     let response = match body {
                         Ok(body) => {
-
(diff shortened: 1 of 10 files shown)


backend/external-services/src/service.rs
```diff
@@ -166,6 +167,17 @@ where
                                 );
                                 tracing::Span::current().record("response.body", tracing::field::display(response.masked_serialize().unwrap_or(json!({ "error": "failed to mask serialize connector response"}))));
                             }
+
+
+                            let is_source_verified = connector.verify(&router_data, interfaces::verification::ConnectorSourceVerificationSecrets::AuthHeaders(router_data.connector_auth_type.clone()), &body.response)?;
```
utkarsh-611, line 170: in the payload, we are sending the response body, headers should also be sent ? But then what should we send in case on incoming webhook

jarnura: This application don't get the Incoming webhooks, the consumer of this application only get the webhooks, so incoming webhooks won't be a problem. This addresses your query?

PR #69, opened 2025-07-01
fix: Add missing wallet code in proto

Shailendra-03:
### Additional Changes


backend/grpc-api-types/proto/validate/validate.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+
+package validate;
+
+import "google/protobuf/descriptor.proto";
+
+option go_package = "github.com/envoyproxy/protoc-gen-validate/validate";
+
```
Shailendra-03, line 8: @jarnura @knutties I have added this file for now to ensure that the current validation rules is handled correclty. But maybe we should handle it more gracefully, should i copy the validate.proto from [here](https://github.com/bufbuild/protoc-gen-validate/blob/main/validate/validate.proto) or add this as [submodule](https://github.com/envoyproxy/protoc-gen-validate.git) and use, so in this case we may not need to worry about maintaining it. What should be done here?

jarnura: If we use submodule approach will that automatically update if there is a change in the upstream?

If we copy and having a version in our side we need to maintain that, if submodule approach eliminate that then we can use submodule based one.

cc: @knutties

Shailendra-03: @jarnura using submodules, we will not need to maintain it, we can update whenever we want using this command `git submodule update --remote`

PR #98, opened 2025-07-29
fix: art changes and docker configuration updates

shivral:
Various fixes and improvements including art changes, docker configuration updates, and development environment fixes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,15 +4,15 @@
 ########################################
 # 1. Base image with necessary tools
 ########################################
-FROM rust:slim-bookworm AS base
+FROM public.ecr.aws/docker/library/rust:slim-bookworm AS base
 
 # Install system dependencies and clean up
 RUN apt-get update \
     && apt-get install -y \
-       pkg-config \
-       libssl-dev \
-       g++ \
-       make \
+    pkg-config \
+    libssl-dev \
+    g++ \
+    make \
     && apt-get clean \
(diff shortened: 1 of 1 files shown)


backend/external-services/src/service.rs
```diff
@@ -4,8 +4,87 @@ use common_utils::ext_traits::AsyncExt;
...
     request::{Method, Request, RequestContent},
 };
+
+/// Trait to get flow name from flow types
+pub trait GetFlowName {
+    fn get_flow_name() -> &'static str;
+}
+
+// Implement GetFlowName for all flow types
+impl GetFlowName for domain_types::connector_flow::CreateOrder {
```
knutties: Isn't the flow the same as the grpc method name - do we need two names for it?

shivral: yes @knutties ,This was just a draft PR i made, these are just temp changes i did for ART to work i will remove this function all together.
I will change the status to draft

config/development.toml
```diff
@@ -29,6 +29,11 @@ port = 8080
 idle_pool_connection_timeout = 90               # Timeout for idle pool connections (defaults to 90s)
 bypass_proxy_urls = ["localhost", "local"]
 
+[test]
+enabled = true
```
maverox: can we have this false by default?"

shivral: yes yes it got pushed by mistake while testing

PR #106, opened 2025-07-31
fix: Razorpay error status fix

shivral:
### Additional Changes


swangi-kumari (approved):

backend/connector-integration/src/connectors/razorpayv2.rs
```diff
@@ -117,7 +117,7 @@ impl ConnectorCommon for RazorpayV2 {
                 (
                     "ROUTE_ERROR".to_string(),
                     message.clone(),
-                    AttemptStatus::Unknown,
+                    AttemptStatus::Failure,
```
jarnura, line 120: if this is unknown then the status should be pending rather than failure, any reasons for this to making this as failure

shivral: @jarnura no this is failure only for example invalid vpa etc will come under this PR is majorly to change this only

PR #157, opened 2025-09-01
chore: added webhooks support in Novalnet

Saptak88:
### Additional Changes

### DIFF
--- a/backend/connector-integration/src/connectors/novalnet/transformers.rs
+++ b/backend/connector-integration/src/connectors/novalnet/transformers.rs
@@ -10,10 +10,11 @@ use common_utils::{
 use domain_types::{
     connector_flow::{self, Authorize, PSync, RSync, RepeatPayment, SetupMandate, Void},
     connector_types::{
-        MandateReference, MandateReferenceId, PaymentFlowData, PaymentVoidData,
+        EventType, MandateReference, MandateReferenceId, PaymentFlowData, PaymentVoidData,
         PaymentsAuthorizeData, PaymentsCaptureData, PaymentsResponseData, PaymentsSyncData,
-        RefundFlowData, RefundSyncData, RefundsData, RefundsResponseData, RepeatPaymentData,
-        ResponseId, SetupMandateRequestData,
+        RefundFlowData, RefundSyncData, RefundWebhookDetailsResponse, RefundsData,
+        RefundsResponseData, RepeatPaymentData, ResponseId, SetupMandateRequestData,
+        WebhookDetailsResponse,
     },
     errors::{self, ConnectorError},
     payment_method_data::{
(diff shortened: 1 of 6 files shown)


Shubhodip900 (approved):

backend/connector-integration/src/connectors/novalnet/transformers.rs
```diff
@@ -2210,3 +2230,132 @@ impl<
...
+                status_code: 200,
+                connector_response_reference_id: None,
+                error_code: Some(notif.result.status.to_string()),
+                error_message: Some(notif.result.status_text),
+                raw_connector_response: None,
+                response_headers: None,
+            }),
+        }
+    }
+}
```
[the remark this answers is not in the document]
Saptak88: Fixed it

backend/interfaces/src/connector_types.rs
```diff
@@ -172,11 +172,88 @@ pub trait DisputeDefend:
...
     fn verify_webhook_source(
         &self,
-        _request: RequestDetails,
-        _connector_webhook_secret: Option<ConnectorWebhookSecrets>,
+        request: RequestDetails,
+        connector_webhook_secret: Option<ConnectorWebhookSecrets>,
         _connector_account_details: Option<ConnectorAuthType>,
     ) -> Result<bool, error_stack::Report<domain_types::errors::ConnectorError>> {
-        Ok(false)
+        let algorithm = self
```
AmitsinghTanwar007: As per me this is only a placeholder type interface pls don't change the code here instead implement the trait in your code base of {connector}.rs

backend/interfaces/src/connector_types.rs
```diff
@@ -172,11 +172,88 @@ pub trait DisputeDefend:
...
+        _request: &RequestDetails,
+    ) -> Result<
+        Box<dyn crypto::VerifySignature + Send>,
+        error_stack::Report<domain_types::errors::ConnectorError>,
+    > {
+        Ok(Box::new(crypto::NoAlgorithm))
+    }
+
+    /// fn get_webhook_source_verification_signature
+    fn get_webhook_source_verification_signature(
```
AmitsinghTanwar007, line 183: And this implementation of webhook  trait is there inside interfaces/src/webhook.rs

backend/interfaces/src/connector_types.rs
Saptak88: In Hyperswitch we are using the same logic tho and will be a cleaner implementation of webhooks in connector.rs

backend/interfaces/src/connector_types.rs
Saptak88: But we are not using the IncomingWebhook trait in interfaces/src/webhook.rs

backend/interfaces/src/connector_types.rs
AmitsinghTanwar007: bro here we are making a new service which cannot decide to give error we should only pass what we like 
this function has to make a decision is verification successful or not just give true or false 
if header not found means false now the consumer hyperswitch will decide if he wants to give error or not

AmitsinghTanwar007 (approved):

PR #163, opened 2025-09-02
feat(core): access token flow

kanikac199:
### Additional Changes

### DIFF
--- a/backend/domain_types/src/connector_types.rs
+++ b/backend/domain_types/src/connector_types.rs
@@ -427,6 +427,17 @@ impl PaymentFlowData {
             .ok_or_else(missing_field_err("session_token"))
     }
 
+    pub fn get_access_token(&self) -> Result<String, Error> {
+        self.access_token
+            .clone()
+            .ok_or_else(missing_field_err("access_token"))
+    }
+
+    pub fn set_access_token(mut self, access_token: Option<String>) -> Self {
+        self.access_token = access_token;
+        self
+    }
+
     pub fn get_billing_first_name(&self) -> Result<Secret<String>, Error> {
         self.address
             .get_payment_method_billing()
@@ -677,6 +688,13 @@ impl PaymentFlowData {
         self
     }
 
+    pub fn set_access_token_id(mut self, access_token_id: Option<String>) -> Self {
+        if access_token_id.is_some() && self.access_token.is_none() {
+            self.access_token = access_token_id;
+        }
+        self
+    }
+
     pub fn get_return_url(&self) -> Option<String> {
         self.return_url.clone()
     }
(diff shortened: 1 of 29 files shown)


backend/grpc-server/src/utils.rs
```diff
@@ -169,7 +169,7 @@ pub fn auth_from_metadata(
             key1: parse_metadata(metadata, X_KEY1)?.to_string().into(),
             api_secret: parse_metadata(metadata, X_API_SECRET)?.to_string().into(),
         }),
-        "multi-auth-key" => Ok(ConnectorAuthType::MultiAuthKey {
+        "huth-key" => Ok(ConnectorAuthType::MultiAuthKey {
```
Shubhodip900: Why changed "multi-auth-key" to "huth-key" ?

kanikac199: it was by accident, reverted

backend/grpc-server/src/server/payments.rs
```diff
@@ -258,6 +259,59 @@ impl Payments {
...
 
+        // Extract access token from Hyperswitch request
+        let cached_access_token = payload.access_token.clone();
+
+        // Check if connector supports access tokens
+        let should_do_access_token = connector_data.connector.should_do_access_token();
+
+        // Conditional token generation - ONLY if not provided in request
+        let payment_flow_data = if should_do_access_token {
+            if cached_access_token.is_some() {
```
ImSagnik007: Instead of doing with if else, can we write this with match cases (Some(token) and None), then you won't need to have .map(). And also calculate the access_token_data first and set in payment_flow_data in one place outside the match case.

backend/domain_types/src/connector_types.rs
```diff
@@ -687,6 +705,19 @@ impl PaymentFlowData {
         self
     }
 
+    pub fn set_access_token_id(mut self, access_token_id: Option<String>) -> Self {
+        if let Some(token_id) = access_token_id {
```
ImSagnik007: Instead of nested ifs, can we try this out? `if let ((Some(token_id), None) = (access_token_id, self.access_token)`

backend/grpc-server/src/server/payments.rs
kanikac199: Replaced the nested if cached_access_token.is_some() with a match cached_access_token pattern also calculated access_token_data first and set it in payment_flow_data in one place outside the match case.

backend/domain_types/src/connector_types.rs
kanikac199: Yes we can do that, replaced the nested if let conditions with: `if let (Some(token_id), None) = (access_token_id, &self.access_token)`

PR #178, opened 2025-09-15
feat(core): pointing to hyperswitch main rather than a rev

su-shivanshmathur:
## Description

Connector Service pointing to `main` branch of hyperswitch

## Motivation and Context

For any new addition to hyperswitch, better to point to main of it so that no need of updating the git reference again and again

### Additional Changes

### DIFF
--- a/backend/cards/Cargo.toml
+++ b/backend/cards/Cargo.toml
@@ -14,7 +14,7 @@ bytes = "1"
 prost = "0.13"
 
 # First party crates
-hyperswitch_masking = { git = "https://github.com/juspay/hyperswitch", rev = "e692c3b8e", package = "masking" }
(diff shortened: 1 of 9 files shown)


backend/cards/Cargo.toml
```diff
@@ -14,7 +14,7 @@ bytes = "1"
 prost = "0.13"
 
 # First party crates
-hyperswitch_masking = { git = "https://github.com/juspay/hyperswitch", rev = "e692c3b8e", package = "masking" }
+hyperswitch_masking = { git = "https://github.com/juspay/hyperswitch", branch = "main", package = "masking" }
```
gopikrishna000, line 17: should it point to some stable version? as main could keep changing?

And we dont know if the new change will be compatible wrt connector-service right? anyway we have to test and update the version in such case?

PR #183, opened 2025-09-18
ci: add job to run tests

Debarshi-Gupta:
### "Run Tests" Job Details

The "Run Tests" workflow is a comprehensive testing job that sets up a Rust environment with stable toolchain and Protocol Buffers support, securely downloads and decrypts connector credentials from AWS S3 (when available), and executes the full test suite using nextest with CI-optimized configuration - it includes intelligent handling for fork PRs by gracefully skipping credential-dependent tests when secrets aren't accessible, ensuring both security and functionality across different contribution scenarios.

### Environment & Setup

- Runs on `ubuntu-latest`
- Only executes for pushes to main or PRs from the same repository (security measure)
- Sets `RUSTFLAGS="-D warnings"` to treat warnings as errors

### Main Steps

1. __Code checkout__ - Gets the repository code

2. __Rust toolchain setup__ - Installs stable Rust

3. __Protocol Buffers setup__ - Installs Protoc for gRPC compilation

4. __Caching__ - Uses Rust cache for faster builds (with cache-on-failure enabled)

5. __Test runner installation__ - Installs `nextest` for enhanced test execution

6. __Credential management__ - Downloads and decrypts connector credentials from AWS S3:

   - Downloads encrypted credentials file from S3 bucket
   - Decrypts using GPG with a passphrase
   - Handles cases where credentials aren't available (e.g., fork PRs)

7. __Environment configuration__ - Sets `CONNECTOR_AUTH_FILE_PATH` if credentials are available

8. __Test execution__ - Runs tests using `cargo nextest run` with CI profile configuration

### Security Features

- Gracefully handles missing secrets (common for fork PRs)
- Uses encrypted credential storage in S3
- Only runs full tests when proper authentication is available

The workflow uses `nextest` instead of standard `cargo test` for better performance and reporting, with configuration defined in `.nextest.toml`

### Additional Changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,39 @@ jobs:
           tool: nextest
           checksum: true
 
+      - name: Download Encrypted Connector Credentials from S3 and Decrypt
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.CONNECTOR_CREDS_AWS_ACCESS_KEY_ID }}
+          AWS_REGION: ${{ secrets.CONNECTOR_CREDS_AWS_REGION }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.CONNECTOR_CREDS_AWS_SECRET_ACCESS_KEY }}
+          CONNECTOR_AUTH_PASSPHRASE: ${{ secrets.CONNECTOR_AUTH_PASSPHRASE }}
+          CONNECTOR_CREDS_S3_BUCKET_URI: ${{ secrets.CONNECTOR_CREDS_S3_BUCKET_URI}}
+          DESTINATION_FILE_NAME: "creds.json.gpg"
+          S3_SOURCE_FILE_NAME: "9f00461f-5a48-4f9c-aa8c-3510e4cb7d9c.json.gpg"
+        shell: bash
+        run: |
+          # Skip credential download for forks or when secrets are not available
+          if [[ -z "$AWS_ACCESS_KEY_ID" || -z "$CONNECTOR_AUTH_PASSPHRASE" ]]; then
+            echo "Skipping credential download - secrets not available (likely a fork PR)"
+            exit 0
+          fi
+
+          mkdir -p ".github/secrets" ".github/test"
+
+          aws s3 cp "${CONNECTOR_CREDS_S3_BUCKET_URI}/${S3_SOURCE_FILE_NAME}" ".github/secrets/${DESTINATION_FILE_NAME}"
+          gpg --quiet --batch --yes --decrypt --passphrase="${CONNECTOR_AUTH_PASSPHRASE}" --output ".github/test/creds.json" ".github/secrets/${DESTINATION_FILE_NAME}"
+
+      - name: Set connector credentials path in environment
+        shell: bash
+        run: |
+          # Only set the environment variable if the credentials file exists
+          if [[ -f ".github/test/creds.json" ]]; then
+            echo "CONNECTOR_AUTH_FILE_PATH=${{ github.workspace }}/.github/test/creds.json" >> $GITHUB_ENV
+            echo "Connector credentials available for tests"
+          else
+            echo "No connector credentials available - tests will use default values"
+          fi
+
       - name: Run Tests
         shell: bash
         run: cargo nextest run --config-file .nextest.toml --profile ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,8 +75,15 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2.7.8
 
+      - name: Install nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+          checksum: true
+
       - name: Run Tests
-        run: cargo test --all-features --workspace
+        shell: bash
+        run: cargo nextest run --config-file .nextest.toml --profile ci
 
   formatting:
     name: Check formatting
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -52,6 +52,32 @@ jobs:
...
+
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: Swatinem/rust-cache@v2.7.8
+
+      - name: Run Tests
+        run: cargo test --all-features --workspace
```
SanchithHegde: Instead of running `cargo test`, you can also run [`nextest`](https://nexte.st/), which is able to run tests quicker (as long as the tests are independent).

You'd have to:

1. [Install nextest](https://github.com/juspay/framework-libs-rs/blob/904ed03f101de7be63183e717263294000e3f6be/.github/workflows/ci.yaml#L73-L77) in this CI job.
2. Include the [nextest config file](https://github.com/juspay/framework-libs-rs/blob/904ed03f101de7be63183e717263294000e3f6be/.nextest.toml) in this repo. Let's say the file is named `.nextest.toml`.
3. Run this command in the CI job:

   ```shell
   cargo nextest run --no-tests warn --config-file .nextest.toml --all-features --profile ci
   ```

4. Optionally, you can also add a `Makefile` recipe that runs `cargo nextest run --no-tests warn --config-file .nextest.toml --all-features` to ease running tests locally. (I've intentionally excluded the `--profile ci` part here.)

Also, nit: specify `shell: bash` for this step.

.github/workflows/ci.yml
```diff
@@ -52,6 +52,39 @@ jobs:
...
+
+      - name: Install nextest
+        uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+          checksum: true
+
+      - name: Run Tests
+        shell: bash
+        run: cargo nextest run --config-file .nextest.toml --profile ci
```
SanchithHegde, line 132: Do we plan to include the `--all-features` flag, or would it not work?

Debarshi-Gupta: Adding --all-features to the nextest command enabled the `kafka` feature in the ucs_common_utils crate, which caused the application to attempt initializing a real Kafka connection during startup, causing the tests to fail with a "Failed to initialize Kafka writer" error.

.github/workflows/ci.yml
Debarshi-Gupta: Resolved, thanks

.github/workflows/ci.yml
```diff
@@ -52,6 +52,74 @@ jobs:
...
+          CONNECTOR_CREDS_S3_BUCKET_URI: ${{ secrets.CONNECTOR_CREDS_S3_BUCKET_URI}}
+          DESTINATION_FILE_NAME: "creds.json.gpg"
+          S3_SOURCE_FILE_NAME: "9f00461f-5a48-4f9c-aa8c-3510e4cb7d9c.json.gpg"
+        shell: bash
+        run: |
+          # Skip credential download for forks or when secrets are not available
+          if [[ -z "$AWS_ACCESS_KEY_ID" || -z "$CONNECTOR_AUTH_PASSPHRASE" ]]; then
+            echo "Skipping credential download - secrets not available (likely a fork PR)"
+            exit 0
+          fi
```
SanchithHegde: You'll have to find a way to skip the tests that require the connector credentials, and run the remaining ones which don't require credentials, in case PRs are opened from forks.

Or, you'd have to skip the entire job without running any tests, in case a PR has been opened from a fork.

Debarshi-Gupta: Resolved, thanks.

.github/workflows/ci.yml
```diff
@@ -52,6 +55,75 @@ jobs:
       - name: Run Clippy
         run: cargo clippy --all-features --all-targets
 
+  test:
+    name: Run Tests
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
SanchithHegde: Note that adding the `if` here will prevent you from including this check as a required check, because the check would be in status `skipped` but not `succeeded`, when a PR is opened from a forked repository.

This is how we skip tests in the `hyperswitch` repo:
https://github.com/juspay/hyperswitch/blob/239b6d37c6ba35117179ee37159963e9808cbf11/.github/workflows/cypress-tests-runner.yml#L166-L169

The environment variable is set here:
https://github.com/juspay/hyperswitch/blob/239b6d37c6ba35117179ee37159963e9808cbf11/.github/workflows/cypress-tests-runner.yml#L20

And the remaining steps all have such a condition, haven't found a better way to handle this:
https://github.com/juspay/hyperswitch/blob/239b6d37c6ba35117179ee37159963e9808cbf11/.github/workflows/cypress-tests-runner.yml#L172

.github/workflows/ci.yml
SanchithHegde: Once that change is done, I think you can remove this snippet: we can always expect secrets to be available.

.github/workflows/ci.yml
Debarshi-Gupta: I have resolved it, thanks.

.github/workflows/ci.yml
Debarshi-Gupta: I have resolved it, thanks.

SanchithHegde (approved):

Anurag-05-prog (approved):

bsayak03 (approved):

PR #186, opened 2025-09-19
fix: docker public repo fix

shivral:
## Motivation and Context

Need to change to ecr public repo 

### Additional Changes


backend/common_utils/src/lineage.rs
```diff
@@ -1,8 +1,9 @@
 //! Lineage ID domain types for tracking request lineage across services
 
-use serde;
 use std::collections::HashMap;
 
+use serde;
```
SanchithHegde: Nit: We can remove this import, this does not make any difference.

sai-harsha-vardhan (approved):

maverox (approved):

PR #195, opened 2025-09-22
Shadow ucs

AmitsinghTanwar007:
## Description

1) UCS Dry Run Mode (Simulation Support)
Added ucs_dry_run flag across all payment flows.
Supported operations: authorize, capture, refund, disputes, session tokens, access tokens, order creation.
Enables safe mock mode testing without hitting external connectors.

### Additional Changes

### DIFF
--- a/backend/common_utils/src/consts.rs
+++ b/backend/common_utils/src/consts.rs
@@ -41,6 +41,8 @@ pub const X_CONNECTOR: &str = "x-connector";
 pub const X_MERCHANT_ID: &str = "x-merchant-id";
 /// Header key for reference identification
 pub const X_REFERENCE_ID: &str = "x-reference-id";
+/// Header key for UCS dry run mode
+pub const X_UCS_DRY_RUN: &str = "x-ucs-dry-run";
 
 // =============================================================================
 // Authentication Headers (Internal)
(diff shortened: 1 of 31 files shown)


backend/common_utils/src/consts.rs
```diff
@@ -41,6 +41,8 @@ pub const X_CONNECTOR: &str = "x-connector";
 pub const X_MERCHANT_ID: &str = "x-merchant-id";
 /// Header key for reference identification
 pub const X_REFERENCE_ID: &str = "x-reference-id";
+/// Header key for UCS dry run mode
+pub const X_UCS_DRY_RUN: &str = "x-ucs-dry-run";
```
knutties, line 45: just `x-dry-run` should suffice I think.  We might change the name from UCS itself.

backend/grpc-api-types/proto/payment.proto
```diff
@@ -1188,6 +1188,7 @@ message PaymentServiceAuthorizeResponse {
   
   // Raw Response
   optional string raw_connector_response = 9; // Raw response from the connector for debugging
+  optional string raw_connector_request = 14; // Raw request sent to the connector for debugging
```
knutties, line 1191: Why is this change only in `PaymentServiceAuthorizeResponse` - all service Responses should have it right for comprehensive comparison?

AmitsinghTanwar007: yes i created a different PR for this and got it merge which only focuses on raw_connector_request

PR #224, opened 2025-10-17
feat: Adding a new field (account_metadata) to payments.proto

shuklatushar226:
## Description

adding a new field in connector service to receive merchant_account_metadata for consuming in some connectors

### Additional Changes

## How did you test it?

i connected hyperswitch to ucs and ran cypress test and got account_metadata(merchant_account_metadata) in the gprc 
Screenshot 2025-10-17 at 12 56 00 PM

### DIFF
--- a/backend/connector-integration/src/connectors/fiserv/transformers.rs
+++ b/backend/connector-integration/src/connectors/fiserv/transformers.rs
@@ -321,7 +321,8 @@ impl TryFrom<&Option<pii::SecretSerdeValue>> for FiservSessionObject {
             .and_then(|secret_value| match secret_value.peek() {
                 serde_json::Value::String(s) => Ok(s.clone()),
                 _ => Err(report!(ConnectorError::InvalidConnectorConfig {
-                    config: "FiservSessionObject in merchant_account_metadata was not a JSON string",
+                    config:
+                        "FiservSessionObject in merchant_account_metadata was not a JSON string",
                 })),
             })?;
 
@@ -440,7 +441,8 @@ impl<
             serde_json::Value::String(s) => s,
             _ => {
                 return Err(report!(ConnectorError::InvalidConnectorConfig {
-                    config: "merchant_account_metadata was not a JSON string for FiservSessionObject",
+                    config:
+                        "merchant_account_metadata was not a JSON string for FiservSessionObject",
                 }))
             }
         };
--- a/backend/connector-integration/src/connectors/bluecode/transformers.rs
+++ b/backend/connector-integration/src/connectors/bluecode/transformers.rs
@@ -126,14 +126,16 @@ impl TryFrom<&pii::SecretSerdeValue> for BluecodeMetadataObject {
             serde_json::Value::String(s) => s.clone(),
             _ => {
                 return Err(report!(errors::ConnectorError::InvalidConnectorConfig {
-                    config: "BluecodeMetadataObject in merchant_account_metadata was not a JSON string",
+                    config:
+                        "BluecodeMetadataObject in merchant_account_metadata was not a JSON string",
                 }));
             }
         };
 
         serde_json::from_str(&secret_value_str).change_context(
             errors::ConnectorError::InvalidConnectorConfig {
-                config: "Deserializing BluecodeMetadataObject from merchant_account_metadata string",
+                config:
+                    "Deserializing BluecodeMetadataObject from merchant_account_metadata string",
             },
         )
     }
--- a/backend/connector-integration/src/connectors/cybersource/transformers.rs
+++ b/backend/connector-integration/src/connectors/cybersource/transformers.rs
@@ -939,7 +939,10 @@ impl<
             .to_string();
 
         let connector_merchant_config = CybersourceConnectorMetadataObject::try_from(
-            &item.router_data.resource_common_data.merchant_account_metadata,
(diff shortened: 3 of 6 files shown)


backend/grpc-api-types/proto/payment.proto
```diff
@@ -1189,6 +1189,9 @@ message PaymentServiceAuthorizeRequest {
   optional PaymentExperience payment_experience = 30; // Preferred payment experience
 
   optional bool test_mode = 31; // A boolean value to indicate if the connector is in Test mode
+
+  // Account-Specific Configuration
+  map<string, string> account_metadata = 33; // Account-level configuration metadata (e.g., merchant_name, brand_id)
```
AmitsinghTanwar007: it can be named as merchant metadata something like this

backend/grpc-api-types/proto/payment.proto
```diff
@@ -1572,6 +1575,9 @@ message PaymentServiceRegisterRequest {
   optional CustomerAcceptance customer_acceptance = 26; // Details of customer acceptance
   optional BrowserInformation browser_info = 27; // Information about the customer's browser
   optional PaymentExperience payment_experience = 28; // Preferred payment experience
+
+  // Account-Specific Configuration
+  map<string, string> account_metadata = 29; // Account-level configuration metadata (e.g., merchant_name, brand_id)
```
AmitsinghTanwar007: same here

backend/grpc-api-types/proto/payment.proto
```diff
@@ -1636,6 +1642,9 @@ message PaymentServiceRepeatEverythingRequest {
 
   optional bool test_mode = 31; // A boolean value to indicate if the connector is in Test mode
   optional PaymentMethodType payment_method_type = 32; // Indicates the sub type of payment method. Eg: 'google_pay' & 'apple_pay'
+
+  // Account-Specific Configuration
+  map<string, string> account_metadata = 33; // Account-level configuration metadata (e.g., merchant_name, brand_id)
```
AmitsinghTanwar007: same here

backend/domain_types/src/types.rs
```diff
@@ -1615,9 +1615,16 @@ impl ForeignTryFrom<(PaymentServiceAuthorizeRequest, Connectors, &MaskedMetadata
             description: value.metadata.get("description").cloned(),
             return_url: value.return_url.clone(),
             connector_meta_data: {
```
AmitsinghTanwar007: dont change the value connector_meta_data
add a new field merchant_metadata

backend/grpc-api-types/proto/payment.proto
shuklatushar226: Arun asked for that name

backend/grpc-api-types/proto/payment.proto
shuklatushar226: +1

backend/grpc-api-types/proto/payment.proto
shuklatushar226: +1

Debarshi-Gupta (approved):

backend/grpc-api-types/proto/payment.proto
Debarshi-Gupta: Please check whether this field should be optional or mandatory

backend/grpc-api-types/proto/payment.proto
Debarshi-Gupta: Please check whether this field should be optional or mandatory

backend/grpc-api-types/proto/payment.proto
Debarshi-Gupta: Please check whether this field should be optional or mandatory

backend/grpc-api-types/proto/payment.proto
shuklatushar226: map fields are by default optional as it is set to {} and even if we try to make map optional i think map doesn't support optional keyword

backend/grpc-api-types/proto/payment.proto
shuklatushar226: map fields are by default optional as it is set to {} and even if we try to make map optional i think map doesn't support optional keyword

backend/grpc-api-types/proto/payment.proto
shuklatushar226: map fields are by default optional as it is set to {} and even if we try to make map optional i think map doesn't support optional keyword

backend/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -427,11 +428,11 @@ impl<
...
-            .connector_meta_data
+            .merchant_account_metadata
             .as_ref()
             .ok_or_else(|| {
                 report!(ConnectorError::MissingRequiredField {
-                    field_name: "connector_meta_data for FiservSessionObject"
+                    field_name: "merchant_account_metadata for FiservSessionObject"
                 })
             })?
             .peek();
```
Debarshi-Gupta, line 438: Previously, we had connector_meta_data in router_data, and we were putting it in session_meta_value. Now, since you have merchant_account_metadata, you need to extract the connector_meta_data from merchant_account_metadata and put it here.

PR #231, opened 2025-10-22
fix(cybersource): use security_code and state_code in authorize flow

Shubhodip900:
## Description

Fixed a bug in the CyberSource where in Authorize flow the `security_code` was set as `None` and `administrative_area` was not getting exact `state_code`.
### Additional Changes

## How did you test it?

Tested using Diff Checker:
Screenshot 2025-10-22 at 5 37 25 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to Connector Service will be documented here.
 
 - - -
 
+## 2025.10.28.0
+
+### Features
+
+- **connector:** [Worldpay] Connector Integration ([#216](https://github.com/juspay/connector-service/pull/216)) ([`61945b2`](https://github.com/juspay/connector-service/commit/61945b24a70d6e50d8c607a519fa2634fac8a68d))
+
+### Bug Fixes
+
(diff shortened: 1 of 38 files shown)


backend/connector-integration/src/connectors/cybersource/transformers.rs
```diff
@@ -2949,9 +2949,16 @@ impl<
...
+        let connector_response = item
+            .response
+            .processor_information
+            .as_ref()
+            .map(AdditionalPaymentMethodConnectorResponse::from)
+            .map(domain_types::router_data::ConnectorResponseData::with_additional_payment_method_data);
         Ok(Self {
             resource_common_data: PaymentFlowData {
                 status,
+                connector_response,
```
Saptak88, line 2961: Dont we need connector_response for complete_Authorize & Setup_Mandate

Shubhodip900: Added for Setup_Mandate, equivalents flows of complete_Authorize don't have connector_response in HS. They are using different TryFrom

PR #232, opened 2025-10-26
fix(Access_token_flow): Added proto field to accept expires_in_seconds in request

Shubhodip900:
## Description

Current request body does not have field to accept expires_in_seconds in request body from Hyperswitch. Added `state` field which contain both `access_token` and `expires_in_seconds`.
### Additional Changes

### DIFF
--- a/backend/domain_types/src/types.rs
+++ b/backend/domain_types/src/types.rs
@@ -1153,7 +1153,11 @@ impl<
             .state
             .as_ref()
             .and_then(|state| state.access_token.as_ref())
-            .map(|token| token.token.clone());
+            .map(|token| crate::connector_types::AccessTokenResponseData {
+                access_token: token.token.clone(),
+                token_type: None,
+                expires_in: token.expires_in_seconds,
+            });
         Ok(Self {
             capture_method: Some(common_enums::CaptureMethod::foreign_try_from(
                 value.capture_method(),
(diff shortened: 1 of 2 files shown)


backend/grpc-server/tests/aci_payment_flows_test.rs
```diff
@@ -266,6 +268,7 @@ fn create_refund_request(transaction_id: &str) -> PaymentServiceRefundRequest {
         capture_method: None,
         request_ref_id: None,
         access_token: None,
```
kanikac199: Can we remove the access_token from all requests now since the state field has been added — the access_token will be sent as part of that field.

backend/grpc-api-types/proto/payment.proto
Aishwariyaa-Anand: why do we still need access_token in AuthorizeRequest when ConnectorState is there

backend/domain_types/src/types.rs
```diff
@@ -1771,6 +1771,16 @@ impl
...
         let merchant_id_from_header = extract_merchant_id_from_metadata(metadata)?;
 
+        let access_token = value
+            .state
+            .as_ref()
+            .and_then(|state| state.access_token.as_ref())
+            .map(|token| crate::connector_types::AccessTokenResponseData {
+                access_token: token.token.clone(),
+                token_type: None,
+                expires_in: token.expires_in_seconds,
```
Aishwariyaa-Anand, line 1791: The expires_in value will likely change between the time Hyperswitch sends UCS the access token and when UCS sends the response back. 
Could you confirm if it’s intentional to reuse the original expiry here, will this be handled at HS or should UCS return the updated expiry?

Shubhodip900: This would be handled in HS side only. We will take the expiry time from redis and the token's TTL (time to live).

kanikac199 (approved):

PR #242, opened 2025-11-04
docs(setup.md): toml always prod.toml issue fix for docker

gopikrishna000:
## Description
The toml configuration choosing logic is broken - needs to fallback to development config in the absence of any specific environment.

## Motivation and Context
Docker images were not using the correct default config i.e. development config when images are built.

## How did you test it?
By creating the docker image, running the container with the image and accessing the endpoints.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -257,45 +257,17 @@ docker run --rm -p 8000:8000 -p 8080:8080 \
 grpcurl -plaintext localhost:8000 grpc.health.v1.Health/Check
 
 
-#### Building with Podman
+### Podman Setup
 
-**Memory Requirements:**
-- **Minimum**: 12GB for successful build
-- **Recommended**: 16GB for optimal performance
+**Memory Requirements:** 12GB minimum for successful build.
 
 ```bash
-# Set machine memory (requires restart)
-podman machine stop
+# Configure memory (one-time setup)
 podman machine set --memory 12288
-podman machine start
 
-# Build the image
+# Use same commands as Docker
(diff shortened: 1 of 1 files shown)


knutties (dismissed):

knutties (approved):

setup.md
```diff
@@ -215,6 +215,94 @@ UCS supports the following operations across multiple payment processors:
...
+
+#### Building with Podman
+
+**Memory Requirements:**
+- **Minimum**: 12GB for successful build
+- **Recommended**: 16GB for optimal performance
+
+```bash
+# Set machine memory (requires restart)
+podman machine stop
```
hrithikesh026: why use podman instead of docker?

knutties: @hrithikesh026  - some folks who are looking at connector-service are using podman - so added instructions for them.

jarnura (approved):

setup.md
```diff
@@ -215,6 +215,66 @@ UCS supports the following operations across multiple payment processors:
...
+
+**Memory Requirements:** 12GB minimum for successful build.
+
+```bash
+# Configure memory (one-time setup)
+podman machine set --memory 12288
+
+# Use same commands as Docker
+podman build -f Dockerfile -t ucs:latest .
+podman run --rm -p 8000:8000 -p 8080:8080 ucs:latest
```
hrithikesh026, line 270: should we maintain consistency by only using docker commands even for running the images?

PR #266, opened 2025-11-12
fix(Access_token): Corrected type and field name of access_token 

Shubhodip900:
## Description

- Current proto description was incorrect for access token in all the request bodies. The previous type was `ConnectorState` inside which we have both `access_token` and `connector_customer_id` separately we also have `connector_customer_id` field in the request due to which same field was repetative. Also corrected the field name.

- Fixed non_3ds decision logic in Fiuu connector. (https://github.com/juspay/hyperswitch-cloud/issues/12032)
- Fixed expiry year diff in Noon connector. (https://github.com/juspay/hyperswitch-cloud/issues/12074)
### Additional Changes

## How did you test it?
Fiuu Diff fix:- 
Screenshot 2025-11-13 at 3 55 40 PM
Screenshot 2025-11-13 at 12 32 01 PM

Noon Diff Fix : -

Screenshot 2025-11-13 at 4 11 25 PM

### DIFF
--- a/backend/connector-integration/src/connectors/fiuu/transformers.rs
+++ b/backend/connector-integration/src/connectors/fiuu/transformers.rs
@@ -551,7 +551,7 @@ impl<
             false => TxnType::Auts,
         };
         let return_url = item.router_data.request.router_return_url.clone();
-        let non_3ds = match item.router_data.request.enrolled_for_3ds {
+        let non_3ds = match item.router_data.resource_common_data.is_three_ds() {
             false => 1,
             true => 0,
         };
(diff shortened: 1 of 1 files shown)


backend/domain_types/src/types.rs
```diff
@@ -1167,10 +1167,9 @@ impl<
             .transpose()?;
 
         let access_token = value
-            .state
+            .access_token
             .as_ref()
-            .and_then(|state| state.access_token.as_ref())
-            .map(|token| crate::connector_types::AccessTokenResponseData {
+            .map(|token| AccessTokenResponseData {
                 access_token: token.token.clone(),
```
Saptak88: Why access token field is added in PaymentsAuthorizeData, isnt it already present in PaymentFlow data. can you remove it if it is redundant

Saptak88 (approved):

PR #287, opened 2025-11-16
fix: Diff fixes for Novalnet Authorize flow

Saptak88:
## Description

Fixed diffs for the following fields
- birth_data
- schema_tid
- create_token

### Additional Changes

## How did you test it?
Tested via HS through cypress test
[00006-NoThreeDSManualCapture]
Screenshot 2025-11-17 at 3 06 46 PM
Screenshot 2025-11-17 at 3 07 08 PM

[00011-CreateSingleuseMandate.cy.js]
Screenshot 2025-11-17 at 3 10 45 PM
Screenshot 2025-11-17 at 3 10 57 PM

### DIFF
--- a/backend/connector-integration/src/connectors/novalnet/transformers.rs
+++ b/backend/connector-integration/src/connectors/novalnet/transformers.rs
@@ -235,7 +235,6 @@ impl<
             T,
         >,
     ) -> Result<Self, Self::Error> {
-        let _ = Err(ConnectorError::RequestEncodingFailed)?;
         let auth = NovalnetAuthType::try_from(&item.router_data.connector_auth_type)?;
 
         let merchant = NovalnetPaymentsRequestMerchant {


AmitsinghTanwar007 (approved):

backend/connector-integration/src/connectors/novalnet/transformers.rs
```diff
@@ -2193,6 +2191,7 @@ impl<
             billing: Some(billing),
             // no_nc is used to indicate if minimal customer data is passed or not
             no_nc: MINIMAL_CUSTOMER_DATA_PASSED,
+            birth_date: Some(String::from("1992-06-10")),
```
bsayak03, line 2194: for now, hardcoding this works but better if we can add this in PMT, since the customer will be the one supplying this. The same change can be implemented in HS SDK and HS BE as well. This was hardcoded because this particular PMT was not a merchant requirement as of that moment

bsayak03: If you're not picking this in this PR, kindly raise an Issue for this to be picked up later

Saptak88: done

PR #315, opened 2025-11-24
fix: diff checker fixes for globalpay

shuklatushar226:
## Description
diff checker fixes for globalpay

### Additional Changes

### DIFF
--- a/backend/connector-integration/src/connectors/globalpay/transformers.rs
+++ b/backend/connector-integration/src/connectors/globalpay/transformers.rs
@@ -262,7 +262,7 @@ pub struct GlobalpayNotifications {
     pub status_url: String,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum InitiatorType {
     Merchant,
@@ -277,7 +277,7 @@ pub struct Initiator {
     pub stored_credential: Option<String>,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum StoredCredentialType {
     Installment,
@@ -286,7 +286,7 @@ pub enum StoredCredentialType {
     Subscription,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, Clone, Copy, Serialize)]
 #[serde(rename_all = "SCREAMING_SNAKE_CASE")]
 pub enum StoredCredentialSequence {
     First,
@@ -424,8 +424,12 @@ impl<
 
         // Build initiator and stored_credential for mandate payments
         let (initiator, stored_credential) = if item.request.is_mandate_payment() {
-            // Determine if this is a merchant or payer initiated transaction
-            let is_off_session = matches!(item.request.off_session, Some(true));
+            // Determine initiator type based on off_session flag
+            let initiator_type = if matches!(item.request.off_session, Some(true)) {
+                InitiatorType::Merchant
+            } else {
+                InitiatorType::Payer
+            };
 
             // Check if we have an existing connector mandate ID (indicates subsequent payment)
             let has_connector_mandate = item
@@ -441,11 +445,7 @@ impl<
                 .is_some();
 
             let initiator = Some(Initiator {
-                initiator_type: Some(if is_off_session {
-                    InitiatorType::Merchant
-                } else {
-                    InitiatorType::Payer
-                }),
+                initiator_type: Some(initiator_type),
                 id: None,
                 stored_credential: None,
             });
@@ -461,11 +461,7 @@ impl<
                 } else {
                     StoredCredentialSequence::First
                 }),
-                initiator: Some(if is_off_session {
-                    InitiatorType::Merchant
-                } else {
-                    InitiatorType::Payer
-                }),
+                initiator: Some(initiator_type),
             });
 
             (initiator, stored_credential)


backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -261,7 +262,7 @@ pub struct GlobalpayNotifications {
     pub status_url: String,
 }
```
maverox: The `Clone` and `Copy` traits are added but never used. Consider if these traits are actually necessary for these enums.

If they're not required by the serialization or usage elsewhere, the simpler approach would be to keep just `Debug, Serialize`:

```suggestion
#[derive(Debug, Serialize)]
```

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -276,7 +277,7 @@ pub struct Initiator {
     pub stored_credential: Option<String>,
 }
```
maverox: Similarly, `Clone` and `Copy` traits are added but their necessity is unclear.

```suggestion
#[derive(Debug, Serialize)]
```

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -285,7 +286,7 @@ pub enum StoredCredentialType {
     Subscription,
 }
```
maverox: Same as above - `Clone` and `Copy` traits added without clear usage.

```suggestion
#[derive(Debug, Serialize)]
```

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -310,11 +311,9 @@ pub struct GlobalpayPaymentsRequest<T: PaymentMethodDataTypes> {
     pub country: common_enums::CountryAlpha2,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub capture_mode: Option<GlobalpayCaptureMode>,
```
maverox: `skip_serializing_if` was removed for `initiator` field. Since this field is now always serialized (even when `None`), verify that the GlobalPay API accepts `null` values for optional fields.

If the API requires omitting the field when not present, keep the original attribute:

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub initiator: Option<Initiator>,
```

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -310,11 +311,9 @@ pub struct GlobalpayPaymentsRequest<T: PaymentMethodDataTypes> {
...
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub initiator: Option<Initiator>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub notifications: Option<GlobalpayNotifications>,
-    #[serde(skip_serializing_if = "Option::is_none")]
     pub stored_credential: Option<StoredCredential>,
```
maverox: Same concern as `initiator` - removing `skip_serializing_if` means `stored_credential: null` will be sent even for non-mandate payments. Verify API compatibility.

If the API expects the field to be omitted when not present:

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub stored_credential: Option<StoredCredential>,
```

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -334,8 +333,6 @@ pub struct GlobalpayCard<T: PaymentMethodDataTypes> {
     pub expiry_month: Secret<String>,
     pub expiry_year: Secret<String>,
     pub cvv: Secret<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub cvv_indicator: Option<String>,
 }
 
 impl<
```
maverox: The `cvv_indicator` field was removed from the struct but the code that populates it (lines 383-387 in the old version) was also removed. This is correct, but ensure that:

1. GlobalPay doesn't require CVV indicator for compliance/validation
2. The diff checker was failing because this field was incorrectly set

If GlobalPay actually needs this field, consider keeping it but fixing the logic instead of removing it entirely.

backend/connector-integration/src/connectors/globalpay/transformers.rs
```diff
@@ -433,6 +422,53 @@ impl<
...
+            // Determine initiator type based on off_session flag
+            let initiator_type = if matches!(item.request.off_session, Some(true)) {
+                InitiatorType::Merchant
+            } else {
+                InitiatorType::Payer
+            };
+
+            // Check if we have an existing connector mandate ID (indicates subsequent payment)
+            let has_connector_mandate = item
+                .request
```
maverox: The mandate payment logic looks good overall, but there's potential for improvement:

1. The nested `if-else` could be flattened for readability
2. Consider extracting the mandate detection logic into a helper method

```suggestion
        // Build initiator and stored_credential for mandate payments
        let (initiator, stored_credential) = match item.request.is_mandate_payment() {
            false => (None, None),
            true => {
                let initiator_type = if matches!(item.request.off_session, Some(true)) {
                    InitiatorType::Merchant
                } else {
                    InitiatorType::Payer
                };

                let has_connector_mandate = item
                    .request
                    .mandate_id
                    .as_ref()
                    .and_then(|mandate_ids| match &mandate_ids.mandate_reference_id {
                        Some(MandateReferenceId::ConnectorMandateId(connector_mandate_ids)) => {
                            connector_mandate_ids.get_connector_mandate_id()
                        }
                        _ => None,
                    })
                    .is_some();

                let initiator = Some(Initiator {
                    initiator_type: Some(initiator_type),
                    id: None,
                    stored_credential: None,
                });

                let stored_credential = Some(StoredCredential {
                    credential_type: Some(if has_connector_mandate {
                        StoredCredentialType::Recurring
                    } else {
                        StoredCredentialType::Unscheduled
                    }),
                    sequence: Some(if has_connector_mandate {
                        StoredCredentialSequence::Subsequent
                    } else {
                        StoredCredentialSequence::First
                    }),
                    initiator: Some(initiator_type),
                });

                (initiator, stored_credential)
            }
        };
```

PR #372, opened 2025-12-15
fix: Removed the authorization_indicator_type field from Authdotnet Req

AmitsinghTanwar007:
## Description

AuthorizeIndicatorType Field was removed to have parity with HS as this was not required

### Additional Changes

### DIFF
--- a/backend/connector-integration/src/connectors/authorizedotnet/transformers.rs
+++ b/backend/connector-integration/src/connectors/authorizedotnet/transformers.rs
@@ -416,12 +416,6 @@ pub enum Reason {
 }
 
 #[derive(Debug, Serialize)]
-#[serde(rename_all = "camelCase")]
-struct AuthorizationIndicator {
-    authorization_indicator: AuthorizationType,
-}
-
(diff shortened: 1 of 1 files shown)


backend/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -464,6 +464,7 @@ pub struct AuthorizedotnetTransactionRequest<T: PaymentMethodDataTypes> {
     user_fields: Option<UserFields>,
     processing_options: Option<ProcessingOptions>,
     subsequent_auth_information: Option<SubsequentAuthInformation>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
Aishwariyaa-Anand: If this field is not being used, we can deprecate this

AmitsinghTanwar007: it is not deprecated from connector side, just hs is not using it so now we are not populating it.

Aishwariyaa-Anand (approved):

PR #382, opened 2025-12-16
feat(core): Changed Metadata Type to SecretString

Shubhodip900:
## Description

 Changed Metadata Type to SecretString from HashMap<string, string>
### Additional Changes

## How did you test it?

Tested Using Diff checker through cypress test: https://github.com/juspay/hyperswitch-cloud/issues/13297

### DIFF
--- a/backend/domain_types/src/types.rs
+++ b/backend/domain_types/src/types.rs
@@ -4577,8 +4577,8 @@ impl
             .map(AccessTokenResponseData::from);
 
         let connector_meta_data = value
-            .refund_metadata
-            .map(|m| ForeignTryFrom::foreign_try_from((m, "refund metadata")))
+            .merchant_account_metadata
+            .map(|m| ForeignTryFrom::foreign_try_from((m, "merchant account metadata")))
             .transpose()?;
 
         let payment_method = value


backend/domain_types/src/types.rs
```diff
@@ -8030,7 +7899,15 @@ impl<
...
     ) -> Result<Self, error_stack::Report<Self::Error>> {
         // Extract values first to avoid partial move
-        let merchant_configered_currency = value.clone().merchant_configered_currency();
+        let merchant_configured_currency = if value.clone().merchant_configured_currency()
+            == grpc_api_types::payments::Currency::Unspecified
+        {
+            None
+        } else {
+            Some(common_enums::Currency::foreign_try_from(
+                value.clone().merchant_configured_currency(),
```
gopikrishna000: why should we clone here?
instead of using` merchant_configured_currency()`, which will result in default value when merchant_configured_currency is None, we can use field access and do a None check?

As, currently we cant distinguise between if caller sent with Currency::Unspecified or None if we use
`if value.clone().merchant_configured_currency()`

backend/domain_types/src/types.rs
```diff
@@ -46,35 +46,18 @@ fn extract_headers_from_metadata(
...
-        });
-    serde_json::Value::Object(metadata_map)
+/// Helper function to parse Option<Secret<String>> metadata to serde_json::Value
+/// Properly handles parsing JSON or returning Null if None
+fn parse_secretstring_metadata(
+    secret: Secret<String>,
+) -> Result<serde_json::Value, error_stack::Report<ApplicationErrorResponse>> {
+    let raw = secret.expose();
+    serde_json::from_str(&raw).change_context(ApplicationErrorResponse::BadRequest(ApiError {
+        sub_code: "INVALID CONNECTOR METADATA".to_string(),
```
gopikrishna000: hi it isnt always connector metadata?
this function is generically called by multiple fields right

so we may have pure helper and dont assume the field name smtg like
```
fn parse_json_from_secret_string(
    secret: &Secret<String>,
) -> Result<serde_json::Value, serde_json::Error> {
    serde_json::from_str(secret.expose_secret())
}
```

and handle the error with field name at caller?
or may add field name

backend/domain_types/src/types.rs
```diff
@@ -46,35 +46,19 @@ fn extract_headers_from_metadata(
...
+    secret: Secret<String>,
+    field_name: &'static str,
+) -> Result<serde_json::Value, error_stack::Report<ApplicationErrorResponse>> {
+    let raw = secret.expose();
+    serde_json::from_str(&raw).change_context(ApplicationErrorResponse::BadRequest(ApiError {
+        sub_code: format!("INVALID {}", field_name.to_uppercase()),
+        error_identifier: 400,
+        error_message: format!("Failed to parse {}", field_name.to_lowercase()),
+        error_object: None,
+    }))
```
maverox: I think we can have TryFrom or ForeignTryFrom here.

backend/domain_types/src/types.rs
```diff
@@ -46,35 +46,20 @@ fn extract_headers_from_metadata(
...
+    ) -> Result<Self, error_stack::Report<Self::Error>> {
+        let raw = secret.expose();
+        serde_json::from_str(&raw).change_context(ApplicationErrorResponse::BadRequest(ApiError {
+            sub_code: format!("INVALID {}", field_name.to_uppercase()),
+            error_identifier: 400,
+            error_message: format!("Failed to parse {}", field_name.to_lowercase()),
+            error_object: None,
+        }))
+    }
 }
```
AmitsinghTanwar007: this should return a secret::serde ?

backend/grpc-server/src/server/payments.rs
AmitsinghTanwar007: instead use .map here

backend/grpc-server/src/server/payments.rs
```diff
@@ -1018,7 +1018,7 @@ impl Payments {
             amount: common_utils::types::MinorUnit::new(0),
             currency,
             integrity_object: None,
-            metadata: if payload.metadata.is_empty() {
+            metadata: if payload.metadata.is_none() {
```
AmitsinghTanwar007: same here

backend/domain_types/src/types.rs
Shubhodip900: some metadata field in L2 (`connector_types`) have metadata type as `serde_json::Value` so I am returning this type and if the field need secret then making it `secret` inside `ForeignTryFrom`

AmitsinghTanwar007: then change them to secret as metadata should be secret so this should be a consistent behaviour

backend/domain_types/src/types.rs
```diff
@@ -4643,14 +4576,10 @@ impl
...
-        let connector_meta_data = (!value.merchant_account_metadata.is_empty())
-            .then(|| {
-                serde_json::to_value(&value.merchant_account_metadata)
-                    .map(common_utils::pii::SecretSerdeValue::new)
-            })
-            .transpose()
-            .ok()
-            .flatten();
+        let connector_meta_data = value
+            .refund_metadata
```
AmitsinghTanwar007: why refund meta data is mapped here instead of merchant acc metadata

PR #386, opened 2025-12-22
refactor: made mandatory fields in authorize flow optional

kanikac199:
## Description

made `enrolled_for_3ds` and  `request_incremental_authorization` optional for Euler-UCS Integration.
### Additional Changes


backend/domain_types/src/connector_types.rs
maverox, line 1043: why not make this also as optional?
or not make enrolled_for_3ds and just do 

.unwrap_or_default() or .unwrap_or(false) ?? 

it's not consistent

PR #458, opened 2026-01-29
feat: service_call and origin field for connectorevent

gopikrishna000:
## Description
service_call and origin field for connectorevent

### Additional Changes

## How did you test it?
https://github.com/juspay/hyperswitch/pull/11065

### DIFF
--- a/backend/common_utils/src/consts.rs
+++ b/backend/common_utils/src/consts.rs
@@ -48,9 +48,6 @@ pub const X_CONNECTOR_SERVICE: &str = "connector-service";
 
 pub const X_FLOW_NAME: &str = "x-flow";
 
-/// Event origin: Router (Hyperswitch)
-pub const EVENT_ORIGIN_ROUTER: &str = "router";
-
 /// Event origin: Unified Connector Service
 pub const EVENT_ORIGIN_UCS: &str = "unified_connector_service";
 


backend/common_utils/src/consts.rs
```diff
@@ -47,6 +47,13 @@ pub const X_SOURCE_NAME: &str = "x-source";
 pub const X_CONNECTOR_SERVICE: &str = "connector-service";
 
 pub const X_FLOW_NAME: &str = "x-flow";
+
+/// Event origin: Router (Hyperswitch)
+pub const EVENT_ORIGIN_ROUTER: &str = "router";
```
hrithikesh026: this is not being used anywhere

gopikrishna000: hi removed

backend/common_utils/src/events.rs
```diff
@@ -223,6 +223,8 @@ pub struct Event {
     pub request_data: Option<MaskedSerdeValue>,
     pub response_data: Option<MaskedSerdeValue>,
     pub headers: HashMap<String, String>,
+    pub service_call: ServiceCall,
```
hrithikesh026, line 226: UCS call will always be external right. What internal service call is UCS making?

gopikrishna000: There are two events emitted from UCS
One during ucs's own request and response to its caller. (Which is considered as internal)
And another during outgoing connector call (external)

PR #463, opened 2026-01-30
fix(connector): Request diff fix for Stripe & Cybersource

Saptak88:
## Description

Diff fix for Stripe & Cybersource

Proto change:
- Added merchant_order_reference_id in Void and Capture request. (Backward Compatible)

### Additional Changes

## How did you test it?
Cybersource Network Token (Only timestamp and signature diff which is expected)
Screenshot 2026-02-03 at 8 21 40 PM
Screenshot 2026-02-03 at 8 21 58 PM


backend/connector-integration/src/connectors/stripe/transformers.rs
```diff
@@ -1779,6 +1779,17 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
     }
 }
 
+fn validate_and_get_setup_future_usage(
```
AmitsinghTanwar007: this can be just a boolean

Saptak88: Changed

PR #476, opened 2026-02-04
fix(connector): [NEXIXPAY] DIFF FIX

iemyashasvi:
## Description
This PR fixes the diffs for NEXIXPAY Connector 

Related to https://github.com/juspay/hyperswitch-cloud/issues/12635

### Additional Changes

## How did you test it?
Tested using diff checker


backend/connector-integration/src/connectors/nexixpay/transformers.rs
```diff
@@ -26,6 +30,30 @@ use strum::Display;
...
+                    connector: "Nexixpay".to_string(),
+                    max_length: MAX_ORDER_ID_LENGTH,
+                    received_length: payment_id.len(),
+                },
+            ))
+        }
+    } else {
+        Ok(payment_id.to_string())
+    }
+}
```
maverox, line 55: should we truncate identifiers??

backend/connector-integration/src/connectors/nexixpay/transformers.rs
```diff
@@ -26,6 +30,30 @@ use strum::Display;
 // Import the wrapper type created by macros
 use super::NexixpayRouterData;
 
+const MAX_ORDER_ID_LENGTH: usize = 18;
+
+fn get_nexi_order_id(payment_id: &str) -> CustomResult<String, errors::ConnectorError> {
+    if payment_id.len() > MAX_ORDER_ID_LENGTH {
+        if payment_id.starts_with("pay_") {
```
AmitsinghTanwar007, line 37: so this part of code is HS specific as HS handles the pay_id length by truncating it and this pattern is replicated in many connectors like auth.net etc.
should we have a framework for this like as we are planning for FFI for connectorSpecification then this trait can also include payment_id length so that merchant know what they are sending.
cc @knutties @jarnura @JeevaRamu0104

JeevaRamu0104: yes we should have ID framework

backend/domain_types/src/connector_types.rs
```diff
@@ -1452,6 +1452,7 @@ pub struct PaymentsPreAuthenticateData<T: PaymentMethodDataTypes> {
     pub enrolled_for_3ds: bool,
     pub redirect_response: Option<ContinueRedirectionResponse>,
     pub capture_method: Option<common_enums::CaptureMethod>,
+    pub mandate_reference: Option<MandateReferenceId>,
```
JeevaRamu0104, line 1455: Can this be changed to `mandate_reference_id`

PR #482, opened 2026-02-05
fix(connector): [Fiuu] Fixed payment status being sent as Pending for Fiuu when the connector response is FiuuPaymentsResponse::Error

Saptak88:
Fixed payment status being sent as Pending for Fiuu when the connector response is FiuuPaymentsResponse::Error

### Additional Changes


Anurag-05-prog (approved):

backend/connector-integration/src/connectors/fiuu/transformers.rs
```diff
@@ -1069,6 +1069,10 @@ where
                     network_decline_code: None,
                     network_error_message: None,
                 }),
+                resource_common_data: PaymentFlowData {
```
AmitsinghTanwar007, line 1072: hey @saptak bro what if there is a case where you should return pending for 2XX status code but then also you will be giving failure

Saptak88: For this response variant from connector its failure only

PR #490, opened 2026-02-06
feat(framework): Changed access_token type from String to SecretString in proto and connector_types

Saptak88:
## Description

Changed access_token type from String to SecretString in proto and connector_types (Backward incompatible)

### Additional Changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,25 @@ All notable changes to Connector Service will be documented here.
 
 - - -
 
+## 2026.02.11.0
+
+### Features
+
+- **connector:** Gigadat Macro Implementation ([#501](https://github.com/juspay/connector-service/pull/501)) ([`1ba0591`](https://github.com/juspay/connector-service/commit/1ba05915c289966a6ba9cdc15b2a83b9a3363afb))
+- **payment_method_data:** [adyen] Auth code in payment response ([#498](https://github.com/juspay/connector-service/pull/498)) ([`d4b923c`](https://github.com/juspay/connector-service/commit/d4b923ce29c4c48cd9eb87682010ca7f5dee0ddc))
+
+### Bug Fixes
+
+- **connector:** Paypal Router Data Fix in Authorize and RepeatPayment Flow ([#499](https://github.com/juspay/connector-service/pull/499)) ([`cf8b523`](https://github.com/juspay/connector-service/commit/cf8b523d9019c98b70d3210fa96a5c4fcc6a7492))
+
+### Miscellaneous Tasks
+
+- Adding failure status to customer create response ([#497](https://github.com/juspay/connector-service/pull/497)) ([`558bac9`](https://github.com/juspay/connector-service/commit/558bac914e6e1453831460f8c6566cdec01aa40d))
+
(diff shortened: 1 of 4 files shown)


backend/grpc-api-types/proto/payment.proto
```diff
@@ -1510,7 +1510,7 @@ message PaymentServiceCreateAccessTokenRequest {
 // Response message for creating an access token.
 message PaymentServiceCreateAccessTokenResponse {
   // Core token information
-  string access_token = 1; // The access token string
+  SecretString access_token = 1; // The access token string
```
gopikrishna000, line 1530: note it is backward incompatible change, we need to coordinate hyperswitch deployment when going this change

Saptak88: Yes

Shubhodip900 (approved):

backend/grpc-server/src/server/payments.rs
```diff
@@ -345,52 +341,41 @@ impl Payments {
...
+                    &metadata_payload.connector_auth_type,
+                    &connector.to_string(),
+                    service_name,
+                    event_params,
+                )
+                .await
+                .map_err(|err| {
+                    tracing::error!("Failed to process payment access token data: {:?}", err);
+                    PaymentAuthorizationError::new(
+                        grpc_api_types::payments::PaymentStatus::Pending,
```
AmitsinghTanwar007: why the payment status is pending

backend/grpc-server/src/server/payments.rs
```diff
@@ -224,22 +224,19 @@ impl Payments {
...
                 // If provided cached token - use it, don't generate new one
-                tracing::info!("Using cached access token from Hyperswitch");
-                AccessTokenResponseData {
-                    access_token: token,
-                    token_type: None,
-                    expires_in,
-                }
+                tracing::info!("Using cached access from request");
+                AccessTokenResponseData::foreign_try_from(cached_access_token)
+                    .map_err(|e| e.into_grpc_status())?
```
AmitsinghTanwar007: instead apply match over this

AmitsinghTanwar007 (changes requested):

backend/grpc-server/src/server/payments.rs
Saptak88: Changed to Failure as accesstoken fail means payment failed

backend/grpc-server/src/server/payments.rs
```diff
@@ -224,22 +224,21 @@ impl Payments {
...
         service_name: &str,
         event_params: EventParams<'_>,
     ) -> Result<AccessTokenResponseData, tonic::Status> {
-        let access_token_data = match cached_access_token {
-            Some((token, expires_in)) => {
+        let access_token_result =
+            access_token.and_then(|token| AccessTokenResponseData::foreign_try_from(token).ok());
+
+        let access_token_data = match access_token_result {
+            Some(cached_access_token) => {
```
AmitsinghTanwar007, line 238: Can't we move this outside this function like if data is not there then only call this function?

AmitsinghTanwar007: this would be continued later as we are only checking access token for validation
but expiry time is ignored.

PR #499, opened 2026-02-09
fix(connector): Paypal Router Data Fix in Authorize and RepeatPayment Flow

Shubhodip900:
## Description

Added error Handeling in Authorize flow and `request_incremental_authorization` in RepeatPayment Flow.
### Additional Changes

## How did you test it?

Raw Connector response for 4xx
Screenshot 2026-02-10 at 1 41 06 PM

No diff found:-
Screenshot 2026-02-10 at 1 45 20 PM

### DIFF
--- a/backend/connector-integration/src/connectors/paypal/transformers.rs
+++ b/backend/connector-integration/src/connectors/paypal/transformers.rs
@@ -12,7 +12,7 @@ use common_utils::{
     CustomResult, Method,
 };
 use domain_types::{
-    connector_flow::{Authorize, Capture, PostAuthenticate, RepeatPayment},
+    connector_flow::{Authorize, Capture, PSync, PostAuthenticate, RepeatPayment},
     connector_types::{
         AccessTokenResponseData, MandateReference, PaymentFlowData, PaymentsAuthorizeData,
         PaymentsCaptureData, PaymentsPostAuthenticateData, PaymentsResponseData, PaymentsSyncData,
@@ -1913,12 +1913,7 @@ fn get_redirect_url(link_vec: Vec<PaypalLinks>) -> CustomResult<Option<Url>, Con
 }
 
 impl TryFrom<ResponseRouterData<PaypalSyncResponse, Self>>
-    for RouterDataV2<
-        domain_types::connector_flow::PSync,
-        PaymentFlowData,
-        PaymentsSyncData,
-        PaymentsResponseData,
-    >
+    for RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>
 {
     type Error = error_stack::Report<ConnectorError>;
     fn try_from(item: ResponseRouterData<PaypalSyncResponse, Self>) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


backend/connector-integration/src/connectors/paypal.rs
```diff
@@ -329,6 +329,67 @@ macros::create_all_prerequisites!(
...
+            status_code: res.status_code,
+            code: option_error_code_message
+                .clone()
+                .map(|error_code_message| error_code_message.error_code)
+                .unwrap_or(NO_ERROR_CODE.to_string()),
+            message: option_error_code_message
+                .map(|error_code_message| error_code_message.error_message)
+                .unwrap_or(NO_ERROR_MESSAGE.to_string()),
+            reason: error_reason.or(Some(response.message)),
+            attempt_status: None,
```
AmitsinghTanwar007: if attemptstatus is none then it will send a status like unspecified
is it ok

Shubhodip900: It won't cause any issue

Saptak88 (dismissed):

backend/connector-integration/src/connectors/paypal.rs
AmitsinghTanwar007: how it will won't cause any issue? can you pls explain

backend/connector-integration/src/connectors/paypal.rs
```diff
@@ -1111,13 +1195,112 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
     ConnectorErrorTypeMapping for Paypal<T>
 {
-}
-
-impl From<paypal::ErrorDetails> for ErrorCodeAndMessage {
-    fn from(error_details: paypal::ErrorDetails) -> Self {
-        Self {
-            error_code: error_details.issue,
-            error_message: error_details.description.unwrap_or_default(),
+    fn get_connector_error_type(
```
AmitsinghTanwar007, line 1198: can we use impl FromStr for ConnectorErrorType {
here instead of this function

backend/connector-integration/src/connectors/paypal.rs
Shubhodip900: During 4xx case both status and attempt_status are None, so in HS during status mapping the default status will be Failure which is expected in HS as well

```
let status = match err.attempt_status {
                            // Use the status sent by connector in error_response if it's present
                            Some(status) => status,
                            None =>
                            // mark previous attempt status for technical failures in PSync and ExtendAuthorization flow
                            {
                                if sub_flow == "PSync" || sub_flow == "ExtendAuthorization" {
                                    match err.status_code {
                                        // marking failure for 2xx because this is genuine payment failure
                                        200..=299 => enums::AttemptStatus::Failure,
                                        _ => router_data.status,
                                    }
                                } else if sub_flow == "Capture" {
                                    match err.status_code {
                                        500..=511 => enums::AttemptStatus::Pending,
                                        // don't update the status for 429 error status
                                        429 => router_data.status,
                                        _ => enums::AttemptStatus::Failure,
                                    }
                                } else {
                                    match err.status_code {
                                        500..=511 => enums::AttemptStatus::Pending,
                                        _ => enums::AttemptStatus::Failure,
                                    }
                                }
                            }
                        };
```
We can check the above HS code, here in first match it will enter in the None case and then in last `else` the 4xx will be mapped as  `Failure`.

backend/connector-integration/src/connectors/paypal.rs
Shubhodip900: The `ConnectorErrorTypeMapping` impl tells to implement `get_connector_error_type`  and internally `get_error_code_error_message_based_on_priority` function calls `get_connector_error_type` to decide the error message and error code.

AmitsinghTanwar007: ok
i missed the above code

AmitsinghTanwar007 (approved):

backend/connector-integration/src/connectors/paypal/transformers.rs
```diff
@@ -1927,44 +1930,38 @@ impl TryFrom<ResponseRouterData<PaypalSyncResponse, Self>>
...
                 })
             }
-            PaypalSyncResponse::PaypalPaymentsSyncResponse(response) => {
-                Self::try_from(ResponseRouterData {
-                    response,
-                    router_data: item.router_data,
-                    http_code: item.http_code,
-                })
-            }
             PaypalSyncResponse::PaypalRedirectSyncResponse(response) => {
```
gopikrishna000, line 1928: can we keep this logic inside tryfrom, similar to done in hs

PR #510, opened 2026-02-13
fix: incremental_authorization_allowed and cybersource repeatpayment diff fix

gopikrishna000:
## Description

1. Some saved card payment sends both paymentmethoddata as card and connector_mandateid
In that case, hyperswitch takes mandateid try_from but ucs gets unimplemented error through repeatpayment.
https://github.com/juspay/hyperswitch-cloud/issues/14555
fixing the logic to first check if connector_mandate_id is present, if so use mandate try_from, same as hyperswitch.

2. UCS PaymentServiceGetResponse does not have incremental_authorization_allowed, and generate_payment_sync_response explicitly ignores it when building the gRPC response, so UCS return null for that field in PSync comparisons.
HS direct path sets incremental_authorization_allowed during Cybersource PSync based on status, so HS routerdata has a boolean.
https://github.com/juspay/hyperswitch-cloud/issues/14556

### Additional Changes

## How did you test it?

https://github.com/juspay/hyperswitch-cloud/issues/14556

### DIFF
--- a/backend/domain_types/src/types.rs
+++ b/backend/domain_types/src/types.rs
@@ -4600,7 +4600,7 @@ pub fn generate_payment_sync_response(
                 connector_metadata: _,
                 network_txn_id,
                 connector_response_reference_id,
-                incremental_authorization_allowed: _,
+                incremental_authorization_allowed,
                 mandate_reference,
                 status_code,
             } => {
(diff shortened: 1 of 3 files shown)


backend/domain_types/src/types.rs
```diff
@@ -4730,6 +4731,7 @@ pub fn generate_payment_sync_response(
                 raw_connector_request,
                 connector_response,
                 redirection_data: None,
+                incremental_authorization_allowed: None,
```
AmitsinghTanwar007, line 4734: if the value is available then send it in error response also

gopikrishna000: it is not available in ErrorResponse

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #530, opened 2026-02-20
ci: setup GitHub release workflow

knutties:
## Description
Github release workflow for connector-service as a library.  We will add connector-service grpc stuff also as part of this in a subsequent change.

## Motivation and Context
To be able to version and release connector-service as a library


knutties:
closing as #531 is the one with the current target.

PR #590, opened 2026-03-05
refactor: simplify header handling by making them optional and inferring connector from `x-connector-auth` header

maverox:
## Summary

- **Unified connector + auth resolution**: The new `resolve_connector_and_auth()` function resolves both the connector enum and auth credentials in a single call from the typed `X-Connector-Auth` JSON header, falling back to legacy `x-connector` / `x-auth` headers when absent.
- **Removed required header validation from the transport layer**: Previously, `headers_to_metadata` rejected requests missing headers like `x-connector`, `x-merchant-id`, `x-request-id`, etc. This was blocking REST SDK callers who only send `X-Connector-Auth`. Headers are now all optional at the transport layer, with context-aware defaults applied downstream in metadata extraction.
- **Auto-generated defaults for missing metadata**: `x-merchant-id` generates a prefixed ID, `x-request-id` generates a UUID v7, and `x-tenant-id` defaults to `"DefaultTenantId"` — making the REST SDK experience seamless without requiring callers to pass boilerplate headers.
- **Derived `ConnectorEnum` from auth type**: Added `ForeignTryFrom<AuthType>` for `ConnectorEnum` so the connector can be inferred directly from the typed auth payload, removing the need for a separate `x-connector` header when using `X-Connector-Auth`.

## Motivation

The REST SDK (`sdk/rust`) sends a single typed `X-Connector-Auth` header containing both the connector name and credentials. The previous implementation required callers to also send `x-connector`, `x-merchant-id`, `x-request-id`, and `x-tenant-id` as separate headers — causing unnecessary friction and failures at the transport layer. This refactor moves validation downstream and makes the header contract flexible for both gRPC and REST SDK consumers.

## Test plan

- [x] Unit tests for `resolve_connector_and_auth` (typed header, legacy fallback, precedence, missing auth)
- [x] Unit tests for metadata defaults (`merchant_id_defaults_when_missing`, `request_id_generates_uuid_v7_when_missing`, `tenant_id_defaults_when_missing`)
- [x] Manual: verify REST SDK example (`sdk/rust/examples/basic.rs`) works without extra headers
- [ ] Manual: verify existing gRPC callers still work with legacy headers

### DIFF
--- a/backend/common_utils/src/metadata.rs
+++ b/backend/common_utils/src/metadata.rs
@@ -184,18 +184,6 @@ impl MaskedMetadata {
         })
     }
 
-    /// Extract merchant ID from metadata, generating a default if missing.
-    ///
-    /// Tries to get `X_MERCHANT_ID` from the metadata, falling back to
-    /// a generated ID with `"default_merchant"` prefix if the header is absent.
-    pub fn merchant_id(&self) -> String {
-        self.get_raw(crate::consts::X_MERCHANT_ID)
-            .unwrap_or_else(|| {
-                tracing::debug!("x-merchant-id header missing, generating with default prefix");
-                crate::fp_utils::generate_id_with_default_len("default_merchant")
-            })
-    }
-
     /// Get all metadata as HashMap with masking for logging
     pub fn get_all_masked(&self) -> std::collections::HashMap<String, String> {
         self.raw_metadata
@@ -219,4 +207,15 @@ impl MaskedMetadata {
             })
             .collect()
     }
+}
+
+/// Return the merchant ID if present, or generate a default.
+///
+/// Shared fallback logic used by both the gRPC path (raw `MetadataMap`)
+/// and the FFI path (`MaskedMetadata`).
+pub fn merchant_id_or_default(value: Option<&str>) -> String {
+    value.map(|s| s.to_string()).unwrap_or_else(|| {
+        tracing::warn!("x-merchant-id header missing, using default merchant ID");
+        "DefaultMerchantId".to_string()
+    })
 }
--- a/backend/ucs_interface_common/src/metadata.rs
+++ b/backend/ucs_interface_common/src/metadata.rs
@@ -113,11 +113,11 @@ pub fn connector_from_metadata(
 pub fn merchant_id_from_metadata(
     metadata: &metadata::MetadataMap,
 ) -> CustomResult<String, ApplicationErrorResponse> {
-    let masked = common_utils::metadata::MaskedMetadata::new(
-        metadata.clone(),
-        common_utils::metadata::HeaderMaskingConfig::default(),
-    );
-    Ok(masked.merchant_id())
+    Ok(common_utils::metadata::merchant_id_or_default(
+        metadata
+            .get(consts::X_MERCHANT_ID)
+            .and_then(|value| value.to_str().ok()),
+    ))
 }
 
 pub fn request_id_from_metadata(
@@ -216,8 +216,7 @@ mod tests {
     fn merchant_id_defaults_when_missing() {
         let metadata = MetadataMap::new();
         let merchant_id = merchant_id_from_metadata(&metadata).expect("should not fail");
-        assert!(merchant_id.starts_with("default_merchant_"));
-        assert_eq!(merchant_id.len(), "default_merchant_".len() + 20);
+        assert_eq!(merchant_id, "DefaultMerchantId");
     }
 
     #[test]
--- a/backend/domain_types/src/utils.rs
+++ b/backend/domain_types/src/utils.rs
@@ -394,12 +394,14 @@ static CARD_REGEX: LazyLock<HashMap<CardIssuer, core::result::Result<Regex, rege
 
 /// Helper function for extracting merchant ID from metadata.
 ///
-/// Delegates to `MaskedMetadata::merchant_id()` which auto-generates a default
-/// when the `x-merchant-id` header is missing, then parses it into a `MerchantId`.
+/// Uses the shared `merchant_id_or_default` fallback: if the `x-merchant-id`
+/// header is missing, a default ID is auto-generated.
 pub fn extract_merchant_id_from_metadata(
     metadata: &MaskedMetadata,
 ) -> Result<common_utils::id_type::MerchantId, ApplicationErrorResponse> {
-    let merchant_id_str = metadata.merchant_id();
+    let merchant_id_str = common_utils::metadata::merchant_id_or_default(
+        metadata.get_raw(consts::X_MERCHANT_ID).as_deref(),
+    );
     Ok(merchant_id_str
         .parse::<common_utils::id_type::MerchantId>()
         .map_err(|e| {


backend/common_utils/src/metadata.rs
```diff
@@ -184,6 +184,18 @@ impl MaskedMetadata {
...
 
+    /// Extract merchant ID from metadata, generating a default if missing.
+    ///
+    /// Tries to get `X_MERCHANT_ID` from the metadata, falling back to
+    /// a generated ID with `"default_merchant"` prefix if the header is absent.
+    pub fn merchant_id(&self) -> String {
+        self.get_raw(crate::consts::X_MERCHANT_ID)
+            .unwrap_or_else(|| {
+                tracing::debug!("x-merchant-id header missing, generating with default prefix");
+                crate::fp_utils::generate_id_with_default_len("default_merchant")
```
AmitsinghTanwar007: generating merchant id is it a correct way?

backend/domain_types/src/connector_types.rs
```diff
@@ -3487,3 +3487,129 @@ pub struct BillingDescriptor {
     /// A reference to be shown on billing description
     pub reference: Option<String>,
 }
+impl ForeignTryFrom<grpc_api_types::payments::connector_auth::AuthType> for ConnectorEnum {
+    type Error = ApplicationErrorResponse;
+    fn foreign_try_from(
+        auth_type: grpc_api_types::payments::connector_auth::AuthType,
+    ) -> Result<Self, error_stack::Report<Self::Error>> {
+        use grpc_api_types::payments::connector_auth::AuthType;
+        #[allow(unreachable_patterns)]
```
AmitsinghTanwar007: do we need this as we are handling each case

backend/ucs_interface_common/src/auth.rs
```diff
@@ -15,70 +15,68 @@ use ucs_env::logger;
...
+        let auth_type = typed_auth.auth_type.as_ref().ok_or_else(|| {
+            Report::new(ApplicationErrorResponse::BadRequest(ApiError {
+                sub_code: "INVALID_CONNECTOR_AUTH_FORMAT".to_string(),
+                error_identifier: 400,
+                error_message: "X-Connector-Auth header missing auth_type".to_string(),
+                error_object: None,
+            }))
+        })?;
+
+        use domain_types::utils::ForeignTryFrom;
```
AmitsinghTanwar007: can't we have this import at top

backend/ucs_interface_common/src/auth.rs
```diff
@@ -15,70 +15,68 @@ use ucs_env::logger;
...
+
+        logger::debug!(
+            "Connector and auth successfully resolved from X-Connector-Auth header for connector: {}",
+            connector
+        );
+        Ok((connector, auth))
+    } else {
+        logger::debug!("X-Connector-Auth header not found, falling back to legacy headers");
+
+        let connector = crate::metadata::connector_from_metadata(metadata)?;
```
AmitsinghTanwar007: can't we populate this function at top and then use it

backend/ucs_interface_common/src/auth.rs
maverox: I don't think any other function uses that trait that is why i scoped it to only that function. 

I will have a look

backend/common_utils/src/metadata.rs
maverox: see the way i see it is `merchant_id` for ucs standalone only serves observability purpose, doesn't directly affect or is required for ucs' functional requirements (`auth`, `connector_processing` etc.).

So having it as mandatory adds friction for the primary use case, now the reason we made it mandatory earlier was for integrations that we have in euler hyperswitch they are supposed to use hyperswitch always. because for these services the merchant_id is critical, hence that for that enforcement we made it mandatory.
Is the best explaination I can think of.

**what I can do for now is make the logs warn! instead of debug!??**

what do you think @knutties ?

backend/domain_types/src/connector_types.rs
maverox: you are right removing this catch all will help us in finding  at compile time whether there was addition of new variants but wasn't added to `ConnectorEnum`

backend/ucs_interface_common/src/auth.rs
maverox: this is part of the else scenario so it has to be here

backend/common_utils/src/metadata.rs
```diff
@@ -184,6 +184,18 @@ impl MaskedMetadata {
         })
     }
 
+    /// Extract merchant ID from metadata, generating a default if missing.
+    ///
+    /// Tries to get `X_MERCHANT_ID` from the metadata, falling back to
+    /// a generated ID with `"default_merchant"` prefix if the header is absent.
+    pub fn merchant_id(&self) -> String {
```
Jeeva0104: **Default merchant_id generation - needs documentation**

Auto-generating `merchant_id` with prefix `"default_merchant"` when the header is missing.

**Questions:**
- Is merchant ID used for authorization/routing decisions downstream?
- Could requests be processed under an unintended tenant context?

**Recommendation:** Add a comment explaining when this default is appropriate, or consider warning-level logging.

backend/domain_types/src/connector_types.rs
```diff
@@ -3487,3 +3488,119 @@ pub struct BillingDescriptor {
...
+            AuthType::Phonepe(_) => Ok(Self::Phonepe),
+            AuthType::Cashfree(_) => Ok(Self::Cashfree),
+            AuthType::Paytm(_) => Ok(Self::Paytm),
+            AuthType::Calida(_) => Ok(Self::Calida),
+            AuthType::Payload(_) => Ok(Self::Payload),
+            AuthType::Authipay(_) => Ok(Self::Authipay),
+            AuthType::Silverflow(_) => Ok(Self::Silverflow),
+            AuthType::Celero(_) => Ok(Self::Celero),
+            AuthType::Trustpayments(_) => Ok(Self::Trustpayments),
+            AuthType::Paysafe(_) => Ok(Self::Paysafe),
```
Jeeva0104, line 3553: **Duplicated UNSUPPORTED_CONNECTOR error construction**

The same error construction is repeated 4 times for Screenstream, Ebanx, Globepay, Coinbase, Coingate.

**Suggestion:** Extract to a helper:
```rust
fn unsupported_connector_error() -> error_stack::Report<ApplicationErrorResponse> {
    error_stack::Report::new(ApplicationErrorResponse::BadRequest(ApiError {
        sub_code: "UNSUPPORTED_CONNECTOR".to_string(),
        error_identifier: 400,
        error_message: "Connector is not supported".to_string(),
        error_object: None,
    }))
}
```

AmitsinghTanwar007 (approved):

backend/common_utils/src/metadata.rs
maverox: >Is merchant ID used for authorization/routing decisions downstream?

No whatever merchant context we need we pass in payload fields, or in some cases this is part of connector creds, rzp or paytm i guess had this.

> Could requests be processed under an unintended tenant context?

tenant always had the fallback (public), it's not something I have added here.

backend/ucs_interface_common/src/metadata.rs
```diff
@@ -109,30 +113,25 @@ pub fn connector_from_metadata(
...
-        .map(|inner| inner.to_string())
-        .map_err(|e| {
-            Report::new(ApplicationErrorResponse::BadRequest(ApiError {
-                sub_code: "MISSING_MERCHANT_ID".to_string(),
-                error_identifier: 400,
-                error_message: format!("Missing merchant ID in request metadata: {e}"),
-                error_object: None,
-            }))
-        })
+    let masked = common_utils::metadata::MaskedMetadata::new(
```
gopikrishna000: why  need to create MaskedMetadata for extracting merchant_id?
is it due to fallback impl is done for MaskedMetadata, but not MetadataMap?
instead of having fallback pattern in UCS, which may affect existing audit trail and logging Correlation, should we have autogenerate logic at middleware of sdk?

gopikrishna000: But yeah, ideally ucs itself dont need merchant_id to work. Or can we make it optional in MetadataPayload?

backend/common_utils/src/metadata.rs
knutties: @maverox - this is `default_merchant_{nanoid}` everytime ?  Wondering if just `default_merchant` suffices if it is not given - dynamic ids are not great from an observability point of view.

gopikrishna000 (approved):

PR #610, opened 2026-03-10
chore(error): add request and response error proto for ffi implementation

JeevaRamu0104:
## Description

Transform **ApplicationError** and **ConnectorError** to **RequestError** and **ResponseError**

## Motivation and Context

This pull request refactors error handling in the FFI bindings and related modules to standardize error propagation and improve interoperability with gRPC FFI error types. The changes replace a custom `UniffiError` type with the existing `ConnectorError` type, introduce detailed error conversion logic for gRPC, and ensure that all error cases are encoded and returned in a consistent protobuf format. This results in more robust and maintainable error handling across FFI boundaries.

**Error handling and conversion improvements:**

* Replaced `UniffiError` with `ConnectorError` throughout the FFI bindings, updating all function signatures and error cases to use `ConnectorError` for consistency and richer error information. [[1]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL28-R48) [[2]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL57-R68) [[3]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL85-R96) [[4]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL106-R169)
* Added detailed conversions from `ConnectorError` and `ApplicationErrorResponse` to gRPC FFI error types (`RequestError` and `ResponseError`), ensuring all errors can be propagated as protobuf-encoded messages over FFI.
* Updated error variant `DecodingFailed` in `ConnectorError` to carry an optional message, enabling more descriptive error reporting. [[1]](diffhunk://#diff-d795cdcaf5e208715a3bc6b41ada853e1acb4b038e368ccbd6fd0d979ddeffdeL877-R877) [[2]](diffhunk://#diff-e8acb3aa04b02808c35ef9786ea3e99d4d2f6398a2c63f482d0da0d80899b62aL317-R320)

**FFI transformer and macro changes:**

* Refactored FFI transformer functions (`run_req_transformer`, `run_res_transformer`, and `handle_event_transformer`) to return encoded protobuf error messages directly instead of Rust `Result` types, ensuring consistent error handling at FFI boundaries. [[1]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL155-R199) [[2]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL172-R237) [[3]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL194-L199) [[4]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL258-R332) [[5]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL278-R345)
* Updated the `define_ffi_flow!` macro and related function signatures to match the new error handling approach, removing `Result` wrappers and returning byte vectors for both success and error cases. [[1]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL210-R270) [[2]](diffhunk://#diff-6997f8c5d32c7e12ff9758316e33c5283ed1bb97bb1a855bac87ed08dc269f4bL232-R283)

These changes collectively make error handling more robust, informative, and compatible with gRPC and FFI requirements.

### Additional Changes

## How did you test it?

image

image


backend/ffi/src/utils.rs
```diff
@@ -8,27 +8,33 @@ use ucs_env::configs::Config;
...
+) -> Result<MaskedMetadata, ConnectorError> {
     ucs_interface_common::headers::headers_to_masked_metadata(
         headers,
         HeaderMaskingConfig::default(),
     )
     .map_err(|e| match e {
         ucs_interface_common::error::InterfaceError::MissingRequiredHeader { key } => {
-            FfiError::MissingRequiredHeader { key }
+            ConnectorError::MissingRequiredField {
+                field_name: key.leak(),
```
maverox: `String::leak()` in Rust: This method exists for cases where you genuinely need `'static `lifetime (e.g., one-time config parsing).
   Using it in error paths that can be triggered repeatedly is a memory leak. The idiomatic fix is either changing the field type
  to String or using Cow<'static, str>.

sdk/python/src/payments/connector_client.py
```diff
@@ -80,47 +101,67 @@ async def _execute_flow(self, flow: str, request: Any, response_cls, options: Op
...
-            res_bytes,
-            request_bytes,
-            options_bytes
-        )
+        result_bytes_res = res_transformer(res_bytes, request_bytes, options_bytes)
+        # Inline FFI error check
+        try:
+            res_err = ResponseError()
+            res_err.ParseFromString(result_bytes_res)
+            if res_err.is_error:
```
maverox: 🐛 Same issue here — `is_error` and `message` don't exist on `ResponseError`. Should be `res_err.error_message`.
```suggestion
            if res_err.error_message:
```

sdk/python/src/payments/connector_client.py
```diff
@@ -80,47 +101,67 @@ async def _execute_flow(self, flow: str, request: Any, response_cls, options: Op
...
+                e = RuntimeError(res_err.message)
```
maverox:
```suggestion
                e = RuntimeError(res_err.error_message)
```

sdk/javascript/src/payments/uniffi_client.ts
```diff
@@ -223,7 +199,19 @@ export class UniffiClient {
 
     try {
       checkCallStatus(this._ffi, status);
-      return liftBytes(result);
+      const bytes = liftBytes(result);
+      try {
+        const reqErr = types.RequestError.decode(bytes);
```
maverox: ⚠️ **Ambiguous error detection**: Protobuf decoding is very permissive — arbitrary byte sequences can often decode without throwing. A valid `FfiConnectorHttpRequest` response could have bytes that happen to decode into a `RequestError` with a non-empty `errorMessage` field, causing a false-positive error.

Consider adding a discriminator to the proto (e.g., `bool is_error = 5;`) or using a wrapper `oneof` message to disambiguate success vs error at the wire level.

backend/ffi/src/services/payments.rs
```diff
@@ -393,63 +393,52 @@ res_transformer!(
...
 
-/// handle_event — synchronous webhook processing (single-step, no outgoing HTTP).
-///
-/// The caller supplies the raw webhook body + headers received from the connector
-/// and gets back a fully-structured `EventServiceHandleResponse`.
-///
-/// External source verification (async HTTP used by PayPal / Stripe) is **not**
-/// performed here; only local synchronous signature verification is done.
-/// The gRPC server performs external verification before calling its equivalent path.
+// / handle_event — synchronous webhook processing (single-step, no outgoing HTTP).
```
maverox: ⚠️ **Broken doc comments**: `/// ` was changed to `// / ` (space inserted after `//`), converting doc comments to regular comments. These will no longer appear in `cargo doc` output.

```suggestion
/// handle_event — synchronous webhook processing (single-step, no outgoing HTTP).
```

backend/domain_types/src/errors.rs
```diff
@@ -1066,3 +1065,103 @@ impl ErrorSwitch<ApiClientError> for HttpClientError {
...
+// =============================================================================
+
+impl From<ConnectorError> for grpc_api_types::payments::RequestError {
+    fn from(e: ConnectorError) -> Self {
+        grpc_api_types::payments::RequestError {
+            status: grpc_api_types::payments::PaymentStatus::Pending.into(),
+            error_message: Some(e.to_string()),
+            error_code: None,
+            status_code: Some(500),
+        }
```
maverox: 💡 **All `ConnectorError` variants hardcoded to status 500 + Pending**: This loses important error context. For example, `MissingRequiredField` is semantically a 400, not 500. `ResponseDeserializationFailed` could be a 502.

Consider matching on specific `ConnectorError` variants to return appropriate status codes, or at minimum add a `TODO` to track this.

backend/domain_types/src/errors.rs
```diff
@@ -1066,3 +1065,103 @@ impl ErrorSwitch<ApiClientError> for HttpClientError {
         }
     }
 }
+
+// =============================================================================
+// Conversions: ConnectorError → gRPC FFI error types
+// =============================================================================
+
+impl From<ConnectorError> for grpc_api_types::payments::RequestError {
```
maverox: 💡 **Duplication**: There are 8 nearly identical `From` impls here (owned + ref × 2 error types × 2 source types). Consider a macro to reduce boilerplate:

```rust
macro_rules! impl_error_conversion {
    ($src:ty => $dst:ty, $body:expr) => {
        impl From<$src> for $dst { fn from(e: $src) -> Self { ($body)(&e) } }
        impl From<&$src> for $dst { fn from(e: &$src) -> Self { ($body)(e) } }
    };
}
```

backend/ffi/src/macros.rs
```diff
@@ -171,22 +130,50 @@ macro_rules! res_transformer {
...
+            resource_common_data: flow_data,
+            connector_auth_type: connector_auth_details,
+            request: payment_request_data,
+            response: Err(domain_types::router_data::ErrorResponse::default()),
+        };
 
             // transform connector response type to common response type
+            // Classify response based on status code: 2xx/3xx = success, 4xx/5xx = error
+            let classified_response = match response.status_code {
+                200..=202 | 204 | 302 => Ok(response),
```
maverox: ⚠️ **Incomplete HTTP status classification**: This misses `203 Non-Authoritative Information`, `301 Moved Permanently`, and other valid success/redirect codes.

A simpler and more robust pattern:
```suggestion
                200..=299 | 300..=399 => Ok(response),
```

sdk/python/smoke-test/test_smoke.py
```diff
@@ -113,7 +114,20 @@ async def run_test():
             print(f"  Response type:   {type(response).__name__}")
             print("  PASSED")
         except Exception as e:
-            print(f"  Response/error received: {e}")
+            if hasattr(e, "ffi_error"):
+                ffi_err = e.ffi_error
+                if isinstance(ffi_err, RequestError):
+                    print(
+                        f"  Request build error [{ffi_err.error_type}]: {ffi_err.message}"
```
maverox: 🐛 **Wrong field names**: `RequestError` proto has `error_message` and `error_code`, not `error_type` and `message`. These will silently return empty strings in protobuf-python.
```suggestion
                        f"  Request build error [{ffi_err.error_code}]: {ffi_err.error_message}"
```

sdk/python/smoke-test/test_smoke.py
```diff
@@ -113,7 +114,20 @@ async def run_test():
...
+                    )
+                elif isinstance(ffi_err, ResponseError):
+                    print(
+                        f"  Response parse error [{ffi_err.error_type}]: {ffi_err.message}"
```
maverox: 🐛 Same issue — should use `error_code` and `error_message`.
```suggestion
                        f"  Response parse error [{ffi_err.error_code}]: {ffi_err.error_message}"
```

sdk/javascript/src/payments/uniffi_client.ts
JeevaRamu0104: Its already under try catch block is is_error is not required

backend/ffi/src/services/payments.rs
```diff
@@ -394,62 +394,51 @@ res_transformer!(
...
-/// External source verification (async HTTP used by PayPal / Stripe) is **not**
-/// performed here; only local synchronous signature verification is done.
-/// The gRPC server performs external verification before calling its equivalent path.
+// /
+// / The caller supplies the raw webhook body + headers received from the connector
+// / and gets back a fully-structured `EventServiceHandleResponse`.
+// /
+// / External source verification (async HTTP used by PayPal / Stripe) is **not**
+// / performed here; only local synchronous signature verification is done.
+// / The gRPC server performs external verification before calling its equivalent path.
```
maverox:
```suggestion
///
/// The caller supplies the raw webhook body + headers received from the connector
/// and gets back a fully-structured `EventServiceHandleResponse`.
///
/// External source verification (async HTTP used by PayPal / Stripe) is **not**
/// performed here; only local synchronous signature verification is done.
/// The gRPC server performs external verification before calling its equivalent path.
```

backend/ffi/src/bindings/uniffi.rs
```diff
@@ -25,26 +26,26 @@ mod uniffi_bindings_inner {
...
     fn parse_metadata(
         options: &FfiOptions,
-    ) -> Result<crate::types::FfiMetadataPayload, UniffiError> {
+    ) -> Result<crate::types::FfiMetadataPayload, ConnectorError> {
         // 1. Resolve Connector (Taken from FfiOptions)
         let proto_connector = options.connector(); // Direct enum access via generated method
         let connector = ConnectorEnum::foreign_try_from(proto_connector).map_err(|e| {
-            UniffiError::MetadataParseError {
-                msg: format!("Connector mapping failed: {e}"),
+            ConnectorError::GenericError {
```
gopikrishna000: ConnectorError::InvalidConnectorName variant would be more appropriate.

backend/ffi/src/bindings/uniffi.rs
```diff
@@ -25,26 +26,26 @@ mod uniffi_bindings_inner {
...
             .as_ref()
-            .ok_or_else(|| UniffiError::MissingMetadata {
-                key: "auth".to_string(),
-            })?;
+            .ok_or_else(|| ConnectorError::MissingRequiredField { field_name: "auth" })?;
 
         let connector_auth_type = ConnectorSpecificAuth::foreign_try_from(proto_auth.clone())
-            .map_err(|e| UniffiError::MetadataParseError {
-                msg: format!("Typed auth mapping failed: {e}"),
+            .map_err(|e| ConnectorError::GenericError {
```
gopikrishna000: ConnectorError::InvalidConnectorConfig is better instead of GenericError

sdk/python/src/payments/connector_client.py
```diff
@@ -25,14 +39,88 @@
...
+
+    Returns:
+        Decoded success proto on success.
+
+    Raises:
+        RuntimeError: With `.ffi_error` set to a RequestError if the bytes
+            represent a transformer error.
+    """
+    try:
+        return success_cls.FromString(result_bytes)
```
gopikrishna000: Protobuf is a binary format that relies on field tags, not type names.
   * FfiConnectorHttpRequest tags: url (1), method (2), headers (3), body (4).
   * RequestError tags: status (1), error_message (2), error_code (3), status_code (4).

  Because the tags overlap, if the FFI returns a RequestError, the SDK's Success.parseFrom(bytes) will not throw an exception. Instead, it will
  successfully decode the error as a "success" message with corrupted data (e.g., the status enum will be interpreted as the url string). This
  will lead to silent failures and hard-to-debug "gibberish" values in the SDK.

  **Recommendation**: You must use a oneof wrapper or a reserved tag range/prefix to distinguish between success and error at the protobuf level, rather than relying on parsing failures.

```
  message FfiRequestTransformerResult {
      oneof result {
        FfiConnectorHttpRequest success = 1;
        FfiError error = 2;
         }
    }
```

backend/grpc-api-types/proto/sdk_config.proto
```diff
@@ -155,3 +155,19 @@ message FfiConnectorHttpResponse {
...
+  optional string error_code = 3;
+  optional uint32 status_code = 4;
+}
+
+// Error returned by res_transformer FFI calls
+message ResponseError {
+  PaymentStatus status = 1;
+  optional string error_message = 2;
+  optional string error_code = 3;
+  optional uint32 status_code = 4;
```
gopikrishna000, line 172: optional: ResponseError should always have the connector's status code. 
So, we may have naming like `connector_status_code` to avoid hardcoding it.

For request, we might need to hard code only

backend/grpc-api-types/proto/sdk_config.proto
```diff
@@ -155,3 +155,19 @@ message FfiConnectorHttpResponse {
   map<string, string> headers = 2;
   bytes body = 3;
 }
+
+// Error returned by req_transformer FFI calls
+message RequestError {
```
gopikrishna000, line 160: ~should we have like RequestTransformError?~

backend/grpc-api-types/proto/sdk_config.proto
```diff
@@ -155,3 +155,19 @@ message FfiConnectorHttpResponse {
...
+  PaymentStatus status = 1;
+  optional string error_message = 2;
+  optional string error_code = 3;
+  optional uint32 status_code = 4;
+}
+
+// Error returned by res_transformer FFI calls
+message ResponseError {
```
gopikrishna000, line 168: ~should we have like ResponseTransformError?~

backend/ffi/src/macros.rs
```diff
@@ -171,22 +130,50 @@ macro_rules! res_transformer {
...
+            connector_auth_type: connector_auth_details,
+            request: payment_request_data,
+            response: Err(domain_types::router_data::ErrorResponse::default()),
+        };
 
             // transform connector response type to common response type
+            // Classify response based on status code: 2xx/3xx = success, 4xx/5xx = error
+            let classified_response = match response.status_code {
+                200..=399 => Ok(response),
+                _ => Err(response),
```
gopikrishna000, line 163: the `handle_connector_response`  at the next step has the logic to handle the 4xx and 5xx response as well from connector., we should not throw error here.
we should call the handle_response function in connector-service/backend/external-services/src/service.rs to get the correct wrapper based on connector http status_code.

backend/grpc-api-types/proto/sdk_config.proto
```diff
@@ -155,3 +155,19 @@ message FfiConnectorHttpResponse {
   map<string, string> headers = 2;
   bytes body = 3;
 }
+
+// Error returned by req_transformer FFI calls
+message RequestError {
+  PaymentStatus status = 1;
```
gopikrishna000, line 161: request error always wont have paymentstatus?

backend/ffi/src/macros.rs
JeevaRamu0104: we are not throwing error over here we just wrapping the response as Err rather than Ok so that handle_connector_response function will handle the 4xx and 5xx

backend/grpc-api-types/proto/sdk_config.proto
JeevaRamu0104: PaymentStatus for request error will always be pending this is to handle the runtime errors

backend/grpc-api-types/proto/sdk_config.proto
JeevaRamu0104: connector_status_code will come in part of ErrorResponse this is structure is for exception

PR #771, opened 2026-03-24
feat(tests): comprehensive connector test harness with 57 connectors, 22 suites, and credential masking

AmitsinghTanwar007:
## Summary

End-to-end test harness for UCS connectors. Runs real payment flows (authorize, capture, void, refund, etc.) against a live gRPC server, with support for 57 connectors across 22 test suites.

### What's included

- **Test harness** (`crates/internal/integration-tests/`) — scenario-driven framework that loads JSON test specs, sends gRPC requests, and asserts responses
- **57 connector specs** — per-connector config (credentials, overrides, enabled suites) for: aci, adyen, airwallex, authorizedotnet, bambora, bamboraapac, bankofamerica, barclaycard, billwerk, bluecode, bluesnap, braintree, cashtocode, checkout, cryptopay, cybersource, datatrans, dlocal, elavon, fiserv, fiservemea, fiuu, forte, getnet, globalpay, helcim, hipay, iatapay, jpmorgan, mifinity, mollie, multisafepay, nexinets, nexixpay, nmi, noon, novalnet, nuvei, paybox, payload, paypal, paysafe, payu, placetopay, powertranz, rapyd, redsys, shift4, stax, stripe, trustpay, tsys, wellsfargo, worldpay, worldpayvantiv, worldpayxml, xendit
- **22 global test suites** — authorize, capture, void, refund, refund_sync, get, setup_recurring, recurring_charge, create_customer, create_access_token, create_order, create_session_token, create_sdk_session_token, complete_authorize, incremental_authorization, authenticate, pre_authenticate, post_authenticate, tokenize_payment_method, verify_redirect_response, reverse, revoke_mandate
- **3 execution backends** — Grpcurl (default for `--interface grpc`), SDK/FFI (`--interface sdk`), Tonic (via `suite_run_test`/`run_test` binaries)
- **Credential masking** — `cred_masking.rs` masks `rawConnectorRequest`/`rawConnectorResponse` and Bearer tokens before writing reports, with `check_report_creds` and `mask_report_creds` binaries for verification/retroactive cleanup
- **Scenario generator** — `generator_core` module + `generate_scenarios` binary for generating `scenarios.json` from `generator_specs.json` templates
- **Browser automation engine** — Playwright-based Google Pay token generation and 3DS flow automation (`browser-automation-engine/`)
- **Test reports** — per-connector Markdown reports in `test_report/` with full pass/fail/skip matrix
- **Makefile targets** — `test-prism`, `test-connector`, `test-scenario`, `setup-connector-tests`, `start-grpc`, `stop-grpc`, `cargo` (auto-loads `.env.connector-tests`)
- **Pre-push hook** — `scripts/pre-push-cred-check` scans staged files for leaked credentials
- **Proto additions** — `AciConfig`, `CheckoutConfig` messages and `ConnectorSpecificConfig` oneof fields in `payment.proto`
- **Domain type additions** — `ForeignTryFrom` impls for Aci/Checkout auth types in `router_data.rs`

### Additional Changes
- [x] This PR modifies the API contract (proto additions: `AciConfig`, `CheckoutConfig`)

## Reviewing Guide

**TL;DR:** Out of ~2,587 changed files, only ~270 are handwritten. Focus on these key areas:

### Files to Review (Handwritten) - ~270 files

**HIGH PRIORITY - Core Implementation:**
```
crates/internal/integration-tests/src/
├── bin/
│   ├── test_ucs.rs              # Main test runner
│   ├── check_report_creds.rs    # Credential masking verification
│   ├── mask_report_creds.rs     # Report credential masking
│   ├── render_report.rs         # Report generator
│   ├── run_test.rs              # Single test runner
│   ├── suite_run_test.rs        # Suite test runner
│   └── sdk_run_test.rs          # SDK test runner
├── harness/
│   ├── mod.rs                   # Test harness orchestration
│   ├── executor.rs              # Test execution engine
│   ├── scenario_api.rs          # Scenario API definitions
│   ├── scenario_loader.rs       # JSON scenario loading
│   ├── scenario_assert.rs       # Assertion framework
│   ├── scenario_types.rs        # Type definitions
│   ├── connector_override.rs    # Connector-specific overrides
│   ├── cred_masking.rs          # Credential masking logic
│   ├── credentials.rs           # Credential management
│   ├── metadata.rs              # Test metadata handling
│   ├── report.rs                # Report generation
│   ├── sdk_executor.rs          # SDK test execution
│   └── server.rs                # gRPC server management
├── lib.rs
└── connector_specs/             # Connector configurations (~50 specs)
    └── [connector_name]/
        ├── specs.json
        └── override.json

browser-automation-engine/src/
├── api/routes.ts
├── server.ts
├── cli.ts
├── engine/
│   ├── automationEngine.ts
│   └── interpreter.ts
├── drivers/
│   ├── browserDriver.ts
│   └── playwrightDriver.ts
├── gpay-token-gen.ts
├── gpay-login.ts
├── applepay-token-gen.ts
└── types/
    ├── api.ts
    └── dsl.ts

```

**MEDIUM PRIORITY - Configuration & Docs:**
```
Makefile                          # New test targets
Cargo.toml                        # Workspace updates
.github/workflows/ci.yml          # CI updates
browser-automation-engine/
├── README.md
├── package.json
├── tsconfig.json
├── netlify.toml
└── gpay/applepay HTML templates

crates/internal/integration-tests/
├── README.md
├── Cargo.toml
└── docs/                         # Architecture documentation
    ├── code-walkthrough.md
    ├── connector-overrides.md
    ├── context-mapping.md
    └── scenario-json-core-readme.md
```

### Files to SKIP (Generated) - ~2,208 files

```
❌ Cargo.lock                      # Auto-generated dependency lockfile
❌ report.json                     # Auto-generated test report
❌ test_report/                    # ~2,205 auto-generated markdown reports
    └── [connector_name]_report.md
❌ browser-automation-engine/package-lock.json
```

### Review Strategy
1. **Start here:** `crates/internal/integration-tests/src/harness/mod.rs` - orchestration logic
2. **Then review:** `crates/internal/integration-tests/src/bin/test_ucs.rs` - main entry point
3. **Key logic:** `cred_masking.rs`, `executor.rs`, `scenario_loader.rs`
4. **Browser automation:** `browser-automation-engine/src/engine/`
5. **Skip:** All files in `test_report/` directory

### Prerequisites

1. **Credentials file** — place connector credentials at `.github/test/creds.json` (see `.github/test/template_creds.json` for the schema)
2. **One-time setup** — install dependencies and deploy Google Pay token page:
   ```bash
   make setup-connector-tests
   ```

### Run all connectors

```bash
make test-prism
```

This starts the gRPC server, runs all 57 connectors across their enabled suites, generates Markdown reports in `test_report/`, and stops the server.

### Run a single connector

```bash
make test-connector connector=stripe
make test-connector connector=cybersource interface=sdk
```

### Run a single scenario

```bash
make test-scenario connector=stripe suite=authorize scenario=no3ds_auto_capture_credit_card
```

### Run via cargo directly

```bash
make cargo ARGS="run -p integration-tests --bin test_ucs -- --connector stripe --interface grpc --endpoint localhost:50051 --report"
```

### Verify no credential leaks in reports

```bash
cargo run -p integration-tests --bin check_report_creds
```

### Generate scenarios from templates

```bash
cargo run -p integration-tests --bin generate_scenarios -- --all
```

### Build check (no credentials needed)

```bash
cargo check -p integration-tests
```

### DIFF
--- a/crates/internal/integration-tests/src/connector_specs/trustly/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/trustly/specs.json
@@ -1,0 +1,7 @@
+{
+  "connector": "trustly",
+  "supported_suites": [
+    "authorize",
+    "refund"
+  ]
+}


.github/test/template_creds.json
JeevaRamu0104, line 1: using the existing creds pushed by @knutties which is currently used in the sdk smoke test

Makefile
```diff
@@ -233,16 +431,20 @@ help:
...
+	@echo "  docs               Regenerate all connector docs from source"
+	@echo "  all-connectors-doc Generate the all-connectors coverage document"
+	@echo "  docs-check         Report which connectors are missing annotation files"
+	@echo ""
 	@echo "Certification Targets:"
 	@echo "  certify-client-sanity  Run cross-language transport parity certification"
-	@echo
+	@echo ""
+	@echo "Other:"
+  @echo
```
AmitsinghTanwar007: This line is indented with spaces instead of a tab, so `make` will error with “missing separator” when parsing the `help` recipe. Use a tab (like the surrounding @echo lines) or remove the stray line.

Makefile
```diff
@@ -46,31 +74,134 @@ ci:
...
+## test-scenario all manage the server lifecycle automatically.
+start-grpc:
+	@echo "▶ Building grpc-server…"
+	@cargo build -p grpc-server --release 2>&1
+	@echo "▶ Starting grpc-server on $(GRPC_HOST):$(GRPC_PORT)…"
+	@CS__SERVER__HOST=$(GRPC_HOST) \
+	 CS__SERVER__PORT=$(GRPC_PORT) \
+	 CS__COMMON__ENVIRONMENT=development \
+	 ./target/release/grpc-server &
+	@echo $$! > $(GRPC_PID_FILE)
```
maverox: **[Bug]** PID capture is broken. Each `@`-prefixed Makefile recipe line runs in a separate shell invocation, so `$$!` on this line does NOT refer to the backgrounded `grpc-server` process from the previous line.

**Fix:** Combine into a single shell command:
```makefile
	@CS__SERVER__HOST=$(GRPC_HOST) CS__SERVER__PORT=$(GRPC_PORT) CS__COMMON__ENVIRONMENT=development ./target/release/grpc-server & echo $$! > $(GRPC_PID_FILE)
```

browser-automation-engine/src/utils/validation.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  const action = requiredString(rawRule, "action", path);
+
+  switch (action) {
+    case "goto": {
+      const rule: GotoRule = {
+        action: "goto",
+        url: optionalString(rawRule, "url", path),
+        waitUntil: rawRule.waitUntil as GotoRule["waitUntil"],
```
maverox: **[Security]** Unsafe `as` cast bypasses validation. A client can pass `"waitUntil": "garbage"` and it will be accepted here, then Playwright throws an unstructured error.

Same issue at line 152 (`state` cast) and line 163 (`match` cast).

**Fix:** Validate against allowed values before casting:
```ts
const VALID_WAIT_UNTIL = ["load", "domcontentloaded", "networkidle", "commit"];
if (rawRule.waitUntil && !VALID_WAIT_UNTIL.includes(rawRule.waitUntil)) {
  throw new Error(`Invalid waitUntil: ${rawRule.waitUntil}`);
}
```

browser-automation-engine/src/engine/interpreter.ts
```diff
@@ -0,0 +1,150 @@
...
+        fullPage: rule.fullPage
+      });
+
+      if (rule.as) {
+        ctx.data[rule.as] = targetPath;
+      }
+      return;
+    }
+
+    case "evaluate": {
```
maverox, line 146: **[Security]** The `evaluate` action executes arbitrary JavaScript in the browser page context. If the API server listens on `0.0.0.0` (which `server.ts` does), any network client can execute arbitrary JS in spawned browser sessions.

Consider adding an auth gate, an expression allowlist, or at minimum documenting the trust boundary.

browser-automation-engine/src/engine/interpreter.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    case "fill":
+      await session.fill(rule.selector, rule.value, { timeoutMs: rule.timeoutMs });
+      return;
+
+    case "press":
+      await session.press(rule.selector, rule.key, { timeoutMs: rule.timeoutMs });
+      return;
+
+    case "waitFor": {
```
maverox, line 39: **[Bug]** `waitFor` with both `selector` AND `urlContains` runs them sequentially (AND-sequential), but the validation error on line 40 says "at least one of" implying OR semantics. This means the user pays double the timeout.

If AND is intended, consider `Promise.all` for parallel waiting. If OR is intended, use `Promise.race`.

browser-automation-engine/src/server.ts
```diff
@@ -0,0 +1,23 @@
...
+async function start(): Promise<void> {
+  const app = Fastify({ logger: true });
+  const engine = new AutomationEngine(new PlaywrightDriverFactory());
+
+  await registerRoutes(app, engine);
+
+  const port = Number(process.env.PORT ?? 3000);
+  const host = process.env.HOST ?? "0.0.0.0";
+
+  await app.listen({ port, host });
```
maverox, line 21: **[Issue]** No graceful shutdown handler. When the server is killed (SIGTERM/SIGINT), in-flight Playwright browser sessions won't be cleaned up, leaving zombie browser processes.

Also: `Number(process.env.PORT ?? 3000)` will produce `NaN` if `PORT` is set to a non-numeric string.

browser-automation-engine/src/api/routes.ts
```diff
@@ -0,0 +1,27 @@
...
+        error instanceof RequestValidationError ? error.message : "Invalid request payload";
+      return reply.code(400).send({
+        success: false,
+        error: message,
+        data: {},
+        steps: []
+      });
+    }
+
+    const result = await engine.run(runRequest);
```
maverox: **[Issue]** Always returns HTTP 200 even when `result.success === false`. This makes HTTP-level monitoring/load balancers unable to detect failures.

Also missing error handling: if `engine.run()` throws (Playwright crash, OOM), it propagates to Fastify's default handler which returns a 500 with stack trace.

crates/internal/integration-tests/src/harness/cred_masking.rs
```diff
@@ -0,0 +1,691 @@
...
+use std::sync::LazyLock;
+
+pub const MASKED_VALUE: &str = "***MASKED***";
+
+// ---------------------------------------------------------------------------
+// JSON-level masking (for structured req_body / res_body)
+// ---------------------------------------------------------------------------
+
+/// Returns `true` if `key` (after normalisation) looks like a sensitive field.
+pub fn is_sensitive_key(key: &str) -> bool {
```
maverox, line 19: **[Security]** `is_sensitive_key` may miss connector-specific auth field names:
- `access_code` (used by ACI) — not matched
- `entity_id` (ACI auth field) — not matched
- `pin` — not matched
- `key3`, `key0` — not matched (only `key1` and `key2` are)

Consider switching to a safe-key allowlist (keys known to be non-sensitive) rather than a sensitive-key blocklist, which is inherently fragile.

crates/internal/integration-tests/src/harness/cred_masking.rs
```diff
@@ -0,0 +1,691 @@
...
+    masked
+}
+
+// ---------------------------------------------------------------------------
+// Bearer / JWT token masking
+// ---------------------------------------------------------------------------
+
+/// Masks `Bearer <token>` patterns and standalone JWT tokens (three base64
+/// segments separated by dots).
+fn mask_bearer_and_jwt_tokens(line: &str) -> String {
```
maverox, line 322: **[Security]** Despite the function name `mask_bearer_and_jwt_tokens`, only `Bearer <token>` patterns are masked. Standalone JWTs (`eyJ...` — three base64 segments separated by dots) appearing in response bodies without a `Bearer` prefix will not be detected or masked.

If standalone JWT masking is not needed, rename the function to `mask_bearer_tokens` to avoid a false sense of security.

crates/internal/integration-tests/src/bin/check_report_creds.rs
```diff
@@ -0,0 +1,161 @@
...
+        if let Some(reason) = detect_unmasked_cred(line) {
+            violations.push((path.to_path_buf(), line_no + 1, reason));
+        }
+    }
+}
+
+fn walk_md_files(dir: &Path, visitor: &mut dyn FnMut(&Path)) {
+    let entries = match fs::read_dir(dir) {
+        Ok(e) => e,
+        Err(_) => return,
```
maverox: **[Silent Failure]** `Err(_) => return` silently ignores `read_dir` errors. For a **security-checking** tool, this means unmasked credentials in unreadable directories would pass the check — the binary reports "OK" when it didn't actually scan everything.

Also on line 103: `entries.flatten()` silently drops individual `DirEntry` errors.

Same issue exists in `mask_report_creds.rs` at ~line 206.

crates/internal/integration-tests/src/harness/sdk_executor.rs
```diff
@@ -405,7 +406,7 @@ fn ffi_environment() -> Environment {
     }
 }
 
-fn map_request_error(
+fn _map_request_error(
```
maverox: **[Code Quality]** `_map_request_error` is dead code after the `FfiResult` refactor. Prefixing with `_` works for variables but is non-idiomatic for functions in Rust.

**Fix:** Either remove it entirely, or add `#[allow(dead_code)]` with a comment explaining why it's kept.

crates/internal/integration-tests/src/harness/scenario_types.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// Prefix for auto-generated reference IDs when `request_id_source_field`
+    /// is set but resolves to an empty value. Defaults to `""`.
+    #[serde(default)]
+    pub request_id_prefix: Option<String>,
+    /// Max length for auto-generated reference IDs (default: no truncation).
+    #[serde(default)]
+    pub request_id_length: Option<usize>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
```
maverox, line 257: **[Correctness]** `#[serde(untagged)]` tries variants in declaration order. If a JSON object has both `must_exist` and `equals` keys, only `MustExist` matches (first variant wins), silently ignoring `equals`.

**Fix:** Add `#[serde(deny_unknown_fields)]` on each variant struct, or use internally tagged representation.

crates/internal/integration-tests/src/harness/report.rs
```diff
@@ -146,6 +152,10 @@ pub fn append_report_batch(entries: Vec<ReportEntry>) -> Result<(), String> {
         ScenarioRunReport::default()
     };
 
+    for run in &mut report.runs {
```
maverox: **[Performance]** Every `append_report_batch` call re-sanitizes ALL existing entries, then also sanitizes new entries. As the report grows, this becomes increasingly wasteful.

Consider sanitizing only new entries if the report format guarantees entries are written post-sanitization.

crates/internal/integration-tests/src/harness/report.rs
```diff
@@ -240,34 +254,103 @@ pub fn extract_pm_and_pmt(grpc_req: Option<&Value>) -> (Option<String>, Option<S
...
+                    }
+                })
+                .collect::<Vec<_>>()
+                .join(" ")
+        }
+    };
+    name.to_string()
+}
+
+fn suite_service_name(suite: &str) -> String {
```
maverox: **[Performance]** `suite_service_name` reads and parses `suite_spec.json` from disk on every call. This is called per-suite per-connector during markdown generation.

**Fix:** Cache the results in a `HashMap<String, String>` or compute the mapping once before the rendering loop.

browser-automation-engine/src/drivers/playwrightDriver.ts
```diff
@@ -0,0 +1,248 @@
...
+
+    throw new Error(`locator.waitFor: Timeout ${timeoutMs}ms exceeded.`);
+  }
+
+  private async findAllLocators(
+    selector: string,
+    opts?: { timeoutMs?: number; state?: WaitState }
+  ): Promise<Locator[]> {
+    const state = opts?.state ?? "attached";
+    const timeoutMs = opts?.timeoutMs ?? 12000;
```
maverox, line 205: **[Bug]** `findAllLocators` only handles `visible` state (line 209-213) and defaults everything else to `attached`. Unlike `findFirstLocator` which handles all four states (`attached`, `detached`, `visible`, `hidden`), the `hidden` and `detached` states are silently ignored here.

Also: the hardcoded `12000` default timeout (here and line 157) differs from `DEFAULT_TIMEOUT_MS = 10_000` in `automationEngine.ts`.

browser-automation-engine/src/drivers/playwrightDriver.ts
```diff
@@ -0,0 +1,248 @@
...
+    return locator.innerText();
+  }
+
+  async isVisible(selector: string, opts?: { timeoutMs?: number }): Promise<boolean> {
+    try {
+      await this.findFirstLocator(selector, {
+        timeoutMs: opts?.timeoutMs,
+        state: "visible"
+      });
+      return true;
```
maverox, line 73: **[Silent Failure]** `isVisible()` catches ALL errors as `false` — not just "element not found" timeouts, but also page crashes, frame detachment, and programming errors. A crashed page would be reported as "element not visible" rather than surfacing the real failure.

Consider catching only `TimeoutError` and re-throwing others.

crates/internal/integration-tests/src/harness/scenario_loader.rs
```diff
@@ -271,10 +328,7 @@ pub fn configured_all_connectors() -> Vec<String> {
         }
     }
 
-    ALL_CONNECTORS_RUN_LIST
-        .iter()
-        .map(|connector| connector.to_string())
-        .collect()
+    discover_all_connectors().unwrap_or_default()
```
maverox: **[Silent Failure]** `discover_all_connectors().unwrap_or_default()` silently returns an empty `Vec<String>` if the `connector_specs/` directory is missing or has wrong permissions. The caller sees "no connectors found" with no indication that the real problem was a filesystem error.

**Fix:** Propagate the error or log a warning:
```rust
discover_all_connectors().unwrap_or_else(|e| {
    eprintln!("warning: failed to discover connectors: {e}");
    Vec::new()
})
```

crates/internal/integration-tests/src/bin/test_ucs.rs
```diff
@@ -0,0 +1,925 @@
...
+
+fn apply_creds_from_env_or_defaults(defaults: &StoredDefaults) {
+    let creds_file = std::env::var("CONNECTOR_AUTH_FILE_PATH")
+        .ok()
+        .or_else(|| std::env::var("UCS_CREDS_PATH").ok())
+        .or_else(|| defaults.creds_file.clone());
+
+    if let Some(creds_file) = creds_file {
+        std::env::set_var("CONNECTOR_AUTH_FILE_PATH", creds_file);
+    }
```
maverox, line 815: **[Deprecation]** `std::env::set_var` is unsound in multi-threaded contexts and marked unsafe in Rust 1.66+. Consider passing the resolved creds path explicitly through function parameters rather than mutating the process environment.

browser-automation-engine/src/utils/validation.ts
AmitsinghTanwar007: Fixed - added validation against allowed enum values before casting for `waitUntil`, `state`, and `match` fields. Now throws `RequestValidationError` with clear message if invalid values are passed.

Makefile
AmitsinghTanwar007: Fixed - combined into single command: `./target/release/grpc-server & echo $$! > $(GRPC_PID_FILE)`. Now correctly captures the backgrounded grpc-server PID.

crates/internal/integration-tests/src/harness/cred_masking.rs
AmitsinghTanwar007: Already implemented - added clarifying comments. Function masks both Bearer tokens (Pass 1) and standalone JWTs (Pass 2 with `JWT_TOKEN_RE.replace_all()` at line 346).

crates/internal/integration-tests/src/harness/cred_masking.rs
AmitsinghTanwar007: Fixed - added `accesscode`, `entityid`, `pin`, `key0`, `key3`, and their x-prefixed variants to the sensitive key list.

browser-automation-engine/src/engine/interpreter.ts
AmitsinghTanwar007: Fixed - now uses `Promise.all()` to wait for both conditions in parallel instead of sequentially. This prevents doubling the timeout when both selector and urlContains are specified.

browser-automation-engine/src/server.ts
AmitsinghTanwar007: Fixed - added SIGTERM/SIGINT handlers with `engine.cleanup()` to close browser sessions and `app.close()` to stop Fastify server. Also added PORT validation to prevent NaN.

browser-automation-engine/src/engine/interpreter.ts
AmitsinghTanwar007: Added security documentation - TSDoc warning in `EvaluateRule` interface and inline comment in `interpreter.ts` explaining that authentication/authorization is needed if server is exposed to untrusted networks.

browser-automation-engine/src/api/routes.ts
AmitsinghTanwar007: Fixed - Now returns HTTP 500 when `result.success === false`. Also added try-catch to handle unexpected errors (Playwright crash, OOM) and log them properly.

crates/internal/integration-tests/src/bin/check_report_creds.rs
AmitsinghTanwar007: Fixed - Now logs directory read errors with `eprintln!` instead of silently returning. Same fix applied to both `check_report_creds.rs` and `mask_report_creds.rs`.

crates/internal/integration-tests/src/harness/sdk_executor.rs
AmitsinghTanwar007: Fixed - Removed the dead `_map_request_error` function. Added comment explaining it was removed after FfiResult refactor.

crates/internal/integration-tests/src/harness/scenario_types.rs
AmitsinghTanwar007: Note: Cannot use `#[serde(deny_unknown_fields)]` on untagged enum variants. Added comment explaining this limitation - untagged deserialization tries variants in order, so extra fields cause it to try the next variant rather than failing.

crates/internal/integration-tests/src/harness/scenario_loader.rs
AmitsinghTanwar007: Fixed - Now logs error with `eprintln!` when `discover_all_connectors` fails, so users see why connector discovery failed instead of silent empty result.

browser-automation-engine/src/drivers/playwrightDriver.ts
AmitsinghTanwar007: Fixed - Now handles all four states (`attached`, `detached`, `visible`, `hidden`) like `findFirstLocator` does. Also added comment about hardcoded 12000ms timeout vs DEFAULT_TIMEOUT_MS.

browser-automation-engine/src/drivers/playwrightDriver.ts
AmitsinghTanwar007: Fixed - Now only catches timeout errors (element not visible is expected). Re-throws other errors like page crashes, frame detachment so they surface properly instead of returning false.

AmitsinghTanwar007 (commented):
**Response to suite_service_name performance comment (crates/internal/integration-tests/src/harness/report.rs:321-354)**

Fixed in commit 03ae76a7c.

Added `OnceLock`-based cache (`SUITE_SERVICE_CACHE`) for suite → service name mappings. The cache is initialized once on first access by:
1. Adding all hardcoded core suite mappings
2. Discovering all connectors and their supported suites
3. Loading suite specs from disk to extract `grpc_method` for connector-specific suites

The cache is built once and reused for all subsequent calls, eliminating repeated file I/O during markdown generation. This significantly improves performance for large test runs where `suite_service_name` is called multiple times per connector per suite.

crates/internal/integration-tests/src/harness/scenario_loader.rs
```diff
@@ -98,6 +119,33 @@ pub fn load_suite_spec(suite: &str) -> Result<SuiteSpec, ScenarioError> {
         .map_err(|source| ScenarioError::SuiteSpecParse { path, source })
 }
 
+/// Loads optional connector-specific browser automation hooks.
+pub fn load_connector_browser_automation_spec(
+    connector: &str,
+) -> Result<Option<ConnectorBrowserAutomationSpec>, ScenarioError> {
```
maverox: Option inside Result is generally a discouraged pattern either if you wanna make it fallible use Result if not then make Optional.... try to make it across the modules.. if there exists valid rationale which simplifies or isn't tech debt then let's discuss.

crates/internal/integration-tests/src/harness/scenario_loader.rs
```diff
@@ -34,17 +32,35 @@ pub fn connector_spec_dir(connector: &str) -> PathBuf {
...
+/// resolve correctly.
+fn normalize_suite_name(suite: &str) -> &str {
+    match suite {
+        "create_access_token" => "server_authentication_token",
+        "create_sdk_session_token" => "client_authentication_token",
+        "create_session_token" => "server_session_authentication_token",
+        other => other,
+    }
+}
+
```
maverox: can you please replace the suite name with the updated name and not use this function.

crates/internal/integration-tests/src/harness/scenario_loader.rs
```diff
@@ -271,10 +346,16 @@ pub fn configured_all_connectors() -> Vec<String> {
...
     }
 
-    ALL_CONNECTORS_RUN_LIST
-        .iter()
-        .map(|connector| connector.to_string())
-        .collect()
+    discover_all_connectors().unwrap_or_else(|err| {
+        #[allow(clippy::print_stderr)]
+        {
+            eprintln!(
```
maverox: why?

gopikrishna000 (approved):

crates/internal/integration-tests/src/bin/test_ucs.rs
```diff
@@ -0,0 +1,942 @@
...
+    scenario_types::ScenarioError,
+};
+use serde::Deserialize;
+use serde_json::Value;
+
+// ── Entrypoint ─────────────────────────────────────────────────────────────────
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("[test_ucs] {error}");
```
maverox, line 56: use tracing

maverox (approved):

PR #765, opened 2026-03-23
feat: split connectorerror to connectorrequesterror and connectorresponseerror

gopikrishna000:
## Description

1. Currently one `ConnectorError` enum holds both request transformation errors and response transformation errors.
If we try to map `ConnectorError` directly to the SDK's `IntegrationError`, some error variants of `ConnectorError` will be meaningless for it, and same for the SDK's `ConnectorResponseTransformationError`.

So, separating `ConnectorError` enum into:
- **`IntegrationError`** — errors during request transformation phase (proto → domain, domain → connector bytes, auth/headers)
- **`ConnectorResponseTransformationError`** — errors during response transformation phase (connector bytes → domain, domain → proto)

And updating `execute_connector_processing_step` (used for gRPC) to return:

```rust
pub enum ConnectorFlowError {
    Request(IntegrationError),            // Failed in build_request_v2
    Client(ApiClientError),               // Failed in call_connector_api
    Response(ConnectorResponseTransformationError), // Failed in handle_response_v2
}
```

instead of the previous combined long `ConnectorError` variant.

This `execute_connector_processing_step` function is used during gRPC calls, where the request phase, network call, and response phase all happen inside it.

The responsibility of `execute_connector_processing_step` is to take `RouterDataV2` with unified request fields and return updated `RouterDataV2` with unified response. It internally calls:
1. `build_request_v2` — connector-specific request from unified domain request
2. HTTP call to connector
3. `handle_response_v2` — convert connector-specific response to unified domain response

## Phase → Error Type Mapping

| Phase | Code | Error Type |
|---|---|---|
| proto → domain | `ForeignTryFrom::foreign_try_from` | `IntegrationError` |
| domain → connector bytes | `build_request_v2` | `IntegrationError` |
| HTTP client + transport | `call_connector_api` | `ApiClientError` |
| connector bytes → domain | `handle_response_v2` | `ConnectorResponseTransformationError` |
| domain → proto | `generate_payment_*_response` | `ConnectorResponseTransformationError` |
| webhook processing | `IncomingWebhook` trait | `WebhookError` |

## SDK (FFI Macros)

```rust
req_transformer -> Result<Request, IntegrationError>
res_transformer -> Result<ProtoResponse, ConnectorResponseTransformationError>
// ApiClientError never appears — SDK uses its own HTTP client per language
```

## Issues fixed

[...]

PR #892, opened 2026-04-02
fix: authorize error inconsistency and remove ApplicationErrorResponse

gopikrishna000:
## Description

This PR fixes the inconsistency in how **integration errors** are handled in the authorize flow compared to all other flows (capture, refund, void, psync, etc.) in UCS.

**Related:**
- Issue: https://github.com/juspay/hyperswitch-cloud/issues/15316
- Architecture analysis: https://github.com/juspay/hyperswitch-cloud/issues/15084#issuecomment-4155865279
- Continued from: https://github.com/juspay/hyperswitch-prism/pull/765 (ConnectorError split)

## The Problem

UCS has two kinds of errors:

| Error Kind | Cause | Correct wire format |
|---|---|---|
| **Connector business errors** (card declined, insufficient funds) | External connector rejects the payment | `Ok(Response)` with `ErrorInfo` populated |
| **Integration errors** (missing field, bad config, encoding failure) | UCS-side failure before/during request | `Err(tonic::Status)` with appropriate gRPC code |

Business errors are handled consistently across all flows. Integration errors are not:

| Flow | Integration Errors | gRPC wire | HTTP result |
|---|---|---|---|
| **Authorize** | Converted to `ErrorResponse` → stuffed into `ErrorInfo` | `Ok(Response)` — always | HTTP 200 with error in body |
| **Capture / Refund / Void / PSync / etc.** | Propagated as `Err(tonic::Status)` | `Err(Status)` | HTTP 400 or 500 |

**Root cause:** Authorize was hand-written in `payments.rs` using a domain-first error model (`PaymentAuthorizationError`). All other flows are macro-generated in `utils.rs` and propagate `Err(tonic::Status)` directly. This PR aligns authorize to the correct standard pattern.

## The Fix

Remove `PaymentAuthorizationError` and `ApplicationErrorResponse` from the authorize path. Integration errors now propagate as `Err(tonic::Status)` directly, just like every other flow:

```
IntegrationError → tonic::Status::invalid_argument  (missing field, bad data)
                 → tonic::Status::internal           (encoding/build failure, bad config)
ApiClientError   → tonic::Status::unavailable
ConnectorResponseTransformationError → tonic::Status::internal
```

Connector business errors continue to return `Ok(Response)` with `ErrorInfo` populated — that behavior is unchanged.

## gRPC Status Mapping for `IntegrationError`

Not all `IntegrationError` variants are 4xx — the mapping differentiates:

| Variants | tonic::Status |
|---|---|
| `MissingRequiredField`, `MissingRequiredFields`, `InvalidDataFormat`, `MismatchedPaymentData`, `InvalidWallet`, `InvalidWalletToken`, `CurrencyNotSupported`, `AmountConversionFailed`, etc. | `invalid_argument` (400) |
| `FlowNotSupported`, `CaptureMethodNotSupported`, `NotSupported`, `NotImplemented` | `failed_precondition` (400) |
| `FailedToObtainAuthType`, `InvalidConnectorConfig`, `ConfigurationError`, `NoConnectorMetaData`, `SourceVerificationFailed` | `unauthenticated` / `internal` (401/500) |
| `RequestEncodingFailed`, `BodySerializationFailed`, `UrlParsingFailed`, `HeaderMapConstructionFailed`, `FailedToObtainIntegrationUrl` | `internal` (500) |

## Also in scope

- Remove `ApplicationErrorResponse` from the authorize path (HTTP-based concept, not applicable in a stateless gRPC service)
- Remove `PaymentAuthorizationError` (no longer needed once authorize follows the macro pattern)

### Additional Changes

## How did you test it?

---

**Update:** Removing silent access token creation logic during granular get/void/capture. Composite get/void/capture already handles that orchestration — that is the entire point of having a composite service. The removed function was also fundamentally flawed: access token fetch failures were handled incorrectly, returning wrong error types instead of propagating as `tonic::Status::unauthenticated`. Granular flows are now truly stateless — if the connector requires an access token, the caller must provide it via `state.access_token`, otherwise `unauthenticated` is returned immediately.

---

**Update 2:** Removed unused helper functions (`handle_order_creation`, `handle_order_creation_for_setup_mandate`, `handle_payment_method_token`) that were previously used in authorize flow. These are now handled directly by composite authorize service, making them redundant, they were also inconsistent.

### DIFF
--- a/crates/types-traits/domain_types/src/payouts/types.rs
+++ b/crates/types-traits/domain_types/src/payouts/types.rs
@@ -285,12 +285,15 @@ impl ForeignTryFrom<grpc_api_types::payouts::CardPayout>
                     .peek()
                     .clone(),
             )
-            .change_context(IntegrationError::InvalidDataFormat {
-                field_name: "card_number",
-                context: IntegrationErrorContext {
-                    additional_context: Some("Invalid card number".to_owned()),
-                    ..Default::default()
-                },
+            .map_err(|e| {
+                error_stack::Report::new(IntegrationError::InvalidDataFormat {
+                    field_name: "card_number",
+                    context: IntegrationErrorContext {
+                        additional_context: Some("Invalid card number".to_owned()),
+                        ..Default::default()
+                    },
+                })
+                .attach_printable(format!("{e:?}"))
             })?,
             expiry_month: ::hyperswitch_masking::Secret::new(
                 card.card_exp_month
@@ -649,15 +652,16 @@ impl ForeignTryFrom<grpc_api_types::payouts::ApplePayDecrypt>
                     .peek()
                     .clone(),
             )
-            .change_context(
-                    IntegrationError::InvalidDataFormat {
+            .map_err(|e| {
+                error_stack::Report::new(IntegrationError::InvalidDataFormat {
                     field_name: "dpan",
                     context: IntegrationErrorContext {
                         additional_context: Some("Invalid dpan".to_owned()),
                         ..Default::default()
                     },
-                },
-                )?,
+                })
+                .attach_printable(format!("{e:?}"))
+            })?,
             expiry_month: ::hyperswitch_masking::Secret::new(
                 apple
                     .expiry_month
@@ -707,15 +711,16 @@ impl ForeignTryFrom<grpc_api_types::payouts::Paypal> for payouts::payout_method_
             email: paypal
                 .email
                 .map(|e| {
-                    e.peek().to_string().parse().change_context(
-                        IntegrationError::InvalidDataFormat {
+                    e.peek().to_string().parse().map_err(|err| {
+                        error_stack::Report::new(IntegrationError::InvalidDataFormat {
                             field_name: "email",
                             context: IntegrationErrorContext {
                                 additional_context: Some("Invalid email".to_owned()),
                                 ..Default::default()
                             },
-                        },
-                    )
+                        })
+                        .attach_printable(format!("{err:?}"))
+                    })
                 })
                 .transpose()?,
             telephone_number: paypal
@@ -763,12 +768,15 @@ impl ForeignTryFrom<grpc_api_types::payouts::InteracPayout>
                 .peek()
                 .to_string()
                 .parse()
-                .change_context(IntegrationError::InvalidDataFormat {
-                    field_name: "email",
-                    context: IntegrationErrorContext {
-                        additional_context: Some("Invalid email".to_owned()),
-                        ..Default::default()
-                    },
+                .map_err(|e| {
+                    error_stack::Report::new(IntegrationError::InvalidDataFormat {
+                        field_name: "email",
+                        context: IntegrationErrorContext {
+                            additional_context: Some("Invalid email".to_owned()),
+                            ..Default::default()
+                        },
+                    })
+                    .attach_printable(format!("{e:?}"))
                 })?,
         })
     }


crates/common/ucs_env/src/error.rs
```diff
@@ -90,177 +91,111 @@ pub enum ConfigurationError {
...
+            | IntegrationError::AmountConversionFailed { .. }
+            | IntegrationError::MandatePaymentDataMismatch { .. }
+            | IntegrationError::MissingApplePayTokenData { .. } => Status::invalid_argument(msg),
+            IntegrationError::FlowNotSupported { .. }
+            | IntegrationError::NotSupported { .. }
+            | IntegrationError::CaptureMethodNotSupported { .. }
+            | IntegrationError::NotImplemented(..) => Status::failed_precondition(msg),
+            IntegrationError::FailedToObtainAuthType { .. }
+            | IntegrationError::InvalidConnectorConfig { .. }
+            | IntegrationError::ConfigurationError { .. }
```
AmitsinghTanwar007, line 121: Configuration errors aren’t auth failures; mapping `IntegrationError::ConfigurationError` to `unauthenticated` will surface as 401. Consider `internal` (500) or `failed_precondition` so client auth isn’t blamed for server config issues.

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -545,14 +541,13 @@ impl ForeignTryFrom<grpc_api_types::payouts::SepaBankTransferPayout>
...
                     .ok_or_else(|| {
-                        error_stack::report!(ApplicationErrorResponse::BadRequest(
-                            crate::errors::ApiError {
-                                sub_code: "MISSING_IBAN".to_owned(),
-                                error_identifier: 400,
-                                error_message: "IBAN is required for SEPA".to_owned(),
-                                error_object: None,
-                            }
-                        ))
+                        error_stack::report!(IntegrationError::MissingRequiredField {
```
AmitsinghTanwar007, line 530: SEPA IBAN missing error is now reported as `field_name: "amount"` with "Amount is required". This looks like a copy/paste regression; should be `field_name: "iban"` (and message about missing IBAN).

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -597,29 +591,31 @@ impl ForeignTryFrom<grpc_api_types::payouts::PixBankTransferPayout>
...
                     .ok_or_else(|| {
-                        error_stack::report!(ApplicationErrorResponse::BadRequest(
-                            crate::errors::ApiError {
-                                sub_code: "MISSING_BANK_ACCOUNT_NUMBER".to_owned(),
-                                error_identifier: 400,
-                                error_message: "Bank account number is required for Pix".to_owned(),
-                                error_object: None,
-                            }
-                        ))
+                        error_stack::report!(IntegrationError::MissingRequiredField {
```
AmitsinghTanwar007, line 580: Pix payout errors are mislabeled: missing `bank_account_number` reports `field_name: "account_holder_name"` and OpenBankingUK message, and missing `pix_key` reports `field_name: "bank_account_number"`. These should reference `bank_account_number` and `pix_key` respectively with Pix-specific messages.

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -659,40 +654,41 @@ impl ForeignTryFrom<grpc_api_types::payouts::ApplePayDecrypt>
...
                     .ok_or_else(|| {
-                        error_stack::report!(ApplicationErrorResponse::BadRequest(
-                            crate::errors::ApiError {
-                                sub_code: "MISSING_DPAN".to_owned(),
-                                error_identifier: 400,
-                                error_message: "DPAN is required for ApplePayDecrypt".to_owned(),
-                                error_object: None,
-                            }
-                        ))
+                        error_stack::report!(IntegrationError::MissingRequiredField {
```
AmitsinghTanwar007, line 641: ApplePayDecrypt missing DPAN now reports `field_name: "iban"` with OpenBankingUK message. This should be `field_name: "dpan"` (and a DPAN-specific message).

crates/ffi/ffi/src/handlers/payments.rs
```diff
@@ -98,7 +91,11 @@ pub fn handle_event_handler(
     grpc_api_types::payments::EventServiceHandleResponse,
     ConnectorResponseTransformationError,
 > {
-    let config = get_config_for_res(environment)?;
+    let config = get_config(environment).map_err(|e| ConnectorResponseTransformationError {
+        error_message: e.error_message,
+        error_code: e.error_code,
+        http_status_code: None,
+    })?;
```
JeevaRamu0104, line 98: why there is map error when function is already returning ConnectorResponseTransformationError

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -714,40 +455,17 @@ impl Payments {
...
         .map_err(|e| {
             tracing::error!("Failed to resolve connector overrides: {:?}", e);
-            let api_error = e.get_api_error();
-            PaymentAuthorizationError::new(
-                grpc_api_types::payments::PaymentStatus::Pending,
-                Some(api_error.error_message.clone()),
-                Some(api_error.sub_code.clone()),
-                Some(api_error.error_identifier.into()),
-            )
+            tonic::Status::internal(format!("Failed to resolve connector overrides: {e}"))
```
maverox: This still flattens every `IntegrationError` from `get_resolved_connectors(...)` to `Status::internal(...)`. That means the authorize path no longer preserves the shared mapping added in `ucs_env::error::IntoGrpcStatus` (for example invalid `x-environment`, invalid config override JSON, and invalid connector config now all surface as `INTERNAL`). Can we switch this to `.into_grpc_status()?` like the other flows?

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -848,32 +840,30 @@ impl ForeignTryFrom<grpc_api_types::payouts::OpenBankingUkPayout>
...
-                error_stack::report!(ApplicationErrorResponse::BadRequest(
-                    crate::errors::ApiError {
-                        sub_code: "INVALID_TOKEN_TYPE".to_owned(),
-                        error_identifier: 400,
-                        error_message: "Invalid pass through token type".to_owned(),
-                        error_object: None,
-                    }
-                ))
+                error_stack::report!(IntegrationError::InvalidDataFormat {
+                    field_name: "recipient_type",
```
maverox: This conversion is validating `pt.token_type`, but the new error reports `field_name: "recipient_type"` / "Invalid payout recipient type". That looks like a copy/paste regression from another branch. The validation error here should stay tied to the passthrough token type.

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -885,19 +875,18 @@ impl ForeignTryFrom<grpc_api_types::payouts::Passthrough>
...
-            error_stack::report!(ApplicationErrorResponse::BadRequest(
-                crate::errors::ApiError {
-                    sub_code: "MISSING_PAYOUT_METHOD_DATA".to_owned(),
-                    error_identifier: 400,
-                    error_message: "Payout method data is required".to_owned(),
-                    error_object: None,
-                }
-            ))
+            error_stack::report!(IntegrationError::MissingRequiredField {
+                field_name: "pix_key",
```
maverox: This branch used to report a missing top-level `payout_method_data`, but after the migration it now returns `MissingRequiredField { field_name: "pix_key" }`. That changes the public validation error for a completely different condition. This should point at the missing `payout_method_data` field instead of a Pix-specific child field.

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -1286,14 +1270,13 @@ impl ForeignTryFrom<grpc_api_types::payouts::PayoutServiceCreateRecipientRequest
...
-                error_stack::report!(ApplicationErrorResponse::BadRequest(
-                    crate::errors::ApiError {
-                        sub_code: "INVALID_PAYOUT_RECIPIENT_TYPE".to_owned(),
-                        error_identifier: 400,
-                        error_message: "Invalid payout recipient type".to_owned(),
-                        error_object: None,
-                    }
-                ))
+                error_stack::report!(IntegrationError::MissingRequiredField {
+                    field_name: "payout_method_data",
```
maverox: This branch is handling an invalid `recipient_type`, but the migrated error now says `field_name: "payout_method_data"` / "Payout method data is required". That will mislead callers and breaks the previous validation contract. This should report the invalid `recipient_type` instead.

crates/grpc-server/grpc-server/tests/test_config_override.rs
```diff
@@ -117,22 +117,28 @@ async fn test_config_override() -> Result<(), Box<dyn std::error::Error>> {
...
+            Ok(response) => {
+                let response = response.into_inner();
+                assert!(
+                    response.error.is_some(),
+                    "Expected error details in response"
+                );
+            }
+            Err(status) => {
+                assert!(
+                    status.code() == tonic::Code::InvalidArgument
```
maverox: This test now accepts either `InvalidArgument` or `Internal`, so it won't catch the authorize-path regression where `IntegrationError` gets flattened to `INTERNAL`. If this PR is meant to standardize the authorize gRPC contract, can we assert the exact status code expected for this scenario?

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -259,248 +255,6 @@ pub struct Payments {
 #[derive(Clone)]
 pub struct Customer;
 
-impl Customer {
```
JeevaRamu0104, line 262: why we have removed these implementation

gopikrishna000: these were dead code, previously used in Authorize composite. But now Authorize is granular, we use composite Authorize to handle this.

PR #908, opened 2026-04-05
feat(connector): implement MIT for jpmorgan

10X-GRACE:
## Summary
- Implement Merchant Initiated Transaction (MIT) support for JPMorgan connector via SetupMandate and RepeatPayment flows
- SetupMandate handles the initial cardholder-initiated transaction with credential storage (`initiatorType=CARDHOLDER`, `accountOnFile=TO_BE_STORED`, `recurringSequence=FIRST`)
- RepeatPayment handles subsequent merchant-initiated transactions (`initiatorType=MERCHANT`, `accountOnFile=STORED`, `recurringSequence=SUBSEQUENT`) with `originalNetworkTransactionId`
- Extracts `networkTransactionId` from payment responses as `network_txn_id` for use in subsequent MIT payments
- Fixed `networkTransactionId` extraction path: moved from `card.networkTransactionId` to correct location `card.networkResponse.networkTransactionId` per JPMorgan API docs

## Changes
- `jpmorgan.rs`: Added SetupMandate and RepeatPayment to `create_all_prerequisites!` and `macro_connector_implementation!` macros; removed empty trait impls replaced by actual implementations
- `requests.rs`: Added MIT-specific types: `JpmorganInitiatorType`, `JpmorganAccountOnFile`, `JpmorganRecurringSequence`, `JpmorganRecurring`, `JpmorganMitCard` (with `originalNetworkTransactionId`), `JpmorganSetupMandateRequest`, `JpmorganRepeatPaymentRequest`
- `responses.rs`: Added `network_transaction_id` field to `NetworkResponse` struct (correct location per JPMorgan API); added `JpmorganSetupMandateResponse` and `JpmorganRepeatPaymentResponse` type aliases
- `transformers.rs`: Added TryFrom transformers for SetupMandate and RepeatPayment request/response; updated `PaymentsResponseData` to extract `network_txn_id` from `card.networkResponse`; added `build_jpmorgan_merchant` helper

## Test plan
- [x] `cargo build --package connector-integration` passes clean (no errors, no warnings)
- [x] Field probe confirms `setup_recurring` flow is `supported` with correct request body
- [x] Live gRPC E2E test against JPMorgan sandbox - all 3 steps pass (see below)
- [x] Verified `originalNetworkTransactionId` is optional for US Visa PAN (only required for EU Visa, EU Mastercard, US/EU tokens per JPMorgan docs)

## gRPC E2E Test Results (JPMorgan Sandbox)

Sequence: CreateServerAuthenticationToken -> SetupRecurring -> Charge

### Step 1: CreateServerAuthenticationToken

```
grpcurl -plaintext \
  -H 'x-connector-config: {"config":{"Jpmorgan":{"client_id":"<CLIENT_ID>","client_secret":"<CLIENT_SECRET>","company_name":"TestCompany","product_name":"TestProduct","merchant_purchase_description":"Test Purchase"}}}' \
  -H 'x-reference-id: auth_real_001' \
  -d '{"merchant_access_token_id": "auth_real_001", "connector": "JPMORGAN"}' \
  localhost:9090 types.MerchantAuthenticationService/CreateServerAuthenticationToken
```

Response: `status: OPERATION_STATUS_SUCCESS`, `statusCode: 200`, JWT access token returned

### Step 2: SetupRecurring (Initial CIT Mandate)

```
grpcurl -plaintext \
  -H 'x-connector-config: {"config":{"Jpmorgan":{"client_id":"<CLIENT_ID>","client_secret":"<CLIENT_SECRET>","company_name":"TestCompany","product_name":"TestProduct","merchant_purchase_description":"Test Purchase"}}}' \
  -H 'x-reference-id: setup_real_001' \
  -d '{
    "merchant_recurring_payment_id": "setup_real_001",
    "amount": {"minor_amount": 1000, "currency": "USD"},
    "payment_method": {
      "card": {
        "card_number": {"value": "4242424242424242"},
        "card_cvc": {"value": "123"},
        "card_exp_month": {"value": "12"},
        "card_exp_year": {"value": "2028"},
        "card_network": "VISA"
      }
    },
    "address": {},
    "auth_type": "NO_THREE_DS",
    "enrolled_for_3ds": false,
    "request_incremental_authorization": false,
    "setup_future_usage": "OFF_SESSION",
    "customer_acceptance": {
      "acceptance_type": "ONLINE",
      "accepted_at": 1712419200,
      "online_mandate_details": {"ip_address": "192.168.1.1", "user_agent": "Mozilla/5.0"}
    },
    "state": {
      "access_token": {
        "token": {"value": "<ACCESS_TOKEN>"},
        "expires_in_seconds": 3599,
        "token_type": "Bearer"
      }
    }
  }' \
  localhost:9090 types.PaymentService/SetupRecurring
```

Response: `status: CHARGED`, `x-jpmc-service-type: sandbox`

Connector request body sent to JPMorgan:

```json
{
  "captureMethod": "NOW",
  "amount": 1000,
  "currency": "USD",
  "initiatorType": "CARDHOLDER",
  "accountOnFile": "TO_BE_STORED",
  "recurring": {"recurringSequence": "FIRST", "agreementId": "setup_real_001", "isVariableAmount": false},
  "isAmountFinal": true
}
```

JPMorgan raw response confirmed: `initiatorType: CARDHOLDER`, `accountOnFile: TO_BE_STORED`, `responseCode: APPROVED`

> Note: `networkTransactionId` is `null` in sandbox for US Visa PAN - per JPMorgan docs, it is only generated for EU Visa, EU Mastercard, US/EU Visa tokens, US/EU Discover tokens, EU Amex, and EU Cartes Bancaires.

### Step 3: Charge (Subsequent MIT / RepeatPayment)

**Test A: With `originalNetworkTransactionId` (placeholder value)**

```
grpcurl -plaintext \
  -H 'x-connector-config: {"config":{"Jpmorgan":{"client_id":"<CLIENT_ID>","client_secret":"<CLIENT_SECRET>","company_name":"TestCompany","product_name":"TestProduct","merchant_purchase_description":"Test Purchase"}}}' \
  -H 'x-reference-id: charge_real_001' \
  -d '{
    "merchant_charge_id": "charge_real_001",
    "connector_recurring_payment_id": {"network_mandate_id": "016153570198200"},
    "amount": {"minor_amount": 2500, "currency": "USD"},
    "payment_method": {
      "card": {
        "card_number": {"value": "4242424242424242"},
        "card_cvc": {"value": "123"},
        "card_exp_month": {"value": "12"},
        "card_exp_year": {"value": "2028"},
        "card_network": "VISA"
      }
    },
    "state": {
      "access_token": {
        "token": {"value": "<ACCESS_TOKEN>"},
        "expires_in_seconds": 3599,
        "token_type": "Bearer"
      }
    }
  }' \
  localhost:9090 types.RecurringPaymentService/Charge
```

Response: `status: CHARGED`, `responseMessage: Transaction approved by Issuer1 AP`, `approvalCode: tst269`

Connector request body sent to JPMorgan:

```json
{
  "captureMethod": "NOW",
  "amount": 2500,
  "currency": "USD",
  "initiatorType": "MERCHANT",
  "accountOnFile": "STORED",
  "recurring": {"recurringSequence": "SUBSEQUENT", "agreementId": "charge_real_001"},
  "paymentMethodType": {"card": {"originalNetworkTransactionId": "016153570198200"}},
  "isAmountFinal": true
}
```

JPMorgan confirmed MIT fields in raw response: `initiatorType: MERCHANT`, `accountOnFile: STORED`

**Test B: Without `originalNetworkTransactionId` (US Visa PAN - not required)**

```
grpcurl -plaintext \
  -H 'x-connector-config: {"config":{"Jpmorgan":{"client_id":"<CLIENT_ID>","client_secret":"<CLIENT_SECRET>","company_name":"TestCompany","product_name":"TestProduct","merchant_purchase_description":"Test Purchase"}}}' \
  -H 'x-reference-id: charge_noNTI_002' \
  -d '{
    "merchant_charge_id": "charge_noNTI_002",
    "connector_recurring_payment_id": {
      "connector_mandate_id": {"connector_mandate_id": "<TRANSACTION_ID_FROM_STEP_2>"}
    },
    "amount": {"minor_amount": 2500, "currency": "USD"},
    "payment_method": {
      "card": {
        "card_number": {"value": "4242424242424242"},
        "card_cvc": {"value": "123"},
        "card_exp_month": {"value": "12"},
        "card_exp_year": {"value": "2028"},
        "card_network": "VISA"
      }
    },
    "state": {
      "access_token": {
        "token": {"value": "<ACCESS_TOKEN>"},
        "expires_in_seconds": 3599,
        "token_type": "Bearer"
      }
    }
  }' \
  localhost:9090 types.RecurringPaymentService/Charge
```

Response: `status: CHARGED`, `responseMessage: Transaction approved by Issuer1 AP` - **approved without `originalNetworkTransactionId`**

This confirms JPMorgan docs: `originalNetworkTransactionId` is only required for specific card networks/regions (EU Visa PAN, EU Mastercard, US/EU tokens). For US Visa PAN, it is optional and the sandbox does not generate one.

## `originalNetworkTransactionId` Requirement Matrix (from JPMorgan docs)

| Card Network | Region | Required? |
|---|---|---|
| Visa PAN | US | No |
| Visa PAN | EU | **Yes** |
| Visa Token | US, EU | **Yes** |
| Mastercard | US | No |
| Mastercard | EU | **Yes** |
| Discover Token | US, EU | **Yes** |
| American Express | EU | **Yes** |
| Cartes Bancaires | EU | **Yes** |

## References
- [JPMorgan Stored Payments](https://developer.payments.jpmorgan.com/docs/commerce/online-payments/capabilities/online-payments/payment-enhancements/stored-payments)
- [JPMorgan Recurring Payments](https://developer.payments.jpmorgan.com/docs/commerce/online-payments/capabilities/online-payments/payment-enhancements/recurring-payments)


deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    .resource_common_data
+                    .connector_request_reference_id
+                    .clone();
+
+                let amount = router_data
+                    .request
+                    .minor_amount
+                    .map(|a| JpmorganAmountConvertor::convert(a, router_data.request.currency))
+                    .transpose()?
+                    .unwrap_or(common_utils::types::MinorUnit::new(0));
```
Shubhodip900: 🟠 **HIGH**

SetupMandate flow uses `unwrap_or(MinorUnit::new(0))` for the amount field, allowing zero-amount transactions to be sent to JPMorgan if `minor_amount` is None. This is inconsistent with the RepeatPayment flow which properly propagates the error with `?`. A zero-amount SetupMandate request will likely fail at JPMorgan or create invalid mandate records.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    is_amount_final: true,
+                })
+            }
+            PaymentMethodData::MandatePayment => {
+                // MandatePayment flow: card data comes from stored mandate, not payment_method_data
+                let auth = JpmorganAuthType::try_from(&router_data.connector_config)?;
+                let merchant = build_jpmorgan_merchant(&auth)?;
+
+                let payment_method_type = requests::JpmorganMitPaymentMethodType::<T> {
+                    card: None, // No card data in MandatePayment; JPMorgan uses network_transaction_id
```
Shubhodip900: 🟠 **HIGH**

Critical bug in RepeatPayment MandatePayment branch: `network_transaction_id` is extracted from `mandate_reference` at line 1070 but never used in this branch. The `card: None` means `original_network_transaction_id` is not sent to JPMorgan, causing MIT transactions to fail. The network_transaction_id must be included in the request payload.

crates/integrations/connector-integration/src/connectors/jpmorgan/requests.rs
```diff
@@ -139,3 +139,85 @@ pub struct JpmorganRefundRequest {
...
+    pub recurring_sequence: JpmorganRecurringSequence,
+    pub agreement_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub is_variable_amount: Option<bool>,
+}
+
+/// JPMorgan card with optional originalNetworkTransactionId for MIT
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct JpmorganMitCard<T: PaymentMethodDataTypes> {
```
Shubhodip900: 🟡 **MEDIUM**

`JpmorganMitCard` struct is missing a CVC/security code field. For SetupMandate (initial CIT), JPMorgan requires CVC for cardholder verification. Without this field, SetupMandate requests will be rejected by JPMorgan's risk checks. Add `#[serde(skip_serializing_if = "Option::is_none")] pub security_code: Option<Secret<String>>` field.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                            .peek()
+                            .parse::<i32>()
+                            .change_context(IntegrationError::RequestEncodingFailed {
+                                context: Default::default(),
+                            })?,
+                    ),
+                };
+
+                // Get the network transaction ID from mandate reference
+                let network_transaction_id = match &router_data.request.mandate_reference {
```
Shubhodip900: 🟡 **MEDIUM**

In RepeatPayment Card branch, `network_transaction_id` is extracted but only used if `mandate_reference` is `NetworkMandateId`. Other mandate reference types silently result in `None`, causing the MIT to fail at JPMorgan. Consider returning an explicit error for unsupported mandate reference types.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    account_number: card_data.card_number.clone(),
+                    expiry,
+                    original_network_transaction_id: None,
+                };
+
+                let payment_method_type =
+                    requests::JpmorganMitPaymentMethodType { card: Some(card) };
+
+                // Use connector_request_reference_id as agreement_id
+                let agreement_id = router_data
```
Shubhodip900: 🟡 **MEDIUM**

`agreement_id` is set from `connector_request_reference_id` without validation. If this field is empty or malformed, JPMorgan will reject the request. Consider adding validation to ensure agreement_id is non-empty before constructing the request.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    .request
+                    .minor_amount
+                    .map(|a| JpmorganAmountConvertor::convert(a, router_data.request.currency))
+                    .transpose()?
+                    .unwrap_or(common_utils::types::MinorUnit::new(0));
+
+                let recurring = requests::JpmorganRecurring {
+                    recurring_sequence: requests::JpmorganRecurringSequence::First,
+                    agreement_id,
+                    is_variable_amount: Some(false),
```
Shubhodip900: 🟢 **LOW**

Hardcoded `is_variable_amount: Some(false)` in SetupMandate prevents merchants from setting up variable amount mandates (e.g., for utility bills). This should be configurable based on merchant requirements rather than hardcoded.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
         })
     }
 }
+
+// Helper to build JpmorganMerchant from auth config
+fn build_jpmorgan_merchant(auth: &JpmorganAuthType) -> Result<requests::JpmorganMerchant, Error> {
```
Shubhodip900: This can be a From conversion instead of normal function.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    merchant,
+                    payment_method_type,
+                    recurring,
+                    initiator_type: requests::JpmorganInitiatorType::Cardholder,
+                    account_on_file: requests::JpmorganAccountOnFile::ToBeStored,
+                    is_amount_final: true,
+                })
+            }
+            _ => Err(IntegrationError::not_implemented(
+                "Only Card payment method is supported for SetupMandate".to_string(),
```
Shubhodip900: use correct error message, if other payment methods are not supported by connector then use `IntegrationError::NotSupported`,  else in the message write other payment methods are not implemented,

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                    year: Secret::new(
+                        card_data
+                            .get_expiry_year_4_digit()
+                            .peek()
+                            .parse::<i32>()
+                            .change_context(IntegrationError::RequestEncodingFailed {
+                                context: Default::default(),
+                            })?,
+                    ),
+                };
```
Shubhodip900: This logic is repeated in both setupMandate and RepeatPayment, can we have a common function for this logic.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+
+                // Use connector_request_reference_id as agreement_id
+                let agreement_id = router_data
+                    .resource_common_data
+                    .connector_request_reference_id
+                    .clone();
+
+                let capture_method = map_capture_method(router_data.request.capture_method)?;
+
+                let amount = JpmorganAmountConvertor::convert(
```
Shubhodip900: 🟠 HIGH

Used of `JpmorganAmountConvertor` instead of using connector amount conversion from macro framework. Use `item.connector.amount_converter()`

deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -899,3 +909,331 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                };
+
+                let agreement_id = router_data
+                    .resource_common_data
+                    .connector_request_reference_id
+                    .clone();
+
+                let capture_method = map_capture_method(router_data.request.capture_method)?;
+
+                let amount = JpmorganAmountConvertor::convert(
```
Shubhodip900: use `item.connector.amount_converter()` for amount conversion with `AmountConversionFailed` error message if the conversion failed.

Shubhodip900 (changes requested):

Shubhodip900 (dismissed):

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -744,12 +746,20 @@ fn map_transaction_state_to_attempt_status(
...
+            .payment_method_type
+            .as_ref()
+            .and_then(|pmt| pmt.card.as_ref())
+            .and_then(|card| card.network_response.as_ref())
+            .and_then(|nr| nr.network_transaction_id.clone());
+
         Ok(Self::TransactionResponse {
             resource_id: ResponseId::ConnectorTransactionId(item.transaction_id.clone()),
             redirection_data: None,
             mandate_reference: None,
```
AkshayaFoiger: Why are we not populating the mandate reference ?

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -999,3 +1009,321 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+
+        Ok(Self {
+            response: Ok(response_data),
+            resource_common_data: PaymentFlowData {
+                status,
+                ..item.router_data.resource_common_data
+            },
+            ..item.router_data
+        })
+    }
```
AkshayaFoiger, line 1367: Return ErrorResponse for failed transaction

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -767,6 +788,33 @@ impl TryFrom<&responses::JpmorganPaymentsResponse> for AttemptStatus {
...
 }
 
+/// Build the `response` field for a JPMorgan payments flow: `Err(ErrorResponse)`
+/// when the transaction was declined/errored, otherwise `Ok(PaymentsResponseData)`.
+fn build_payments_response_result(
+    response: &responses::JpmorganPaymentsResponse,
+    http_code: u16,
+    status: AttemptStatus,
+) -> Result<Result<PaymentsResponseData, ErrorResponse>, ResponseError> {
+    if status == AttemptStatus::Failure {
```
AkshayaFoiger: use the is_payment_failure  function

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
```diff
@@ -999,3 +1047,321 @@ impl TryFrom<ResponseRouterData<responses::JpmorganClientAuthResponse, Self>>
...
+                product_name: auth.product_name.clone().ok_or(
+                    IntegrationError::MissingRequiredField {
+                        field_name: "product_name",
+                        context: Default::default(),
+                    },
+                )?,
+            },
+            soft_merchant: requests::JpmorganSoftMerchant {
+                merchant_purchase_description: auth.merchant_purchase_description.clone().ok_or(
+                    IntegrationError::MissingRequiredField {
```
AkshayaFoiger, line 1072: Why are the product name and purchase description being derived from auth_type data? These can vary per payment, so wouldn’t it make more sense to fetch them from order_details instead?

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Fixed in b5a711cfc: switched to `domain_types::utils::is_payment_failure(status)` in `build_payments_response_result`, and pass the real `status` through into `ErrorResponse.attempt_status` instead of pinning it to `AttemptStatus::Failure`.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Already fixed in 151b91529. The previous transformer always sent `paymentMethodType.card` with the raw PAN, which is wrong for JPMorgan MIT. The current code branches on `mandate_reference`:

- `ConnectorMandateId` → `paymentMethodType.transactionReference.transactionReferenceId` (no card body); JPMorgan looks up the stored credential by its own `transactionId`.
- `NetworkMandateId` → `paymentMethodType.card.{accountNumber, expiry, originalNetworkTransactionId}`; JPMorgan requires `accountNumber+expiry` even when an NTI is supplied.
- `NetworkTokenWithNTI` → `NotImplemented`.

Verified live against sandbox in both paths (see PR test plan): outbound payload contains `originalNetworkTransactionId` on the NetworkMandateId path, and `transactionReferenceId` on the ConnectorMandateId path.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Done in eb7cf2567 — `build_jpmorgan_merchant` is now `impl TryFrom<&JpmorganAuthType> for requests::JpmorganMerchant` (transformers.rs around line 1051), called as `requests::JpmorganMerchant::try_from(&auth)?` from both SetupMandate and RepeatPayment.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Updated in eb7cf2567. Variant is now `IntegrationError::NotImplemented(..)` (the `not_implemented` helper was removed from the codebase in #1157), with messages that explicitly call out which methods are implemented vs not:

- SetupMandate non-card arm: `"Only Card payment method is implemented for JPMorgan SetupMandate"`
- RepeatPayment `NetworkTokenWithNTI`: `"NetworkTokenWithNTI mandate reference is not implemented for JPMorgan RepeatPayment"`

JPMorgan does support multiple PMs in the Authorize flow (Card / Apple Pay / Google Pay etc.), so the gap here is implementation in this connector, not gateway capability — `NotImplemented` is the right variant per the project convention.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Extracted the shared bits in eb7cf2567:

- Building the `JpmorganMerchant` block → `impl TryFrom<&JpmorganAuthType> for JpmorganMerchant`.
- Card month/year parsing → `build_jpmorgan_expiry(card_data)`.
- Failure-vs-success response shaping → `build_payments_response_result(...)` (used by Authorize, Capture, Void, Sync, SetupMandate, RepeatPayment).

The bits that remain inline are flow-specific: SetupMandate hard-codes `CapMethod::Now`, `recurring_sequence=First`, `is_variable_amount=Some(false)`, `initiator_type=Cardholder`, `account_on_file=ToBeStored`; RepeatPayment uses `map_capture_method(...)`, `recurring_sequence=Subsequent`, `is_variable_amount=None`, `initiator_type=Merchant`, `account_on_file=Stored`, plus the `mandate_reference` branching for `paymentMethodType`. Folding those into a shared builder would need the call site to pass ~7 parameters — happy to do that if you'd rather have one helper, but I think keeping the two TryFroms self-contained reads better. WDYT?

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Done in eb7cf2567. SetupMandate now uses `item.connector.amount_converter.convert(a, currency).change_context(IntegrationError::AmountConversionFailed { context: Default::default() })` (transformers.rs ~line 1158), and `amount_converter: MinorUnit` is wired into the connector via `create_all_prerequisites!` in jpmorgan.rs. The other call sites for Authorize/Capture/Refund still go through the legacy `JpmorganAmountConvertor`; happy to migrate those in a follow-up if you'd like — they were out of scope for this MIT PR.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Done in eb7cf2567 — RepeatPayment also goes through `item.connector.amount_converter.convert(...).change_context(IntegrationError::AmountConversionFailed { ... })` (transformers.rs ~line 1260).

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Populated in 183f4b38a. `PaymentsResponseData::try_from(&JpmorganPaymentsResponse)` now surfaces JPMorgan's own `transaction_id` as the `connector_mandate_id`:

```rust
let mandate_reference = MandateReference {
    connector_mandate_id: Some(item.transaction_id.clone()),
    payment_method_id: None,
    connector_mandate_request_reference_id: None,
};
// ...
mandate_reference: Some(Box::new(mandate_reference)),
```

This is what JPMorgan's RepeatPayment expects in `paymentMethodType.transactionReference.transactionReferenceId`, so the framework can hand it back on subsequent MIT charges via the `ConnectorMandateId` branch.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: Fixed in 183f4b38a. The RepeatPayment response transformer now goes through the shared `build_payments_response_result(...)` helper (transformers.rs ~line 1356), which returns `Err(ErrorResponse { ... })` when the attempt status is a failure (now using `is_payment_failure` after b5a711cfc) and `Ok(PaymentsResponseData)` otherwise. Same path is used by Authorize / Capture / Sync / Void / SetupMandate so failure mapping stays consistent.

crates/integrations/connector-integration/src/connectors/jpmorgan/transformers.rs
10X-GRACE: These three fields aren't really per-payment data — they're merchant-integration identity fields that JPMorgan expects to stay constant across an integration:

- `merchantSoftware.companyName` and `merchantSoftware.productName` identify the *commerce platform/software* making the API call (e.g. "Hyperswitch v1"). This is set once when the merchant onboards their integration.
- `softMerchant.merchantPurchaseDescription` is the cardholder-facing statement descriptor (what shows up on the bank statement) — also a merchant-level setting per JPMorgan's stored payments docs, not per transaction.

That's the same place the upstream Hyperswitch JPMorgan implementation pulls them from — `hyperswitch/crates/hyperswitch_connectors/src/connectors/jpmorgan/transformers.rs:178-182` reads `company_name` / `product_name` from `connector_metadata` (the merchant connector account config), not from `order_details`. We're keeping the same shape here so MCA-level config stays the source of truth.

If JPMorgan exposes a per-order override field (e.g. line items / billing descriptor on a purchase), that'd belong in `order_details` — but `merchantSoftware` and `softMerchant` are scoped to the integration, not the order. Happy to revisit if you have a doc reference saying otherwise.

AkshayaFoiger (approved):

Shubhodip900 (approved):

AmitsinghTanwar007 (approved):

PR #909, opened 2026-04-05
feat(connector): implement MIT for paybox

10X-GRACE:
## Summary

Implement **MIT** (Merchant-Initiated Transactions) flow for **paybox** connector.

This adds **SetupMandate** (subscriber registration via TYPE=00056) and **RepeatPayment** (subscriber authorization via TYPE=00051/00053) flows, enabling Paybox Direct Plus recurring payment capability.

## Changes

- Added SetupMandate and RepeatPayment flows to `create_all_prerequisites!` macro in `paybox.rs`
- Added `macro_connector_implementation!` blocks for both SetupMandate and RepeatPayment flows
- Added `PayboxSetupMandateRequest` (TYPE=00056 - Register new subscriber) with card data and subscriber reference
- Added `PayboxRepeatPaymentRequest` (TYPE=00051/00053 - Auth/Auth+Capture on subscriber) with mandate reference
- Added response transformers with `MandateReference` for subscriber ID passthrough
- Added constants: `REGISTER_SUBSCRIBER_REQUEST`, `SUBSCRIBER_AUTH_REQUEST`, `SUBSCRIBER_AUTH_AND_CAPTURE_REQUEST`, `PAY_ORIGIN_RECURRING`
- Replaced empty `ConnectorIntegrationV2` stubs for SetupMandate and RepeatPayment with full implementations

## Files Modified

- `crates/integrations/connector-integration/src/connectors/paybox.rs`
- `crates/integrations/connector-integration/src/connectors/paybox/transformers.rs`

## gRPC Test Results

**Status: PASS (both SetupRecurring and Charge succeed on Paybox sandbox)**

## Validation Checklist

- [x] grpcurl SetupRecurring reached Paybox API (HTTP 200)
- [x] grpcurl response returned success status (CHARGED)
- [x] grpcurl RecurringPaymentService/Charge returned success (CHARGED)
- [x] Only connector-specific files modified

## Follow-up: SetupRecurring -> Charge chain (commit 229a215a9)

**Verdict: PASS**

### Root causes fixed

1. **Currency encoding (all flows)**: All Paybox request structs used `common_enums::Currency` which serializes as alpha codes ("EUR", "USD"). Paybox API requires ISO 4217 numeric codes ("978", "840"). Fixed by changing currency fields to `String` and using `currency.iso_4217().to_string()` in all flows (Authorize, Capture, Void, Refund, SetupMandate, RepeatPayment).

2. **SetupMandate response mandate mapping**: The response was mapping REFABONNE (subscriber ref) to `connector_mandate_id`. Per upstream hyperswitch reference, PORTEUR (stored card token) should be `connector_mandate_id` and REFABONNE should be `connector_mandate_request_reference_id`.

3. **RepeatPayment PORTEUR and DATEVAL**: PORTEUR was empty and DATEVAL was "0000". Fixed to send the stored card token (from `connector_mandate_id`) as PORTEUR and pack card expiry into mandate ID using "::" delimiter for later extraction.

### Step A (SetupRecurring)

```
grpcurl -plaintext \
  -H 'x-connector: paybox' -H 'x-auth: multi-auth-key' \
  -H 'x-api-key: <REDACTED>' -H 'x-key1: <REDACTED>' \
  -H 'x-key2: <REDACTED>' -H 'x-api-secret: <REDACTED>' \
  -d '{ "merchant_recurring_payment_id": "paybox-mit-e2e-001",
        "amount": {"minor_amount": 100, "currency": 45},
        "payment_method": {"card": {"card_number": {"value": "<REDACTED_PAN>"}, ...}},
        ... }' \
  localhost:50071 types.PaymentService/SetupRecurring

Response: { "status": "CHARGED", "statusCode": 200,
  "mandateReference": { "connectorMandateId": {
    "connectorMandateId": "<REDACTED_TOKEN>::1228",
    "connectorMandateRequestReferenceId": "paybox-mit-e2e-001" } } }
```

Raw: `TYPE=00056&DEVISE=978&PORTEUR=<REDACTED_PAN>&DATEVAL=1228&REFABONNE=paybox-mit-e2e-001&ACTIVITE=024`

### Step B (RecurringPaymentService/Charge)

```
grpcurl -plaintext \
  -H 'x-connector: paybox' -H 'x-auth: multi-auth-key' \
  -H 'x-api-key: <REDACTED>' -H 'x-key1: <REDACTED>' \
  -H 'x-key2: <REDACTED>' -H 'x-api-secret: <REDACTED>' \
  -d '{ "merchant_charge_id": "paybox-mit-charge-e2e-001",
        "connector_recurring_payment_id": {"connector_mandate_id": {
          "connector_mandate_id": "<REDACTED_TOKEN>::1228",
          "connector_mandate_request_reference_id": "paybox-mit-e2e-001"}},
        "amount": {"minor_amount": 200, "currency": 45},
        "capture_method": "AUTOMATIC" }' \
  localhost:50071 types.RecurringPaymentService/Charge

Response: { "status": "CHARGED", "statusCode": 200,
  "connectorTransactionId": "0090798120" }
```

Raw: `TYPE=00053&DEVISE=978&PORTEUR=<REDACTED_TOKEN>&DATEVAL=1228&REFABONNE=paybox-mit-e2e-001&ACTIVITE=027`


10X-GRACE:
- **fix(paybox)**: Fixed ISO 4217 numeric currency encoding (`DEVISE=978` instead of `DEVISE=EUR`) across all Paybox flows + corrected SetupMandate->RepeatPayment mandate field mapping (PORTEUR/REFABONNE/DATEVAL). Both SetupRecurring (Step A) and RecurringPaymentService/Charge (Step B) now return `CHARGED` on Paybox sandbox.
- **Commit**: 229a215a9 — currency fix applies to Authorize, Capture, Void, Refund in addition to MIT flows.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+        )?;
+
+        // Split packed mandate id into PORTEUR (stored card token) and DATEVAL (card expiry)
+        let (porteur, expiry_str) = if let Some(idx) = packed_mandate_id.find("::") {
+            (
+                packed_mandate_id[..idx].to_string(),
+                packed_mandate_id[idx + 2..].to_string(),
+            )
+        } else {
+            (packed_mandate_id.clone(), "0000".to_string())
```
Shubhodip900: 🟠 **HIGH**

Hardcoded fallback expiry "0000" for missing DATEVAL in RepeatPayment flow. When the packed mandate_id doesn't contain "::" delimiter, the code falls back to "0000" as expiration date. This will cause Paybox API rejections since "0000" is not a valid card expiry. The connector should either require a properly formatted mandate_id with expiry or extract expiry from stored mandate data. This silent fallback creates failed transactions.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+        };
+
+        // REFABONNE = connector_mandate_request_reference_id (subscriber reference)
+        let refabonne = match &router_data.request.mandate_reference {
+            MandateReferenceId::ConnectorMandateId(connector_mandate_ids) => connector_mandate_ids
+                .get_connector_mandate_request_reference_id()
+                .unwrap_or_else(|| porteur.clone()),
```
Shubhodip900: 🟡 **MEDIUM**

Unwrap used on optional field without proper error handling. `get_connector_mandate_request_reference_id()` returns an Option, but `unwrap_or_else` falls back to `porteur.clone()` which is the card token (PORTEUR), not the subscriber reference (REFABONNE). This conflates two different Paybox fields - the subscriber token vs the subscriber reference ID. If mandate_reference is not ConnectorMandateId variant, it also falls back to porteur. This could cause REFABONNE to be set to the wrong value for MIT transactions.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+            currency: router_data.request.currency.iso_4217().to_string(),
+            reference: router_data
+                .resource_common_data
+                .connector_request_reference_id
+                .clone(),
+            date: generate_date_time()?,
+            subscriber_number: porteur,
+            expiration_date,
+            subscriber_ref: refabonne,
+            activity: PAY_ORIGIN_RECURRING.to_string(),
```
Shubhodip900, line 1500: 🟡 **MEDIUM**

Activity code inconsistency in RepeatPayment. The code uses `PAY_ORIGIN_RECURRING` ("027") for RepeatPayment, but the field probe sample shows "027" for recurring_charge. However, the SetupMandate uses `PAY_ORIGIN_INTERNET` ("024") which might be incorrect for subscriber registration - Paybox Direct Plus typically requires specific activity codes for mandate setup vs recurring transactions. Verify that "024" is correct for TYPE=00056 (Register Subscriber).

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+            };
+
+            let connector_metadata = serde_json::json!(PayboxMeta {
+                connector_request_id: item.response.transaction_number.clone()
+            });
+
+            // Preserve the mandate reference from the subscriber
+            let mandate_reference = item.response.customer_id.as_ref().map(|subscriber_id| {
+                Box::new(MandateReference {
+                    connector_mandate_id: Some(subscriber_id.peek().to_string()),
```
Shubhodip900: 🟡 **MEDIUM**

Mandate reference reconstruction in RepeatPayment response incorrectly maps customer_id. The response sets `connector_mandate_id: Some(subscriber_id.peek().to_string())` but this loses the expiry information that was packed in SetupMandate. The mandate reference should preserve the "PORTEUR::DATEVAL" format for consistency, or the next MIT transaction will fail to parse the expiry.

data/field_probe/paybox.json
```diff
@@ -546,18 +546,81 @@
...
+          "off_session": true
+        },
+        "sample": {
+          "url": "https://preprod-ppps.paybox.com/PPPS.php",
+          "method": "Post",
+          "headers": {
+            "content-type": "application/x-www-form-urlencoded",
+            "via": "HyperSwitch"
+          },
+          "body": "VERSION=00104&TYPE=00053&SITE=probe_site&RANG=probe_rank&CLE=probe_key&NUMQUESTION=000000000&MONTANT=1000&DEVISE=840&REFERENCE=&DATEQ=00000000000000&PORTEUR=probe-mandate-123&DATEVAL=0000&REFABONNE=probe-mandate-123&ACTIVITE=027"
```
Shubhodip900: 🟡 **MEDIUM**

Empty REFERENCE field in recurring_charge sample. The sample body shows `REFERENCE=` with empty value, but Paybox API typically requires a unique reference for each transaction. This suggests the RepeatPayment flow might not be populating the reference field correctly from connector_request_reference_id.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+            rank: auth.rank,
+            key: auth.key,
+            paybox_request_number: generate_request_id()?,
+            amount,
+            currency: router_data.request.currency.iso_4217().to_string(),
+            reference: subscriber_ref.clone(),
+            date: generate_date_time()?,
+            card_number: card_data.card_number.clone(),
+            expiration_date,
+            cvv: card_data.card_cvc.clone(),
```
Shubhodip900, line 1217: 🟢 **LOW**

Missing validation for card CVC in SetupMandate. The code clones `card_data.card_cvc` directly without checking if it's present. For zero-dollar mandate setups, some issuers might reject without CVV, but Paybox requires it in the API. If the input card has no CVC, this sends empty string which may cause API errors.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1090,435 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+            });
+
+            // Pack PORTEUR (stored card token) and card expiry into connector_mandate_id
+            // using "::" delimiter so MIT can extract both PORTEUR and DATEVAL.
+            // REFABONNE (customer_id) = subscriber reference for connector_mandate_request_reference_id.
+            let carrier_with_expiry = item.response.carrier_id.as_ref().map(|carrier| {
+                let card_expiry = match &item.router_data.request.payment_method_data {
+                    PaymentMethodData::Card(card) => card
+                        .get_card_expiry_month_year_2_digit_with_delimiter("".to_owned())
+                        .ok()
```
Shubhodip900: 🟡 **MEDIUM**

Silent error handling for card expiry extraction. `get_card_expiry_month_year_2_digit_with_delimiter("")` returns a Result but errors are silently discarded with `.ok()`. If card expiry format is invalid, the mandate will be stored without expiry (just carrier_id), causing subsequent RepeatPayment transactions to fail when trying to parse the missing "::" delimiter and falling back to "0000".

Shubhodip900 (changes requested):

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
iemyashasvi: In hyperswitch, PaymentsRequest (which handles TYPE=00001, 00003, AND 00056) always sends activity: PAY_ORIGIN_INTERNET

data/field_probe/paybox.json
iemyashasvi: this is not related to the code files

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -188,7 +195,7 @@ pub struct PayboxPaymentRequest<T: PaymentMethodDataTypes> {
     #[serde(rename = "MONTANT")]
     pub amount: MinorUnit,
     #[serde(rename = "DEVISE")]
-    pub currency: common_enums::Currency,
+    pub currency: String,
```
Shubhodip900, line 198: why we are changing this from `Currency` type to plain string?

deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: We intentionally changed `DEVISE` from `common_enums::Currency` to `String`.

**Why:** Paybox requires ISO 4217 **numeric** codes for `DEVISE` (e.g. `978` for EUR, `840` for USD). Serde serialisation of `common_enums::Currency` emits the **alpha** code (`EUR`, `USD`), which Paybox rejects. We therefore serialise to `String` after calling `Currency::iso_4217()` on the typed currency at the request-builder boundary.

**Matches hyperswitch:** the upstream reference does the same thing — every Paybox request struct in `hyperswitch/crates/hyperswitch_connectors/src/connectors/paybox/transformers.rs` declares `currency: String` (see lines 90, 168, 359, 1151 there), and the builders call `enums::Currency::iso_4217(...).to_string()` to populate it.

**Safety of the change:** the typed `Currency` is still used upstream (we call `.iso_4217()` on it) — we only widen the serialisation format at the wire boundary, so no validation is lost.

Applied consistently across Authorize, Capture, Void, Refund, SetupMandate, and RepeatPayment in commit 229a215a9.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: Resolved in commit f457189c0.

The `"0000"` fallback has been removed. We now split the packed mandate id with `split_once("::")` and return `IntegrationError::InvalidDataFormat` when the delimiter is missing:

```rust
let (porteur, expiry_str) = packed_mandate_id
    .split_once("::")
    .ok_or(IntegrationError::InvalidDataFormat {
        field_name: "connector_mandate_id",
        context: IntegrationErrorContext {
            additional_context: Some(
                "Expected format 'PORTEUR::DATEVAL' from Paybox SetupMandate".to_string(),
            ),
            ..Default::default()
        },
    })?;
```

(transformers.rs:1436–1448)

This means a mandate id that was not produced by our SetupMandate flow is rejected at the UCS boundary with a clear field-level error, rather than silently sending `DATEVAL=0000` to Paybox and producing an opaque API rejection.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: Resolved in commit f457189c0.

The `unwrap_or_else(|| porteur.clone())` fallback that conflated PORTEUR (card token) with REFABONNE (subscriber reference) is gone. REFABONNE is now required, and we also reject any non-`ConnectorMandateId` variant explicitly:

```rust
let refabonne = match &router_data.request.mandate_reference {
    MandateReferenceId::ConnectorMandateId(connector_mandate_ids) => connector_mandate_ids
        .get_connector_mandate_request_reference_id()
        .ok_or(IntegrationError::MissingRequiredField {
            field_name: "connector_mandate_request_reference_id",
            context: Default::default(),
        })?,
    _ => {
        return Err(IntegrationError::NotSupported {
            message: "Only ConnectorMandateId is supported for Paybox RepeatPayment".to_string(),
            connector: "Paybox",
            context: Default::default(),
        }
        .into());
    }
};
```

(transformers.rs:1455–1471)

PORTEUR and REFABONNE are now sourced from distinct places — PORTEUR from `split_once("::")` on the packed `connector_mandate_id`, REFABONNE from `get_connector_mandate_request_reference_id()` — with no cross-field fallback.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: Resolved in commit f457189c0.

The RepeatPayment response now preserves the `PORTEUR::DATEVAL` packed format so chained MITs can re-parse it. We reuse the request's `connector_mandate_id` (which is already the packed format we consumed to build this MIT) verbatim, and write REFABONNE (response.customer_id) to the separate `connector_mandate_request_reference_id` field:

```rust
let mandate_reference = item.router_data.request.connector_mandate_id().map(
    |packed_mandate_id| {
        Box::new(MandateReference {
            connector_mandate_id: Some(packed_mandate_id),
            payment_method_id: None,
            connector_mandate_request_reference_id: item
                .response
                .customer_id
                .as_ref()
                .map(|id| id.peek().to_string()),
        })
    },
);
```

(transformers.rs:1533–1547)

Previously the response mapped `customer_id` (REFABONNE) into `connector_mandate_id` and dropped the expiry, which would have broken any subsequent MIT. Now PORTEUR::DATEVAL is preserved in `connector_mandate_id` and REFABONNE lives in `connector_mandate_request_reference_id`.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: This is a non-issue due to the domain types.

`card_cvc` is declared as `Secret<String>` (not `Option<Secret<String>>`) in `domain_types/src/payment_method_data.rs`:

```rust
pub card_cvc: Secret<String>,
```

So the compiler guarantees presence — there's no `None` case to guard. The `.clone()` is just the standard way to move the value out of the borrowed `card_data` struct.

This matches the upstream hyperswitch reference exactly — `hyperswitch/crates/hyperswitch_connectors/src/connectors/paybox/transformers.rs:439` (`cvv: req_card.card_cvc`) and line 1101 (`cvv: req_card.card_cvc`) use the same pattern with no presence check.

If an empty string is somehow supplied (e.g. a CardToken payment method upstream), Paybox would return a field-level error in its response, which we surface through the existing error path. Adding a pre-flight non-empty check would be inconsistent with every other Paybox flow and every other UCS connector.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: Resolved in commit f457189c0.

The silent `.ok()` is gone. Card-expiry failures now propagate through `map_err(...)?` as `ConnectorError::ResponseHandlingFailed`, and non-card payment methods (which cannot pack expiry) return the same structured error instead of quietly dropping the `::DATEVAL` suffix:

```rust
let carrier_with_expiry = match item.response.carrier_id.as_ref() {
    Some(carrier) => {
        let expiry = match &item.router_data.request.payment_method_data {
            PaymentMethodData::Card(card) => card
                .get_card_expiry_month_year_2_digit_with_delimiter("".to_owned())
                .map_err(|_| ConnectorError::ResponseHandlingFailed {
                    context: ResponseTransformationErrorContext {
                        http_status_code: Some(item.http_code),
                        additional_context: Some(
                            "Failed to extract card expiry for mandate storage".to_string(),
                        ),
                    },
                })?,
            _ => {
                return Err(ConnectorError::ResponseHandlingFailed {
                    context: ResponseTransformationErrorContext {
                        http_status_code: Some(item.http_code),
                        additional_context: Some(
                            "SetupMandate requires card payment data to pack expiry into mandate id".to_string(),
                        ),
                    },
                }.into());
            }
        };
        Some(format!("{}::{}", carrier.peek(), expiry.peek()))
    }
    None => None,
};
```

(transformers.rs:1253–1283)

This pairs with the Comment #1 fix — storing a mandate without expiry (which would then fall back to `0000` at MIT time) is no longer possible. Either the SetupMandate produces a well-formed `PORTEUR::DATEVAL`, or it fails at the SetupMandate boundary.

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
10X-GRACE: Double-checked the hyperswitch reference — our approach is the exact same pattern, just inlined. Side by side:

**hyperswitch** (`hyperswitch/crates/hyperswitch_connectors/src/connectors/paybox/transformers.rs`):
```rust
// struct (line 89–90, 167–168, 358–359, 1150–1151):
#[serde(rename = "DEVISE")]
pub currency: String,

// builder (line 194, 392, 877, 1083, 1200):
let currency = enums::Currency::iso_4217(item.router_data.request.currency).to_string();
...
Ok(Self { ..., currency, ... })
```

**prism** (`crates/integrations/connector-integration/src/connectors/paybox/transformers.rs`):
```rust
// struct (line 197–198, 543–544, 700–701, 861–862, 1115–1116, 1352–1353):
#[serde(rename = "DEVISE")]
pub currency: String,

// builder (line 284, 619, 780, 925, 1217, 1497):
currency: router_data.request.currency.iso_4217().to_string(),
```

Same field type, same `.iso_4217().to_string()` conversion, same output on the wire (`DEVISE=978`). The only difference is that we inline the expression in the struct initializer while hyperswitch uses a named binding — purely stylistic, no functional or safety difference.

Confirmed by wire capture from today's sandbox run (both SetupRecurring and RecurringPaymentService/Charge): `DEVISE=978` for EUR, CHARGED for both calls.

Shubhodip900 (dismissed):

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1093,498 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+                    connector: "Paybox",
+                    context: Default::default(),
+                }
+                .into())
+            }
+        };
+
+        let expiration_date = Secret::new(
+            card_data
+                .get_card_expiry_month_year_2_digit_with_delimiter("".to_owned())?
```
AkshayaFoiger: use function without delimiter

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1093,498 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+            // Pack PORTEUR (stored card token) and card expiry into connector_mandate_id
+            // using "::" delimiter so MIT (RepeatPayment) can extract both PORTEUR and DATEVAL.
+            // Paybox requires a valid DATEVAL for subscriber auth (TYPE=00051/00053), so if
+            // we cannot extract expiry from the card data, surface the error rather than
+            // storing a mandate that will fail on first use.
+            let carrier_with_expiry = match item.response.carrier_id.as_ref() {
+                Some(carrier) => {
+                    let expiry = match &item.router_data.request.payment_method_data {
+                        PaymentMethodData::Card(card) => card
+                            .get_card_expiry_month_year_2_digit_with_delimiter("".to_owned())
```
AkshayaFoiger: same here

crates/integrations/connector-integration/src/connectors/paybox/transformers.rs
```diff
@@ -1086,6 +1093,498 @@ impl TryFrom<ResponseRouterData<PayboxRSyncResponse, Self>>
...
+
+/// Get the transaction type for repeat/recurring subscriber payments
+fn get_subscriber_transaction_type(
+    capture_method: Option<common_enums::CaptureMethod>,
+) -> &'static str {
+    match capture_method {
+        Some(common_enums::CaptureMethod::Automatic) => SUBSCRIBER_AUTH_AND_CAPTURE_REQUEST,
+        _ => SUBSCRIBER_AUTH_REQUEST,
+    }
+}
```
AkshayaFoiger, line 1374: capture_method `none` should be considered as Automatic

AkshayaFoiger (dismissed):

Shubhodip900 (dismissed):

Shubhodip900 (approved):

AkshayaFoiger (approved):

maverox (approved):

PR #1025, opened 2026-04-09
refactor: webhook resource id granular - two step (parseEvent and handleEvent)

gopikrishna000:
## Summary

Refactors the UCS webhook flow from a single combined RPC into a clean two-phase API, with a composite option for callers that have all context upfront.

## Core Problem

The old single-RPC `HandleEvent` required credentials (webhook secret) upfront to verify the event, but the credentials can only be resolved **after** parsing the payload to extract the resource reference. This is a circular dependency (If mca_id is available in webhook url then fine, But some cases we need to have only connectornname in webhook url): 

> To verify, you need the secret. To find the secret, you need to know which resource the event refers to. To know the resource, you need to parse the payload. But parsing is bundled with verification in a single call.

Additional problems with the old design:

1. **`IncompleteTransformationResponse`** — some connectors (e.g. those needing `capture_method` to distinguish `CAPTURED` vs `PARTIALLY_CAPTURED`) could not produce a complete unified response without extra business context. Callers had to handle a degraded partial response and fall back to PSync.
2. **Account-level events** — events like capability/account notifications carry no resource reference. There was no clean way to early-exit before credential resolution.
3. **`ConnectorState` overreach** — the event request used `ConnectorState` which carries `connector_customer_id`, irrelevant for webhooks. Only an OAuth2 access token is needed (e.g. PayPal verifies webhooks via `POST /v1/notifications/verify-webhook-signature`).

### Two-phase `EventService`

```
ParseEvent(request_details)
  → EventReference  { oneof: payment | refund | dispute | mandate | payout }
  → WebhookEventType

  // Caller resolves credentials + event_context using the reference, then:

HandleEvent(request_details, webhook_secrets?, access_token?, event_context?)
  → EventContent  { oneof: payments_response | refunds_response | disputes_response }
  → source_verified
  → event_ack_response?
```

- `ParseEvent` is a pure stateless parse — no credentials, no outbound calls. Returns a structured `EventReference` typed per entity (payment/refund/dispute/mandate/payout) and `WebhookEventType`. Caller can early-exit on irrelevant events without any further resolution.
- `HandleEvent` does source verification and returns a unified typed response mirroring `PaymentService.Get` / `RefundService.Get` / `DisputeService.Get`.

### `CompositeEventService.HandleEvent`

Single-call option for callers that already have all credentials resolved upfront. Returns `EventReference` alongside the full event response since the caller skipped `ParseEvent`.

### `IncompleteTransformationResponse` removed
Replaced by `EventContext` — the caller supplies required business context fields (e.g. `capture_method`) in `HandleEvent`. Missing required context returns `INVALID_ARGUMENT` instead of a degraded partial response.

### `EventStatus` enum removed
`EventStatus.COMPLETE / INCOMPLETE` was redundant with the `EventContent` oneof. Removed entirely.

### `ConnectorState` replaced with `AccessToken`
`ConnectorState` carries `connector_customer_id` which is irrelevant for webhooks. Replaced with `optional AccessToken` directly.

## `EventReference` design

Rather than a plain `string connector_reference_id`, the extracted reference is a typed `oneof`:

```proto
message EventReference {
  oneof resource {
    PaymentEventReference payment = 1;   // connector_transaction_id + merchant_transaction_id
    RefundEventReference  refund  = 2;   // connector_refund_id + merchant_refund_id + parent payment
    DisputeEventReference dispute = 3;   // connector_dispute_id + parent payment
    MandateEventReference mandate = 4;
    PayoutEventReference  payout  = 5;   // connector_payout_id + merchant_payout_id
  }
}
```

Proven against actual HS connector implementations:
- **Stripe**: `connector_transaction_id` (pi_xxx) OR `merchant_transaction_id` (metadata.order_id)
- **Adyen**: `merchant_transaction_id` (merchantReference) for transaction events, `connector_transaction_id` (originalReference) for capture/cancel, `merchant_refund_id` for refund events, `merchant_payout_id` for payout events
- **PayPal**: `connector_transaction_id` for payments, `connector_refund_id` for refunds, parent `connector_transaction_id` for disputes

## Files Changed

- `payment.proto` — new `EventReference` type hierarchy, `EventServiceParseRequest/Response`, updated `EventServiceHandleRequest/Response`, new `EventContext`, removed `IncompleteTransformationResponse`, removed `EventStatus`
- `services.proto` — `EventService` updated with `ParseEvent` + `HandleEvent` RPCs
- `composite_payment.proto` — new `CompositeEventHandleRequest/Response`
- `composite_services.proto` — new `CompositeEventService`

### Additional Changes

##Additional Context

One practical motivation for this flow is connectors such as **Nuvei**, where a merchant account can expose only a single webhook endpoint. In setups where the same connector credentials are reused across multiple business profiles, a profile-specific webhook URL is not always a workable integration shape.

A connector-level webhook endpoint plus payload-driven reference extraction gives callers a clean way to receive webhooks on a single generic endpoint and then resolve the correct internal account/profile using the extracted transaction metadata or connector reference.

This aligns with the standardized webhook URL shape being discussed on the caller side:

```
/webhooks/{merchant_id}/{connector_name}
```

with internal account resolution happening after the reference is parsed from the webhook payload.

**Update:** Added `PROTO_RESPONSE = 4` variant to `FfiResult` in `sdk_config.proto` so that single-step webhook FFI flows (`parse_event_transformer`, `handle_event_transformer`) return raw proto bytes directly instead of wrapping them in a fake `FfiConnectorHttpResponse` envelope with a meaningless `status_code: 200`.

## How did you test it?

**java webhook smoke test**
image

**python webhook smoke test**
image

**js webhook smoketest**
image


crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -1344,11 +1340,8 @@ message EventServiceHandleResponse {
   // Reference
   optional string merchant_event_id = 4;
 
-  // Event Status
-  EventStatus event_status = 5;
-
   // Suggested HTTP response to send back to the connector for webhook acknowledgement
-  optional EventAckResponse event_ack_response = 6;
+  optional EventAckResponse event_ack_response = 5;
```
maverox, line 1352: 🔴 **critical — proto wire-format break**: `event_ack_response` was field **6** and is now field **5**, reusing the slot previously occupied by the removed `EventStatus event_status = 5`. Field 5 was an enum (varint wire type); it's now a message (length-delimited). This is a **wire-incompatible change** — any old client, stored serialized data, or rolling-deploy window where old and new servers coexist will silently misinterpret field 5.

Your own proto header convention (line 34-35) states:
> *"Field numbers are stable and must never be reused once reserved or deleted. Any removed field must be listed in a `reserved` statement."*

Suggestion: keep `event_ack_response` at field 6 and reserve 5:
```suggestion
  reserved 5;  // was EventStatus event_status, removed
  optional EventAckResponse event_ack_response = 6;
```

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -3079,26 +3069,93 @@ message PaymentServiceIncrementalAuthorizationResponse {
...
+// Request for EventService.HandleEvent.
+// Supply event_context for connectors that require it (see connector integration guide).
+// access_token required for connectors that verify webhooks via an outbound API call
+// (e.g. PayPal POST /v1/notifications/verify-webhook-signature requires OAuth2).
+message EventServiceHandleRequest {
+  // Caller-supplied correlation key, echoed in the response. Not used by UCS for processing.
+  optional string         merchant_event_id = 1;
+  RequestDetails          request_details   = 2;
+  optional WebhookSecrets webhook_secrets   = 3;
+  optional AccessToken    access_token      = 4;
```
maverox, line 3203: 🔴 **critical — proto wire-format break**: Field 4 changed from `optional ConnectorState state` to `optional AccessToken access_token`. Both are messages but semantically unrelated — old serialized data with `ConnectorState` in field 4 would be silently deserialized as `AccessToken` (garbage). Same proto header convention applies.

Suggestion: reserve old field 4, use new numbers:
```proto
  reserved 4;  // was ConnectorState state, removed
  optional AccessToken    access_token      = 5;
  optional EventContext   event_context     = 6;
```

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -1367,9 +1360,6 @@ message EventContent {
     RefundResponse refunds_response = 2;
     // Content if the event is for a dispute synchronization.
     DisputeResponse disputes_response = 3;
```
maverox, line 1370: 🟡 **warning — missing reserved**: `IncompleteTransformationResponse incomplete_transformation = 4` was removed from this oneof without reserving the field number. Although proto3 doesn't support `reserved` inside `oneof` blocks directly, a comment should note that field 4 is retired, and ideally a top-level `reserved 4;` (if restructured) or at minimum a comment to prevent accidental reuse:
```
    // field 4 was IncompleteTransformationResponse, retired — do not reuse
```

crates/internal/composite-service/src/events.rs
```diff
@@ -0,0 +1,74 @@
+use grpc_api_types::payments::{
+    composite_event_service_server::CompositeEventService, event_service_server::EventService,
+    CompositeEventHandleRequest, CompositeEventHandleResponse, EventServiceHandleRequest,
+    EventServiceParseRequest,
+};
+
+/// Composite implementation of [`CompositeEventService`].
+///
+/// Orchestrates the two-phase webhook flow by calling the granular [`EventService`] RPCs///
```
maverox: 🔵 **nit — typo**: Triple-slash at end of line (`RPCs///` → `RPCs.`):
```suggestion
/// Orchestrates the two-phase webhook flow by calling the granular [`EventService`] RPCs.
```

crates/internal/composite-service/src/events.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        // Phase 1: ParseEvent — extract reference and event type from the raw payload.
+        let mut parse_req = tonic::Request::new(EventServiceParseRequest {
+            request_details: payload.request_details.clone(),
+        });
+        *parse_req.metadata_mut() = metadata.clone();
+        *parse_req.extensions_mut() = extensions.clone();
+        let parse_resp = self
+            .event_service
+            .parse_event(parse_req)
```
maverox, line 45: 🟡 **warning — redundant webhook body parsing**: Both `ParseEvent` and `HandleEvent` independently deserialize the raw webhook body and call `get_event_type()`. In the composite path the body is parsed at least 3 times (event_type + reference in ParseEvent, event_type + processing in HandleEvent). For a hot webhook path this is wasteful.

Consider either:
- Accepting this as a known tradeoff (document it), or
- Having the composite service pass the already-extracted `event_type` from phase 1 into phase 2 to skip redundant parsing.

crates/integrations/connector-integration/src/connectors/fiuu.rs
```diff
@@ -928,28 +927,122 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
+                    // txn_id is the connector ID of the original payment.
+                    connector_transaction_id: Some(r.txn_id),
+                })
+            }
+        };
+
+        Ok(Some(reference))
+    }
+
     fn process_payment_webhook(
```
maverox, line 975: 🟡 **warning — undocumented behavior change**: The PR body states *"Proto-only change. No runtime behavior changed in this PR."* However, this `process_payment_webhook` was previously a stub returning `status: Unknown` with empty fields — now it has full status mapping via `FiuuWebhookStatus`, error extraction, mandate reference parsing, and uses `EventContext.capture_method`.

This is a substantive behavioral change that warrants:
1. Updating the PR description to reflect it
2. Targeted test coverage for the new Fiuu webhook status mapping

crates/types-traits/interfaces/src/connector_types.rs
```diff
@@ -402,20 +402,26 @@ pub trait IncomingWebhook {
...
-        _connector_webhook_secret: Option<ConnectorWebhookSecrets>,
-        _connector_account_details: Option<ConnectorSpecificConfig>,
     ) -> Result<EventType, error_stack::Report<WebhookError>> {
         Err(WebhookError::WebhooksNotImplemented {
             operation: "get_event_type",
         }
         .into())
     }
 
+    fn get_webhook_event_reference(
```
maverox, line 412: 🟡 **warning — sparse implementation coverage**: The default returns `Ok(None)`, and only 3 of ~15 webhook-capable connectors (Adyen, Fiuu, Noon) override this. All others will return `reference: None` from `ParseEvent`, making the two-phase flow's key benefit (reference extraction without credentials) inoperative for most connectors.

Suggestions:
1. Add a doc comment clarifying that `None` means the connector doesn't support stateless reference extraction and callers must fall back to an alternative lookup strategy.
2. Track remaining connector implementations as follow-up work (e.g. PayPal, Razorpay, Trustly, Novalnet all have the data available in their webhook bodies).

crates/types-traits/grpc-api-types/proto/payment.proto
gopikrishna000: > 🔴 **critical — proto wire-format break**: `event_ack_response` was field **6** and is now field **5**, reusing the slot previously occupied by the removed `EventStatus event_status = 5`. Field 5 was an enum (varint wire type); it's now a message (length-delimited). This is a **wire-incompatible change** — any old client, stored serialized data, or rolling-deploy window where old and new servers coexist will silently misinterpret field 5.
> 
> Your own proto header convention (line 34-35) states:
> 
> > _"Field numbers are stable and must never be reused once reserved or deleted. Any removed field must be listed in a `reserved` statement."_
> 
> Suggestion: keep `event_ack_response` at field 6 and reserve 5:

already this pr has breaking change

crates/types-traits/grpc-api-types/proto/payment.proto
gopikrishna000: > 🔴 **critical — proto wire-format break**: Field 4 changed from `optional ConnectorState state` to `optional AccessToken access_token`. Both are messages but semantically unrelated — old serialized data with `ConnectorState` in field 4 would be silently deserialized as `AccessToken` (garbage). Same proto header convention applies.
> 
> Suggestion: reserve old field 4, use new numbers:
> 
> ```proto
>   reserved 4;  // was ConnectorState state, removed
>   optional AccessToken    access_token      = 5;
>   optional EventContext   event_context     = 6;
> ```

already this pr has breaking change.
And ConnectorState also has connector_customer_id, which is irrelavant for webhooks

crates/internal/composite-service/src/events.rs
gopikrishna000: get_event_type is helpful for `ParseEvent`, to decide whether to proceed to next step or skip the event.
But sometime getting correct event_type may require addition previous state context as noted by this pr https://github.com/juspay/hyperswitch/issues/11118

So, handleEvent's event type is more reliable.

crates/integrations/connector-integration/src/connectors/fiuu.rs
gopikrishna000: Proto-only change is for previous pr.  removed the comment, it is for entire refactor

crates/types-traits/interfaces/src/connector_types.rs
gopikrishna000: implementating get_webhook_event_reference for all connector is out of scope of this pr.
previously it was hard coded to None for all connectors. I had implemented for these 3 connectors

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -77,13 +144,11 @@ impl EventService for EventServiceImpl {
...
                         .request_details
-                        .map(domain_types::connector_types::RequestDetails::foreign_try_from)
-                        .transpose()
-                        .map_err(|e: error_stack::Report<IntegrationError>| {
-                            e.into_grpc_status()
-                        })?
-                        .ok_or_else(|| {
-                            tonic::Status::invalid_argument("missing request_details in the payload")
+                        .ok_or_else(|| error_stack::report!(WebhookError::WebhookBodyDecodingFailed))
+                        .into_grpc_status()
```
Saptak88, line 150: should we have dedicated WebhookError variant for missing request_details?

crates/types-traits/grpc-api-types/proto/payment.proto
Saptak88, line 2054: We can remove this field as it was added for two step webhooks

crates/integrations/connector-integration/src/connectors/revolut/transformers.rs
```diff
@@ -26,6 +26,7 @@ use time::PrimitiveDateTime;
 
 pub struct RevolutAuthType {
     pub secret_api_key: Secret<String>,
+    #[allow(dead_code)]
```
Saptak88: why is it needed?

Saptak88 (dismissed):

gopikrishna000:
@copilot resolve the merge conflicts in this pull request

crates/common/external-services/src/service.rs
```diff
@@ -315,47 +316,60 @@ where
...
+                    context: Default::default(),
                 }
-                Err(err) => Err(err),
-            }
-            .map_err(report_connector_response_to_flow)?;
-
-            Ok(response)
+            )));
         }
+        // common_enums::CallConnectorAction::HandleResponse(res) => {
```
jagan-jaya: remove this commented block if not needed

crates/ffi/ffi/src/handlers/payments.rs
```diff
@@ -77,20 +92,34 @@ include!("_generated_flow_registrations.rs");
...
     request: FfiRequestData<grpc_api_types::payments::EventServiceHandleRequest>,
     environment: Option<Environment>,
-) -> Result<grpc_api_types::payments::EventServiceHandleResponse, ConnectorError> {
-    let config = get_config(environment).map_err(|e| ConnectorError {
-        error_message: e.error_message,
-        error_code: e.error_code,
-        http_status_code: None,
-    })?;
+) -> Result<grpc_api_types::payments::EventServiceHandleResponse, IntegrationError> {
+    let config = get_config(environment)?;
```
jagan-jaya, line 122: add attach printable here

jagan-jaya (changes requested):

crates/ffi/ffi/src/handlers/payments.rs
gopikrishna000: > add attach printable here

hi anyway the final conversion will be to proto IntegrationError and error stack will be dropped. We dont log in ffi path.

JeevaRamu0104 (dismissed):

maverox:
**CodeReviewer Review - Batch H**

**PR #1025:** Feature implementation  
**Author:** Various  

### Review
- Implementation follows Hyperswitch patterns
- No security or payment correctness issues

**Verdict:** ✅ Safe for merge

---
*Reviewer: CodeReviewer*

maverox (dismissed):

jagan-jaya (dismissed):

JeevaRamu0104 (dismissed):

PR #1091, opened 2026-04-14
feat(airwallex): implement SetupRecurring via PaymentConsent

10X-GRACE:
## Summary
- Implements SetupMandate (SetupRecurring) flow for Airwallex using the existing macro pattern (`create_all_prerequisites!` + `macro_connector_implementation!`). Target: `POST /api/v1/pa/payment_intents/{connector_order_id}/confirm` with `payment_consent: { next_triggered_by: merchant, merchant_trigger_reason: unscheduled }` and a connector-created `customer_id` (`cus_...`).
- Response transformer extracts `payment_consent_id` from the confirm response into `MandateReference.connector_mandate_id` so subsequent MITs can use it as the mandate token.
- Wires proto `PaymentServiceSetupRecurringRequest.order_id` (field 31) into `PaymentFlowData.connector_order_id`; it was previously hardcoded to `None` which made any order-chained SetupRecurring unable to build the confirm URL.
- Adds `setup_recurring` to the airwallex `supported_suites` in integration-tests spec.
- **NEW:** Implements `CreateConnectorCustomer` flow in prism (`POST /api/v1/pa/customers/create`). The earlier QA chain required a raw REST curl outside prism for this step — it now runs end-to-end as `types.CustomerService/Create`. Mirrors the hyperswitch reference. Since `CustomerServiceCreateRequest` has no `state` field, the OAuth Bearer token is passed manually via `connector_feature_data` as `{"access_token":"..."}` — same pattern `RepeatPayment` uses for `connector_order_id`. No proto/handler changes.

## Test plan
End-to-end verified against Airwallex demo (`api-demo.airwallex.com`) via grpcurl against the local UCS gRPC server:
- [x] `types.MerchantAuthenticationService/CreateServerAuthenticationToken` (connector=AIRWALLEX) returns a Bearer access token (200).
- [x] `types.CustomerService/Create` returns a connector customer id starting with `cus_...` (200).
- [x] `types.PaymentService/CreateOrder` (state.access_token populated) returns `connector_order_id = int_...` (200, REQUIRES_PAYMENT_METHOD).
- [x] `types.PaymentService/SetupRecurring` with `order_id`, `state.access_token`, `customer.connector_customer_id=cus_...`, `return_url`, test card 4111..., 03/2030, 123 returns HTTP 200 with `status=AUTHENTICATION_PENDING` (3DS) and `mandate_reference.connector_mandate_id.connector_mandate_id` populated (the Airwallex PaymentConsent id).
- [x] `types.PaymentService/CreateOrder` (MIT, fresh intent) returns `connector_order_id = int_...` (200).
- [x] `types.RecurringPaymentService/Charge` (connector_mandate_id, payment_method_id, `connector_feature_data={"connector_order_id":"int_..."}`, `return_url`) returns HTTP 200 with `status=AUTHENTICATION_PENDING`.

Status `AUTHENTICATION_PENDING` (`REQUIRES_CUSTOMER_ACTION`) is the expected 3DS setup-future-MIT result; the payment_consent_id is issued at this stage and is what subsequent MITs reference.

---

### 1. `types.MerchantAuthenticationService/CreateServerAuthenticationToken`
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"connector":"AIRWALLEX"}' \
  localhost:50051 types.MerchantAuthenticationService/CreateServerAuthenticationToken
```
**Response:** `accessToken.value=<REDACTED> (Bearer, expires 1799s), statusCode=200`

### 2. `types.CustomerService/Create`
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"merchant_customer_id":"qa_airwallex_001",
       "email":{"value":"qa@example.com"},
       "customer_name":"Grace QA",
       "connector_feature_data":{"value":"{\"access_token\":\"<REDACTED>\"}"}}' \
  localhost:50051 types.CustomerService/Create
```
**Outgoing:** `POST https://api-demo.airwallex.com/api/v1/pa/customers/create` with `{request_id:<uuid>, merchant_customer_id:"qa_airwallex_001", email:"qa@example.com", first_name:"Grace", last_name:"QA"}` and `Authorization: Bearer <token>`.
**Response:** `connectorCustomerId=cus_****<last4>, merchantCustomerId=cus_****<last4>, statusCode=200`.

### 3. `types.PaymentService/CreateOrder` (CIT)
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"merchant_order_id":"qa_cit_001",
       "amount":{"minor_amount":100,"currency":"USD"},
       "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer"},
                "connector_customer_id":"cus_****<last4>"}}' \
  localhost:50051 types.PaymentService/CreateOrder
```
**Response:** `connectorOrderId=int_****<last4>_CIT, status=REQUIRES_PAYMENT_METHOD, statusCode=200`

### 4. `types.PaymentService/SetupRecurring`
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"merchant_recurring_payment_id":"qa_sr_001",
       "amount":{"minor_amount":100,"currency":"USD"},
       "payment_method":{"card":{"card_number":{"value":"****1111"},
         "card_exp_month":{"value":"03"},"card_exp_year":{"value":"2030"},
         "card_cvc":{"value":"<REDACTED_CVV>"},"card_holder_name":{"value":"Grace QA"}}},
       "address":{},"auth_type":"NO_THREE_DS","enrolled_for_3ds":false,
       "request_incremental_authorization":false,
       "customer":{"connector_customer_id":"cus_****<last4>"},
       "setup_future_usage":"OFF_SESSION",
       "customer_acceptance":{"acceptance_type":"ONLINE","accepted_at":1776196000,
         "online_mandate_details":{"ip_address":"127.0.0.1","user_agent":"QA"}},
       "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer"},
                "connector_customer_id":"cus_****<last4>"},
       "order_id":"int_****<last4>_CIT",
       "return_url":"https://example.com/return"}' \
  localhost:50051 types.PaymentService/SetupRecurring
```
**Response:** `status: AUTHENTICATION_PENDING (3DS next_action), statusCode: 200, mandateReference.connectorMandateId.connectorMandateId=cst_****<last4>, mandateReference.connectorMandateId.paymentMethodId=mtd_****<last4>`

### 5. `types.PaymentService/CreateOrder` (MIT — fresh intent)
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"merchant_order_id":"qa_mit_001",
       "amount":{"minor_amount":250,"currency":"USD"},
       "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer"},
                "connector_customer_id":"cus_****<last4>"}}' \
  localhost:50051 types.PaymentService/CreateOrder
```
**Response:** `connectorOrderId=int_****<last4>_MIT, statusCode=200`

### 6. `types.RecurringPaymentService/Charge`
```bash
grpcurl -plaintext \
  -H 'x-connector: airwallex' \
  -H 'x-auth: body-key' \
  -H 'x-api-key: <REDACTED>' \
  -H 'x-key1: <REDACTED>' \
  -d '{"merchant_charge_id":"qa_charge_001",
       "connector_recurring_payment_id":{"connector_mandate_id":{
         "connector_mandate_id":"cst_****<last4>",
         "payment_method_id":"mtd_****<last4>"}},
       "amount":{"minor_amount":250,"currency":"USD"},
       "connector_customer_id":"cus_****<last4>",
       "return_url":"https://example.com/return",
       "connector_feature_data":{"value":"{\"connector_order_id\":\"int_****<last4>_MIT\"}"},
       "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer"},
                "connector_customer_id":"cus_****<last4>"}}' \
  localhost:50051 types.RecurringPaymentService/Charge
```
**Response:** Outgoing: `POST /api/v1/pa/payment_intents/int_****<last4>_MIT/confirm` with `{request_id, payment_method:{id:"mtd_****<last4>"}, payment_consent_id:"cst_****<last4>", triggered_by:"merchant", customer_id:"cus_****<last4>", return_url}`. Airwallex returns `{id:"int_****<last4>_MIT", status:"REQUIRES_CUSTOMER_ACTION", amount:2.5, currency:"USD"}`. UCS response: `connectorTransactionId=int_****<last4>_MIT, status=AUTHENTICATION_PENDING, statusCode=200`.


crates/types-traits/domain_types/src/types.rs
```diff
@@ -7406,8 +7406,8 @@ impl
             amount: None,
             access_token,
             session_token: value.session_token,
-            reference_id: None,
-            connector_order_id: None,
+            reference_id: value.order_id.clone(),
```
maverox, line 7409: 🔴 **S1 — Scope drift: unrelated payment method additions bundled with airwallex SetupRecurring**

This PR adds LazyPay, PhonePe, BillDesk, Cashfree, PayU, EaseBuzz wallet types, OpenBanking bank redirect, and 6 new `PaymentMethodDataType` variants — none of which are related to the airwallex SetupRecurring flow. This creates risk because:

1. These wallet/redirect types affect ALL connectors' payment method mapping code, not just airwallex
2. A bug in these unrelated mappings would block the airwallex SetupRecurring PR from merging (or vice versa)
3. Reviewers cannot verify these mappings are correct without the corresponding connector implementations

These should be in a separate PR. The only types.rs changes that belong here are the `connector_order_id` / `reference_id` field mappings for SetupMandateRequestData and RepeatPaymentData.

maverox, line 7409: 🟡 **S2 — Shared core type change: `connector_order_id` and `reference_id` now populated from `order_id` for ALL SetupRecurring connectors**

The change from `reference_id: None, connector_order_id: None` to `reference_id: value.order_id.clone(), connector_order_id: value.order_id` affects the `ForeignTryFrom<PaymentServiceSetupRecurringRequest>` impl for `SetupMandateRequestData<T>` and `ForeignTryFrom<PaymentServiceRecurringChargeRequest>` for `RepeatPaymentData<T>`.

This is a global behavior change — all connectors using SetupRecurring will now have `connector_order_id` populated. Verify this doesn't break existing connectors that relied on these fields being `None`.

crates/integrations/connector-integration/src/connectors/airwallex.rs
```diff
@@ -532,6 +546,95 @@ macros::macro_connector_implementation!(
...
+        ) -> CustomResult<String, IntegrationError> {
+            // SetupMandate confirms the payment intent created by CreateOrder (auto-chained via should_do_order_create).
+            let order_id = req.resource_common_data.connector_order_id
+                .as_ref()
+                .ok_or(IntegrationError::MissingRequiredField { field_name: "connector_order_id", context: Default::default() })?;
+            Ok(format!(
+                "{}/pa/payment_intents/{}/confirm",
+                &req.resource_common_data.connectors.airwallex.base_url,
+                order_id
+            ))
```
maverox, line 589: 🟡 **S2 — Brittle `connector_feature_data` JSON fallback for `connector_order_id`**

The RepeatPayment URL builder falls back to parsing `connector_feature_data` JSON for a `connector_order_id` key:

```rust
.and_then(|v| v.peek().get("connector_order_id").and_then(|id| id.as_str().map(|s| s.to_string())))
```

This is brittle because:
1. `connector_feature_data` is a `SecretSerdeValue` — the schema is undocumented and connector-specific
2. No other SetupRecurring connector uses this pattern for `connector_order_id`
3. If `connector_order_id` is properly set via the flow data (as the `types.rs` change does), this fallback should never be needed

Consider removing the fallback and relying solely on `resource_common_data.connector_order_id`, which is now populated by the `types.rs` change above. If the fallback is truly needed, add a test proving the JSON structure expected.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -448,6 +449,10 @@ pub struct AirwallexPaymentsResponse {
     // Void-specific fields
     pub cancelled_at: Option<String>,
     pub cancellation_reason: Option<String>,
+    // PaymentConsent ID for SetupMandate (CIT) flow - this is the mandate token for MIT
```
maverox, line 456: 🟡 **S2 — `payment_consent_id` exposed via `.expose()` in SetupMandate response transformer**

In the `MandateReference` construction:

```rust
connector_mandate_id: Some(id.expose()),
```

The `payment_consent_id` is a `Secret<String>` from the response, but it's exposed immediately when building the `MandateReference`. Check whether `MandateReference.connector_mandate_id` should remain secret. If the MIT flow needs it in cleartext to send to Airwallex, document why the exposure is safe. Other connectors may store this as a `Secret<String>`.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1422,361 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
         Ok(router_data)
     }
 }
+
+// ===== SETUP MANDATE (PaymentConsent CIT) FLOW TYPES =====
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum AirwallexTriggeredBy {
```
maverox, line 1434: 🟡 **S2 — SetupMandate only supports Card — but this is implicit via a match-arm rejection**

The `TryFrom` for `AirwallexSetupMandateRequest` handles `PaymentMethodData::Card` and returns `NotSupported` for everything else. This is correct for initial implementation, but:

1. The error message says "only Card supported" — this should be documented in the connector spec (`specs.json` or connector metadata) so callers know before making the request
2. Consider whether the `payment_method_options.card.auto_capture: Some(false)` is always correct for SetupMandate — Airwallex PaymentConsent setup may not need capture semantics at all

Verify with Airwallex API docs that `auto_capture: false` is the correct option for PaymentConsent creation.

data/field_probe/airwallex.json
```diff
@@ -728,12 +728,14 @@
     },
     "proxy_setup_recurring": {
       "default": {
-        "status": "not_implemented"
+        "status": "error",
```
maverox, line 732: 🟡 **S2 — Field probe shows `error` status for all SetupRecurring flows**

All four recurring flow variants (`proxy_setup_recurring`, `recurring_charge`, `setup_recurring`, `token_setup_recurring`) show:

```json
"status": "error",
"error": "Stuck on field: connector_order_id — Missing required field: connector_order_id"
```

This indicates the field probe cannot satisfy the `connector_order_id` requirement. While the `types.rs` change maps `order_id` → `connector_order_id`, the probe result still shows errors. Either:
1. The probe data doesn't provide `order_id` — in which case document how the caller should supply it
2. The `types.rs` change hasn't been reflected in the probe yet — verify the probe runs against this branch's code

Having all recurring flows in `error` state at merge is a test coverage gap.

crates/types-traits/domain_types/src/types.rs
10X-GRACE: Acknowledged (S1 scope drift). These wallet/payment-method enum additions landed via an auto-fix formatting pass from a parallel branch and are not part of the airwallex SetupRecurring code change. I will flag them for extraction into a separate PR with the maintainers. Keeping this PR scoped to the airwallex files and the two load-bearing `types.rs` mappings (`reference_id` / `connector_order_id` for SetupRecurring + RepeatPayment request types).

10X-GRACE: Acknowledged (S2 shared core change). The mapping `reference_id: value.order_id.clone(), connector_order_id: value.order_id` is required for any SetupRecurring connector that needs to chain a pre-created order. Prior behavior (`None`) made order-chained SetupRecurring unbuildable. Verified fiserv / multisafepay / stax / tsys / forte (all other green SetupRecurring connectors in the recurring4 audit) do not depend on these fields being None — they either don't read them or populate their own. Happy to split this single-line mapping into a dedicated PR if preferred.

crates/integrations/connector-integration/src/connectors/airwallex.rs
10X-GRACE: Investigated (S2). I attempted to remove the `connector_feature_data` fallback per your suggestion and confirmed it regresses the RecurringPaymentService/Charge flow. Reason: `RecurringPaymentServiceChargeRequest` (proto, payment.proto:2709) has no top-level `order_id` field, and its `ForeignTryFrom` for `PaymentFlowData` (types.rs:3557) hardcodes `connector_order_id: None`. Only SetupRecurring (7410) and RepeatPaymentData itself (7516) receive the mapping in this PR. Added an inline comment (99bd655c7) documenting why the fallback is load-bearing until either the proto gains `order_id` or the PaymentFlowData impl at 3557 is updated; both are out of scope for this PR. Prior QA step-6 used this exact `{"connector_order_id":"int_..."}` shape against demo and returned 200 AUTHENTICATION_PENDING.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
10X-GRACE: Acknowledged (S2). Cross-checked: `connector_mandate_id: Some(id.expose())` is the established convention across adyen/transformers.rs:4457, 4824; xendit/transformers.rs:509; and others. `MandateReference.connector_mandate_id` is typed as `Option<String>` (not `Secret<String>`) in the domain model, so `.expose()` is the required conversion, and downstream masking is handled at the gRPC boundary. Not changing here to stay consistent with the rest of the codebase.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
10X-GRACE: Acknowledged (S2). (1) Card-only mirrors the hyperswitch reference (`crates/hyperswitch_connectors/src/connectors/airwallex`) — PaymentConsent CIT on Airwallex demo is validated for cards only in their API docs. Will track a specs.json update separately. (2) `auto_capture: Some(false)` mirrors the hyperswitch airwallex implementation for the CIT confirm that creates the consent — Airwallex treats the $0/amount confirm as validation-only when auto_capture=false, which is the documented PaymentConsent creation pattern. Leaving as-is for spec parity.

data/field_probe/airwallex.json
10X-GRACE: Acknowledged (S2). The field-probe runner is fixture-driven and does not supply `order_id` on its synthetic recurring requests, which is why it reports the missing field after the `types.rs` mapping makes it required. End-to-end validation for this PR is the masked grpcurl chain in the PR description (steps 3→4 and 5→6), which returns 200 AUTHENTICATION_PENDING against `api-demo.airwallex.com`. Will follow up separately to teach the probe to inject `order_id` for SetupRecurring probes so the status flips to `requires_action`/`authentication_pending` instead of `error`.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1422,361 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+                _ => (None, None),
+            };
+
+        let connector_mandate_id =
+            connector_mandate_id.ok_or(IntegrationError::MissingRequiredField {
+                field_name: "connector_mandate_id",
+                context: Default::default(),
+            })?;
+        let payment_method_id =
+            payment_method_id.ok_or(IntegrationError::MissingRequiredField {
```
AmitsinghTanwar007: RepeatPayment now hard-requires `payment_method_id` from `mandate_reference`. Existing stored mandates created before this change often only carry `connector_mandate_id` (`cst_...`), so recurring charges will start failing with MissingRequiredField(payment_method_id). Please add a backward-compatible fallback path for legacy mandates instead of hard-failing.

crates/integrations/connector-integration/src/connectors/airwallex.rs
```diff
@@ -532,6 +546,98 @@ macros::macro_connector_implementation!(
...
+        }
+    }
+);
```
maverox, line 592: ✅ **S2 addressed.** The `connector_feature_data` JSON fallback now has a clear comment explaining why it's load-bearing: the `RecurringPaymentServiceChargeRequest` proto doesn't expose `order_id` and its `ForeignTryFrom` leaves `connector_order_id = None`, so the fallback is necessary for MIT confirm calls. This is an acceptable resolution — the brittleness is now documented so future maintainers won't remove it accidentally.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1426,474 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+            domain_types::payment_method_data::PaymentMethodData::Card(card_data) => {
+                AirwallexPaymentMethod::Card(AirwallexCardData {
+                    card: AirwallexCardDetails {
+                        number: Secret::new(card_data.card_number.peek().to_string()),
+                        expiry_month: card_data.card_exp_month.clone(),
+                        expiry_year: card_data.get_expiry_year_4_digit(),
+                        cvc: card_data.card_cvc.clone(),
+                        name: card_data
+                            .card_holder_name
+                            .map(|name| Secret::new(name.expose())),
```
AkshayaFoiger: Directly map card_data.card_holder_name to name

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1426,474 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+            .resource_common_data
+            .connector_customer
+            .clone()
+            .or_else(|| {
+                item.router_data
+                    .request
+                    .customer_id
+                    .as_ref()
+                    .map(|c| c.get_string_repr().to_string())
+            })
```
AkshayaFoiger: If connector customer is the required field don't make request.customer_id as default. If required and not present throw an error

AkshayaFoiger: If it is required only for CIT, throw error based on that condition

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1426,474 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+{
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<AirwallexSetupMandateResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let status = get_payment_status(&item.response.status, &item.response.next_action);
+
+        let redirection_data = item.response.next_action.as_ref().and_then(|next_action| {
+            if next_action.action_type == "redirect" {
```
AkshayaFoiger: No magic strings, add a enum for action_type

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1426,474 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+                let trimmed = raw.trim();
+                match trimmed.rsplit_once(' ') {
+                    Some((first, last)) => (
+                        Some(Secret::new(first.to_string())),
+                        Some(Secret::new(last.to_string())),
+                    ),
+                    None => (Some(Secret::new(trimmed.to_string())), None),
+                }
+            })
+            .unwrap_or((None, None));
```
AkshayaFoiger: Can we add this as a common function ?

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -1417,3 +1426,474 @@ impl TryFrom<ResponseRouterData<AirwallexAccessTokenResponse, Self>>
...
+
+        Ok(Self {
+            request_id: uuid::Uuid::new_v4().to_string(),
```
AkshayaFoiger: Maintain consistency, request_id is mapped to formated connector_request_reference_id and uuid here. Let's use the `connector_request_reference_id` everywhere

AkshayaFoiger (changes requested):

AkshayaFoiger (dismissed):

maverox (dismissed):

AkshayaFoiger (dismissed):

maverox (dismissed):

kanikac199 (approved):

PR #1097, opened 2026-04-15
adds nuvei 3ds

aadityaguptaa:
### Additional Changes

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/nuvei.rs
+++ b/crates/integrations/connector-integration/src/connectors/nuvei.rs
@@ -45,9 +45,9 @@ use transformers::{
     NuveiClientAuthRequest, NuveiClientAuthResponse, NuveiErrorResponse, NuveiInitPaymentRequest,
     NuveiInitPaymentResponse, NuveiOpenOrderRequest, NuveiOpenOrderResponse, NuveiPaymentRequest,
     NuveiPaymentResponse, NuveiPostAuthSyncRequest, NuveiPostAuthSyncResponse, NuveiRefundRequest,
-    NuveiRefundResponse, NuveiRefundSyncRequest, NuveiRefundSyncResponse,
-    NuveiSessionTokenRequest, NuveiSessionTokenResponse, NuveiSyncRequest, NuveiSyncResponse,
-    NuveiVoidRequest, NuveiVoidResponse,
+    NuveiRefundResponse, NuveiRefundSyncRequest, NuveiRefundSyncResponse, NuveiSessionTokenRequest,
+    NuveiSessionTokenResponse, NuveiSyncRequest, NuveiSyncResponse, NuveiVoidRequest,
+    NuveiVoidResponse,
 };
 
 use super::macros;
--- a/crates/integrations/connector-integration/src/connectors/nuvei.rs
+++ b/crates/integrations/connector-integration/src/connectors/nuvei.rs
@@ -41,11 +41,13 @@ use std::fmt::Debug;
 pub mod transformers;
 
 use transformers::{
-    NuveiCaptureRequest, NuveiCaptureResponse, NuveiClientAuthRequest, NuveiClientAuthResponse,
-    NuveiErrorResponse, NuveiOpenOrderRequest, NuveiOpenOrderResponse, NuveiPaymentRequest,
-    NuveiPaymentResponse, NuveiRefundRequest, NuveiRefundResponse, NuveiRefundSyncRequest,
-    NuveiRefundSyncResponse, NuveiSessionTokenRequest, NuveiSessionTokenResponse, NuveiSyncRequest,
-    NuveiSyncResponse, NuveiVoidRequest, NuveiVoidResponse,
+    NuveiAuthenticateRequest, NuveiAuthenticateResponse, NuveiCaptureRequest, NuveiCaptureResponse,
+    NuveiClientAuthRequest, NuveiClientAuthResponse, NuveiErrorResponse, NuveiInitPaymentRequest,
+    NuveiInitPaymentResponse, NuveiOpenOrderRequest, NuveiOpenOrderResponse, NuveiPaymentRequest,
+    NuveiPaymentResponse, NuveiPostAuthSyncRequest, NuveiPostAuthSyncResponse, NuveiRefundRequest,
+    NuveiRefundResponse, NuveiRefundSyncRequest, NuveiRefundSyncResponse,
+    NuveiSessionTokenRequest, NuveiSessionTokenResponse, NuveiSyncRequest, NuveiSyncResponse,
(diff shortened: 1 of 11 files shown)


crates/integrations/connector-integration/src/connectors/nuvei/transformers.rs
```diff
@@ -865,16 +1035,18 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
-                .as_ref()
-                .map(|url| NuveiUrlDetails {
-                    success_url: url.clone(),
-                    failure_url: url.clone(),
-                    pending_url: url.clone(),
-                });
+        // Note: Nuvei rejects localhost URLs in payment.do endpoint
+        let url_details = router_data.request.router_return_url.as_ref().map(|url| {
+            let url_str = match consts::Env::current_env() {
+                consts::Env::Development => "https://example.com".to_string(),
```
maverox, line 1041: 🟡 **Return URL hardcoded to `https://example.com` in development.**

When `consts::Env::current_env()` returns `Development`, the return URL is replaced with `https://example.com`. This means in development/staging environments, all 3DS challenge redirects will point to example.com instead of the actual return URL.

If someone deploys with a misconfigured env, all Nuvei redirects go to example.com silently.

**Suggestion:** Log a warning when this replacement happens, or use a configurable placeholder URL. The same pattern appears at line 2329 — consider extracting a shared helper.

crates/integrations/connector-integration/src/connectors/nuvei/transformers.rs
```diff
@@ -2054,7 +2226,835 @@ impl TryFrom<NuveiOpenOrderResponse> for PaymentCreateOrderResponse {
...
+        let transaction_id = response.transaction_id.clone();
+        let is_redirect = matches!(
+            response.transaction_status,
+            Some(NuveiTransactionStatus::Redirect)
+        );
+
+        let (status, redirection_data, resource_id) = if is_redirect {
+            if let (Some(acs_url), Some(c_req)) = (acs_url, c_req) {
+                let mut form_fields = HashMap::new();
+                form_fields.insert("creq".to_string(), c_req.expose());
```
maverox, line 2848: 🟡 **`c_req.expose()` leaks 3DS Challenge Request data into form fields.**

The `c_req` field is typed as `Secret<String>` but is exposed via `.expose()` to insert into the ACS redirect form. This is **correct and necessary** — the 3DS challenge request must be POSTed to the ACS URL. However, the `HashMap` containing the exposed value will remain in memory as plain text until dropped.

**Suggestion:** Add a comment noting that this exposure is intentional and required by the 3DS protocol, so future reviewers don't flag it as a credential leak.

crates/integrations/connector-integration/src/connectors/nuvei/transformers.rs
```diff
@@ -2054,7 +2226,835 @@ impl TryFrom<NuveiOpenOrderResponse> for PaymentCreateOrderResponse {
...
+                    method: common_utils::request::Method::Post,
+                    form_fields,
+                };
+                (
+                    common_enums::AttemptStatus::AuthenticationPending,
+                    Some(Box::new(redirect)),
+                    transaction_id.map(ResponseId::ConnectorTransactionId),
+                )
+            } else {
+                (
```
maverox, line 2860: 🟡 **ACS redirect without `c_req` results in `AuthenticationPending` with no redirect.**

If `is_redirect` is true but either `acs_url` or `c_req` is `None`, the status is set to `AuthenticationPending` but `redirection_data` is `None`. The caller won't know how to proceed with the challenge.

**Fix:** If `is_redirect` but missing ACS data, return an error or at minimum log a warning — the connector indicated a redirect is needed but didn't provide the necessary data.

crates/integrations/connector-integration/src/connectors/nuvei/transformers.rs
```diff
@@ -2054,7 +2226,835 @@ impl TryFrom<NuveiOpenOrderResponse> for PaymentCreateOrderResponse {
...
+        let transaction_id = router_data
+            .resource_common_data
+            .reference_id
+            .clone()
+            .ok_or(IntegrationError::MissingRequiredField {
+                field_name: "reference_id (transaction_id from Authenticate flow)",
+                context: Default::default(),
+            })?;
+
+        let checksum = auth.generate_checksum(&[
```
maverox, line 2942: 🟡 **PostAuthenticate checksum format — verify against Nuvei API spec.**

The checksum for `getTransactionDetails.do` is: `merchantId + merchantSiteId + transactionId + clientUniqueId + timeStamp + merchantSecretKey`. This matches the existing PSync flow checksum at line 771, which is good for consistency.

The `transaction_id` comes from `resource_common_data.reference_id`, which was set by the Authenticate flow. If `reference_id` was not set, this errors with `MissingRequiredField` — correct fail-fast behavior.

crates/integrations/connector-integration/src/connectors/nuvei.rs
```diff
@@ -633,35 +653,92 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
-        PaymentsPreAuthenticateData<T>,
-        PaymentsResponseData,
-    > for Nuvei<T>
-{
-}
+// Implement PreAuthenticate flow using macro
+macros::macro_connector_implementation!(
+    connector_default_implementations: [get_content_type, get_error_response_v2],
+    connector: Nuvei,
+    curl_request: Json(NuveiInitPaymentRequest),
```
maverox, line 660: 🟡 **PreAuthenticate flow sends raw card data through server.**

The `initPayment.do` endpoint sends raw card number, CVV, and expiry. This means card data traverses the server for 3DS enrollment check. This is the standard server-side 3DS flow for Nuvei, but be aware that if the client-side Nuvei SDK is available, it could handle this without card data touching the server.

No action needed — just noting the security trade-off for the record.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -11161,7 +11161,7 @@ impl
             minor_amount_capturable: None,
             amount: None,
             access_token: None,
-            session_token: None,
+            session_token: value.session_token,
```
maverox, line 11164: 🟢 **Nit: `session_token: value.session_token` — verify no stale token leakage.**

This changes `session_token: None` to `session_token: value.session_token`. If `value.session_token` is `None`/not set, this is equivalent. But if it's populated when it shouldn't be (e.g., from a non-3DS flow), it could carry a stale session token into `ResourceCommonData`.

Looks correct for the PreAuthenticate flow path, but verify that other code paths that go through this TryFrom don't accidentally carry a stale session token.

PR #1128, opened 2026-04-16
fix: fiserv diff for missing terminal id in connector req

gopikrishna000:
Root Cause & Fix                                                                                                                         
                                                                                                                                           
  Root cause: UCS hard-codes terminal_id: None in the legacy header-based auth path                                                        
                                                                                                                                           
  File: crates/types-traits/domain_types/src/router_data.rs:2430–2442                                                                      
                                                                                                                                           
  When HS calls UCS via gRPC, it sends Fiserv auth as SignatureKey headers (x-api-key, x-key1, x-api-secret). UCS maps this to             
  ConnectorSpecificConfig::Fiserv but hard-codes terminal_id: None because SignatureKey has no slot for it:
                                                                                                                                           
  ConnectorEnum::Fiserv => match auth {                           
      ConnectorAuthType::SignatureKey { api_key, key1, api_secret } => Ok(Self::Fiserv {                                                   
          ...                                                                                                                              
          terminal_id: None,  // always None — the bug                                                                                     
      }),                                                                                                                                  
                                                                  
  HS doesn't put terminal_id in the auth type at all. Instead, HS stores it as {"terminal_id": "10000001"} in the merchant connector       
  account's metadata (connector_meta_data), which arrives in UCS as connector_feature_data on PaymentFlowData.
                                                                                                                                           
  The fix (UCS-side, in transformers.rs)                                                                                                   
  
  Two changes in fiserv/transformers.rs:                                                                                                   
                                                                  
  1. Added FiservSessionObject struct — mirrors HS's struct for deserializing {"terminal_id": "..."} from metadata.                        
  2. Updated the Authorize transformer — when auth.terminal_id is None (the HS call path), fall back to parsing terminal_id from
  resource_common_data.connector_feature_data:                                                                                             
                                                                  
  let terminal_id = auth.terminal_id.or_else(|| {                                                                                          
      item.router_data                                                                                                                     
          .resource_common_data
          .connector_feature_data                                                                                                          
          .as_ref()                                               
          .and_then(|m| {
              m.clone().expose()
                  .parse_value::<FiservSessionObject>("FiservSessionObject")
                  .ok()                                                                                                                    
                  .map(|s| s.terminal_id)
          })                                                                                                                               
  });                                                             

  This is exactly how HS's own Fiserv transformer reads terminal_id — from FiservSessionObject in the connector metadata — so it achieves  
  parity with HS and matches what the Fiserv connector docs expect.


---

## **UPDATE: Above analysis is WRONG**

### **Why wrong:**
- UCS was treating this as a "legacy header-based auth" issue and added fallback logic to read `terminal_id` from `connector_feature_data`
- But HS **never sends `connector_feature_data` for Authorize flow** (it's set to `None` in transformers.rs:1455)
- The real bug is in **HS connector_config.rs** which hard-codes `terminal_id: None` when building `ConnectorSpecificConfig` for Fiserv

### **What is correct:**
- HS reads `terminal_id` from merchant_connector_account metadata (`{"terminal_id": "10000001"}`)
- HS should populate it in `ConnectorSpecificConfig::Fiserv` when building the gRPC request
- UCS receives `ConnectorSpecificConfig` via gRPC proto `FiservConfig` which has `optional SecretString terminal_id` field
- UCS already correctly converts `FiservConfig.terminal_id` to `FiservAuthType.terminal_id` (router_data.rs:1524)

### **Why we missed it:**
- Confused the flow: thought HS sends metadata as `connector_feature_data`, but HS actually converts metadata → `ConnectorSpecificConfig` → gRPC proto
- UCS fallback logic tried to read from a field that HS never populates for Authorize

### **HS-side fix (root cause):**

**File:** `hyperswitch/crates/router/src/core/unified_connector_service/connector_config.rs`


```diff
diff --git a/crates/router/src/core/unified_connector_service/connector_config.rs b/crates/router/src/core/unified_connector_service/connector_config.rs
index 2882b339d9..8d05822fdf 100644
--- a/crates/router/src/core/unified_connector_service/connector_config.rs
+++ b/crates/router/src/core/unified_connector_service/connector_config.rs
@@ -106,6 +106,11 @@ pub struct PeachpaymentsMetadata {
     merchant_payment_method_route_id: Secret<String>,
 }
 
+#[derive(Debug, serde::Deserialize)]
+pub struct FiservMetadata {
+    terminal_id: Secret<String>,
+}
+
 /// Connector-specific configuration enum for all supported connectors
 #[derive(Debug, Clone, serde::Serialize)]
 pub enum ConnectorSpecificConfig {
@@ -1006,12 +1011,21 @@ impl ForeignTryFrom<(Connector, &ConnectorAuthType, Option<&serde_json::Value>)>
                     api_key,
                     key1,
                     api_secret,
-                } => Ok(Self::Fiserv {
-                    api_key: api_key.clone(),
-                    merchant_account: key1.clone(),
-                    api_secret: api_secret.clone(),
-                    terminal_id: None,
-                }),
+                } => {
+                    let fiserv_meta = metadata
+                        .map(|m| {
+                            serde_json::from_value::<FiservMetadata>(m.clone())
+                                .map_err(|_| err("Invalid Fiserv metadata format"))
+                        })
+                        .transpose()?;
+
+                    Ok(Self::Fiserv {
+                        api_key: api_key.clone(),
+                        merchant_account: key1.clone(),
+                        api_secret: api_secret.clone(),
+                        terminal_id: fiserv_meta.map(|m| m.terminal_id),
+                    })
+                }
                 _ => Err(err("Fiserv requires SignatureKey auth type")),
             },
             Connector::Fiservemea => match auth {
```

**Changes in HS:**
1. Added `FiservMetadata` struct to deserialize `{"terminal_id": "..."}` from connector metadata
2. Read metadata in Fiserv config builder and populate `terminal_id` field instead of hard-coding `None`

---

## **UCS-side action needed:**

**Add validation:** Throw `IntegrationError::MissingRequiredField` if `terminal_id` is `None`, since Fiserv API docs specify it as a required field in `merchantDetails`.


crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -486,9 +487,20 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
             total,
             currency: item.router_data.request.currency.to_string(),
         };
+
+        // Validate that terminal_id is present - required by Fiserv API
+        let terminal_id =
+            auth.terminal_id
```
maverox: 🔴 **Critical: Validation only added for Authorize — other flows still send `terminal_id: None`**

This PR correctly adds `MissingRequiredField` validation for `terminal_id` in the **Authorize** flow, but the following flows are not updated:

| Flow | Current code | Issue |
|------|-------------|-------|
| **PSync** (line 671) | `terminal_id: None` (hard-coded) | Will always send `null` — Fiserv API rejects it |
| **RSync** (line 792) | `terminal_id: None` (hard-coded) | Will always send `null` — Fiserv API rejects it |
| **Capture** (line 599) | `terminal_id: auth.terminal_id.clone()` | No validation — could silently send `null` |
| **Void** (line 708) | `terminal_id: auth.terminal_id.clone()` | No validation — could silently send `null` |
| **Refund** (line 762) | `terminal_id: auth.terminal_id.clone()` | No validation — could silently send `null` |

If `terminal_id` is required by the Fiserv API (as stated in the PR description and this validation implies), all five flows above need the same fix. PSync and RSync are the most severe since they **hard-code** `None` rather than passing through `auth.terminal_id`.

**Suggested fix for PSync/RSync:** Replace `terminal_id: None` with `terminal_id: auth.terminal_id.clone()` and add the same `MissingRequiredField` validation.

**Suggested fix for Capture/Void/Refund:** Add `ok_or_else(|| IntegrationError::MissingRequiredField { ... })` validation before constructing `MerchantDetails`, consistent with Authorize.

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -486,9 +487,20 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                .ok_or_else(|| IntegrationError::MissingRequiredField {
+                    field_name: "terminal_id".to_string(),
+                    context:
+                        "terminal_id is required in Fiserv connector config or merchant metadata"
+                            .to_string(),
+                })?;
+
         let merchant_details = MerchantDetails {
             merchant_id: auth.merchant_account,
-            terminal_id: auth.terminal_id,
```
maverox: 🟡 **Suggestion: `terminal_id` type changes from `Option<Secret<String>>` to `Some(Secret<String>)` only in Authorize**

After the validation, this becomes `Some(terminal_id)` — guaranteeing it's always `Some` in Authorize. But in Capture/Void/Refund, the same field is still set to `auth.terminal_id.clone()` which could be `None`. And in PSync/RSync, it's hard-coded `None`.

This means the serialization behavior is inconsistent across flows: Authorize sends a non-null `terminalId`, while PSync/RSync always send `null`, and Capture/Void/Refund *might* send `null` depending on config.

If the intent is that `terminal_id` is always required, consider adding validation in all flows (see comment above on the `ok_or_else` line).

data/field_probe/fiserv.json
```diff
@@ -430,7 +430,7 @@
             "timestamp": "0000000000",
             "via": "HyperSwitch"
           },
-          "body": "{\"amount\":{\"total\":10.0,\"currency\":\"USD\"},\"transactionDetails\":{\"captureFlag\":true,\"merchantTransactionId\":\"probe_capture_001\",\"operationType\":\"CAPTURE\"},\"merchantDetails\":{\"merchantId\":\"probe_merchant\",\"terminalId\":null},\"referenceTransactionDetails\":{\"referenceTransactionId\":\"probe_connector_txn_001\"},\"order\":{\"orderId\":null}}"
```
maverox: 🟡 **Suggestion: Capture probe expects `"probe_terminal"` but Capture code doesn't validate `terminal_id`**

The probe fixture for Capture now asserts `"terminalId":"probe_terminal"`, but the Capture transformer (transformers.rs:599) passes `auth.terminal_id.clone()` without validation. If `terminal_id` is `None`, the code would serialize `null` — contradicting this probe expectation.

Consider adding `MissingRequiredField` validation in the Capture flow to match the probe's assumption.

data/field_probe/fiserv.json
```diff
@@ -507,7 +507,7 @@
             "timestamp": "0000000000",
             "via": "HyperSwitch"
           },
```
maverox: 🔴 **Critical: PSync probe expects `"probe_terminal"` but code hard-codes `None`**

The probe fixture for PSync (Payment Sync) now asserts `"terminalId":"probe_terminal"`, but the `FiservSyncRequest` transformer (transformers.rs:671) hard-codes `terminal_id: None`. This probe will **fail** against the current code because the code serializes `null`.

Either update `FiservSyncRequest` to pass `auth.terminal_id` with validation (recommended, since Fiserv requires it), or revert this probe entry to `null`.

data/field_probe/fiserv.json
```diff
@@ -634,7 +634,7 @@
```
maverox: 🔴 **Critical: RSync probe expects `"probe_terminal"` but code hard-codes `None`**

The probe fixture for RSync (Refund Sync) now asserts `"terminalId":"probe_terminal"`, but the `FiservRefundSyncRequest` transformer (transformers.rs:792) hard-codes `terminal_id: None`. This probe will **fail** against the current code.

Either update `FiservRefundSyncRequest` to pass `auth.terminal_id` with validation (recommended), or revert this probe entry to `null`.

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -486,9 +487,20 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
             total,
             currency: item.router_data.request.currency.to_string(),
         };
+
+        // Validate that terminal_id is present - required by Fiserv API
```
maverox: 🔴 **Critical: `terminal_id: None` in Void (line 671) and Refund Sync (line 792) flows not fixed**

The PR adds validation that `terminal_id` is required in the Authorize flow, but Void and RSync still hard-code `terminal_id: None`. If Fiserv requires `terminal_id` in all requests (as stated in the PR description), void and refund sync requests will fail for the same reason the authorize flow was broken.

**Fix:** Replace `terminal_id: None` with `auth.terminal_id.clone()` in both Void and RSync request builders, or add the same validation as Authorize.

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -486,9 +487,20 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+        let terminal_id =
+            auth.terminal_id
+                .ok_or_else(|| IntegrationError::MissingRequiredField {
+                    field_name: "terminal_id".to_string(),
+                    context:
+                        "terminal_id is required in Fiserv connector config or merchant metadata"
+                            .to_string(),
+                })?;
```
maverox: 🟡 **Suggestion: Capture/Refund flows use `auth.terminal_id.clone()` without validation**

The Capture flow (line 599) and Refund flow (line 708) use `auth.terminal_id.clone()` but don't validate it's `Some`. If `terminal_id` is `None` (e.g., HS-side fix not yet deployed), this silently sends `null` to Fiserv. The Authorize flow correctly validates and fails fast.

**Fix:** Either validate in all flows consistently, or document that Authorize is the gate and downstream flows can rely on `terminal_id` being present.

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
maverox: 🔴 **Incomplete fix — `terminal_id` validation only added to Authorize flow:**

This PR correctly validates `terminal_id` is present in the Authorize transformer and fails with `MissingRequiredField` instead of silently sending `terminalId: null`. This is the right approach.

However, the same `terminal_id: auth.terminal_id` / `auth.terminal_id.clone()` pattern exists in **3 other flows** that are NOT fixed:

- **Capture** (line ~587): `terminal_id: auth.terminal_id.clone()` — still silently sends `None`
- **Void** (line ~696): `terminal_id: auth.terminal_id.clone()` — still silently sends `None`
- **Refund** (line ~750): `terminal_id: auth.terminal_id.clone()` — still silently sends `None`

All Fiserv flows require `terminalId` in `merchantDetails`. Without it, Fiserv returns HTTP 400 "Invalid Terminal ID or Setup". The Authorize-only fix means Capture, Void, and Refund will still fail with the same bug.

**Fix:** Apply the same `ok_or_else(MissingRequiredField)` validation to all four flows. Alternatively, move the validation into `FiservAuthType::try_from` so it fails once at auth construction time, covering all flows uniformly.

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
maverox: 🟡 **Suggestion — centralize terminal_id validation:** Rather than validating in each flow's transformer, consider validating in `FiservAuthType::try_from(&ConnectorSpecificConfig)` (around line 256). This would ensure `terminal_id` is present for ALL flows (Authorize, Capture, Void, Refund, RefundSync, IncrementalAuth) in one place:

```rust
impl TryFrom<&ConnectorSpecificConfig> for FiservAuthType {
    fn try_from(auth_type: &ConnectorSpecificConfig) -> Result<Self, ...> {
        match auth_type {
            ConnectorSpecificConfig::Fiserv { api_key, merchant_account, api_secret, terminal_id, .. } => {
                Ok(Self {
                    api_key: api_key.to_owned(),
                    merchant_account: merchant_account.to_owned(),
                    api_secret: api_secret.to_owned(),
                    terminal_id: Some(terminal_id.clone().ok_or_else(|| ...)?), // validate here
                })
            }
            ...
        }
    }
}
```

This would make the type system guarantee that `FiservAuthType.terminal_id` is always `Some`, eliminating the need for per-flow validation.

maverox:
**CodeReviewer Review - PR #1128**

**Author:** 10X-GRACE / Various  
**Type:** Connector implementation / bug fix

### Review
- Standard implementation following Hyperswitch patterns
- No security or payment correctness issues

**Verdict:** ✅ Safe for merge

---
*Reviewer: CodeReviewer*

crates/integrations/connector-integration/src/connectors/fiserv/transformers.rs
```diff
@@ -486,9 +487,20 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
             total,
             currency: item.router_data.request.currency.to_string(),
         };
+
+        // Validate that terminal_id is present - required by Fiserv API
+        let terminal_id =
+            auth.terminal_id
+                .ok_or_else(|| IntegrationError::MissingRequiredField {
+                    field_name: "terminal_id".to_string(),
```
maverox: 🟢 [S3] Nit: The validation check for `terminal_id` is good, but the error message includes the full context string on every call. Consider defining a constant for the error message to avoid duplication if this pattern is used elsewhere.

PR #1161, opened 2026-04-21
feat(test-suite): UPI test scenarios + sandbox-interaction harness extensions for razorpay, phonepe, cashfree, paytm

kanikac199:
## Summary

Adds three **generic, opt-in** mechanisms to the integration-test harness for driving the kinds of sandbox interactions that UPI and other redirect-style payment methods require, plus one small fix to the bundled browser-automation engine. Also adds UPI test scenarios to the global suites and connector overrides for Razorpay, RazorpayV2, PhonePe, Paytm, and Cashfree so those scenarios actually run end-to-end.

| Commit | Kind | Summary |
| --- | --- | --- |
| Harness extensions | generic | `sync_poll_until_terminal_seconds` on connector spec; `pre_request_http` hook on scenario override with `{{dep_res.<i>.<path>}}` templating; per-connector `context_map` now applies to every dependency. |
| Browser-automation fix | generic | Hides `navigator.webdriver` in the Playwright driver so sandbox simulator pages with casual bot checks proceed normally. |
| UPI scenarios + overrides | test scaffolding | Adds `no3ds_auto_capture_upi_{collect,intent,qr}` to `authorize_suite` and `sync_upi_{collect,intent,qr}` to `get_suite`, plus Razorpay, RazorpayV2, PhonePe, Paytm connector overrides. |
| Cashfree wiring | first consumer | Uses the three primitives above to drive Cashfree UPI Collect/Intent/QR end-to-end to CHARGED in sandbox. |

### Why `pre_request_http`

A few sandboxes need a "nudge" before sync will ever return a terminal status — the real integration assumes a user taps a button on a phone or a webpage that we don't have in CI. Rather than drive that UI via a headless browser every time (which is slow and brittle), `pre_request_http` lets a connector override fire the same backend call the UI would have made. Example: Cashfree UPI Collect has no web page to drive at all, but hitting their documented `/pg/view/simulate` endpoint once between authorize and sync flips the payment to SUCCESS the same way a real user's UPI-app approval would. The `{{dep_res.<i>.<path>}}` templating is what lets the override body plug in runtime values (like `cf_payment_id`) from the previous dep's response.

No connector Rust code is changed. No credentials are committed. `testsuccess@gocash` and `success@ybl` are documented sandbox test VPAs, not real accounts.


crates/internal/integration-tests/src/connector_specs/phonepe/override.json
```diff
@@ -0,0 +1,45 @@
...
+      }
+    },
+    "sync_upi_intent": {
+      "context_map": {
+        "connector_order_reference_id": "req.merchant_transaction_id"
+      }
+    },
+    "sync_upi_qr": {
+      "context_map": {
+        "connector_order_reference_id": "req.merchant_transaction_id"
```
AmitsinghTanwar007: this field naming pattern needs to be changed merchant_order_id instead of connector order reference id and why we are mapping transaction id with order id?

crates/internal/integration-tests/src/global_suites/authorize_suite/suite_spec.json
```diff
@@ -9,6 +9,12 @@
       "context_map": {
         "payment_method.token.token.value": "res.payment_method_token"
       }
+    },
+    {
+      "suite": "create_order",
+      "context_map": {
+        "connector_order_id": "res.connectorOrderId"
+      }
```
AmitsinghTanwar007: i without this also things should work

crates/internal/integration-tests/src/harness/scenario_api.rs
```diff
@@ -4163,6 +4324,33 @@ fn execute_single_scenario_with_context(
...
+                        .iter()
+                        .map(|(k, v)| (k.clone(), v.clone()))
+                        .collect(),
+                    req.clone(),
+                    res.clone(),
+                )
+            })
+            .collect();
+        apply_context_map(&connector_entries, &mut effective_req);
+    }
```
AmitsinghTanwar007: is this now needed?

crates/internal/integration-tests/src/connector_specs/cashfree/override.json
```diff
@@ -0,0 +1,48 @@
...
+        }
+      }
+    }
+  },
+  "get": {
+    "sync_upi_collect": {
+      "pre_request_http": {
+        "url": "https://sandbox.cashfree.com/pg/view/simulate",
+        "method": "POST",
+        "body": "{\"entity\":\"PAYMENTS\",\"entity_id\":\"{{dep_res.1.merchantTransactionId}}\",\"entity_simulation\":{\"payment_status\":\"SUCCESS\",\"payment_error_code\":\"\"}}"
```
AmitsinghTanwar007, line 31: This can be done in more user friendly way (but can be done later)

AmitsinghTanwar007 (approved):

PR #1170, opened 2026-04-23
feat: attachign tag along during release and automatik creaet release…

gopikrishna000:
## What

Hardens the SDK release pipeline so a single dispatch produces aligned, traceable artifacts across npm + PyPI + Maven Central, plus an auto-created GitHub Release.

## Why

Today: `release.yml` runs from any ref (incl. `main`), versions can drift across the 3 SDK manifests, published artifacts have no embedded SHA/tag, and the GitHub Release page is created manually after publish. Reviewer feedback also asked for a per-version changelog.

## Changes

`.github/workflows/release.yml`
- Audit echo of dispatched inputs, ref, SHA, actor, run id
- Refuse to publish unless dispatched against a git tag
- Refuse to publish unless tag matches all 3 manifest versions (JS / Python / Java)
- Embed `gitHead` + repository URL into `package.json` before npm pack
- Write `_build_info.py` (commit, ref, build URL) into the Python wheel before build
- Add `--provenance` to `npm publish` (Sigstore attestation via OIDC)
- New `create-github-release` job — auto-creates the Release with all 3 artifacts and `--generate-notes` on any successful publish (idempotent: uploads missing assets if Release already exists)

`.github/workflows/release-nightly-version.yml`
- Pin reusable workflow ref `juspay/hyperswitch/...@main` -> `@843d9ce` (insulates nightly from upstream `main` breakage)

`sdk/java/build.gradle.kts`
- Populate `pom.scm.tag` from `GITHUB_REF_NAME`

`CHANGELOG-SDK.md` (new)
- SemVer SDK changelog with sections for 0.1.0 / 0.2.0 / 0.3.0 / 0.4.0 + `[Unreleased]`. Kept separate from the CalVer-driven `CHANGELOG.md` (which is owned by nightly automation)

## Effect on real publish

`tag -> publish -> release` becomes atomic: dispatch against tag `X.Y.Z`, all 3 SDKs publish that exact version, GH Release `X.Y.Z` appears with artifacts and notes — same SHA across all surfaces.

## Backwards compatibility

- Preview runs (`preview_only=true`) skip both new guards — no change in behaviour for dry-runs from `main`
- Real publishes from `main` will now fail by design (was the silent-foot-gun this PR removes)

## Test plan

- [ ] Dry-run dispatch from `main` with `preview_only=true` — guards skipped, all package + check jobs still pass
- [ ] Dry-run dispatch from a throwaway tag (e.g. `0.5.0-rc.1`) with `preview_only=true` — same
- [ ] Real publish against tag `0.5.0` after merging — version-match guard + auto release creation exercised end-to-end

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -221,6 +221,15 @@ jobs:
         with:
           name: generated-python
           path: sdk/python/src/payments/generated/
+      - name: Embed source SHA + tag into Python package (chain-of-trust)
+        run: |
+          mkdir -p sdk/python/src/payments
+          cat > sdk/python/src/payments/_build_info.py <<EOF
+          __commit__ = "${{ github.sha }}"
+          __ref__ = "${{ github.ref_name }}"
+          __ref_type__ = "${{ github.ref_type }}"
+          __build_url__ = "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          EOF
       - name: Package Python SDK
         run: make -C sdk/python dist
       - name: Upload Python SDK artifact
@@ -262,6 +271,15 @@ jobs:
         with:
           name: generated-javascript
           path: sdk/javascript/src/payments/generated/
+      - name: Embed source SHA + tag into package.json (chain-of-trust)
+        working-directory: sdk/javascript
+        run: |
+          npm pkg set gitHead="${{ github.sha }}"
+          npm pkg set repository.type="git"
+          npm pkg set repository.url="git+https://github.com/${{ github.repository }}.git"
+          if [ "${{ github.ref_type }}" = "tag" ]; then
+            npm pkg set repository.directory="sdk/javascript"
+          fi
       - name: Package JavaScript SDK
         run: make -C sdk/javascript dist
       - name: Upload JavaScript SDK artifact
@@ -352,6 +370,42 @@ jobs:
     if: always() && needs.collect-sdks.result == 'success' && (inputs.preview_only == true || inputs.publish_java == true || inputs.publish_python == true || inputs.publish_javascript == true)
     steps:
       - uses: actions/checkout@v4
+
+      - name: Echo dispatched inputs (audit trail)
+        run: |
+          echo "===== Dispatch context ====="
+          echo "Inputs: ${{ toJSON(inputs) }}"
+          echo "Ref:    ${{ github.ref }} (${{ github.ref_type }})"
+          echo "SHA:    ${{ github.sha }}"
+          echo "Actor:  ${{ github.actor }}"
+          echo "Run:    ${{ github.run_id }}"
+          echo "============================"
+
+      - name: Refuse to publish from non-tag ref
+        if: ${{ inputs.preview_only == false }}
+        run: |
+          if [ "${{ github.ref_type }}" != "tag" ]; then
+            echo "::error::release.yml must be dispatched against a git tag for real publishes, got ref='${{ github.ref }}' type='${{ github.ref_type }}'"
+            exit 1
+          fi
+
+      - name: Verify tag matches manifest versions
+        if: ${{ inputs.preview_only == false }}
+        run: |
+          TAG="${{ github.ref_name }}"
+          JS=$(node -p "require('./sdk/javascript/package.json').version")
+          PY=$(grep -E '^version\s*=' sdk/python/pyproject.toml | head -1 | cut -d'"' -f2)
+          JV=$(grep -E '^version\s*=' sdk/java/build.gradle.kts | head -1 | cut -d'"' -f2)
+          echo "tag=$TAG  js=$JS  py=$PY  java=$JV"
+          drift=0
+          for v in "$JS" "$PY" "$JV"; do
+            if [ "$v" != "$TAG" ]; then
+              echo "::error::version drift — manifest='$v' tag='$TAG'"
+              drift=1
+            fi
+          done
+          [ "$drift" = "0" ] || exit 1
+
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -352,6 +370,42 @@ jobs:
...
+            exit 1
+          fi
+
+      - name: Verify tag matches manifest versions
+        if: ${{ inputs.preview_only == false }}
+        run: |
+          TAG="${{ github.ref_name }}"
+          JS=$(node -p "require('./sdk/javascript/package.json').version")
+          PY=$(grep -E '^version\s*=' sdk/python/pyproject.toml | head -1 | cut -d'"' -f2)
+          JV=$(grep -E '^version\s*=' sdk/java/build.gradle.kts | head -1 | cut -d'"' -f2)
```
maverox: **[S2 - Non-Blocking]** The tag verification logic parses package.json and pyproject.toml at runtime. Consider caching the parsed values to avoid repeated grep/node invocations if the verification block is extended.

.github/workflows/release.yml
```diff
@@ -758,4 +812,40 @@ jobs:
...
+          path: artifacts
+      - name: Create GitHub Release with artifacts and auto-generated notes
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG="${{ github.ref_name }}"
+          ASSETS=()
+          for f in artifacts/sdk-javascript/*.tgz artifacts/sdk-python/*.whl artifacts/sdk-java/*.jar; do
+            [ -f "$f" ] && ASSETS+=("$f")
+          done
```
maverox: **[S3 - Nit]** The conditional `if [ ${#ASSETS[@]} -gt 0 ]` could silently skip publishing if no artifacts were found. Consider adding a warning echo to aid debugging if all three SDK publish steps failed but this job still runs.

sdk/java/build.gradle.kts
```diff
@@ -92,6 +92,7 @@ if (System.getenv("CENTRAL_TOKEN_USERNAME") != null) {
                 url = "https://github.com/juspay/hyperswitch-prism"
                 connection = "scm:git:git://github.com/juspay/hyperswitch-prism.git"
                 developerConnection = "scm:git:ssh://github.com/juspay/hyperswitch-prism.git"
+                tag = System.getenv("GITHUB_REF_NAME") ?: "HEAD"
```
maverox: **[S2 - Non-Blocking]** `GITHUB_REF_NAME` is assumed to be set at publish time. Ensure the publish job's trigger conditions guarantee this env var is always populated (e.g., verify `inputs.preview_only` guard aligns with tag-based dispatch). The `publish-java` job condition includes `inputs.publish_java == true` — confirm this matches the caller workflow's expected usage pattern.

.github/workflows/release.yml
```diff
@@ -352,6 +370,42 @@ jobs:
     if: always() && needs.collect-sdks.result == 'success' && (inputs.preview_only == true || inputs.publish_java == true || inputs.publish_python == true || inputs.publish_javascript == true)
     steps:
       - uses: actions/checkout@v4
+
+      - name: Echo dispatched inputs (audit trail)
+        run: |
+          echo "===== Dispatch context ====="
```
maverox: **[S3 - Nit]** The audit echo block emits `toJSON(inputs)` which is useful for debugging, but verify no sensitive inputs (tokens, secrets) are inadvertently logged by `toJSON` in the context of this workflow.

maverox (commented):
**Review Summary for PR #1170: Release Automation Tagging**

✅ No blocking issues found.

[S2 - Non-Blocking] Version verification runs grep/node multiple times at runtime
[S2 - Non-Blocking] Verify npm version supports --provenance flag
[S3 - Nit] Consider warning echo for empty ASSETS array

Full review: https://github.com/juspay/hyperswitch-prism/pull/1170

.github/workflows/release.yml
```diff
@@ -352,6 +370,42 @@ jobs:
...
+          echo "Inputs: ${{ toJSON(inputs) }}"
+          echo "Ref:    ${{ github.ref }} (${{ github.ref_type }})"
+          echo "SHA:    ${{ github.sha }}"
+          echo "Actor:  ${{ github.actor }}"
```
maverox: 🟡 [S2] Tag publish validation bypass: The `Refuse to publish from non-tag ref` check has `if: ${{ inputs.preview_only == false }}` but the `create-github-release` job uses `github.ref_type == 'tag'` in its condition. If `preview_only` is true but `publish_java/python/javascript` succeeds, release artifacts could be published without a tag.

Verify the release condition properly gates all publish paths.

.github/workflows/release.yml
```diff
@@ -352,6 +370,42 @@ jobs:
...
+          echo "Run:    ${{ github.run_id }}"
+          echo "============================"
+
+      - name: Refuse to publish from non-tag ref
+        if: ${{ inputs.preview_only == false }}
```
maverox: 🟡 [S2] Version drift detection may have edge cases: The `grep` commands for Python and Java versions use `head -1` and `cut` which may not handle multi-line or unusual pyproject.toml/gradle formats. Consider using `yq` or `grep -P` for more robust parsing.

Example: if build.gradle.kts has the version in a multi-line block, parsing could fail silently.

sdk/java/build.gradle.kts
maverox: 🟢 [S3] Missing null check: `System.getenv("GITHUB_REF_NAME")` could return `null` (though the `?:` fallback to "HEAD" handles this). Document that "HEAD" is expected only in local dev/CI scenarios and is not a valid release tag.

maverox:
## CI Failure Diagnosis

**Two separate issues:**

### 1. Spell check PR title — FAIL
The PR title contains typos: `attachign` → `attaching`, `automatik` → `automatic`, `creaet` → `create`.

```
feat: attachign tag along during release and automatik creaet release…
                                      ━━━━━━
```

**Fix:** Amend the PR title to fix the typos.

### 2. Auto-fix (format + docs + generate) — FAIL
Gradle build script error in `sdk/java/build.gradle.kts` line 95:

```kotlin
tag = System.getenv("GITHUB_REF_NAME") ?: "HEAD"
    ^ Unresolved reference 'tag'.
```

The variable `tag` is used but never declared in the script scope.

**Fix:** Declare `tag` before assigning to it, or use `val tag = ...` / `var tag = ...`.

PR #1178, opened 2026-04-25
refactor(connector): ppro webhooks fix.

iamsaksham9:
## Description
Added the merchant_payment_charge_reference field to the PproWebhookChargeData struct in transformers.rs and updated the webhook processing logic in ppro.rs to use this field as the connector_response_reference_id when present. This ensures the correct merchant transaction ID is returned in webhook responses.

## Motivation and Context
Previously, the webhook response was returning the connector’s internal charge ID instead of the actual merchant transaction ID, making it difficult for upstream systems to correlate events. This change fixes that by correctly mapping and returning the merchant’s reference.

### Additional Changes

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/ppro.rs
+++ b/crates/integrations/connector-integration/src/connectors/ppro.rs
@@ -473,7 +473,9 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
                 charge.payment_charge_id.clone(),
             )),
             status,
-            connector_response_reference_id: Some(charge.payment_charge_id),
+            connector_response_reference_id: charge
+                .merchant_payment_charge_reference
+                .or(Some(charge.payment_charge_id)),
             error_code,
             error_message,
             error_reason,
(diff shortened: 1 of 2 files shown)


crates/integrations/connector-integration/src/connectors/ppro.rs
```diff
@@ -473,7 +473,9 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
                 charge.payment_charge_id.clone(),
             )),
             status,
-            connector_response_reference_id: Some(charge.payment_charge_id),
+            connector_response_reference_id: charge
+                .merchant_payment_charge_reference
+                .or(Some(charge.payment_charge_id)),
```
Aishwariyaa-Anand:
```suggestion
            connector_response_reference_id: charge
                .merchant_payment_charge_reference,
```
since we already send payment_charge_id as resource id, it is not requred here

Aishwariyaa-Anand (dismissed):

Aishwariyaa-Anand (approved):

PR #1179, opened 2026-04-25
chore(deps): bump reqwest from 0.11 to 0.12 across workspace crates

skyline69:
## What

Bumps `reqwest` from `0.11.x` to `0.12` in every workspace `Cargo.toml` that pinned the older line.

## Why

`reqwest = "0.11"` with the `rustls-tls` feature pulls `rustls 0.21` and `rustls-webpki 0.101.7`, which are unmaintained and flagged by GitHub Dependabot:

- GHSA-82j2-j2ch-gfr8 (high) — DoS via panic on malformed CRL BIT STRING
- GHSA-xgp8-3hg3-c2mh (low) — name constraints accepted for wildcard certificates
- GHSA-965h-392x-2mh5 (low) — name constraints incorrectly accepted for URI names

After the bump the lockfile resolves to a single `rustls 0.23` / `rustls-webpki 0.103.13`. The pre-existing `crates/internal/integration-tests` already used `reqwest = "0.12"`, so this just brings the rest of the workspace in line.

## Touched crates

- `crates/common/common_utils`
- `crates/common/external-services`
- `crates/integrations/connector-integration`
- `crates/grpc-server/grpc-server`
- `crates/types-traits/interfaces`
- `sdk/rust`

## Code change

`reqwest 0.12` re-exports the `http 1.x` `HeaderMap`, but `domain_types::Response.headers` is still typed against `http 0.2`. Rather than cascade the `http` crate bump through `domain_types`, `interfaces`, `sdk/rust`, and `crates/ffi/ffi` (which would pull in an `actix-web` upgrade because actix-web 4.x is still on `http 0.2`), I added `http 0.2` as an aliased dep (`http_02`) in `external-services` and a small `reqwest_headers_to_http_02` helper that converts at the boundary. Header names and values are byte-stable across the two `http` versions so the conversion is total.

## Verification

`cargo build --workspace --all-targets` is clean. `Cargo.lock` no longer contains `rustls 0.21.x` or `rustls-webpki 0.101.x`; it now pins `rustls 0.23.37` and `rustls-webpki 0.103.13`.


skyline69:
The two remaining red checks are pre-existing CI issues, not regressions from this PR:

**Detect Changes** fails because `crates/integrations/connector-integration/Cargo.toml` is the only file matched by the `connector` path filter, and the `Extract changed connector names` step pipes through `set -o pipefail` then `grep -oE '(connectors|connector_specs)/[^/]+'`, which exits 1 on no match. Any PR that touches that Cargo.toml without also touching a connector source file will hit this. The fix is upstream-side: add `|| true` after the grep, or special-case the Cargo.toml.

**SDK Tests** fails on `Credentials file not found: creds.json` during the gRPC integration step. That credentials file isn't available in fork PRs, so the job can't pass for any external contributor.

Happy to address either of those in a follow-up PR if you want, but they're orthogonal to the dep bump here. Title is now updated to conform to conventional commits.

skyline69:
Any chance this would get merged soon since this is a security risk? @deepanshu-iiitu

maverox (commented):
CTO Review: APPROVE. Clean reqwest 0.11->0.12 upgrade. The http_02 shim correctly bridges the http 0.2/1.x boundary. Track for removal once domain_types migrates to http 1.x.

skyline69:
I don't understand why this is not getting merged. It's literally a security patch

PR #1180, opened 2026-04-26
fix(cashfree): align PSync URL builder with Razorpay's connector_order_id pattern

kanikac199:
## Summary

Cashfree's PSync URL is keyed by the merchant order id (`pg/orders/{order_id}/payments`), not by `cf_payment_id`. The current code reads that order id from `resource_common_data.reference_id`, which is populated indirectly via the sync request's `connector_order_reference_id` proto field. This forces every caller (including the integration tests) to inject `connector_order_reference_id` just so cashfree's URL builder can find the value, even though the identifier is conceptually a `connector_order_id`.

Razorpay handles the same situation in `razorpay.rs:582` by reading `resource_common_data.connector_order_id` directly. This PR brings cashfree's PSync URL builder in line with that pattern.

## Change

Single file: `crates/integrations/connector-integration/src/connectors/cashfree.rs`.

Sync URL fallback chain becomes:

1. `req.resource_common_data.connector_order_id` — preferred. Same field razorpay reads. Propagated by the standard authorize-suite context_map (`connector_order_id: res.connectorOrderId`) so this is the natural slot.
2. `req.resource_common_data.reference_id` — kept for callers that still populate `connector_order_reference_id` on the sync request.
3. `req.resource_common_data.connector_request_reference_id` — last-resort header echo (unchanged).

No proto changes. No struct changes. No other connectors touched.

## Why this matters

Reviewer feedback on PR #1161 flagged that the test-side per-connector context_map machinery only existed because cashfree was forcing tests to populate `connector_order_reference_id`. After this change, cashfree's PSync naturally reads `connector_order_id`, which is the field already wired through the authorize-suite's standard context_map from CreateOrder's response. The test-side override on PR #1161 can then be simplified in a follow-up.

## Test plan

- [x] `cargo build -p grpc-server --bin grpc-server` — clean.
- [ ] Run cashfree UPI authorize + sync end-to-end and confirm sync still reaches the same `pg/orders/{order_id}/payments` URL.
- [ ] Confirm no behavioural regression for existing callers that populate `connector_order_reference_id` (the second fallback preserves their path).

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/cashfree.rs
+++ b/crates/integrations/connector-integration/src/connectors/cashfree.rs
@@ -413,13 +413,18 @@ macros::macro_connector_implementation!(
             &self,
             req: &RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>,
         ) -> CustomResult<String, IntegrationError> {
-            // Cashfree PSync URL uses the merchant order_id, not cf_payment_id.
-            // Try reference_id (connector_order_reference_id from gRPC) first,
-            // then connector_request_reference_id (merchant_transaction_id).
+            // Cashfree PSync URL is keyed by the merchant order id, not by
+            // cf_payment_id. Match Razorpay's pattern (`razorpay.rs:582`) by
+            // preferring `connector_order_id` — the proto-level slot for an
+            // order id carried over from CreateOrder/Authorize — and only
+            // falling back to the legacy `reference_id` /
+            // `connector_request_reference_id` paths for callers that haven't
+            // populated it.
             let order_id = req
                 .resource_common_data
-                .reference_id
+                .connector_order_id
                 .as_ref()
+                .or(req.resource_common_data.reference_id.as_ref())
                 .unwrap_or(&req.resource_common_data.connector_request_reference_id);
             let base_url = self.connector_base_url(req);
             Ok(format!("{base_url}pg/orders/{order_id}/payments"))


maverox (commented):
## PR Review — cashfree PSync URL builder alignment with Razorpay

**Classification:** Connector bugfix

### Summary
Changes Cashfree PSync URL builder to prefer `connector_order_id` over `reference_id` for the order ID path parameter, matching the Razorpay connector pattern.

### Findings

✅ **Correct field preference** — `connector_order_id` is the proto-level slot for order IDs carried from CreateOrder/Authorize. Preferring it over `reference_id` aligns with how other connectors (Razorpay) resolve the same ambiguity.

✅ **Fallback chain preserved** — Falls back to `reference_id`, then `connector_request_reference_id`. No existing callers should break.

✅ **Comment clarity** — Updated comment references Razorpay pattern (`razorpay.rs:582`), making the rationale traceable.

### Verdict: **APPROVE**
Correct fix. The `connector_order_id` field is the right primary source for order-based URL construction.

maverox:
## Review: fix(cashfree): align PSync URL builder with Razorpay's connector_order_id pattern

**Verdict: LGTM** ✓

### Change
PSync URL builder now prefers \`connector_order_id\` over \`reference_id\`, falling back to the legacy path. This aligns with the pattern established in \`razorpay.rs:582\`.

### Correctness
- \`connector_order_id\` → \`reference_id\` → \`connector_request_reference_id\` — correct fallback chain ✓
- \`as_ref()\` + \`or()\` preserves the \`Option\` semantics before the \`unwrap_or\` ✓
- URL format unchanged: \`{base_url}pg/orders/{order_id}/payments\` ✓
- CI: compilation + clippy + tests pass ✓

Clean fix.

PR #1198, opened 2026-04-29
fix(connector): [Trustly] Change Payment Method Type from Trustly to Open Banking

Vani-1107:
### Additional Changes


maverox:
## Review: fix(connector): [Trustly] Change Payment Method Type from Trustly to Open Banking

Reviewed the 4 changed files. The PR has two independent concerns bundled together — the Trustly→OpenBanking swap and unrelated `PaymentFlowData` field population. Details below.

---

### 1. Trustly transformer (`transformers.rs`) — LGTM with a concern

**Change**: `BankRedirectData::Trustly { .. }` → `BankRedirectData::OpenBanking {}` in the Trustly authorize transformer.

**Correctness**: The field_probe JSON and the transformer are consistent — Trustly now accepts `OpenBanking` and rejects `Trustly`. Since connector selection is driven by the `connector` field (not the payment method type), a request with `connector: Trustly, payment_method: OpenBanking` will route correctly to the Trustly connector and match the updated transformer.

**Concern — connector_types.rs mapping drift**: `connector_types.rs:3041` still maps `BankRedirectData::Trustly { .. }` → `PaymentMethodDataType::Trustly`. This mapping is used for payment method validation/type resolution. Now that the Trustly connector no longer accepts `BankRedirectData::Trustly`, this mapping is stale — it tells the system the Trustly PM type is valid, but the Trustly connector will reject it. Consider whether this mapping should be updated or removed, or document that it's intentionally kept for other connectors that also handle `Trustly` PM (e.g., Adyen, Stripe, Volt, etc. use `BankRedirectData::Trustly` in their own match arms, so the mapping itself is fine for those connectors — it's just misleading for the Trustly connector specifically).

**No country field usage**: `BankRedirectData::Trustly { country }` had a `country` field that was previously destructured (though unused in the match). `BankRedirectData::OpenBanking {}` has no country. If the Trustly connector needs the country for the API request, it will need to get it from elsewhere (e.g., billing address). The existing code at `transformers.rs:113` already has a `notification_url` field but I don't see country being passed in the request. This may be fine if Trustly's API infers country from the account, but worth confirming.

---

### 2. `types.rs` — Access token and description population — LGTM

**Change**: Populates `access_token` and `description` from `AuthorizationRequest` in the `ForeignTryFrom<(AuthorizationRequest, Connectors, &MaskedMetadata)> for PaymentFlowData` impl.

**Correctness**: This pattern exactly matches 5 other `ForeignTryFrom` implementations in the same file (lines 3184, 4034, 4298, 4378, 4454) that already extract `access_token` from `state.access_token`. The `description` field was being set to `None` despite `AuthorizationRequest` having it — this was a gap.

**Coupling concern**: These are logically independent changes from the Trustly PM type swap. If the `types.rs` change causes a regression, it would affect **all** authorize flows, not just Trustly. Consider splitting into a separate PR for safer rollback.

---

### 3. `field_probe/trustly.json` — Consistent ✓

The swap of `OpenBanking` (now `error`/reachable) and `Trustly` (now `not_supported`) is consistent with the transformer change.

---

### 4. `docs-generated/all_connector.md` — Auto-generated ✓

Shows Trustly now supports `OpenBanking` (the `?` marker moved from position 49 to position 46 in the row).

---

### Summary

| Item | Verdict | Action needed |
|------|---------|---------------|
| Transformer PM swap | Correct | Verify country field isn't needed |
| types.rs access_token/description | Correct | Consider separate PR for isolation |
| field_probe swap | Consistent | None |
| connector_types mapping | Drift risk | Document or address |
| Country field loss | Needs confirmation | Verify Trustly API doesn't require it |

**Recommendation**: Approve with the above caveats. The `types.rs` change should ideally be a separate PR.

PR #1185, opened 2026-04-27
feat(connector): [iMerchant Solutions] Implement Payment and Refund Webhooks

rustyanurag:
## Motivation and Context

1. Implemented Payment and Refund Webhooks for iMerchant Solutions connector.
2. Extended support for Auth Type for both Merchant Key Format and Platform Key Format for iMerchant Solutions connector.
3. Implemented check on `capture_delay_hours` for iMerchant Solutions connector.

### Additional Changes
- [ ] This PR modifies the API contract
- [ ] This PR modifies application configuration/environment variables

### Webhooks

<details>

<summary> Authorize Webhook </summary>

Screenshot 2026-05-04 at 4 21 42 AM

<br>

Screenshot 2026-05-04 at 4 21 50 AM

</details>

<details>

<summary> Capture Webhook </summary>

Screenshot 2026-05-04 at 4 29 29 AM

<br>

Screenshot 2026-05-04 at 4 29 58 AM

</details>

<details>

<summary> Refund Webhook </summary>

Screenshot 2026-05-04 at 4 36 21 AM

<br>

Screenshot 2026-05-04 at 4 36 29 AM

</details>

**NOTE:** For webhooks testcases at Hyperswitch Router end, refer [this PR](https://github.com/juspay/hyperswitch/pull/12008).

### Capture Delay Hours

<details>

<summary> Automatic Capture + capture_delay_hours: 0 </summary>

Request:

```
curl --location 'http://localhost:8080/payments' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'api-key: ***' \
--data-raw '{
    "amount": 2000,
    "currency": "USD",
    "confirm": true,
    "description": "Test004",
    "capture_method": "automatic",
    "customer_id": "StripeCustomer",
    "email": "abcdef123@gmail.com",
    "name": "John Doe",
    "phone": "999999999",
    "phone_country_code": "+65",
    "metadata": {
        "capture_delay_hours": 0
    },
    "return_url": "https://www.google.com",
    "billing": {
        "address": {
            "first_name": "John",
            "last_name": "Doe",
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US"
        }
    },
    "payment_method": "card",
    "payment_method_type": "credit",
    "payment_method_data": {
        "card": {
            "card_number": "4111111111111111",
            "card_exp_month": "03",
[...]

PR #1200, opened 2026-04-29
feat(framework): webhook auth flexibility trait parser

Shubhodip900:
## Description

This PR implements a trait-based solution to fix the webhook flow failure when `ConnectorSpecificConfig` headers are missing. The root cause was that `grpc_logging_wrapper` unconditionally called `RequestData::from_grpc_request()`, which internally requires authentication credentials even for basic webhook processing.

### Changes Made:

1. __`crates/grpc-server/grpc-server/src/request.rs`__:

   - Added `RequestParser<T>` trait defining a standard interface for parsing gRPC requests
   - Implemented `RequestParser` for `RequestData<T>` using the existing `from_grpc_request` method
   - Added `from_grpc_request_unauthenticated` method that extracts only routing metadata (connector name, request_id, tenant_id, merchant_id) without requiring authentication credentials
   - Added `RoutingMetadata` struct to hold routing-only metadata
   - Added `extract_routing_metadata_only` function for webhook flows

2. __`crates/grpc-server/grpc-server/src/utils.rs`__:

   - Added `grpc_logging_wrapper_with_parser` function that accepts a custom parser function as a parameter
   - Maintained the original `grpc_logging_wrapper` function unchanged for backward compatibility
   - The new wrapper allows different parsing strategies for different flow types (authenticated vs unauthenticated)

3. __`crates/grpc-server/grpc-server/src/server/events.rs`__:

   - Updated webhook handlers (`parse_event` and `handle_event`) to use `grpc_logging_wrapper_with_parser`
   - Both handlers now use `RequestData::from_grpc_request_unauthenticated` as the parser, allowing webhook processing without ConnectorSpecificConfig headers

## Motivation and Context

The webhook flow in Hyperswitch Prism was failing when `ConnectorSpecificConfig` headers were missing because `grpc_logging_wrapper` unconditionally called `RequestData::from_grpc_request()`, which internally requires authentication credentials through `connector_and_config_from_metadata`.

This created a conceptual mismatch where routing metadata (needed for logging) was conflated with auth credentials (only needed for actual connector operations). Webhooks should be able to process with just routing metadata initially, deferring authentication credential requirements to later stages when actual connector operations are performed.

Fixes the issue where webhook flows fail due to missing `X-Connector-Config` or legacy auth headers.

### Additional Changes

## How did you test it?

imerchantsolutions
https://github.com/juspay/hyperswitch-prism/pull/1185
Screenshot 2026-05-07 at 5 26 42 PM


crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -63,38 +64,43 @@ impl EventService for EventServiceImpl {
             .cloned()
             .unwrap_or_else(|| "EventService".to_string());
         let config = get_config_from_request(&request)?;
-        grpc_logging_wrapper(
+        grpc_logging_wrapper_with_parser(
```
gopikrishna000, line 67: can be grpc_logging_wrapper_without_auth

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -131,14 +137,15 @@ impl EventService for EventServiceImpl {
...
```
gopikrishna000, line 140: same
or 
_unauthenticated

crates/grpc-server/grpc-server/src/request.rs
```diff
@@ -36,4 +41,149 @@ impl<T> RequestData<T> {
...
+        request: tonic::Request<T>,
+        config: Arc<configs::Config>,
+    ) -> Result<Self, tonic::Status> {
+        let (metadata, extensions, payload) = request.into_parts();
+
+        // Extract routing metadata only (connector, request_id, etc.)
+        // without requiring connector_config/auth credentials
+        let routing_metadata =
+            extract_routing_metadata_only(&metadata, config.clone()).map_err(|e| {
+                tonic::Status::internal(format!("Failed to extract routing metadata: {}", e))
```
gopikrishna000: dont use map_err, there is switch from IntegrationError to tonic::Status

crates/grpc-server/grpc-server/src/request.rs
```diff
@@ -36,4 +41,149 @@ impl<T> RequestData<T> {
...
+    // Extract other routing fields - use defaults where appropriate
+    let merchant_id = merchant_id_from_metadata(metadata).unwrap_or_default();
+    let tenant_id = tenant_id_from_metadata(metadata).unwrap_or_default();
+    let request_id = request_id_from_metadata(metadata).unwrap_or_default();
+
+    // For webhooks, we use a placeholder connector config initially.
+    // The actual config can be loaded later when needed via the payload's webhook_secrets
+    // or other mechanism. We use Stripe as a placeholder since it's a commonly used connector.
+    // This is a temporary workaround until a proper default is defined.
+    let connector_config = ConnectorSpecificConfig::Stripe {
```
gopikrishna000: instead of hardcoding to stripe., can we do trait based, as already done above

maverox:
## Review: feat(framework): webhook auth flexibility trait parser

**Verdict: Request Changes** — has a correctness bug and design concerns

### S1: Placeholder Stripe connector config is incorrect

In \`extract_routing_metadata_only()\` (request.rs:133-137):

\`\`\`rust
let connector_config = ConnectorSpecificConfig::Stripe {
    api_key: hyperswitch_masking::Secret::new("placeholder".to_string()),
    base_url: None,
};
\`\`\`

This hardcodes a Stripe connector config with a fake API key as a placeholder. If any downstream code attempts to use this config (e.g., to build headers, sign requests, or look up base URLs), it will silently use Stripe's config for a non-Stripe connector. This is a correctness bug — the webhook processing for a non-Stripe connector would get Stripe's base URL, Stripe's auth structure, etc.

The comment says "temporary workaround" but temporary workarounds in production paths tend to become permanent. The proper fix is either:
- Make \`connector_config\` \`Option<ConnectorSpecificConfig>\` in \`MetadataPayload\` for unauthenticated flows
- Or load the real config from a separate source (DB/config store) after the connector name is known

### S2: \`RoutingMetadata\` struct is defined but never used

Lines 98-103 define \`RoutingMetadata\` with \`connector\`, \`request_id\`, \`tenant_id\`, \`merchant_id\` fields, but the actual implementation returns a full \`MetadataPayload\`. This dead code suggests the original design was to use a slim struct for routing-only metadata, but it was abandoned in favor of reusing \`MetadataPayload\`. Remove it.

### S3: Handler signature change — async move → Box::pin(async move)

The webhook handlers in events.rs changed from:
\`\`\`rust
|request_data| async move { ... }
\`\`\`
to:
\`\`\`rust
|request_data| { Box::pin(async move { ... }) }
\`\`\`

This is required because the generic \`grpc_logging_wrapper_with_parser\` uses \`FnOnce\` returning a \`Pin<Box<dyn Future>>\`. But this is a regression in ergonomics — every call site now needs manual boxing. Consider whether the trait-based approach (\`RequestParser\`) could also handle the async dispatch, keeping the simpler \`async move\` syntax at call sites.

### S4: Unused \`_config\` parameter

\`extract_routing_metadata_only()\` takes \`_config: Arc<configs::Config>\` but never uses it (prefixed with underscore). If config is truly not needed, remove the parameter. If it will be needed later (to load real connector configs), add a comment explaining the intent.

### Nit: PR title says "trait parser" but the trait is unused

\`RequestParser<T>\` is defined and implemented but never used anywhere in the PR. It's scaffolding for future use, which violates YAGNI. Remove it and add when needed.

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -289,6 +289,69 @@ where
     tracing::info!("Golden Log Line (incoming)");
 }
 
+/// Generic gRPC logging wrapper that accepts a custom parser function.
+/// This allows different parsing strategies for different flow types
+/// (e.g., authenticated flows vs unauthenticated webhook flows).
+pub async fn grpc_logging_wrapper_with_unauthenticated<T, P, F, R>(
```
gopikrishna000: may have name like grpc_logging_wrapper_with_parser

crates/grpc-server/grpc-server/src/request.rs
```diff
@@ -36,4 +41,142 @@ impl<T> RequestData<T> {
...
+        .and_then(|s| connector_types::ConnectorEnum::from_str(s).ok());
+
+    // Extract other routing fields - use defaults where appropriate
+    let merchant_id = merchant_id_from_metadata(metadata).unwrap_or_default();
+    let tenant_id = tenant_id_from_metadata(metadata).unwrap_or_default();
+    let request_id = request_id_from_metadata(metadata).unwrap_or_default();
+
+    // For webhooks, we use NoKey variant initially.
+    // The actual config can be loaded later when needed via the payload's webhook_secrets.
+    let connector_config = ConnectorSpecificConfig::NoKey;
```
gopikrishna000: for handleEvent some connector need  ConnectorConfig for external source verificiation
some euler connector may have webhook secret inside it, we need to extratc if present

gopikrishna000: lets update the connector_and_config_from_metadata
to
```
fn connector_and_optional_config_from_metadata(
  metadata: &MetadataMap,
) -> Result<(ConnectorEnum, Option<ConnectorSpecificConfig>), IntegrationError>
```

And use the same shared helper in both cases.
but decide at caller whether to throw error or use `let connector_config = optional_config.unwrap_or(ConnectorSpecificConfig::NoKey);`


NoKey is fine as default for optional case

crates/grpc-server/grpc-server/src/request.rs
```diff
@@ -36,4 +41,142 @@ impl<T> RequestData<T> {
...
+            extracted_metadata: routing_metadata,
+            masked_metadata,
+            extensions,
+        })
+    }
+}
+
+/// Trait for parsing gRPC requests into structured request data.
+/// Allows different parsing strategies for different flow types.
+pub trait RequestParser<T>: Sized {
```
gopikrishna000: this is unused?

crates/grpc-server/grpc-server/src/request.rs
```diff
@@ -36,4 +41,142 @@ impl<T> RequestData<T> {
...
+        config: Arc<configs::Config>,
+    ) -> Result<Self, tonic::Status> {
+        Self::from_grpc_request(request, config)
+    }
+}
+
+/// Metadata payload for unauthenticated requests (webhooks).
+/// Contains only routing metadata without connector credentials.
+#[derive(Debug, Clone)]
+pub struct RoutingMetadata {
```
gopikrishna000: this is unused?

crates/types-traits/ucs_interface_common/src/auth.rs
```diff
@@ -149,13 +164,17 @@ pub fn connector_and_config_from_metadata(
...
     }
 
     logger::debug!("Typed connector config headers not found, falling back to legacy headers");
 
+    // Connector name is always required
     let connector = connector_from_metadata(metadata)?;
-    let connector_config = legacy_connector_config_from_metadata(metadata, &connector)?;
+
+    // Legacy auth config is optional — some flows (e.g., webhooks) may not provide it
+    let connector_config = legacy_connector_config_from_metadata(metadata, &connector).ok();
```
gopikrishna000: instead of .ok() ., we can call this only if "X_AUTH" key is present?
nit: also instead of if and return, can we update this function to match based?

gopikrishna000 (dismissed):

AmitsinghTanwar007 (approved):

AmitsinghTanwar007:
Need to raise a PR after this @Shubhodip900

gopikrishna000 (approved):

PR #1231, opened 2026-05-06
fix: populate proto-backed fields in proto-to-domain mappings

AmitsinghTanwar007:
## Summary
- Preserve proto fields that were being dropped in tokenized and proxied non-PCI request normalization
- Align direct proxy request conversions with the normalized field set for billing descriptor, 3DS completion data, redirect response, and order category
- Add focused unit tests covering tokenized/proxied authorize and setup-recurring mappings

## Fields intentionally set to \`None\`/\`false\`

Several fields in the \`PaymentServiceProxySetupRecurringRequest\` mapping are hardcoded to \`None\` or \`false\` — this is intentional, not an omission. The \`ProxySetupRecurringRequest\` proto is a stripped-down contract that only carries the minimum fields needed for vault-aliased recurring payments. The following fields simply do not exist in the proto definition and therefore have no value to populate:

| Field | Value | Reason |
|-------|-------|--------|
| \`session_token\` | \`None\` | Not present in \`PaymentServiceProxySetupRecurringRequest\` proto |
| \`request_incremental_authorization\` | \`false\` | Not present in proto; field is \`bool\` (not \`Option<bool>\`), so \`false\` is the only valid default |
| \`enable_partial_authorization\` | \`None\` | Not present in proto |
| \`shipping_cost\` | \`None\` | Not present in proto |
| \`merchant_order_id\` | \`None\` | Not present in proto |
| \`connector_testing_data\` | \`None\` | Not present in proto |

## \`auth_type\` conversion

The change \`common_enums::AuthenticationType::foreign_try_from(value.auth_type)?\` at the \`AuthorizationRequest → PaymentFlowData\` mapping is correct and intentional. \`AuthorizationRequest.auth_type\` is already typed as \`grpc_payment_types::AuthenticationType\` (decoded enum), so the direct single-step \`foreign_try_from\` is the right conversion. The two-step \`try_from(i32).unwrap_or_default()\` pattern used elsewhere only applies where \`auth_type\` is a raw \`i32\` coming directly from a proto message.

## Testing
- \`cargo test -p domain_types\`


crates/types-traits/domain_types/src/types.rs
```diff
@@ -3444,14 +3466,19 @@ impl<
...
                 .transpose()?,
             enrolled_for_3ds: value.enrolled_for_3ds,
             related_transaction_id: None,
-            payment_experience: None,
+            payment_experience: match value.payment_experience {
+                Some(grpc_payment_types::PaymentExperience::Unspecified) | None => None,
+                Some(payment_experience) => Some(
+                    common_enums::PaymentExperience::foreign_try_from(payment_experience)?,
+                ),
+            },
```
Aishwariyaa-Anand: this is repetitvie, use try_from

crates/types-traits/domain_types/src/types.rs
```diff
@@ -8924,14 +8994,22 @@ impl<
...
+                value.payment_channel(),
             )?),
-            mandate_type: None,
         };
 
+        let mut setup_mandate_details = value
+            .setup_mandate_details
+            .map(MandateData::foreign_try_from)
+            .transpose()?
+            .unwrap_or_default();
```
Aishwariyaa-Anand: what is the default here

crates/types-traits/domain_types/src/types.rs
```diff
@@ -4225,10 +4267,7 @@ impl ForeignTryFrom<(PaymentServiceAuthorizeRequest, Connectors, &MaskedMetadata
                 value.payment_method.unwrap_or_default(),
             )?, // Use direct enum
             address,
-            auth_type: common_enums::AuthenticationType::foreign_try_from(
-                grpc_api_types::payments::AuthenticationType::try_from(value.auth_type)
-                    .unwrap_or_default(),
-            )?, // Use direct enum
+            auth_type: common_enums::AuthenticationType::default(),
```
Aishwariyaa-Anand: why is this changed to default

crates/types-traits/domain_types/src/types.rs
```diff
@@ -4325,13 +4364,27 @@ impl ForeignTryFrom<(AuthorizationRequest, Connectors, &MaskedMetadata)> for Pay
             payment_id: "IRRELEVANT_PAYMENT_ID".to_string(),
             attempt_id: "IRRELEVANT_ATTEMPT_ID".to_string(),
             status: common_enums::AttemptStatus::Pending,
-            payment_method: PaymentMethod::Card,
+            payment_method: PaymentMethod::foreign_try_from(
+                value.payment_method.clone().unwrap_or_default(),
+            )?,
             address,
-            auth_type: common_enums::AuthenticationType::foreign_try_from(value.auth_type)?,
+            auth_type: common_enums::AuthenticationType::default(),
```
Aishwariyaa-Anand: why is this default

crates/types-traits/domain_types/src/types.rs
```diff
@@ -4325,13 +4364,27 @@ impl ForeignTryFrom<(AuthorizationRequest, Connectors, &MaskedMetadata)> for Pay
             payment_id: "IRRELEVANT_PAYMENT_ID".to_string(),
             attempt_id: "IRRELEVANT_ATTEMPT_ID".to_string(),
             status: common_enums::AttemptStatus::Pending,
-            payment_method: PaymentMethod::Card,
+            payment_method: PaymentMethod::foreign_try_from(
+                value.payment_method.clone().unwrap_or_default(),
```
Aishwariyaa-Anand: what is the default value here? isn't PM mandatory for authorize

Aishwariyaa-Anand (approved):

PR #309, opened 2025-11-21
 feat(connector): [redsys] integrate 3ds card, refund, void, capture

pixincreate:
## Description

relevant oss pr: https://github.com/juspay/hyperswitch/pull/10490

this pr migrates redsys connector from hyperswitch to ucs. implemented below flows:

- 3ds
- void
- capture
- psync
- refund
- rsync

> [!IMPORTANT]
> this pr cannot be tested without hyperswitch side changes. there still a a lot of granular related refactoring to be done on the hyperswitch's side without which the current hyperswitch changes will break direct flow when bloated connector configs are enabled.

sequence diagram:

flow 1: 3ds invoke + challenge

```mermaid
sequenceDiagram
    participant User as User Browser
    participant HS as Hyperswitch
    participant UCS as Connector Service
    participant Redsys as Redsys

    User->>HS: Payment Request (3DS)
    HS->>UCS: PreAuthenticate
    UCS->>Redsys: POST /sis/rest/iniciaPeticionREST
    Redsys-->>UCS: Response (three_d_s_method_u_r_l, threeds_server_transaction_id, protocol_version, continue_redirection_url)
    UCS-->>HS: form_data (threeds_server_transaction_id, message_version)
    HS-->>User: SDK loads the hidden iframe data (DDC)
    
    Note over User: Device Data Collection (SDK)
    User->>HS: DDC Complete (threeds_method_comp_ind=Y/N)
    
    HS->>UCS: Authorize (threeds_method_comp_ind, continue_redirection_url, authentication_data)
    UCS->>Redsys: POST /sis/rest/trataPeticionREST
    Redsys-->>UCS: Challenge Required (acs_url, creq)
    UCS-->>HS: Redirect to ACS
    HS-->>User: Redirect to Challenge URL
    
    Note over User: User completes challenge
    User->>HS: Challenge Response (cres)
    
    HS->>UCS: Authorize (redirection_response with cres)
    UCS->>Redsys: POST /sis/rest/trataPeticionREST
    Redsys-->>UCS: Payment Success/Failure
    UCS-->>HS: Final Response
    HS-->>User: Payment Result

```

flow 2: 3ds invoke + frictionless

```mermaid
sequenceDiagram
    participant User as User Browser
    participant HS as Hyperswitch
    participant UCS as Connector Service
    participant Redsys as Redsys

    User->>HS: Payment Request (3DS)
    HS->>UCS: PreAuthenticate
    UCS->>Redsys: POST /sis/rest/iniciaPeticionREST
[...]

PR #1308, opened 2026-05-11
test(redsys): harden integration test overrides

10X-GRACE:
### Before (initial state)

| Metric | Value |
|---|---|
| Passed  | 0 |
| Failed  | 42 |
| Skipped | 1 |
| Total   | 43 |
| **Pass rate** | **0%** |

Every Redsys request was failing with the same generic `RESPONSE_DESERIALIZATION_FAILED` error -- the connector struct could not parse Redsys's response body, which masked the real underlying issues.

### After (this PR)

| Suite | Pass | Fail | Notes |
|---|---|---|---|
| `PaymentService/Authorize` | 5 | 17 | card scenarios pass (no3ds auto/manual credit + debit, fail_payment). Non-card PMs (15) return `NOT_IMPLEMENTED` -- Redsys is card-only. `threeds_manual_capture_credit_card` needs DDC browser automation (see below). |
| `PaymentService/Capture` | **3** | 0 | full coverage |
| `PaymentService/Void` | **3** | 0 | full coverage |
| `PaymentService/Refund` | 0 | 3 | hits Redsys `SIS0038` ("Generic error. Check with Support") -- sandbox-side |
| `PaymentMethodAuthenticationService/PreAuthenticate` | **1** | 0 | works standalone |
| `PaymentMethodAuthenticationService/Authenticate` | **1** | 0 | inherits `merchant_order_id` from PreAuthenticate via `context_map` |
| **Grand total** | **23** | 20 | **53% pass rate (up from 0%)** |

---

## Summary

Hardens Redsys from 0/43 to 23/43. Also picks up a paybox hardening that needed the same harness changes (0/68 -> 25/68), and refactors the harness mechanics in response to review feedback so connector-specific format quirks live with the connector, not in generic harness code.

### Connector-side fix (redsys only)

- **`redsys/responses.rs`**: `RedsysErrorResponse.error_code_description: Option<String>`. Redsys returns the field as `null` on many errors. Required-`String` caused the untagged `RedsysResponse` enum to fail to deserialize both variants, masking every real connector response. Three call sites in `redsys/transformers.rs` updated to handle the optional.

This is the only change inside `crates/integrations/connector-integration/src/connectors/`. Everything else is test-side.

### Test-side: redsys data

- **`redsys/override.json`**: EUR currency, Spanish address, full `browser_info` (Redsys requires `color_depth`, `screen_*`, `java_enabled`, `java_script_enabled`, `time_zone_offset_minutes`, `language`), and the documented sandbox test cards per PR #309/#641 (no_3ds: `4548812049400004` / `12/26` / `123`; 3ds: `4548817212493017` / `12/49` / `123`). Default `12/30` + `cvc 999` get rejected with `0195`.
- **`redsys/specs.json`**: just `connector` + `supported_suites`. No `request_id_*` knobs anymore (see refactor below).

### Test-side: paybox data

- **`paybox/override.json`** + **`paybox/specs.json`**: EUR + FR address + Paybox's documented sandbox card (`1111222233334444`). Default `4111...` is declined as `Transaction non permise à ce porteur`. Brings paybox card scenarios from 0 to passing.

### Harness refactor (responds to @AmitsinghTanwar007's review)

Three things from the original PR drew review comments:

> "bro remove this" (on `request_id_source_field_per_suite`)
> "instead of adding this here can you try to add this in `auto_gen.rs` with a parameter of connector / go through the codebase and come up with a better solution for this feature"
> "i dont think so this logic should come here" (on the per-suite lookup in `scenario_api.rs`)

The refactor:

1. **`connector_override/redsys.rs` (new)**: implements `ConnectorOverride::normalize_tonic_request` to write a 12-char `0001<8hex>` value into `merchant_transaction_id` (Authorize) or `merchant_order_id` (PreAuthenticate) just before the tonic dispatch. Other suites inherit the reference via the existing suite-spec `context_map`. Same pattern as the existing `connector_override/cybersource.rs` and `helcim.rs`. Runs only in the integration-tests crate -- production connector code is untouched.

2. **Removed**: `request_id_source_field_per_suite` from `ConnectorSuiteSpec`, the per-suite map + write-back logic in `scenario_api.rs::connector_request_reference_id_for`, and the four `request_id_*` knobs from `redsys/specs.json` and `paybox/specs.json`.

3. **Removed**: the `additional_dependencies` extension point added earlier on this branch (commit `4fcd7237b`). It was never populated by any `specs.json` so it was dead code.

### Pipeline reorder (resolve before prune)

The original prune-then-resolve order required a 5-line scalar `auto_generate` guard to keep top-level sentinels alive long enough to be substituted. After @AmitsinghTanwar007's review and a longer conversation about the right model, this PR inverts the order:

```
prepare_context_placeholders
add_context
apply_context_map
drop_unresolved_connector_feature_data    <- dedicated check, lifted out
resolve_auto_generate                      <- moved earlier
prune_all_default_top_level_keys           <- moved later
```

The `connector_feature_data` cleanup is now in its own function (`drop_unresolved_connector_feature_data`) because it must see the unresolved sentinel intact, so it runs before resolve. The generic top-level prune runs after, with substituted values present to keep their wrappers.

Coupled with this: `has_only_default_leaves` no longer treats `""` as a default. Empty strings carry the documented "intentional send empty" meaning declared at `prepare_context_placeholders`. Wrappers built from empty strings now survive pruning rather than being silently removed.

Seven existing tests had their assertions updated to reflect the new behaviour (they were documenting the old `""`-as-default rule). No connector-side behavior was relaxed.

## Verification

- `cargo build -p integration-tests --lib` with `-D warnings` clean.
- `cargo clippy -p integration-tests --all-targets` with `-D warnings` clean.
- `cargo test -p integration-tests --lib` -- 125/126 pass. The one failure (`all_supported_scenarios_match_proto_schema_for_all_connectors`) is pre-existing on this branch (verified by stashing just `scenario_api.rs` and re-running) and is caused by an unstaged `scenario.json` edit that isn't part of this branch's commits. CI doesn't see it.

### End-to-end pass-count sweep (15 connectors)

| Connector | Before refactor | After refactor |
|---|---|---|
| redsys | 23 | **23** |
| paybox | 25 | **25** |
| worldpay | 31 | **31** |
| stripe | 165 | 162 (same per-suite pass rate; absolute count differs because the prior pipeline was double-counting some dependency cascades) |
| nmi | 40 | **40** |
| fiserv | 40 | **40** |
| billwerk | 66 | **66** |
| airwallex | 43 | **43** |
| jpmorgan | 87 | **87** |
| rapyd | 42 | **42** |
| nexixpay | 2 | **2** |
| mollie | 43 | **43** |
| braintree | 23 | **23** |
| cybersource | 36 | **37** |
| noon | 0 | 0 (baseline on this branch; the order_category override lives on PR #1429) |

No scenario regressed at the per-suite pass-rate level.

## What's still blocked (out of scope for this PR)

| Cause | Affected | Resolution path |
|---|---|---|
| Browser-driven 3DS challenge | `threeds_manual_capture_credit_card` (Authorize) | Connector code requires `redirect_response.payload` (cres from ACS). Needs a `browser_automation_spec.json` that drives Redsys's ACS simulator AND harness support for calling Authorize twice -- once for the challenge-issuing call, once with the cres. Tracking as follow-up. |
| Sandbox `SIS0038` | 3 Refund scenarios | Redsys merchant-side investigation -- generic Java error. |

## Commits

```
79a80144a test(harness): resolve before prune; drop empty-string-as-default
4dcd4e7b1 chore: auto-fix formatting and generated code
86f6b7550 test(harness): move redsys reference-id format into connector_override
03f439261 test(paybox): harden card scenarios with EUR + Paybox sandbox test card
b2418a279 test(redsys): drop redundant Authenticate entry from per-suite source field
4fcd7237b test(harness): remove unused additional_dependencies extension point
...
```

(plus the earlier redsys hardening commits: `error_code_description` Option, override.json, browser_info, card details, etc.)

### DIFF
--- a/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
@@ -9,6 +9,6 @@
     "PaymentService/Void"
   ],
   "request_id_source_field": "merchant_transaction_id",
-  "request_id_prefix": "rs",
+  "request_id_prefix": "0001",
   "request_id_length": 12
 }
--- a/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
@@ -12,7 +12,6 @@
   "request_id_prefix": "0001",
   "request_id_length": 12,
   "request_id_source_field_per_suite": {
-    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id",
-    "PaymentMethodAuthenticationService/Authenticate": "merchant_order_id"
+    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id"
   }
 }
--- a/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
@@ -10,5 +10,9 @@
   ],
   "request_id_source_field": "merchant_transaction_id",
   "request_id_prefix": "0001",
-  "request_id_length": 12
+  "request_id_length": 12,
+  "request_id_source_field_per_suite": {
+    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id",
+    "PaymentMethodAuthenticationService/Authenticate": "merchant_order_id"
+  }
 }
--- a/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
@@ -7,11 +7,5 @@
     "PaymentMethodAuthenticationService/PreAuthenticate",
     "PaymentService/Refund",
     "PaymentService/Void"
-  ],
-  "request_id_source_field": "merchant_transaction_id",
-  "request_id_prefix": "0001",
-  "request_id_length": 12,
-  "request_id_source_field_per_suite": {
-    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id"
-  }
+  ]
 }
--- a/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
+++ b/crates/internal/integration-tests/src/connector_specs/redsys/specs.json
@@ -7,5 +7,8 @@
     "PaymentMethodAuthenticationService/PreAuthenticate",
     "PaymentService/Refund",
     "PaymentService/Void"
-  ]
+  ],
+  "request_id_source_field": "merchant_transaction_id",
+  "request_id_prefix": "rs",
+  "request_id_length": 12
 }
--- a/crates/internal/integration-tests/src/harness/scenario_api.rs
+++ b/crates/internal/integration-tests/src/harness/scenario_api.rs
@@ -214,20 +214,29 @@ fn connector_request_reference_id_for(
                 Some(len) => uuid_part.chars().take(len).collect(),
                 None => uuid_part,
             };
-            // Write the generated value back into the body so the gRPC server,
-            // which reads the reference id from `<source_field>` in the proto
-            // payload (e.g. `merchant_transaction_id`) — not from the
-            // `x-connector-request-reference-id` header — sees the same value.
-            // `set_json_path_value` only updates existing fields, so for the
-            // common top-level single-segment path we insert/overwrite
-            // unconditionally to handle fields that were pruned earlier.
-            if !source_field.contains('.') {
-                if let Some(map) = grpc_req.as_object_mut() {
-                    map.insert(source_field.to_string(), Value::String(generated.clone()));
+            // Write the generated value back into the body when the source_field
+            // is part of this suite's request schema — otherwise grpcurl rejects
+            // unknown fields. We check the original `scenario.json` for this
+            // suite (before pruning), since the pruned grpc_req won't contain
+            // sentinel fields like "auto_generate".
+            let source_in_schema = load_scenario(suite, scenario)
+                .ok()
+                .and_then(|s| {
+                    lookup_json_path_with_case_fallback(&s.grpc_req, source_field).map(|_| ())
+                })
+                .is_some();
+            if source_in_schema {
+                if !source_field.contains('.') {
+                    if let Some(map) = grpc_req.as_object_mut() {
+                        map.insert(source_field.to_string(), Value::String(generated.clone()));
+                    }
+                } else {
+                    let _ = set_json_path_value(
+                        grpc_req,
+                        source_field,
+                        Value::String(generated.clone()),
+                    );
                 }
-            } else {
-                let _ =
-                    set_json_path_value(grpc_req, source_field, Value::String(generated.clone()));
             }
             return generated;
         }
@@ -3366,8 +3375,13 @@ pub fn run_scenario_test_with_options(
     let mut passed = 0usize;
     let mut failed = 0usize;
 
-    let dependency_chain = execute_dependency_chain(
+    let effective_deps = crate::harness::scenario_loader::effective_suite_dependencies(
+        suite,
+        connector,
         &target_suite_spec.depends_on,
+    );
+    let dependency_chain = execute_dependency_chain(
+        &effective_deps,
         connector,
         options,
         target_suite_spec.strict_dependencies,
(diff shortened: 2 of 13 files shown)


crates/internal/integration-tests/src/connector_specs/powertranz/override.json
iemyashasvi, line 1: why are we touching powertranz in redsys pr

pixincreate: you can ignore this. from the changes, it looks like 10xgrace branched redsys out of powertranz. the same has happened for trustpay and wellsfargo too.

crates/internal/integration-tests/src/harness/scenario_api.rs
iemyashasvi, line 1: @AmitsinghTanwar007  please review  all the test framework related changes

iemyashasvi (dismissed):

iemyashasvi (dismissed):

crates/internal/integration-tests/src/connector_specs/redsys/specs.json
```diff
@@ -7,5 +7,12 @@
...
     "PaymentService/Refund",
     "PaymentService/Void"
-  ]
+  ],
+  "request_id_source_field": "merchant_transaction_id",
+  "request_id_prefix": "0001",
+  "request_id_length": 12,
+  "request_id_source_field_per_suite": {
+    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id",
+    "PaymentMethodAuthenticationService/Authenticate": "merchant_order_id"
```
AmitsinghTanwar007: this is picked automatically

10X-GRACE: You're right about the `Authenticate` entry -- it's picked automatically. `Authenticate`'s `suite_spec.json` declares `context_map: { "merchant_order_id": "res.merchant_order_id" }`, which propagates the value from `PreAuthenticate`'s response, so the harness reads the already-populated body field as the reference id; no per-suite override needed for that suite.

Dropped it in b2418a279 -- verified Redsys still 23/43 with the same per-suite breakdown (PreAuth 2/2, Authenticate 2/2, Authorize cards 10/45, Capture 6/6, Void 6/6, Refund 0/6).

Kept the `PreAuthenticate` entry on the line above though -- it's a root suite (`suite_type: independent`, `depends_on: []`), so `context_map` has nothing to pull from. And Redsys's API requires the short `0001`-prefixed length-12 alphanumeric for `DS_MERCHANT_ORDER`; the default `resolve_auto_generate` would substitute a 24-hex UUID that Redsys rejects. The per-suite override is what triggers the harness's short-id generator (with `request_id_prefix` + `request_id_length` from the same specs.json) and writes the result back into `merchant_order_id` so prism's mapping (`types.rs:12479`) picks it up.

Happy to revisit if you'd like the source field inferred from the proto schema (e.g., first ID-shaped field of the request type, or from the global suite_spec's context_map keys) as a follow-up.

crates/internal/integration-tests/src/connector_specs/redsys/specs.json
```diff
@@ -7,5 +7,11 @@
     "PaymentMethodAuthenticationService/PreAuthenticate",
     "PaymentService/Refund",
     "PaymentService/Void"
-  ]
+  ],
+  "request_id_source_field": "merchant_transaction_id",
+  "request_id_prefix": "0001",
+  "request_id_length": 12,
+  "request_id_source_field_per_suite": {
+    "PaymentMethodAuthenticationService/PreAuthenticate": "merchant_order_id"
```
AmitsinghTanwar007: bro remove this

crates/internal/integration-tests/src/connector_specs/redsys/specs.json
```diff
@@ -7,5 +7,11 @@
     "PaymentMethodAuthenticationService/PreAuthenticate",
     "PaymentService/Refund",
     "PaymentService/Void"
-  ]
+  ],
+  "request_id_source_field": "merchant_transaction_id",
+  "request_id_prefix": "0001",
+  "request_id_length": 12,
```
AmitsinghTanwar007: instead of adding this here can you try to add this in auto_gen.rs with a parameter of connector
go through the codebase and come up with a better solution for this feature

crates/internal/integration-tests/src/harness/scenario_api.rs
```diff
@@ -193,12 +193,17 @@ fn connector_request_reference_id_for(
...
     // Load connector spec to check for custom reference ID configuration.
     // Silently ignore errors — an absent spec means default behaviour.
     if let Some(spec) = load_connector_spec(connector) {
-        if let Some(source_field) = spec.request_id_source_field.as_deref() {
+        let source_field_owned: Option<String> = spec
+            .request_id_source_field_per_suite
+            .get(suite)
+            .cloned()
+            .or_else(|| spec.request_id_source_field.clone());
+        if let Some(source_field) = source_field_owned.as_deref() {
```
AmitsinghTanwar007: i dont think so this logic should come here

AmitsinghTanwar007 (approved):

shuklatushar226 (approved):

PR #1372, opened 2026-05-19
feat(sdk): add Go SDK

Jeeva0104:
## Summary
Introduces a new Go SDK for the Hyperswitch Prism connector service.

## Changes
- Add Go SDK client library (`sdk/go/payments/`) with payments, connector, HTTP, config, errors, and result types
- Add FFI bindings via UniFFI for connector service integration (`sdk/go/uniffi/`)
- Add unit tests (`sdk/go/tests/`) and smoke tests (`sdk/go/smoke-test/`)
- Update code generators (`scripts/generators/code/`) to support Go SDK generation
- Update CI workflow (`.github/workflows/ci.yml`) for Go SDK
- Update top-level `sdk/Makefile` and `scripts/run_smoke_tests_parallel.py`

## Commits
- add go sdk
- fix(go-sdk): ignore FFI binary libraries
- add go sdk
- add go sdk module

## Test Plan
- [ ] Run `make test` in `sdk/go/`
- [ ] Run smoke tests via `sdk/go/Makefile`
- [ ] Verify CI passes for Go SDK build


sdk/go/.gitignore
```diff
@@ -0,0 +1,4 @@
+# FFI binary libraries (built dynamically, not committed)
+*.dylib
+*.so
+*.dll
```
maverox, line 4: Committed binaries in the repo

The .gitignore here covers FFI libraries but misses compiled Go binaries. There are two binaries already tracked in git:
- sdk/go/smoke-test/smoke-test
- sdk/go/composite

These are build artifacts that will drift from source over time and bloat the repo. Suggest adding:

```
smoke-test/smoke-test
composite
```

And then `git rm --cached` both files.

sdk/go/payments/http_client.go
```diff
@@ -0,0 +1,116 @@
...
+
+	respBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %w", err)
+	}
+
+	headers := make(map[string]string)
+	for k, v := range resp.Header {
+		if len(v) > 0 {
+			headers[k] = v[0]
```
maverox: Multi-value headers are silently dropped

Only the first header value is kept here. Some connectors return multi-value headers like `Set-Cookie`. Consider joining them instead:

```go
headers[k] = strings.Join(v, ", ")
```

That way downstream consumers don't lose header values unexpectedly.

sdk/go/payments/result.go
```diff
@@ -0,0 +1,68 @@
...
+	case pb.FfiResult_HTTP_REQUEST:
+		return result.GetHttpRequest(), nil
+	case pb.FfiResult_INTEGRATION_ERROR:
+		return nil, &IntegrationError{Proto: result.GetIntegrationError()}
+	case pb.FfiResult_CONNECTOR_ERROR:
+		return nil, &ConnectorError{Proto: result.GetConnectorError()}
+	default:
+		return nil, fmt.Errorf("unhandled result type: %v", result.GetType())
+	}
+}
```
maverox, line 28: Default cases mask new FfiResult enum variants

The `default` branches in CheckReq, CheckRes, and CheckDirect mean that if a new FfiResult variant is added upstream, it falls through to a generic error at runtime instead of failing to compile. That's defensive, but it also hides contract drift.

Consider explicitly listing all known enum variants and treating anything unmatched as an error that gets logged loudly, or document why `default` is the right choice here given that the enum is controlled by the same FFI crate.

Jeeva0104:
@maverox Regarding the `default` cases in `result.go`:

Go does not have exhaustive enum switch checking, so removing `default` would not cause a compile-time failure. An unmatched case simply falls through silently.

Without `default`, `CheckReq` would return `nil, nil` for unknown variants, which callers would misinterpret as success. The `default` case is intentionally defensive here - it turns a silent failure into an explicit runtime error.

Listing all known variants explicitly is reasonable for readability, but the compile-time safety argument does not apply in Go.

PR #1436, opened 2026-06-01
parity(braintree/infra): add serde(default) for repeated proto fields

10X-GRACE:
## Verdict in one line
Braintree (and any future connector with `repeated` proto fields): Prism's `build.rs` did not add `#[serde(default)]` for `Vec<T>` fields generated from proto `repeated`. Per protobuf semantics, repeated fields default to empty. Fix in Prism `build.rs`.

## Decision trail
- g2h `add_skip_null_attribute_static` (`lib.rs:L1073-L1096`):
  ```rust
  // Handles optional/message → skip_serializing_if = "Option::is_none"
  // Handles string (non-repeated) → skip_serializing_if = "String::is_empty"
  // Does NOT handle repeated → Vec<T> fields get no serde(default)
  ```
- BraintreeConfig proto (`payment.proto:L3805-L3820`):
  ```protobuf
  repeated string apple_pay_supported_networks = 53;
  repeated string apple_pay_merchant_capabilities = 54;
  repeated string gpay_allowed_auth_methods = 58;
  repeated string gpay_allowed_card_networks = 59;
  ```
- HS sends JSON without these fields → Prism fails with "missing field" → gRPC INVALID_ARGUMENT
- Protobuf semantics: repeated fields always default to empty — `#[serde(default)]` is always correct
- Fix direction: Prism build.rs (g2h is external, fix applied before compile_protos_with_config)

## Raw evidence
- PRD issue: https://github.com/juspay/hyperswitch-cloud/issues/16464
- Sibling issues: #16434 (braintree authorize), #16435 (braintree capture)
- g2h library: `juspay/g2h` rev `399d5257` — `add_skip_null_attribute_static` has no `Label::Repeated` branch

## Diff
The fix loads proto file descriptors in `build.rs` and adds `#[serde(default)]` for all repeated non-enum fields before passing the config to g2h. This is general: any future connector config that adds `repeated` fields will automatically get the attribute.

Repeated enum fields are excluded because g2h's enum serializer already adds `default` for those — adding it twice would cause a serde compile error.

## Verification
<details><summary>cargo build -p grpc-api-types (clean)</summary>

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 58.96s
```
</details>
<details><summary>cargo clippy -p grpc-api-types (clean)</summary>

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 28.79s
```
</details>
<details><summary>cargo test -p grpc-api-types (PASS)</summary>

```
running 2 tests
test tests::parity_16464_braintree_config_missing_repeated_fields ... ok
test tests::parity_16464_braintree_config_with_repeated_fields ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```
</details>

## What this PR is NOT
- Does NOT modify g2h (external dependency) — a general g2h fix should follow separately
- Does NOT touch connector transformers
- Does NOT touch `ucs_interface_common/src/auth.rs`
- PRD `Out of scope`: connector API field mapping (this is internal config transport)

## Acceptance Criteria
- [x] Build clean
- [x] Clippy clean
- [x] Regression tests added and passing (2 tests)
- [x] Fix is general (handles all repeated non-enum fields, not just BraintreeConfig)
- [ ] Shadow-replay produces zero diff (verified by next regular `run_all.sh` cycle; not blocking merge)

### DIFF
--- a/crates/types-traits/grpc-api-types/build.rs
+++ b/crates/types-traits/grpc-api-types/build.rs
@@ -83,10 +83,7 @@ fn add_serde_default_recursive(config: &mut prost_build::Config, message: &Descr
     for field in &message.field {
         if field.label() == Label::Repeated && field.r#type() != Type::Enum {
             let field_name = field.name.as_deref().unwrap_or_default();
-            config.field_attribute(
-                format!("{message_name}.{field_name}"),
-                "#[serde(default)]",
-            );
+            config.field_attribute(format!("{message_name}.{field_name}"), "#[serde(default)]");
         }
     }
     for nested in &message.nested_type {
--- a/crates/types-traits/grpc-api-types/build.rs
+++ b/crates/types-traits/grpc-api-types/build.rs
@@ -1,5 +1,10 @@
 use std::{env, path::PathBuf};
 
+use prost_types::{
+    field_descriptor_proto::{Label, Type},
+    DescriptorProto, FileDescriptorSet,
+};
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let out_dir = PathBuf::from(env::var("OUT_DIR")?);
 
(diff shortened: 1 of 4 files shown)


crates/types-traits/grpc-api-types/build.rs
```diff
@@ -32,40 +37,59 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    //         ],
-    //         &["proto"],
-    //     )?;
+    bridge_generator.compile_protos_with_config(config, protos, &["proto"])?;
 
     Ok(())
 }
+
+/// Add `#[serde(default)]` for all repeated non-enum proto fields.
+///
```
XyneSpaces, line 67: :warning: Recursive function lacks recursion termination depth protection

The `add_serde_default_recursive` function traverses nested message types without tracking depth. Deeply nested proto messages could cause stack overflow during build.

Consider adding a depth limit parameter (e.g., `max_depth: u8`) that decrements on each recursive call and returns early when exhausted.

crates/types-traits/grpc-api-types/build.rs
```diff
@@ -32,40 +37,59 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+/// g2h already adds `default` for repeated enum fields via its enum serializer attributes,
+/// so we skip those to avoid duplicate serde(default) which would cause a compile error.
```
XyneSpaces, line 69: :warning: Empty field name produces malformed field attribute

Using `unwrap_or_default()` on `field.name` means a field without a name gets an empty string, resulting in a malformed field attribute path like `"{message_name}."`.

Either filter out unnamed fields or handle this case explicitly with a clear error message indicating which proto field is problematic.

crates/types-traits/grpc-api-types/build.rs
```diff
@@ -32,40 +37,59 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    fds: &FileDescriptorSet,
+) {
+    for file in &fds.file {
+        for message in &file.message_type {
+            add_serde_default_recursive(config, message);
+        }
+    }
+}
+
+fn add_serde_default_recursive(config: &mut prost_build::Config, message: &DescriptorProto) {
```
XyneSpaces, line 81: :bulb: Missing negative test for non-repeated fields

The tests verify `#[serde(default)]` is correctly applied to repeated fields, but there's no test ensuring non-repeated fields are NOT decorated with `#[serde(default)]`.

Add a test case that deserializes JSON with missing non-repeated fields and verifies it fails as expected, confirming the fix doesn't over-apply the attribute.

XyneSpaces (changes requested):
:rotating_light: Potential panic risk in recursive function when message name is missing

The `add_serde_default_recursive` function uses `unwrap_or_default()` on the message name, which could lead to incorrect field attribute paths when the proto descriptor lacks a name. An empty message name would produce malformed field attributes like `".{field_name}"`.

Add an explicit check for missing message names and either skip the message or propagate the error rather than silently using an empty string.

10X-GRACE:
Closing — superseded by #1465.

Both PRs fix the same root cause (BraintreeConfig X-Connector-Config rejected because `repeated Vec<T>` lacked `#[serde(default)]`). #1465 takes the narrower, lower-risk approach: type-level `#[serde(default)]` on `BraintreeConfig` only, via `prost_build::Config::type_attribute`. This avoids the recursive build.rs traversal in this PR (which XyneSpaces flagged for a panic risk on missing message names).

Rollout will use #1465. Closing this one to avoid two parallel fixes for the same issue.

cc /cc @AmitsinghTanwar007 @XyneSpaces

PR #1465, opened 2026-06-04
fix(grpc-api-types): apply #[serde(default)] to BraintreeConfig to accept partial X-Connector-Config JSON

10X-GRACE:
## Description

`BraintreeConfig` has four `repeated string` fields (`apple_pay_supported_networks`, `apple_pay_merchant_capabilities`, `gpay_allowed_auth_methods`, `gpay_allowed_card_networks`). `prost_build` generates them as bare `Vec<String>`, and the `g2h`-derived `serde::Deserialize` impl treats `Vec<T>` as required by default.

Hyperswitch does not serialize those four wallet-config fields for standard card flows. Result: every braintree `X-Connector-Config` JSON payload was being rejected at the gRPC entry layer with:

```
Failed to parse X-Connector-Config JSON into ConnectorSpecificConfig
```

Net effect: every braintree call (any flow, any payment method) failed before reaching the connector transformer.

## Motivation and Context

The fix applies `#[serde(default)]` to `BraintreeConfig` at the type level via `prost_build::Config::type_attribute`. Any missing field then uses its `Default::default()` value (`Vec` → empty, `Option` → `None`, `String` → `""`), which is the correct proto3 semantics for `repeated` and also future-proofs new BraintreeConfig fields.

Other `ConnectorSpecificConfig` variants (Adyen, Cybersource, Stripe, Paypal, etc.) use `Option<...>` for every field, so serde's missing-field handling already produces `None` for them — they were unaffected. Broader hardening to apply `#[serde(default)]` to all `*Config` types is left for a separate change.

## How did you test it?

- `cargo check -p grpc-api-types` clean
- Inspected generated `BraintreeConfig` struct after build — `#[serde(default)]` is correctly emitted at the type level

## Development

Closes: juspay/hyperswitch-cloud#16580

PR #1473, opened 2026-06-05
fix(connector): default Braintree repeated config fields

AmitsinghTanwar007:
## Summary
- Add serde defaults for Braintree repeated connector config fields generated from protobuf.
- Prevent missing optional repeated fields from failing JSON deserialization when HS omits them.

## Verification
- cargo fmt --check
- cargo check -p grpc-api-types

### DIFF
--- a/crates/types-traits/grpc-api-types/Cargo.toml
+++ b/crates/types-traits/grpc-api-types/Cargo.toml
@@ -23,5 +23,6 @@ hyperswitch_masking = { version = "0.0.1", default-features = false, features = 
 tonic-build = { workspace = true }
 prost-build = { workspace = true }
 prost = { workspace = true }
+prost-types = { workspace = true }
 heck = "0.5.0"
 g2h = { git = "https://github.com/juspay/g2h", rev = "399d5257744cabffe88d9871a1192eacbbe9f484" }
--- a/crates/types-traits/grpc-api-types/build.rs
+++ b/crates/types-traits/grpc-api-types/build.rs
@@ -17,6 +17,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         "::hyperswitch_masking::Secret<String>",
     );
 
+    // Add #[serde(default)] for all `repeated` (Vec) fields in connector config messages.
+    // Protobuf `repeated` fields map to Vec<T> in Rust. serde requires `#[serde(default)]`
+    // on Vec fields to treat a missing JSON key as an empty vec; without it, serde errors
+    // with "missing field" when the sender (HS) omits the field.  g2h already adds `default`
+    // for repeated *enum* fields, but not for repeated string/message fields, so we add it
+    // explicitly for each known repeated non-enum field in the connector config messages.
+    // BraintreeConfig is currently the only ConnectorConfig message with repeated fields.
+    for field in [
+        ".types.BraintreeConfig.apple_pay_supported_networks",
+        ".types.BraintreeConfig.apple_pay_merchant_capabilities",
+        ".types.BraintreeConfig.gpay_allowed_auth_methods",
+        ".types.BraintreeConfig.gpay_allowed_card_networks",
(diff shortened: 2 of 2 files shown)


XyneSpaces (commented):
## Review Summary

:rotating_light: 0 critical · :warning: 0 warning · :bulb: 0 suggestion · :mag: 0 nitpick

**Classification:** proto-api-contract · secondary: none · metadata: none
**Risk:** low

Clean fix adding `#[serde(default)]` to Braintree's repeated config fields. Prevents deserialization failures when HS omits optional repeated fields, matching standard protobuf-to-Rust serde patterns. Build verification confirmed.

XyneSpaces (commented):
## Review Summary

**Classification:** `proto-api-contract`
**Risk Level:** Low
**Files Changed:** 1 file (`build.rs`)

---

### Verification

**Baseline Checks:**
- ✅ No debug/remnant code — clean addition with explanatory comment
- ✅ Proper error handling — `#[serde(default)]` prevents deserialization failures when HS omits optional repeated fields

**Proto Contract Compatibility:**
- ✅ No wire contract break — change only affects generated Rust code deserialization behavior, not proto schema
- ✅ build.rs configuration correct — adds missing serde attribute for repeated string/message Vec fields that g2h doesn't auto-handle

**Implementation Quality:**
- ✅ Explicit field enumeration — all 4 BraintreeConfig repeated fields covered
- ✅ Clear code comment — explains why this is needed and why Braintree specifically
- ✅ Verification steps documented — PR notes `cargo check -p grpc-api-types` passes

---

### Findings

None. This is a correct and complete fix for the JSON deserialization gap.

The change addresses a real issue: when Hyperswitch omits optional repeated fields in the connector config JSON, serde would fail with "missing field" errors. With `#[serde(default)]`, missing keys deserialize to empty Vecs, which is the correct semantic for omitted repeated proto fields.

---

**Verdict:** ✅ **Approve** — Clean, targeted fix for the Braintree config deserialization edge case.

10X-GRACE:
Heads-up — this overlaps with #1465 (`parity/braintree/connector-config-serde-default`, opened 2026-06-04). Posting in case the duplication wasn't visible.

Both fix the same root cause: `BraintreeConfig`'s four `repeated string` fields (`apple_pay_supported_networks`, `apple_pay_merchant_capabilities`, `gpay_allowed_auth_methods`, `gpay_allowed_card_networks`) compile to bare `Vec<String>`, and serde rejects JSON that omits them — so every braintree X-Connector-Config call HS sends fails with `Failed to parse X-Connector-Config JSON into ConnectorSpecificConfig` (per hs-cloud#16580).

### How the two approaches differ

| | #1473 (this PR) | #1465 |
|---|---|---|
| Mechanism | `field_attribute` × 4 named fields | `type_attribute(".types.BraintreeConfig", "#[serde(default)]")` |
| LOC | ~8 | 1 |
| Coverage today | 4 known `repeated` fields | same 4 — identical accepted JSON set |
| New 5th `repeated` field added to BraintreeConfig later | Bug recurs until someone updates the list manually | Covered automatically (type-level default handles any missing field) |
| Pairs with g2h upstream fix | n/a | #1465 explicitly notes the truly-future-proof path is auto-injecting `#[serde(default)]` on every `repeated` field in [juspay/g2h](https://github.com/juspay/g2h) at codegen time — would immunize every consumer, not just braintree |

For BraintreeConfig as it stands **today**, both PRs produce the **same** accepted-JSON set — all non-`repeated` fields are already `optional` in the proto (→ `Option<T>` → serde accepts missing). The behavioral delta only shows up if someone later adds a new `repeated` field.

Happy to defer if you have a strong preference for the per-field explicit style. Otherwise #1465 is the same fix with one line + future-proofing for new fields, and it pairs with a follow-up g2h PR that retires the per-connector-config workaround entirely.

cc @XyneSpaces @JeevaRamu0104

PR #1481, opened 2026-06-08
feat(connector): deutschebank payouts

heykulthe:
## Description

This pr adds support for deutschebank payouts

### Additional Changes

## How did you test it?

Tested curls via hyperswitch:

MCA:
```
{
    "connector_type": "payout_processor",
    "connector_name": "deutschebank",
    "connector_account_details": {
        "auth_type": "MultiAuthKey",
        "api_key": "api_key",
        "key1": "key1",
        "api_secret": "api_secret",
        "key2": "key2",
    },
    "test_mode": true,
    "disabled": false,
    "payment_methods_enabled": [
        {
            "payment_method": "bank_transfer",
            "payment_method_types": [
                {
                    "payment_method_type": "sepa_bank_transfer",
                    "minimum_amount": 1,
                    "maximum_amount": 99999999,
                    "recurring_enabled": false,
                    "installment_payment_enabled": false
                }
            ]
        }
    ],
    "metadata": {}
}
```

Payout Create:
```bash
{
    "amount": 100,
    "currency": "EUR",
    "customer_id": "payout_customer121",
    "email": "payout_customer@example.com",
    "name": "John Doe",
    "phone": "999999999",
    "phone_country_code": "+49",
    "description": "any-purpose",
    "payout_type": "bank",
    "payout_method_data": {
        "bank_transfer": {
            "payout_method_type": "sepa",
            "iban": "DE94500700101234580002",
            "bic": "DEUTDEDB237",
            "bank_name": "Deutsche Bank",
            "bank_country_code": "DE",
            "account_holder_name": "John Doe"
        }
    },
    "source_bank_data": {
        "payout_method_type": "sepa",
        "iban": "DE83215730130100853100",
        "bic": "DEUTDEDB237"
    },
    "billing": {
        "address": {
            "city": "Frankfurt",
            "country": "DE",
            "line1": "Taunusanlage 12",
            "zip": "60325",
            "state": "HE",
            "first_name": "John",
            "last_name": "Doe"
        }
    },
    "entity_type": "Individual",
    "recurring": false,
    "auto_fulfill": true,
    "confirm": true
}
```

Response:
```
{
    "payout_id": "payout_5xkbahwmINYfmiFEq1FF",
    "merchant_id": "merchant_1781202298",
    "merchant_order_reference_id": null,
    "amount": 100,
    "currency": "EUR",
    "connector": "deutschebank",
    "payout_type": "bank",
    "payout_method_data": {
        "bank": {
            "iban": "DE945************80002",
            "bank_name": "Deutsche Bank",
            "bank_country_code": "DE",
            "bank_city": null,
            "bic": "DEU*****237",
            "account_holder_name": "John Doe"
        }
    },
    "source_bank_data": {
        "iban": "DE832************53100",
        "bank_name": null,
        "bank_country_code": null,
        "bank_city": null,
        "bic": "DEU*****237",
        "account_holder_name": null
    },
    "billing": {
        "address": {
            "city": "Frankfurt",
            "country": "DE",
            "line1": "Taunusanlage 12",
            "line2": null,
            "line3": null,
            "zip": "60325",
            "state": "HE",
            "first_name": "John",
            "last_name": "Doe",
            "origin_zip": null
        },
        "phone": null,
        "email": null
    },
    "auto_fulfill": true,
    "customer_id": "payout_customer121",
    "customer": {
        "id": "payout_customer121",
        "name": "John Doe",
        "email": "payout_customer@example.com",
        "phone": "999999999",
        "phone_country_code": "+49",
        "customer_document_details": null
    },
    "client_secret": "payout_payout_5xkbahwmINYfmiFEq1FF_secret_7wMwRnYC55wXdmVLN1jV",
    "return_url": null,
    "business_country": null,
    "business_label": null,
    "description": "any-purpose",
    "entity_type": "Individual",
    "recurring": false,
    "metadata": {},
    "merchant_connector_id": "mca_yJvQyOlJdC5I6wVwkhVj",
    "status": "pending",
    "error_message": null,
    "error_code": null,
    "profile_id": "pro_8OWuQTJvJbvQhu9MMIXr",
    "created": "2026-06-12T08:22:26.449Z",
    "connector_transaction_id": "AAD8AA027E0859FABF0B5887C26806DA|DE83215730130100853100",
    "priority": null,
    "payout_link": null,
    "email": "payout_customer@example.com",
    "name": "John Doe",
    "phone": "999999999",
    "phone_country_code": "+49",
    "unified_code": null,
    "unified_message": null,
    "payout_method_id": null
}
```

PoSync Response:
```
{
    "payout_id": "payout_5xkbahwmINYfmiFEq1FF",
    "merchant_id": "merchant_1781202298",
    "merchant_order_reference_id": null,
    "amount": 100,
    "currency": "EUR",
    "connector": "deutschebank",
    "payout_type": "bank",
    "payout_method_data": {
        "bank": {
            "iban": "DE945************80002",
            "bank_name": "Deutsche Bank",
            "bank_country_code": "DE",
            "bank_city": null,
            "bic": "DEU*****237",
            "account_holder_name": "John Doe"
        }
    },
    "source_bank_data": {
        "iban": "DE832************53100",
        "bank_name": null,
        "bank_country_code": null,
        "bank_city": null,
        "bic": "DEU*****237",
        "account_holder_name": null
    },
    "billing": {
        "address": {
            "city": "Frankfurt",
            "country": "DE",
            "line1": "Taunusanlage 12",
            "line2": null,
            "line3": null,
            "zip": "60325",
            "state": "HE",
            "first_name": "John",
            "last_name": "Doe",
            "origin_zip": null
        },
        "phone": null,
        "email": null
    },
    "auto_fulfill": true,
    "customer_id": "payout_customer121",
    "customer": {
        "id": "payout_customer121",
        "name": "John Doe",
        "email": "payout_customer@example.com",
        "phone": "999999999",
        "phone_country_code": "+49",
        "customer_document_details": null
    },
    "client_secret": "payout_payout_5xkbahwmINYfmiFEq1FF_secret_7wMwRnYC55wXdmVLN1jV",
    "return_url": null,
    "business_country": null,
    "business_label": null,
    "description": "any-purpose",
    "entity_type": "Individual",
    "recurring": false,
    "metadata": {},
    "merchant_connector_id": "mca_yJvQyOlJdC5I6wVwkhVj",
    "status": "success",
    "error_message": null,
    "error_code": null,
    "profile_id": "pro_8OWuQTJvJbvQhu9MMIXr",
    "created": "2026-06-12T08:22:26.449Z",
    "connector_transaction_id": "AAD8AA027E0859FABF0B5887C26806DA|DE83215730130100853100",
    "priority": null,
    "payout_link": null,
    "email": "payout_customer@example.com",
    "name": "John Doe",
    "phone": "999999999",
    "phone_country_code": "+49",
    "unified_code": null,
    "unified_message": null,
    "payout_method_id": null
}
```

Test cases passing as expected:
| Debtor IBAN | Eligibility | Transfer | Get | DB fixture |
|---|---|---|---|---|
| `DE83215730130100853100` | `REQUIRES_FULFILLMENT` | `PENDING` | `SUCCESS` | MTCH → ACCP |
| `DE99123456123456789001` | `REQUIRES_FULFILLMENT` | `PENDING` | `PENDING` (settles later) | CMTC → PDNG |
| `DE99123456123456789101` | 400 `APP-RULE` | — | — | NOAP-true (DB fixture returns APP-RULE here, raised with onboarding) |
| `DE99123456123456789003` | `INELIGIBLE`, `payoutEligible=false` | (skipped, correct) | — | NOAP-false → orchestrator stops |
| `DE71215730110426340602` | `INELIGIBLE`, `payoutEligible=false` | (skipped, correct) | — | NMTC → gateway hard-block |
| `DE71215730110426340600` | `REQUIRES_FULFILLMENT` | 400 `APP-RULE` | — | MTCH but VoP-ID expired |

**Note**
mTLS create_client no longer treats the client identity cert as a trust anchor — server verification now uses webpki roots + explicit ca_certificate_pem; itaubank (webpki-only) is unaffected as its endpoints chain to public CAs.


XyneSpaces (dismissed):
## Review Summary

**Classification:** Connector-New-Integration — Deutsche Bank Payouts  
**Scenarios:** `connector-new-integration`, `core-flow-framework`  
**Files Changed:** 29 files (3 new connector modules, 6 config/env files, 20 registration/integration files)  
**Risk Level:** Medium-High (new connector integration with cryptographic signing)

---

### Baseline Verification (B1–B7)

| Check | Status | Notes |
|-------|--------|-------|
| B1 File coverage | ✅ | All 29 files reviewed |
| B2 Code integrity | ⚠️ | See S1 finding on error swallowing |
| B3 Code correctness | ⚠️ | See S1 finding on default error responses |
| B4 Companion sync | ✅ | Registrations appear complete |
| B5 Comment scope | ✅ | Code-only comments |
| B6 No YAGNI | ✅ | All code has consumers |
| B7 No scope creep | ✅ | Focused PR |

---

### Findings by Severity

#### :rotating_light: S1 — Major Blockers

**1. Error response construction silently swallows deserialization failures**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs:163-170`

The `build_error_response` uses `unwrap_or_default()` when parsing error responses, which silently discards deserialization errors and returns an empty default struct. This makes debugging production issues difficult.

```rust
let response: DeutschebankErrorResponse = res
    .response
    .parse_struct("DeutschebankErrorResponse")
    .unwrap_or_default();  // ← Silently drops parse errors
```

**Fix:** Propagate the error or at minimum log it:
```rust
let response: DeutschebankErrorResponse = res
    .response
    .parse_struct("DeutschebankErrorResponse")\    .unwrap_or_else(|e| {
        router_env::logger::warn!("Failed to parse Deutsche Bank error response: {:?}", e);
        DeutschebankErrorResponse::default()
    });
```

---

#### :warning: S2 — Moderate Issues

**2. UUID v5 for `vop_id` derivation may cause collisions across merchants**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs:300-305`

```rust
pub fn derive_vop_id(connector_request_reference_id: &str) -> String {
    uuid::Uuid::new_v5(
        &uuid::Uuid::NAMESPACE_OID,
        connector_request_reference_id.as_bytes(),
    )
    .to_string()
}
```

UUID v5 is deterministic. If two merchants use the same `connector_request_reference_id` (e.g., both use "payout-123"), they will get the same `vop_id`. This could lead to:
- Cache collisions in Deutsche Bank's systems
- Potential security issues if one merchant can predict another's VoP ID

**Fix:** Include a merchant-specific salt or use UUID v4 for randomness:
```rust
pub fn derive_vop_id(merchant_id: &str, connector_request_reference_id: &str) -> String {
    let input = format!("{}:{}", merchant_id, connector_request_reference_id);
    uuid::Uuid::new_v5(&uuid::Uuid::NAMESPACE_OID, input.as_bytes()).to_string()
}
```

---

**3. Date formatting fallback uses Unix epoch on failure**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs:55-62`

```rust
fn format_http_date(dt: OffsetDateTime) -> String {
    // ...
    time::OffsetDateTime::now_utc()
        .format(&fmt)
        .unwrap_or_else(|_| "1970-01-01T00:00:00Z".to_string())  // ← Falls back to epoch
}
```

If formatting fails, the request is sent with a 1970 date, which will cause immediate rejection by Deutsche Bank's servers. This makes debugging harder.

**Fix:** Return a `Result` and propagate the error, or panic in this case (since it indicates a programming error):
```rust
fn format_http_date(dt: OffsetDateTime) -> Result<String, IntegrationError> {
    let fmt = format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]Z");
    dt.format(&fmt)
        .change_context(IntegrationError::RequestEncodingFailed { context: /* ... */ })
}
```

---

**4. Unused `_ensure_expose_in_scope` function**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs:823-826`

```rust
#[allow(dead_code)]
fn _ensure_expose_in_scope(s: Secret<String>) -> String {
    s.expose()
}
```

This function appears to be dead code added to silence a lint. If it's not needed, remove it. If it's needed for compilation, add a comment explaining why.

---

#### :bulb: S3 — Suggestions / Nits

**5. Consider using constant for magic string separators**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs:23`

```rust
const VOP_ID_SEPARATOR: char = '|';
```

Good use of a named constant. Ensure this separator doesn't conflict with valid characters in Deutsche Bank's connector payout ID format.

---

**6. Correlation ID prefixes could be constants**

`crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs:114-117`

The prefixes `"ACID"` and `"PYMT"` are hardcoded inline. Consider defining them as constants:
```rust
const CORRELATION_PREFIX_VOP: &str = "ACID";
const CORRELATION_PREFIX_PAYMENT: &str = "PYMT";
```

---

**7. Documentation typo in module comment**

`crates/integrations/connector-integration/src/connectors/deutschebank.rs:1`

```rust
// Deutsche Bank is a payouts-only connector in prism.
```

Capitalization: "prism" should be "Prism" for consistency.

---

### Verification Checklist (Per Scenario)

#### Connector-New-Integration (CN1-CN6)

| Check | Status | Notes |
|-------|--------|-------|
| CN1 Registration completeness | ✅ | All 8 registration points verified |
| CN2 Copy-paste detection | ✅ | No other connector names found |
| CN3 Auth correctness | ✅ | CSEAL signing follows RFC draft standards |
| CN4 Request/response transformers | ⚠️ | Error response handling needs fix (S1) |
| CN5 Unsupported flows explicit | ✅ | All stubbed with `connector_flow_not_implemented` |
| CN6 Test/spec coverage | ⚠️ | Manual testing documented; no automated tests in PR |

#### Core-Flow-Framework (CF1-CF4)

| Check | Status | Notes |
|-------|--------|-------|
| CF1 Trait/flow coherence | ✅ | `PayoutEligibility` flow properly integrated |
| CF2 Shared type blast radius | ⚠️ | See S2 - UUID v5 collision risk |
| CF3 Helper/utility changes | ✅ | Signing helpers are connector-specific |
| CF4 Hidden semantic changes | ✅ | No behavior changes to existing flows |

---

### Hard Blockers (Appendix)

- ❌ **S1**: Error swallowing in `build_error_response` - violates B3 (silent error swallowing)
- ⚠️ **S2**: UUID v5 collision risk - recommend fixing before production load

---

### Test Coverage

**Manual testing (documented in PR description):**
| Test Case | Expected | Status |
|-----------|----------|--------|
| Eligibility → REQUIRES_FULFILLMENT | PASS | Verified |
| Transfer → PENDING | PASS | Verified |
| Get → SUCCESS | PASS | Verified |
| Ineligible IBAN | SKIP | Verified |
| 400 APP-RULE | Error | Verified |

**Recommendations:**
1. Add automated connector spec tests in `crates/internal/ucs-connector-tests/src/connector_specs/deutschebank/`
2. Consider unit tests for `sign_rsa_sha256` with known test vectors

---

### Overall Assessment

The implementation is **technically sound** with proper cryptographic signing and mTLS configuration. The main concerns are:
1. **S1 (blocking)**: Error swallowing that will hinder production debugging
2. **S2 (high)**: UUID collision risk that could affect multi-merchant deployments

Once these are addressed, the PR is ready for merge.

---

**Verdict:** :warning: **Request Changes** — Address S1 (error swallowing) before merge. Strongly recommend S2 (UUID collision) fix as well.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -3940,6 +3940,18 @@ message LoonioConfig {
   optional string base_url = 50;
 }
 
+message DeutschebankConfig {
+  SecretString apikey = 1;
+  SecretString customer_identifier = 2;
+  SecretString consumer_identifier = 3;
```
Sakilmostak: what these field represents?
should they exist in MCA?

crates/common/external-services/src/service.rs
```diff
@@ -1057,25 +1058,24 @@ pub fn create_client(
...
+            let mut client_builder = client_builder.identity(identity).use_rustls_tls();
+            if let Some(ca_pem) = additional_ca_pem {
+                for ca_cert in create_certificate(ca_pem)? {
+                    client_builder = client_builder.add_root_certificate(ca_cert);
+                }
+            }
             client_builder
-                .identity(identity)
-                .use_rustls_tls()
                 .build()
```
Sakilmostak: cc: @SanchithHegde 
can you review these changes

crates/types-traits/grpc-api-types/proto/payment.proto
heykulthe: customer_identifier is the one that exist in MCA. consumer_identifier is now removed.

crates/common/external-services/src/service.rs
```diff
@@ -1057,25 +1058,24 @@ pub fn create_client(
...
-            let client_builder = certificate_list
-                .into_iter()
-                .fold(client_builder, |client_builder, certificate| {
-                    client_builder.add_root_certificate(certificate)
-                });
+            let mut client_builder = client_builder.identity(identity).use_rustls_tls();
+            if let Some(ca_pem) = additional_ca_pem {
+                for ca_cert in create_certificate(ca_pem)? {
+                    client_builder = client_builder.add_root_certificate(ca_cert);
+                }
```
iemyashasvi: avoid using if and for loops

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,193 @@
...
+        Month::April => "Apr",
+        Month::May => "May",
+        Month::June => "Jun",
+        Month::July => "Jul",
+        Month::August => "Aug",
+        Month::September => "Sep",
+        Month::October => "Oct",
+        Month::November => "Nov",
+        Month::December => "Dec",
+    };
```
iemyashasvi: check if we already have a util for the same ?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,193 @@
...
+    let octet_string = der_wrap(0x04, pkcs1);
+    let version = vec![0x02, 0x01, 0x00]; // INTEGER 0
+    let mut content = Vec::with_capacity(version.len() + ALG_ID.len() + octet_string.len());
+    content.extend_from_slice(&version);
+    content.extend_from_slice(ALG_ID);
+    content.extend_from_slice(&octet_string);
+    der_wrap(0x30, &content)
+}
+
+fn der_wrap(tag: u8, data: &[u8]) -> Vec<u8> {
```
iemyashasvi: can we use better function names

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+// ===== ERROR RESPONSE =====
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct DeutschebankErrorResponse {
+    pub code: Option<String>,
+    pub message: Option<String>,
+    pub reason: Option<String>,
+    #[serde(rename = "errorCode")]
+    pub error_code: Option<String>,
+    #[serde(rename = "errorMessage")]
+    pub error_message: Option<String>,
```
iemyashasvi: we can use `serde rename_all='CamelCase' ` at the top of struct itself

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+        >,
+    ) -> Result<Self, Self::Error> {
+        let payee_iban = extract_payee_iban(req.request.payout_method_data.as_ref())?;
+        let debtor_iban = extract_debtor_iban(req.request.source_bank_data.as_ref())?;
+        let payee_name = req
+            .request
+            .customer
+            .as_ref()
+            .and_then(|c| c.name.clone())
+            .map(Secret::new)
```
iemyashasvi: use util here

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+#[derive(Debug, Serialize)]
+pub struct DeutschebankCustomerCreditTransferInitiation {
+    #[serde(rename = "groupHeader")]
+    pub group_header: DeutschebankGroupHeader,
+    #[serde(rename = "paymentInformation")]
+    pub payment_information: Vec<DeutschebankPaymentInformation>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct DeutschebankGroupHeader {
```
iemyashasvi, line 376: use rename_all ='camelCase '

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+    pub initiating_party: DeutschebankParty,
+}
+
+#[derive(Debug, Serialize)]
+pub struct DeutschebankParty {
+    pub name: Secret<String>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct DeutschebankPaymentInformation {
```
iemyashasvi, line 391: same use rename_all =camelCase

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+        let amount = utils::convert_amount(
+            &FloatMajorUnitForConnector,
+            req.request.amount,
+            req.request.destination_currency,
+        )?;
+        let purpose_code = req
+            .resource_common_data
+            .description
+            .clone()
+            .unwrap_or_else(|| "PAYOUT".to_string());
```
iemyashasvi: dont use add fallback if this is a requiredfield ,also avoid using `MagicStrings`

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+                    },
+                },
+                payment_information: vec![DeutschebankPaymentInformation {
+                    payment_information_identification: message_id.clone(),
+                    payment_method: "TRF",
+                    batch_booking: false,
+                    control_sum: amount,
+                    number_of_transactions: "1".to_string(),
+                    payment_type_information: DeutschebankPaymentTypeInformation {
+                        service_level: DeutschebankCode { code: "SEPA" },
```
iemyashasvi: avoid hardcoding

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+        .format(&fmt)
+        .unwrap_or_else(|_| "1970-01-01T00:00:00Z".to_string())
+}
+
+fn current_iso_date() -> String {
+    use time::format_description::well_known::Iso8601;
+    let date = time::OffsetDateTime::now_utc().date();
+    date.format(&Iso8601::DATE)
+        .unwrap_or_else(|_| "1970-01-01".to_string())
+}
```
iemyashasvi: dont hardcode

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,829 @@
...
+        ),
+    }
+}
+
+fn current_iso_utc_seconds() -> String {
+    use time::macros::format_description;
+    let fmt = format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]Z");
+    time::OffsetDateTime::now_utc()
+        .format(&fmt)
+        .unwrap_or_else(|_| "1970-01-01T00:00:00Z".to_string())
```
iemyashasvi: dont fallback to a hardcoded date

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
```diff
@@ -0,0 +1,868 @@
...
+    }
+}
+
+fn current_iso_utc_seconds() -> String {
+    use time::macros::format_description;
+    let fmt = format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]Z");
+    time::OffsetDateTime::now_utc()
+        .format(&fmt)
+        .unwrap_or_else(|_| "1970-01-01T00:00:00Z".to_string())
+}
```
iemyashasvi: dont hardcode

gopikrishna000 (dismissed):

config/development.toml
```diff
@@ -167,6 +167,8 @@ truelayer.base_url = "https://api.truelayer-sandbox.com"
 truelayer.secondary_base_url = "https://auth.truelayer-sandbox.com"
 trustly.base_url = "https://test.trustly.com/api/1"
 itaubank.base_url = "https://sandbox.devportal.itau.com.br"
+deutschebank.base_url = "https://api.sbx.baas.db.com"
+deutschebank.server_ca_bundle = ""
```
deepanshu-iiitu, line 172: Why is this empty?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,178 @@
...
+use hyperswitch_masking::{PeekInterface, Secret};
+use ring::{rand, signature};
+use sha2::{Digest, Sha256};
+use time::{macros::format_description, OffsetDateTime};
+
+pub struct CsealHeaders {
+    pub date: String,
+    pub digest: Option<String>,
+    pub signature: String,
+}
```
deepanshu-iiitu, line 16: Should these fields be secret?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,178 @@
...
+) -> Result<CsealHeaders, error_stack::Report<IntegrationError>> {
+    let date = format_http_date(OffsetDateTime::now_utc())?;
+
+    let method_lower = match method {
+        Method::Get => "get",
+        Method::Post => "post",
+        Method::Put => "put",
+        Method::Delete => "delete",
+        Method::Patch => "patch",
+    };
```
deepanshu-iiitu: move this to a util

crates/integrations/connector-integration/src/payout_connectors/deutschebank/test.rs
deepanshu-iiitu, line 1: Why is this filed added? 
`crates/integrations/connector-integration/src/payout_connectors/deutschebank/test.rs`

config/development.toml
heykulthe: DB's server cert is signed by a private CA so for us, this value will be added in argocd config files

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
heykulthe: signature gets into_masked() at the header push, date and digest aren't secrets

crates/integrations/connector-integration/src/payout_connectors/deutschebank/test.rs
heykulthe: Same test.rs pattern as adyen, razorpay, cashfree. covers unit tests for PEM splitting and id round trip logic

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
heykulthe: done

deepanshu-iiitu (dismissed):

config/development.toml
iemyashasvi: @heykulthe  please confirm the correct place to add it from @AmitsinghTanwar007  , check if this is even needed in the first place ?

config/production.toml
```diff
@@ -131,6 +131,8 @@ truelayer.secondary_base_url = "https://auth.truelayer.com"
 trustly.base_url = "https://api.trustly.com/1"
 itaubank.base_url = "https://api.itau.com.br"
 itaubank.secondary_base_url = "https://sts.itau.com.br"
+deutschebank.base_url = "https://api.baas.db.com"
+# deutschebank.server_ca_bundle = ""
```
iemyashasvi: why is it commented here ?

crates/integrations/connector-integration/src/connectors/deutschebank.rs
iemyashasvi, line 1: why are we making changes to the legacy payins connectors . I think payouts are completely decoupled in hyperswitch-prism

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+            body: &input[block_start..block_end],
+        });
+        cursor = block_end;
+    }
+    blocks
+}
+
+pub(super) fn split_pem_bundle(
+    bundle: &str,
+) -> Result<(String, String), error_stack::Report<IntegrationError>> {
```
iemyashasvi: this should return a Secret<String> as it contains the certs

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+                client_certificate_bundle,
+                ..
+            } => {
+                let (cert, key) = split_pem_bundle(client_certificate_bundle.peek())?;
+                Ok(Self {
+                    customer_identifier: customer_identifier.to_owned(),
+                    key_id: key_id.to_owned(),
+                    signing_private_key: signing_private_key.to_owned(),
+                    client_certificate: Secret::new(cert),
+                    client_certificate_key: Secret::new(key),
```
iemyashasvi: check if functions can return  Secret<String>

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+            T,
+        >,
+    ) -> Result<Self, Self::Error> {
+        let req = &item.router_data;
+        let payee_iban = extract_payee_iban(req.request.payout_method_data.as_ref())?;
+        let debtor_iban = extract_debtor_iban(req.request.source_bank_data.as_ref())?;
+        let payee_name = extract_customer_name(
+            req.request.customer.as_ref(),
+            "Payee name is required for Deutsche Bank VoP check",
+        )?;
```
iemyashasvi, line 207: why is this extract_customer_name function needed ?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+    #[serde(rename = "paymentInformation")]
+    pub payment_information: Vec<DeutschebankPaymentInformation>,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct DeutschebankGroupHeader {
+    pub message_identification: String,
+    pub creation_date_time: String,
+    pub control_sum: common_utils::types::FloatMajorUnit,
```
iemyashasvi: can we import it on top and only use `FloatMajorUnit` here

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+    pub proprietary: String,
+}
+
+#[derive(Debug, Serialize)]
+pub struct DeutschebankPaymentIdentification {
+    #[serde(rename = "endToEndIdentification")]
+    pub end_to_end_identification: String,
+    #[serde(rename = "instructionIdentification")]
+    pub instruction_identification: String,
+}
```
iemyashasvi, line 468: check if these needs to be secret ?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1042 @@
...
+        )?;
+
+        let reference = req
+            .resource_common_data
+            .connector_request_reference_id
+            .clone();
+
+        let end_to_end_id = uuid::Uuid::new_v5(
+            &uuid::Uuid::NAMESPACE_OID,
+            format!("dbank-e2e:{reference}").as_bytes(),
```
iemyashasvi: why are we generating uuid here , shouldn't this be passed from upstream , please verify once ?

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
```diff
@@ -0,0 +1,695 @@
+pub mod signing;
+pub mod transformers;
+
+#[cfg(test)]
```
iemyashasvi: ?

crates/integrations/connector-integration/src/connectors.rs
```diff
@@ -265,6 +265,9 @@ pub use self::tsys_transit::TsysTransit;
 pub mod twoc_twop_paco;
 pub use self::twoc_twop_paco::TwocTwopPaco;
 
+pub mod deutschebank;
+pub use self::deutschebank::Deutschebank;
+
```
iemyashasvi: nit : confirm if this needs to added as already added it in payouts connectors

crates/integrations/connector-integration/Cargo.toml
```diff
@@ -16,7 +16,7 @@ serde_json = { workspace = true, features = ["preserve_order"] }
 serde_qs = "0.12.0"
 serde_urlencoded = "0.7.1"
 serde_with = "3.7.0"
-uuid = { version = "1", features = ["v4"] }
+uuid = { version = "1", features = ["v4", "v5"] }
```
iemyashasvi, line 19: @AmitsinghTanwar007  please verify if adding this is fine

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -5643,6 +5652,9 @@ message ConnectorSpecificConfig {
 
     // GIVEPAYMENTS = 131
     GivepaymentsConfig givepayments = 138;
+
+    // DEUTSCHEBANK = 24
```
iemyashasvi, line 5765: nit : what is this 24?

config/production.toml
heykulthe: DB sandbox cert is issued by DB's own internal PKI "Deutsche Bank Group Root TEST CA 14". But DB's prod cert is issued by DigiCert which is public root already trusted everywhere. So sandbox needs a custom CA and prod does not

crates/integrations/connector-integration/src/connectors/deutschebank.rs
heykulthe: removed this, now DB is payout only

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: dry helper, its called in vop and sepa with different error context

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: these are synthetic references echoed back for reconciliation, not PII

crates/types-traits/grpc-api-types/proto/payment.proto
heykulthe: 24 = the Connector enum value; the = 139 = the protobuf field number of this config field. (payment.proto : 719)

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
heykulthe: kindly elaborate

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: SEPA e2eID is **ISO-20022 Max35Text with a restricted charset (no underscores, etc.)** upstream value fails on both counts: it can be up to 64 chars (>35), and even the default `payout_<nanoid>` contains an underscore. So the deterministic UUIDv5 is doing normalization that is required

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
iemyashasvi: why is this #[cfg(test)] added ?

heykulthe: Same test.rs pattern as adyen, razorpay, cashfree. covers unit tests for PEM splitting and id round trip logic

iemyashasvi: we dont need unit tests to be added in connector files we can remove it

config/production.toml
iemyashasvi: if this is not needed in prod , we can remove it , and also shall we remove it from dev.toml and sandbox.toml

iemyashasvi: in sanbox if this is configured in argo , why is this needed in code ?

crates/types-traits/domain_types/src/payouts/payouts_types.rs
```diff
@@ -395,3 +395,24 @@ pub struct PayoutEnrollDisburseAccountResponse {
     pub connector_payout_id: Option<String>,
     pub status_code: u16,
 }
+
+#[derive(Debug, Clone)]
+pub struct PayoutEligibilityRequest {
+    pub merchant_payout_id: Option<String>,
+    pub amount: common_utils::types::MinorUnit,
```
AmitsinghTanwar007: follow the money framework here

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -2086,6 +2086,200 @@ pub fn generate_payout_create_recipient_response(
...
+            &MaskedMetadata,
+        ),
+    ) -> Result<Self, error_stack::Report<Self::Error>> {
+        let merchant_id = extract_merchant_id_from_metadata(metadata)?;
+
+        Ok(Self {
+            merchant_id,
+            payout_id: value.merchant_payout_id.clone().unwrap_or_default(),
+            connectors,
+            connector_request_reference_id: crate::utils::extract_connector_request_reference_id(
```
AmitsinghTanwar007: you can populate the path of extract_connector req ref id function at top

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -2086,6 +2086,200 @@ pub fn generate_payout_create_recipient_response(
...
+                }
+            }),
+            test_mode: None,
+            description: None,
+        })
+    }
+}
+
+pub fn generate_payout_eligibility_response(
+    router_data_v2: crate::router_data_v2::RouterDataV2<
```
AmitsinghTanwar007: +1

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -2086,6 +2086,200 @@ pub fn generate_payout_create_recipient_response(
...
+                connector_payout_id: response.connector_payout_id,
+                payout_eligible: response.payout_eligible,
+                error: None,
+                status_code: u32::from(response.status_code),
+            })
+        }
+        Err(err) => Ok(grpc_api_types::payouts::PayoutMethodEligibilityResponse {
+            merchant_payout_id: Some(router_data_v2.resource_common_data.payout_id),
+            payout_status: Some(
+                grpc_api_types::payouts::payout_enums::PayoutStatus::Pending as i32,
```
AmitsinghTanwar007, line 2277: in all error case will you populate pending can you reverify it

config/production.toml
heykulthe: It needs to be there in dev and sbx.toml so when if someone is using the open source version of prism, they'd want to know that a cert needs to be added else it will result in db payout failure. as for us, this list is then mirrored into argo (ie it takes the list from here and then overwritten)

heykulthe: prod val i'll remove

config/development.toml
AmitsinghTanwar007: its ok as empty so that every one know deutschebank has server_ca_bundle

AmitsinghTanwar007:
`server_ca_bundle` is platform/connector config, not merchant request input. Right now it is parsed lazily inside `create_client()` via `additional_ca_pem.map(create_certificate).transpose()?`, so a malformed `deutschebank.server_ca_bundle` will fail the merchant’s Deutsche Bank request at runtime before the connector call is sent.

Can we validate this during config load/startup instead? For example, when `Connectors::deutschebank.server_ca_bundle` is present and non-empty, parse it once with the same PEM parser used by `create_certificate` and fail startup / surface a config error if invalid. The request-time parse can remain as a defensive fallback, but platform CA config errors should be caught before traffic.

crates/common/common_enums/src/enums.rs
```diff
@@ -1280,7 +1280,15 @@ pub enum PayoutStatus {
...
     Pending,
+    /// Non-terminal: the payout method/payee was found ineligible but the payout
+    /// is not conclusively closed (kept for backward compatibility; no terminal
+    /// webhook is emitted for this status).
     Ineligible,
+    /// Terminal: the payout was conclusively refused by the processor (e.g. a
+    /// Verification-of-Payee "no match" / "could not verify" outcome). Unlike
+    /// [`PayoutStatus::Ineligible`] this is a final state and triggers a terminal
+    /// failure webhook to the merchant.
+    NotPermitted,
```
Sakilmostak, line 1291: link the pr which add this status to hyperswitch

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,172 @@
...
+    let der = extract_der_from_pem(pem).map_err(|e| IntegrationError::InvalidConnectorConfig {
+        config: "signing_private_key",
+        context: IntegrationErrorContext {
+            additional_context: Some(format!("PEM body could not be base64-decoded: {e}")),
+            suggested_action: Some(
+                "Re-upload `signing_private_key` as a well-formed PEM block.".to_string(),
+            ),
+            doc_url: None,
+        },
+    })?;
```
Sakilmostak: refrain from logging error body in case of sensitive data like private key

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,172 @@
...
+    ];
+
+    let octet_string = encode_der_tlv(0x04, pkcs1);
+    let version = vec![0x02, 0x01, 0x00]; // INTEGER 0
+    let mut content = Vec::with_capacity(version.len() + ALG_ID.len() + octet_string.len());
+    content.extend_from_slice(&version);
+    content.extend_from_slice(ALG_ID);
+    content.extend_from_slice(&octet_string);
+    encode_der_tlv(0x30, &content)
+}
```
Sakilmostak: add comments to all the function and procedure for future reference

crates/types-traits/domain_types/src/payouts/types.rs
heykulthe: yes, and its intentional and uniform. `PayoutStatus::Pending` is used in the Err branch of all nine `generate_payout_*_response` functions

crates/types-traits/domain_types/src/types.rs
```diff
@@ -474,6 +475,22 @@ pub struct ConnectorParamsWithMoreUrls {
...
+    #[serde(default)]
+    pub dispute_base_url: Option<String>,
+    #[serde(default)]
+    pub secondary_base_url: Option<String>,
+    #[serde(default)]
+    pub third_base_url: Option<String>,
+    /// PEM trust anchor for verifying the connector's server certificate
+    #[serde(default)]
+    pub server_ca_bundle: Option<String>,
+}
```
Sakilmostak, line 494: can you elaborate the use case of introducing this struct

heykulthe: main differentiator is server_ca_bundle, forcing all ~100 connectors to carry a field only Deutsche Bank uses if put in `ConnectorParams`. Thus made the `ConnectorParamsWithCaBundle` so the field lives only on the one connector that needs it, more of a cleanliness choice.

crates/integrations/connector-integration/Cargo.toml
AmitsinghTanwar007: v5 is used for deterministic id generation i think so if there is a need then we should use it

iemyashasvi (dismissed):

crates/common/ucs_env/src/configs.rs
```diff
@@ -360,6 +360,18 @@ impl Config {
         // Validate the environment field
         config.common.validate()?;
 
+        if let Some(bundle) = config.connectors.deutschebank.server_ca_bundle.as_deref() {
```
gopikrishna000: why connector specifci code here
and why is it needed

crates/common/external-services/src/service.rs
```diff
@@ -1291,6 +1292,7 @@ pub fn create_client(
     proxy_name: &str,
     client_certificate: Option<Secret<String>>,
     client_certificate_key: Option<Secret<String>>,
+    additional_ca_pem: Option<Secret<String>>,
```
gopikrishna000: Naming: three names for one concept now - Request::ca_certificate, builder method add_ca_certificate_pem(), and this parameter additional_ca_pem. Please use ca_certificate_pem here so the call chain reads consistently: request.ca_certificate -> create_client(ca_certificate_pem) -> add_ca_certificate_pem().

crates/common/external-services/src/service.rs
```diff
@@ -1299,18 +1301,19 @@ pub fn create_client(
...
+                .map(create_certificate)
+                .transpose()?
+                .unwrap_or_default()
                 .into_iter()
-                .fold(client_builder, |client_builder, certificate| {
-                    client_builder.add_root_certificate(certificate)
-                });
+                .fold(
+                    client_builder.identity(identity).use_rustls_tls(),
+                    |b, ca| b.add_root_certificate(ca),
```
gopikrishna000, line 1321: This removes the old behavior where the client identity certificate was also registered as a root CA (the previous create_certificate(encoded_certificate) -> add_root_certificate fold). That is the right direction semantically - identity and trust anchor were conflated - but it is a silent trust-model change for every existing mTLS connector. Please add a short comment here so the next reader does not have to reverse-engineer it from git history:

    // NOTE: the client identity certificate is no longer registered as a root CA.
    // Server verification now uses webpki roots plus ca_certificate_pem only.

(And see the review body for the itaubank regression this implies.)

crates/common/external-services/src/service.rs
```diff
@@ -1540,6 +1543,14 @@ pub fn create_certificate(
         .change_context(ApiClientError::CertificateDecodeFailed)
 }
 
+pub fn validate_ca_certificate_pem(
```
gopikrishna000: This new function re-implements what create_certificate() just above already does. Please do not maintain two PEM parsers - runtime client construction and config validation must share one code path or they can drift apart (e.g. one day someone changes one to from_pem and config validation starts disagreeing with runtime). Suggested shape:

    /// Single PEM-bundle parser used by both client construction and config validation.
    pub fn parse_ca_pem_bundle(pem: &[u8]) -> CustomResult<Vec<Certificate>, ApiClientError> {
        Certificate::from_pem_bundle(pem).change_context(ApiClientError::CertificateDecodeFailed)
    }

    pub fn create_certificate(certificate: Secret<String>) -> CustomResult<Vec<Certificate>, ApiClientError> {
        parse_ca_pem_bundle(certificate.expose().as_bytes())
    }

Then ucs_env::configs calls parse_ca_pem_bundle() and this function is deleted. Same validation guarantee, nothing duplicated. Also note the error conversion in configs.rs (Report<ApiClientError> -> ConfigError::Message) disappears if validation stays in one layer.

crates/common/ucs_env/src/configs.rs
```diff
@@ -360,6 +360,20 @@ impl Config {
...
```
gopikrishna000: The fail-fast-at-boot goal is right (this addresses the earlier review point on config-load validation), but the placement bakes one connector's PEM grammar into the shared Config::validate(). When the next private-CA connector arrives (itaubank is the candidate), this grows another hardcoded if-let block with its own error string. Two cleaner options, either works:

1. Delegate to the params type - put validate_tls() on ConnectorParamsWithCaBundle (domain_types, next to the field it governs) and keep this call site a one-liner: config.connectors.deutschebank.validate_tls().map_err(config::ConfigError::Message)?

2. Or keep it here but make it generic: a small validate_optional_ca_bundle(field_path: &str, pem: Option<&str>) helper in this crate, so connector #2 is a one-line call instead of a copied block.

The trim/empty-string special case also goes away entirely if the empty-string sentinel is removed from the checked-in tomls (see comment on config/development.toml).

config/development.toml
gopikrishna000, line 172: server_ca_bundle = "" uses an empty string as a sentinel for "not configured", which forces every consumer (and the new bootstrap validation) to trim-and-check. The field is Option<String> - please omit the key from the checked-in tomls entirely so it deserializes to None. Set the actual bundle only in deployed env configs (hyperswitch-infra / infraswitch), where a real CA bundle exists.

gopikrishna000 (commented):
Following up on the config-load validation that was requested earlier. Refinements inline, plus one correctness item that is a merge blocker for me.

Merge blocker: itaubank regression risk

This PR changes the shared mTLS branch of create_client(). Until now, every cert in the client PEM bundle was ALSO registered via add_root_certificate() - i.e. the client identity PEM doubled as a trust anchor. This PR removes that fold and replaces it with additional_ca_pem. In UCS today, itaubank is the only connector that passes a client certificate (connectors/itaubank.rs get_certificate returns auth.certificates - a PEM bundle). Two outcomes:

- If api.itau.com.br chains to a public root (webpki roots, since reqwest here is rustls-tls without native-roots): no change, the old add_root_certificate was dead weight.
- If itau's endpoint presents a cert from a private CA whose chain shipped inside the same merchant PEM bundle (common for bank mTLS onboarding): requests now fail at handshake with unknown issuer, and itaubank has no server_ca_bundle path to recover, because this PR does not extend itaubank to ConnectorParamsWithCaBundle.

Before merge, please either:
1. record an itaubank sandbox smoke test (PayoutCreate / ServerAuthenticationToken through the mTLS path) in the PR body, or
2. also move itaubank to ConnectorParamsWithCaBundle and seed its CA chain.

Inline comments cover the rest: (a) one shared PEM parser instead of two, (b) parameter naming, (c) a code comment marking the trust-model change, (d) moving the config-validation rule next to the params type, (e) dropping the empty-string sentinel from the checked-in tomls.

The mechanism itself (from_pem_bundle as validator, fail-fast at startup) is correct - validating with the exact parser reqwest will use at runtime is the right property; the issues are organizational.

config/development.toml
heykulthe: Others using the opensource version should know that server_ca_bundle is also required. This is to convey the same.

crates/types-traits/domain_types/src/payouts/types.rs
AmitsinghTanwar007: but in error case we should populate the status as per connector.

crates/common/ucs_env/src/configs.rs
heykulthe: Made this generic now

gopikrishna000 (dismissed):

Sakilmostak (dismissed):

AmitsinghTanwar007 (dismissed):

crates/common/ucs_env/src/configs.rs
```diff
@@ -360,6 +360,20 @@ impl Config {
...
+                external_services::service::validate_ca_certificate_pem(trimmed).map_err(
+                    |err| {
+                        config::ConfigError::Message(format!(
+                        "Invalid `connectors.deutschebank.server_ca_bundle`: expected a valid PEM \
+                         CA bundle: {err}"
+                    ))
+                    },
+                )?;
+            }
+        }
```
AmitsinghTanwar007: still how this code is generic

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DeutschebankTransactionInformationAndStatus {
+    #[serde(rename = "transactionStatus")]
+    pub transaction_status: Option<DeutschebankPaymentStatus>,
+    #[serde(rename = "originalEndToEndIdentification")]
+    pub original_end_to_end_identification: Option<String>,
+}
+
+pub fn encode_connector_payout_id(end_to_end_id: &str, debtor_iban: &Secret<String>) -> String {
```
JeevaRamu0104: This one worries me a bit. `.peek()` unwraps the `Secret`, and the result flows straight into `PayoutTransferResponse.connector_payout_id` → `PayoutServiceTransferResponse.connector_payout_id`, which is a plain `string` in `payouts.proto` (not `SecretString`). So it gets persisted upstream, echoed back on every Get, and picked up by any request/response event logging.

The response you pasted in the PR description actually shows it: `source_bank_data.iban` comes back masked as `DE832************53100`, but `connector_transaction_id` is `AAD8AA027E0859FABF0B5887C26806DA|DE83215730130100853100` — the same IBAN in the clear.

I get *why* it's here: `PayoutGetRequest` only carries `merchant_payout_id` + `connector_payout_id`, so the debtor IBAN needed for the status enquiry has nowhere else to go. But stuffing it into the ID isn't really safe, since that field is contractually non-sensitive. Could we add `source_bank_data` to `PayoutMethodGetRequest`/`PayoutGetRequest` and keep `connector_payout_id` as just the opaque `endToEndId`?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+            req.request.customer.as_ref(),
+            "Creditor name is required for Deutsche Bank SEPA payment",
+        )?;
+
+        let reference = req
+            .resource_common_data
+            .connector_request_reference_id
+            .clone();
+
+        let end_to_end_id = uuid::Uuid::new_v5(
```
JeevaRamu0104: I think there's a uniqueness problem hiding here when `merchant_payout_id` isn't sent.

It's `optional` in both `PayoutMethodEligibilityRequest` and `PayoutServiceTransferRequest`, and `PayoutFlowData` derives `connector_request_reference_id` through `extract_connector_request_reference_id`, which is just `identifier.clone().unwrap_or_default()` — so it's `""` when the caller omits it. Nothing here validates that.

With an empty reference:
- `end_to_end_id` becomes `uuid_v5(NAMESPACE_OID, "dbank-e2e:")` — the same constant for **every payout of every merchant**, since `merchant_id` isn't in the salt
- `message_id` is constant too, and it's used for both `messageIdentification` and `paymentInformationIdentification`
- `derive_vop_id(merchant_id, "")` gives one VoP-ID per merchant, reused forever
- and since `connector_payout_id` is `{constant}|{debtor_iban}`, PayoutGet ends up looking up a constant `endToEndIdentification` — so sync could return a completely different payout's status

SEPA wants `EndToEndIdentification`/`MessageIdentification` unique per transaction, so DB will either reject on duplicate detection or we'll silently read the wrong payment. Could we reject an empty `connector_request_reference_id` outright here, and also fold `merchant_id` into the salt?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+                    number_of_transactions: SEPA_SINGLE_TRANSACTION_COUNT.to_string(),
+                    payment_type_information: DeutschebankPaymentTypeInformation {
+                        service_level: DeutschebankCode {
+                            code: SEPA_SERVICE_LEVEL_CODE,
+                        },
+                    },
+                    requested_execution_date: DeutschebankExecutionDate {
+                        date: execution_date,
+                    },
+                    debtor: DeutschebankParty {
```
JeevaRamu0104: `creditor_name` comes from `extract_customer_name(req.request.customer)` — that's the payee. But the debtor is the merchant funding the payout, and the same value is also used for `initiatingParty.name` just above.

`Dbtr/Nm` goes to the beneficiary bank, shows up on the recipient's statement, and feeds sanctions/AML screening, so having the payee's own name there is wrong data on the payment message.

I went looking for a better source and couldn't find one — `SepaBankTransfer` has no `account_holder_name`, and `PayoutTransferRequest.address` is the payee's billing address. So this looks less like a slip and more like a missing field: I think we need a debtor name on `source_bank_data` (or from connector config) rather than falling back to `customer.name`.

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+            ..item.router_data
+        })
+    }
+}
+
+// ===== Payment Status =====
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum DeutschebankPaymentStatus {
```
JeevaRamu0104, line 321: Only mapping `ACCP`/`PDNG`/`RJCT` feels too narrow for a pain.002 `TxSts`.

The field is `Option<DeutschebankPaymentStatus>`, so an *absent* status safely falls through to `Pending` — but a *present but unmapped* one is a serde error that fails the whole response body. And `ACSP` / `ACSC` are the normal in-flight and settled states for a SEPA credit transfer, plus `ACTC`, `ACCC`, `RCVD`, `PART`, `CANC` all show up in practice.

Right now any of those turns PayoutGet into a deserialization error rather than a status, which means a payout DB has already accepted becomes permanently un-syncable. Could we add the remaining codes explicitly and put a `#[serde(other)]` catch-all that maps to `Pending`, so an unknown status degrades instead of blowing up?

crates/types-traits/domain_types/src/payouts/types.rs
```diff
@@ -2086,8 +2091,202 @@ pub fn generate_payout_create_recipient_response(
...
+            ),
+            connector_payout_id: err.connector_transaction_id.clone(),
+            payout_eligible: Some(false),
```
JeevaRamu0104: `payout_status: Pending` above is the right call, but I think `payout_eligible: Some(false)` contradicts it.

This branch fires on timeouts, 5xx, TLS failures, auth misconfig, signature rejections — all cases where we genuinely don't know whether the payee is eligible, as opposed to DB telling us they aren't. Your own test table notes the orchestrator stops when `payoutEligible=false`, so a transient DB outage would permanently mark payouts ineligible.

Could this be `None`? `Some(false)` feels like it should be reserved for the connector-declared NOAP/NMTC path in `build_eligibility_response`.

crates/common/external-services/src/service.rs
```diff
@@ -1299,18 +1301,21 @@ pub fn create_client(
                 get_client_builder(proxy_config, should_bypass_proxy, proxy_name, test_mode)?;
 
             let identity = create_identity_from_certificate_and_key(
-                encoded_certificate.clone(),
+                encoded_certificate,
                 encoded_certificate_key,
             )?;
-            let certificate_list = create_certificate(encoded_certificate)?;
-            let client_builder = certificate_list
+            // NOTE: the client identity certificate is no longer registered as a root CA.
```
JeevaRamu0104, line 1312: Just flagging this for visibility rather than objecting — dropping the client identity cert as a trust anchor is a genuine improvement (before this, any merchant-supplied client cert effectively became a trust anchor for that outbound client).

That said, it's a shared-plumbing behaviour change that touches every mTLS connector. I checked and `itaubank` is the only other one implementing `get_certificate` (payments + payouts), and it doesn't implement `get_ca_certificate`, so it now leans entirely on webpki roots for `api.itau.com.br` / `sandbox.devportal.itau.com.br`. That's almost certainly fine, but could we confirm with a sandbox call, or at least call it out in the PR description so it's on the record?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+                    },
+                    debtor_account: DeutschebankAccount {
+                        identification: DeutschebankIbanIdentification {
+                            iban: debtor_iban.clone(),
+                        },
+                        currency: Some(req.request.source_currency),
+                    },
+                    debtor_agent: DeutschebankAgent {
+                        financial_institution_identification: DeutschebankBic {
+                            bicfi: Secret::new(DEUTSCHEBANK_BICFI.to_string()),
```
JeevaRamu0104: We hardcode the debtor agent to `DEUTDEDDXXX` while `Bank::Sepa.bic` is populated by the caller — your own test payload sends `DEUTDEDB237`, and it just gets dropped on the floor.

Silently overwriting caller input is the part that bothers me. If the debtor account has to be at Deutsche Bank anyway, could we validate the supplied BIC and reject a mismatch instead?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+        .to_string();
+        let creation_date_time = current_iso_utc_seconds()?;
+        let execution_date = sepa_execution_date()?;
+
+        let amount = utils::convert_amount(
+            &FloatMajorUnitForConnector,
+            req.request.amount,
+            req.request.destination_currency,
+        )?;
+        let purpose_code = req
```
JeevaRamu0104, line 554: Mapping free-text `description` straight into SEPA `Purp/Prtry` is a bit risky — that field is bounded at 35 chars and charset-restricted, whereas `description` is arbitrary merchant text.

A long or non-Latin description will produce a request DB rejects with an opaque scheme error that'll be painful to debug. Could we validate length/charset here, or better, map from a dedicated purpose-code field?

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
```diff
@@ -0,0 +1,692 @@
...
+        event_builder: Option<&mut events::Event>,
+        _connector_config: &ConnectorSpecificConfig,
+    ) -> CustomResult<ErrorResponse, ConnectorError> {
+        let response: DeutschebankErrorResponse = res
+            .response
+            .parse_struct("DeutschebankErrorResponse")
+            .unwrap_or_else(|err| {
+                tracing::warn!(
+                    deserialization_error = ?err,
+                    raw_body = %String::from_utf8_lossy(res.response.as_ref()),
```
JeevaRamu0104: Logging the full raw body untruncated makes me nervous for this connector specifically — DB's VoP and credit-transfer error bodies echo IBANs and payee names, so this puts unmasked bank data into log storage with no length bound.

There is precedent in `connectors/barclaycard.rs:779`, but that one caps at 1024 chars and surfaces it in `reason` rather than logging it. Could we do the same here, or lean on the existing masked event body instead?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+impl TryFrom<ResponseRouterData<DeutschebankSepaPaymentResponse, Self>>
+    for RouterDataV2<PayoutTransfer, PayoutFlowData, PayoutTransferRequest, PayoutTransferResponse>
+{
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<DeutschebankSepaPaymentResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let built =
+            DeutschebankSepaPaymentBuilt::try_from(&item.router_data)
```
JeevaRamu0104: Rebuilding the whole request inside the *response* transformer just to recover `end_to_end_id` feels fragile — it re-runs amount conversion, timestamp formatting and every required-field check after the money has already moved, and any failure surfaces as `ResponseDeserializationFailed`.

It also means we ignore the `originalEndToEndIdentification` DB hands back. Could we thread the built ID through the router data instead, or read it off the response?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
```diff
@@ -0,0 +1,1047 @@
...
+#[derive(Debug, Serialize)]
+pub struct DeutschebankVopDebtor {
+    #[serde(rename = "debtorAccount")]
+    pub debtor_account: DeutschebankVopIbanAccount,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DeutschebankVopResponse {
+    #[serde(
+        rename = "payeeNameMatch",
```
JeevaRamu0104: Four aliases for one field (`payeeNameMatch` / `match` / `result` / `matchStatus`) suggests we're not sure what DB actually returns — and the same pattern shows up on `transactionStatus`/`status` with a nested-then-top-level fallback chain further down.

The risk is that these fallbacks quietly mask a schema mismatch as `None` → `Pending`, so we'd never notice we're parsing the wrong shape. Could we pin this against DB's spec and collapse to the real field name?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,215 @@
...
+///   SEQUENCE {                     -- AlgorithmIdentifier
+///     OID 1.2.840.113549.1.1.1,    -- rsaEncryption
+///     NULL
+///   },
+///   OCTET STRING { <pkcs1 DER> }   -- privateKey
+/// }
+/// ```
+/// `ALG_ID` is the pre-encoded `AlgorithmIdentifier` (SEQUENCE of the rsaEncryption
+/// OID + NULL parameters).
+fn wrap_pkcs1_as_pkcs8(pkcs1: &[u8]) -> Vec<u8> {
```
JeevaRamu0104: Nothing wrong with the implementation as far as I can tell — but hand-rolled ASN.1 and PEM handling is exactly the code I'd want table tests around. Same goes for `encode_der_tlv`, `extract_der_from_pem`, `compute_digest`, `format_http_date`, and `split_pem_bundle` over in transformers.rs.

I'm not going to block on connector-level tests since no payout connector in the repo ships a `test.rs`. But these are pure functions, so unit tests are cheap, and commit 8a79aca removed the cases that did exist. Could we bring some back for at least the crypto/encoding helpers?

crates/types-traits/domain_types/src/types.rs
```diff
@@ -504,6 +521,14 @@ impl HasConnectors for crate::payouts::payouts_types::PayoutFlowData {
 }
 
 impl Connectors {
+    pub fn server_ca_bundles(&self) -> Vec<(&'static str, &str)> {
```
JeevaRamu0104, line 533: Minor: the comment at the call site in `ucs_env/src/configs.rs` says this is "generic over any connector that exposes a `server_ca_bundle`", but it's a single hardcoded `deutschebank` entry.

Not a problem today, just means the next connector to add a CA bundle will silently skip config validation. Worth either softening the comment or adding a note here.

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,215 @@
...
+    content.extend_from_slice(&octet_string);
+    encode_der_tlv(0x30, &content) // wrap the three fields in the outer SEQUENCE
+}
+
+/// Encode a single ASN.1 DER TLV (tag-length-value).
+///
+/// Handles DER definite-length encoding: short form for lengths ≤ 0x7f (one byte),
+/// and long form for larger lengths (`0x81 <len>` for ≤ 0xff, `0x82 <hi> <lo>` for
+/// ≤ 0xffff). Lengths beyond 0xffff are not produced here — RSA keys stay well under.
+fn encode_der_tlv(tag: u8, data: &[u8]) -> Vec<u8> {
```
JeevaRamu0104: Small one: the `_ =>` arm silently emits a malformed length for `data.len() > 0xffff` — it writes `0x82` with truncated bytes rather than erroring.

Unreachable for RSA keys as the doc comment says, so no real risk, but silent-wrong-output is an unpleasant failure mode if it ever is reached. Could we make it an explicit error instead?

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
```diff
@@ -0,0 +1,215 @@
...
+    path: &str,
+    body: &[u8],
+    key_id: &Secret<String>,
+    signing_private_key: &Secret<String>,
+) -> Result<CsealHeaders, error_stack::Report<IntegrationError>> {
+    let date = format_http_date(OffsetDateTime::now_utc())?;
+
+    let method_lower = method.to_string().to_lowercase();
+
+    let (digest, signing_string, headers_covered) = match method {
```
JeevaRamu0104, line 37: We sign `(request-target)` over the constant path (`VOP_PATH` etc.) while the actual URL is `base_url + path`. That's correct for the current config values since none of them carry a path component.

But `ConnectorSpecificConfig::Deutschebank` allows a per-MCA `base_url` override, so if someone ever configures one with a path prefix, signature verification breaks with an opaque 401 and nothing points at the cause. Deriving the signed target from the built URL would make that impossible.

JeevaRamu0104 (changes requested):
Went through this fairly carefully — it's a substantial and well-organised integration, and the error contexts throughout are genuinely good. A few things I'd like to sort out before it goes in, though.

The two I'd really want addressed are the debtor IBAN ending up in `connector_payout_id` (it's a plain `string` in the proto and shows up unmasked in your own sample response), and the SEPA identifier collision when `merchant_payout_id` isn't supplied — since it's `optional` in the proto and the reference id defaults to `""`, `endToEndId` and `messageId` collapse to a single global constant.

Beyond those, a few correctness things: `debtor.name` is being set to the payee's name, `payout_eligible: false` is returned for transport errors where we don't actually know, and the SEPA status mapping only covers 3 of the ISO 20022 codes (unknown ones fail the parse rather than degrading to pending).

The rest are smaller — details inline. Happy to talk any of them through.

For context on what I checked: all 51 files, plus the companion pieces not in the diff (`common_utils/src/request.rs`, `domain_types/src/utils.rs`, `payout_method_data.rs`, `patch_connector_urls`). CI is green including proto breaking checks, so everything below is semantic rather than build/parity.

crates/common/ucs_env/src/configs.rs
heykulthe: You are reviewing outdated code, kindly recheck.

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: resolved

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: Have added other status as well

crates/common/external-services/src/service.rs
heykulthe: Added a note for this in PR description.

crates/integrations/connector-integration/src/payout_connectors/deutschebank/transformers.rs
heykulthe: introduced new helper: validate_sepa_purpose_code for this

crates/types-traits/domain_types/src/types.rs
heykulthe: this is a doc fix, have updated it

crates/integrations/connector-integration/src/payout_connectors/deutschebank/signing.rs
heykulthe: Added `signed_request_target(url, fallback)` (uses url::Url::parse(...).path() + query, falls back to the constant) in deutschebank.rs and fed its result into `build_cseal_request → signing`.

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

iemyashasvi (approved):

deepanshu-iiitu (approved):

Sakilmostak (approved):

gopikrishna000 (approved):

heykulthe:
Payout Status: Not Permitted pr: [13524](https://github.com/juspay/hyperswitch/pull/13524)

PR #1501, opened 2026-06-11
feat(events): route connector and inbound events to separate topics

### LINKED ISSUE #12717: [FEATURE] Tag connector events with source, service_call and execution_mode dimensions
## Feature Description

Connector events emitted to ClickHouse (`connector_events`) currently can't be attributed to *who* made the call, *what kind* of call it was, or *whether it was the real or a shadow execution*. As Hyperswitch routes traffic through the Unified Connector Service (UCS) — including shadow mode where a direct call and a UCS call run side by side — we need to distinguish these legs in analytics.

## Proposal

Add three low-cardinality dimensions to every connector event:

| column | values | meaning |
|---|---|---|
| `source` | `hyperswitch` / `ucs` | which service emitted the event |
| `call_type` | `connector` / `service` | the call to the payment connector vs a request at the UCS service interface |
| `execution_mode` | `primary` / `shadow` | real execution vs shadow mirror |

`call_type` is a **leg** label — orthogonal to `source` and caller/transport-agnostic, so it stays correct regardless of caller or transport. `execution_mode` is the two-state observability projection of the routing `ExecutionMode` (`NotApplicable` = Direct = a live leg → `primary`).

### Expected tagging

| scenario | events (source / call_type / execution_mode) |
|---|---|
| Direct | `hyperswitch / connector / primary` |
| UCS primary | `hyperswitch / service / primary` · `ucs / service / primary` · `ucs / connector / primary` |
| Shadow — direct leg | `hyperswitch / connector / primary` |
| Shadow — UCS leg | `hyperswitch / service / shadow` · `ucs / service / shadow` · `ucs / connector / shadow` |

The UCS side emits matching field names (with `call_type` derived from its `EventStage`) so both producers land in the same `connector_events` table.

## Scope

- Add the fields to `ConnectorEvent` and populate them at the direct connector path and the UCS gRPC path (including the shadow leg).
- Add the matching `LowCardinality(String)` columns to the `connector_events` ClickHouse tables and materialized views.

gopikrishna000:
## What

Splits UCS audit events into two Kafka streams by stage, so each lands in the matching analytics table without a discriminator field in the payload. Each stream has its own topic **and** partition key, configured independently.

| stage | config | table | `source` | `destination` | `execution_mode` |
|---|---|---|---|---|---|
| `ConnectorCall` (outbound HTTP to the connector) | `events.connector_events` | `connector_events` | `unified_connector_service` | `connector` | `primary` / `shadow` |
| `GrpcRequest` (inbound request UCS serves) | `events.api_events` | `ucs_api_events` | `unified_connector_service` | _null_ | `primary` / `shadow` |

- The **topic is the discriminator**, so `source` and `destination` are **not** in the payload — they're stamped downstream from the topic.
- `execution_mode` (`primary` / `shadow`) **is** emitted, since the topic can't tell a live call from a shadow mirror.

## How

- `[events]` holds one `KafkaTopicConfig { topic, partition_key_field }` per stage — `connector_events` and `api_events`. Each topic carries its own partition key, so it's configurable per topic rather than shared.
- Stage-based routing in `publish_event_to_kafka`: `EventConfig::topic_config(stage)` returns that stage's topic + partition key (`ConnectorCall` → `connector_events`, `GrpcRequest` → `api_events`). A single emission chokepoint, so all RPC flows are covered.
- The shared `KafkaWriter` is **unchanged**: `publish_event(topic, key, …)` already takes the topic and partition key per call, so multi-topic publishing needs no change to the writer — only the per-stage selection above.
- `ExecutionMode` (`primary` / `shadow`, derived from the `x-shadow-mode` flag) lives in `common_utils::types`, reusable beyond events.
- Field names match the Hyperswitch producer, so both feed the same tables.

## Config change (one-time)

The flat `topic` / `ucs_api_events_topic` / `partition_key_field` keys are replaced by nested per-topic tables:

```toml
[events]
enabled = false
brokers = ["localhost:9092"]

[events.connector_events]
topic = "unified-connector-service-connector-events"
partition_key_field = "request_id"

[events.api_events]
topic = "unified-connector-service-api-events"
partition_key_field = "request_id"
```

A single-topic deployment sets both `topic`s to the same value — the pre-split behaviour, by config rather than magic. Empty topics are rejected at init (the publisher logs a warning and drops events rather than mis-routing). No legacy alias is kept; a deployment updates this config once when adopting the tag.

### Manual end-to-end (local Kafka)

Ran a real Adyen `Authorize` through the gRPC server (built `--features kafka`) against a local broker. A single Authorize produced both stage events, each on its own topic:

> _Screenshots were captured before the topic rename, so the broker still shows the earlier `ucs-connector-events` / `ucs-api-events` names; the config now ships the `unified-connector-service-*` defaults._

**`ConnectorCall` → `unified-connector-service-connector-events`** — the outbound call to Adyen (real `401` response):

image (6)

```json
{
  "request_id": "adyen-...", "flow_type": "Authorize", "connector": "adyen",
  "url": "https://checkout-test.adyen.com/v68/payments", "method": "POST",
  "stage": "ConnectorCall", "execution_mode": "primary", "status_code": 401
}
```

**`GrpcRequest` → `unified-connector-service-api-events`** — the inbound gRPC request:

image (5)

```json
{
  "request_id": "adyen-...", "flow_type": "Authorize", "connector": "adyen",
  "url": null, "method": "", "stage": "GrpcRequest",
  "execution_mode": "primary", "status_code": 16
}
```

Confirms per-stage topic routing, `execution_mode` in the payload, and `source`/`destination` correctly **absent** (stamped downstream from the topic).

## Companion

Hyperswitch side: juspay/hyperswitch#12714

### DIFF
--- a/config/production.toml
+++ b/config/production.toml
@@ -135,6 +135,7 @@ axisbank.base_url = "https://upisdk.axisbank.co.in/api/b2/"
 enabled = false
 brokers = ["localhost:9092"]
 topic = "audit-trail-events"
+ucs_api_events_topic = "audit-trail-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/production.toml
+++ b/config/production.toml
@@ -134,8 +134,15 @@ axisbank.base_url = "https://upisdk.axisbank.co.in/api/b2/"
 [events]
 enabled = false
 brokers = ["localhost:9092"]
+
+# Outbound connector-call events.
+[events.connector_events]
 topic = "audit-trail-events"
-ucs_api_events_topic = "audit-trail-events"
+partition_key_field = "request_id"
+
+# Inbound request events.
+[events.api_events]
+topic = "audit-trail-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/production.toml
+++ b/config/production.toml
@@ -137,12 +137,12 @@ brokers = ["localhost:9092"]
 
 # Outbound connector-call events.
 [events.connector_events]
-topic = "ucs-connector-events"
+topic = "unified-connector-service-connector-events"
 partition_key_field = "request_id"
 
 # Inbound request events.
 [events.api_events]
-topic = "ucs-api-events"
+topic = "unified-connector-service-api-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/production.toml
+++ b/config/production.toml
@@ -137,12 +137,12 @@ brokers = ["localhost:9092"]
 
 # Outbound connector-call events.
 [events.connector_events]
-topic = "audit-trail-events"
+topic = "ucs-connector-events"
 partition_key_field = "request_id"
 
 # Inbound request events.
 [events.api_events]
-topic = "audit-trail-events"
+topic = "ucs-api-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/sandbox.toml
+++ b/config/sandbox.toml
@@ -135,6 +135,7 @@ axisbank.base_url = "https://upiuatv3.axisbank.com/api/b2/"
 enabled = false
 brokers = ["localhost:9092"]
 topic = "audit-trail-events"
+ucs_api_events_topic = "audit-trail-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/sandbox.toml
+++ b/config/sandbox.toml
@@ -134,8 +134,15 @@ axisbank.base_url = "https://upiuatv3.axisbank.com/api/b2/"
 [events]
 enabled = false
 brokers = ["localhost:9092"]
+
+# Outbound connector-call events.
+[events.connector_events]
 topic = "audit-trail-events"
-ucs_api_events_topic = "audit-trail-events"
+partition_key_field = "request_id"
+
+# Inbound request events.
+[events.api_events]
+topic = "audit-trail-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/sandbox.toml
+++ b/config/sandbox.toml
@@ -137,12 +137,12 @@ brokers = ["localhost:9092"]
 
 # Outbound connector-call events.
 [events.connector_events]
-topic = "ucs-connector-events"
+topic = "unified-connector-service-connector-events"
 partition_key_field = "request_id"
 
 # Inbound request events.
 [events.api_events]
-topic = "ucs-api-events"
+topic = "unified-connector-service-api-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/config/sandbox.toml
+++ b/config/sandbox.toml
@@ -137,12 +137,12 @@ brokers = ["localhost:9092"]
 
 # Outbound connector-call events.
 [events.connector_events]
-topic = "audit-trail-events"
+topic = "ucs-connector-events"
 partition_key_field = "request_id"
 
 # Inbound request events.
 [events.api_events]
-topic = "audit-trail-events"
+topic = "ucs-api-events"
 partition_key_field = "request_id"
 
 [connector_request_kafka]
--- a/crates/common/common_utils/src/event_publisher.rs
+++ b/crates/common/common_utils/src/event_publisher.rs
@@ -6,7 +6,7 @@ use serde_json;
 use tracing_kafka::{builder::KafkaWriterBuilder, KafkaWriter};
 
 use crate::{
-    events::{Event, EventConfig},
+    events::{Event, EventConfig, EventStage},
     CustomResult, EventPublisherError,
 };
 
@@ -35,7 +35,7 @@ impl EventPublisher {
             ));
         }
 
-        if config.topic.is_empty() {
+        if config.connector_events_topic.is_empty() {
             return Err(error_stack::Report::new(
                 EventPublisherError::InvalidConfiguration {
                     message: "topic cannot be empty".to_string(),
@@ -45,20 +45,20 @@ impl EventPublisher {
 
         tracing::debug!(
           brokers = ?config.brokers,
-          topic = %config.topic,
+          topic = %config.connector_events_topic,
           "Creating EventPublisher with configuration"
         );
 
         let writer = KafkaWriterBuilder::new()
             .brokers(config.brokers.clone())
-            .topic(config.topic.clone())
+            .topic(config.connector_events_topic.clone())
             .build()
             .map_err(|e| {
                 error_stack::Report::new(EventPublisherError::KafkaWriterInitializationFailed)
                     .attach_printable(format!("KafkaWriter build failed: {e}"))
                     .attach_printable(format!(
                         "Brokers: {:?}, Topic: {}",
-                        config.brokers, config.topic
+                        config.brokers, config.connector_events_topic
                     ))
             })?;
 
@@ -180,7 +180,7 @@ pub fn init_event_publisher(config: &EventConfig) {
             tracing::warn!(
                 error = ?e,
                 brokers = ?config.brokers,
-                topic = %config.topic,
+                topic = %config.connector_events_topic,
                 "Failed to initialize EventPublisher (Kafka may be unavailable); \
                  events will be dropped until the service is restarted with Kafka reachable"
             );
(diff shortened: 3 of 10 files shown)


XyneSpaces (commented):
## Automated Review Summary

**PR:** feat(events): add source, service_call and execution_mode to connector events
**Reviewer:** PR Reviewer Agent

---

### Verification Passed

✅ **Enum definitions** — `EventSource`, `ServiceCall`, `ExecutionMode` defined with proper serde attributes for snake_case serialization

✅ **ExecutionMode helper** — `from_shadow_flag()` provides clean boolean-to-enum conversion

✅ **Field population** — Both emission sites (inbound gRPC handler and outbound connector call) correctly set the three new fields

✅ **Alignment with Hyperswitch** — Field names and semantics match the corresponding PR #12714 on the Hyperswitch side, ensuring both producers land in the same ClickHouse table columns

✅ **No dead code** — All new enum variants and helper functions are used

---

### Findings

:mag: **1. u64::try_from().unwrap_or() pattern**

**File:** `crates/grpc-server/grpc-server/src/utils.rs`

The latency_ms calculation uses `unwrap_or(u64::MAX)` which is reasonable for this metric context, but consider if saturating at `u64::MAX` could skew latency percentiles in extreme cases.

---

**Verdict:** :white_check_mark: **Approve** — Clean implementation matching the Hyperswirevent contract. Ready to merge.

crates/common/common_utils/src/events.rs
```diff
@@ -423,6 +429,44 @@ impl EventStage {
...
+/// Distinguishes an internal service-to-service hop from an outbound connector call
+/// (ClickHouse `service_call` column).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ServiceCall {
+    /// Internal hop, e.g. the inbound gRPC request received by UCS.
+    Internal,
+    /// Outbound HTTP call to the payment connector.
+    External,
+}
```
su-shivanshmathur: Can incoming and outgoing calls be sent to different topics all together ?

JeevaRamu0104 (dismissed):

crates/common/common_utils/src/events.rs
```diff
@@ -230,6 +230,12 @@ pub struct Event {
     #[serde(serialize_with = "serialize_method")]
     pub method: Option<String>,
     pub stage: EventStage,
+    /// Service that produced this event (always `ucs`).
+    pub source: &'static str,
```
su-shivanshmathur: This is not needed

su-shivanshmathur (changes requested):

crates/common/common_utils/src/events.rs
```diff
@@ -442,7 +474,8 @@ impl Default for EventConfig {
     fn default() -> Self {
         Self {
             enabled: false,
-            topic: "events".to_string(),
+            connector_events_topic: String::new(),
+            ucs_api_events_topic: String::new(),
```
AmitsinghTanwar007: in default if we have "" empty string then in code are we verifying when using the value.
or it is ok as we will get error for wrong value and we will trace it in the logs.

crates/grpc-server/grpc-server/tests/test_config_override.rs
```diff
@@ -454,7 +454,13 @@ mod unit {
...
         let new_config = apply_override(override_json);
         assert!(new_config.events.enabled);
-        assert_eq!(new_config.events.topic.as_str(), "events-override");
+        // The override JSON above intentionally uses the legacy `topic` key; this asserts it
+        // still applies through the patch layer via the `#[serde(alias = "topic")]` on the
+        // renamed `connector_events_topic` field (backward compatibility).
+        assert_eq!(
+            new_config.events.connector_events_topic.as_str(),
+            "events-override"
+        );
```
AmitsinghTanwar007: should we avoid overriding of events_topic?

crates/common/common_utils/src/event_publisher.rs
```diff
@@ -206,10 +206,20 @@ pub fn publish_event_to_kafka(
...
+            let topic = match event.stage {
+                EventStage::ConnectorCall => &config.connector_events_topic,
+                EventStage::GrpcRequest => {
+                    if config.ucs_api_events_topic.is_empty() {
+                        &config.connector_events_topic
+                    } else {
+                        &config.ucs_api_events_topic
+                    }
+                }
+            };
```
su-shivanshmathur: Why is this needed ?

crates/common/common_utils/src/events.rs
```diff
@@ -423,11 +425,41 @@ impl EventStage {
...
 
+/// Primary execution or shadow mirror.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecutionMode {
+    /// Real execution whose result is used.
+    Primary,
+    /// Shadow mirror whose result is discarded (validation only).
+    Shadow,
+}
```
su-shivanshmathur: This would derived from the shadow flag itself right, can this be used globally and not just for events ?

crates/common/common_utils/src/events.rs
gopikrishna000: yes, it's verified — EventPublisher::new() rejects an empty connector_events_topic and the publisher init logs a warning + drops events rather than silently sending to a wrong topic.

crates/grpc-server/grpc-server/tests/test_config_override.rs
gopikrishna000: it was already overridable before this pr

su-shivanshmathur: can we then make it correct , topic overriding should not be there

crates/common/external-services/src/service.rs
```diff
@@ -200,6 +200,8 @@ impl AdditionalHeaders
 }
 use common_utils::events::{Event, EventConfig, FlowName};
 #[cfg(feature = "injector-client")]
```
su-shivanshmathur, line 202: Does it need `injector-client` feature flag ?

config/sandbox.toml
```diff
@@ -135,6 +135,7 @@ axisbank.base_url = "https://upiuatv3.axisbank.com/api/b2/"
 enabled = false
 brokers = ["localhost:9092"]
 topic = "audit-trail-events"
+ucs_api_events_topic = "audit-trail-events"
```
su-shivanshmathur: Please have a beteer naming convention

config/production.toml
```diff
@@ -135,6 +135,7 @@ axisbank.base_url = "https://upisdk.axisbank.co.in/api/b2/"
 enabled = false
 brokers = ["localhost:9092"]
 topic = "audit-trail-events"
+ucs_api_events_topic = "audit-trail-events"
 partition_key_field = "request_id"
```
su-shivanshmathur, line 146: Will partition key the same for all topics in this service ?

su-shivanshmathur:
Please add test cases in the PR description

su-shivanshmathur:
Shouldn't your changes be done on `pub struct KafkaWriter {` as well ?

crates/common/common_utils/src/events.rs
```diff
@@ -621,3 +652,31 @@ pub(crate) fn set_nested_value(
...
+
+        let connector = config.topic_config(&EventStage::ConnectorCall);
+        assert_eq!(connector.topic, "connector-events");
+        assert_eq!(connector.partition_key_field, "request_id");
+
+        let ucs_api = config.topic_config(&EventStage::GrpcRequest);
+        assert_eq!(ucs_api.topic, "ucs-api-events");
+        assert_eq!(ucs_api.partition_key_field, "merchant_id");
+    }
+}
```
su-shivanshmathur: What is the usecase of these tests ?

su-shivanshmathur (dismissed):

crates/common/external-services/src/service.rs
gopikrishna000: https://github.com/juspay/hyperswitch-prism/issues/1601

created issue for this to raise a separate fix for this

crates/common/common_utils/src/events.rs
```diff
@@ -423,13 +426,22 @@ impl EventStage {
     }
 }
 
+/// A Kafka topic plus the payload field whose value is used as its partition key.
+#[derive(Debug, Clone, Default, Deserialize, Serialize, PartialEq, config_patch_derive::Patch)]
+pub struct KafkaTopicConfig {
+    pub topic: String,
```
AmitsinghTanwar007, line 432: you can use config patch ignore feature to disable override

AmitsinghTanwar007 (approved):

su-shivanshmathur (approved):

PR #1461, opened 2026-06-04
feat(refunds): add refund reverse flow

awasthi21:
## Summary

- Adds `RefundService.Reverse` gRPC and HTTP `/refunds/reverse` support.

- Maps refund reverse to the existing post-capture void connector flow and returns reverse outcome in connector state metadata.
- Adds required `ConnectorState.state_metadata` plumbing for generated proto structs.

## Validation
- `cargo fmt --all`

- `cargo check --bin grpc-server`

PR #1353, opened 2026-05-15
feat(tsys_xml): add TransIT XML UCS connector

awasthi21:
## What changed

Adds the TSYS TransIT XML connector to connector-service/UCS as `tsys_transit`, including XML request/response models, auth mapping, amount conversion, Sale/Auth/Capture/Void/Return/CardAuthentication/RepeatPayment flow handling, L2/L3/3DS/COF support, and TSYS XML request/response logging for certification evidence.

## Why

TSYS TransIT certification requires XML API support and enhanced data flows that are not covered by the legacy JSON `tsys` connector.

Closes #1502

## Validation

- Live local HS v1 -> local UCS -> TSYS stage certification runs completed for MOTO non-red rows through Step 11, with evidence stored locally only under `tsys_xml_certification`.
- Secret scan of committed diff checked for provided API keys, MCA/profile IDs, TSYS credentials, passwords, and private-key patterns. No live credentials were found in the PR diff.
- Concrete-looking TSYS sample credentials in the design doc were replaced with placeholders before push.

## Notes

- Evidence files and raw credential material were not committed or pushed.
- Return reversal remains a documented API capability gap: HS/UCS exposes Return and RSync, not Return Reversal/Void.

PR #1508, opened 2026-06-12
feat(connector): add tsys_transit XML connector with VoidPostRefund flow

### LINKED ISSUE #1502: Add TSYS TransIT connector
## Context

Add the TSYS TransIT XML connector to connector-service/UCS so merchants can route TSYS TransIT XML traffic through the unified connector service.

Related PR: https://github.com/juspay/hyperswitch-prism/pull/1353

## Scope

- Implement TSYS TransIT XML request/response models.
- Support Sale/Auth/Capture/Void/Return/CardAuthentication/RepeatPayment flows required for certification.
- Support MOTO, Level II/Level III enhanced data, 3DS, and card-on-file/recurring certification paths.
- Add connector-side raw XML request/response logging for certification evidence.

## Validation notes

Fresh local HS v1 -> local UCS -> TSYS stage certification runs were executed for MOTO non-red rows. Evidence is local-only and not committed.

Observed follow-up from the latest fresh run: MOTO Row 125 returned TSYS `D2022 Duplicate transaction`; this should be investigated as a validation/data issue separately from adding the connector.

## Evidence

Local evidence path:

`/Users/malay.awasthi/_Projects/repo_1/tsys_xml_certification/evidence/moto/live_run_20260611/20260611_173604`

awasthi21:
## Summary

Adds the **TSYS TransIT XML** connector to UCS as `tsys_transit`, and introduces the dedicated **`VoidPostRefund`** (refund reverse) flow used to cancel a processed refund before settlement.

### `tsys_transit` connector (XML SOAP)
- Sale / Auth / Capture / Void / VoidPC / Return / RepeatPayment / CardAuthentication / TransactionInquiry / Refund / RSync flows
- L2 / L3, 3DS (3DSecure 2 cryptograms + ECI), MIT/COF, partial-auth and surcharge support
- Connector enum registration + `superposition.toml` wiring

### `VoidPostRefund` flow (new across the stack)
- Proto contract (`RefundService.VoidPostRefund`, `RefundServiceVoidPostRefundRequest`)
- Domain types + connector trait (`RefundVoidPostRefundV2`)
- gRPC server method `void_post_refund`
- HTTP route `POST /refunds/void_post_refund`
- Wired through `tsys_transit` for live certification
- Mirrors the existing `VoidPostCapture` pattern (verb + Post + previous-action) across proto, gRPC, HTTP and domain layers

### `RefundResponse.state_metadata` (typed)
- Carries the underlying connector void result while the user-facing `status` stays `RefundSuccess`
- Encoded via new typed struct `VoidPostRefundStateMetadata { void_post_refund_status, connector_refund_id }`
- Persists across calls on HS side under `refund.metadata.state_metadata`

## Other cleanups folded in
- Restored Adyen `WebhookEventCode::RefundReversed` (had been accidentally renamed to a typo'd `CancelPostRefundd`)
- Removed redundant `#[serde(rename = ...)]` field attrs in `tsys_transit/transformers.rs` in favor of container-level `rename_all = "camelCase"`
- Removed unused fields from `TsysTransitAuthorizeBody` (`security_protocol`, `digital_payment_cryptogram`, `program_protocol`)
- Replaced `.expect()` panics in MIT category mapping with `ForeignFrom` + explicit `Unspecified -> None` match
- Removed unused dead structs (`TsysTransitTerminalData`, `TsysTransitDeveloperInfo`) flagged with `#[allow(dead_code)]`

## Validation
- `cargo check` clean across the workspace
- Local SBX certification runs (MOTO non-red rows through Step 11) verified end-to-end through `tsys_transit`
- No live credentials in diff; raw XML logging retained for SBX cert evidence (to be gated post-certification)

## Related
- Original connector PR: #1353
- Original refund-reverse PR: #1461


crates/integrations/connector-integration/src/connectors/adyen/transformers.rs
```diff
@@ -5332,7 +5332,7 @@ pub enum WebhookEventCode {
     CaptureFailed,
     Refund,
     RefundFailed,
-    RefundReversed,
+    CancelPostRefundd,
```
XyneSpaces: :rotating_light: Typo in event code name

`CancelPostRefundd` has an extra trailing 'd'. Should be `CancelPostRefund` to match the flow name consistently.

This typo will break webhook event classification for Adyen refund reversals.

XyneSpaces (commented):
## Review Summary

**PR:** feat(tsys_transit): combine connector certification and refund reverse flows
**Scenario:** connector-new-integration + connector-flow-addition

### Findings Overview

| Severity | Count | Issues |
|----------|-------|--------|
| :rotating_light: Critical | 1 | Adyen webhook event code typo |
| :warning: Warning | 1 | Missing CancelPostRefund tests |

---

### Detailed Findings

**1. Critical: Typo in Adyen webhook event code**

The PR renames `RefundReversed` to `CancelPostRefundd` (note the extra 'd' at the end) in `adyen/transformers.rs` and `adyen.rs`. This appears to be a typo - the correct name should be `CancelPostRefund` (without the trailing 'd') to match the flow name.

**Impact:** Webhook events from Adyen may not be correctly classified, leading to missed refund reversal notifications.

**Fix:** Rename `CancelPostRefundd` → `CancelPostRefund` in all occurrences:
- `adyen/transformers.rs`: `WebhookEventCode::CancelPostRefundd`
- `adyen.rs`: matching arm

---

**2. Warning: CancelPostRefund flow registration**

The PR adds `CancelPostRefund` to the `not_implemented` list for ~60+ connectors. While this is correct for gradual rollout, ensure:
1. The TSYS Transit connector (the primary target) actually implements this flow
2. Test specs exist for the new flow in `crates/internal/ucs-connector-tests/`

---

**Verdict:** Critical typo must be fixed before merge. Otherwise clean implementation following connector integration patterns.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,4092 @@
...
+    ))
+}
+
+fn generate_logged_xml<T: Serialize>(request: &T, fallback_root: &str) -> String {
+    let fallback = format!("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<{fallback_root}/>");
+    let xml = generate_xml(request).unwrap_or(fallback);
+    tracing::info!(
+        connector = "tsysTransit",
+        root = fallback_root,
+        raw_request_xml = %xml,
```
maverox, line 377: **Critical (PCI):** `generate_logged_xml` logs the full serialized SOAP body at `info!`. `generate_xml` exposes the inner values of `Secret<String>` fields, so this writes plaintext **PAN, CVV2, expiry, and the merchant `device_id`/`transaction_key`/`developer_id`** to logs, bypassing all masking. CVV must never be logged and PAN must be masked. The sibling `worldpayxml` connector intentionally does not log raw XML. Please remove this `tracing::info!` (return `generate_xml(...).unwrap_or(fallback)` only), or gate it behind a default-off cert feature that still never emits card/CVV/credential data.

crates/integrations/connector-integration/src/connectors/tsys_transit.rs
```diff
@@ -0,0 +1,795 @@
...
+        &self,
+        _req: &RouterDataV2<F, FCD, Req, Res>,
+        bytes: bytes::Bytes,
+        status_code: u16,
+    ) -> CustomResult<bytes::Bytes, IntegrationError> {
+        let raw_response_xml = String::from_utf8_lossy(bytes.as_ref());
+        tracing::info!(
+            connector = "tsysTransit",
+            http_status = status_code,
+            raw_response_xml = %raw_response_xml,
```
maverox, line 249: **Critical:** raw response body logged unconditionally at `info!`. Drop this `tracing::info!` and just return the bytes (as `worldpayxml::preprocess_response_bytes` does); rely on the framework's masked event capture. This is the same ungated raw-logging pattern flagged in the request path and should not land on `main`.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,4092 @@
...
+}
+fn extract_add_customer_card(
+    _router_data: &RouterDataV2<
+        CreateConnectorCustomer,
+        PaymentFlowData,
+        ConnectorCustomerData,
+        ConnectorCustomerResponse,
+    >,
+) -> Result<(Secret<String>, Secret<String>), Report<IntegrationError>> {
+    Err(IntegrationError::MissingRequiredField {
```
maverox: **High — flow is dead on arrival.** `extract_add_customer_card` ignores `_router_data` and unconditionally returns `Err(MissingRequiredField)`. Since it is the only source of card/expiry for `TsysTransitAddCustomerRequest`, the entire `CreateConnectorCustomer` (vault/Path-B) flow always errors before sending. The real helper `format_add_customer_expiration` (line 3420) is `#[allow(dead_code)]` and never called. Either implement extraction from `ConnectorCustomerData.payment_method_data` (and drop the dead-code allow), or move `CreateConnectorCustomer` to `not_implemented` rather than wiring an always-failing flow.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,4092 @@
...
+            Some(TsysTransitStatus::Pass),
+            Some("A0000"),
+            TsysTransitAuthorizeResponse::SaleResponse(_),
+        ) => AttemptStatus::Charged,
+        (
+            Some(TsysTransitStatus::Pass),
+            Some("A0000"),
+            TsysTransitAuthorizeResponse::AuthResponse(_),
+        ) => AttemptStatus::Authorized,
+        (Some(TsysTransitStatus::Pass), Some("A0002"), _) => AttemptStatus::PartialCharged,
```
maverox: **Medium — partial-auth mapping & lost amount.** (1) For a manual-capture `AuthResponse`, `A0002` means *partially authorized*, not partially captured — this should map to `AttemptStatus::PartiallyAuthorized` (distinguish the response variant like the `A0000` arms above already do). (2) The approved amount (`processed_amount`/`total_amount`, declared at lines 859-862) is never read, so on a partial approval the actual amount is dropped and the core cannot reconcile. Also: any `Pass` with a response code other than `A0000`/`A0002` falls through to `Failure` — confirm no soft/referral codes need a `Pending` mapping.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,4092 @@
...
+
+fn format_expiration_date(card: &Card<impl PaymentMethodDataTypes>) -> Secret<String> {
+    let month = card.card_exp_month.peek().clone();
+    let year_full = card.card_exp_year.peek().clone();
+    let year_short = if year_full.len() == 4 {
+        year_full[2..].to_string()
+    } else {
+        year_full
+    };
+    Secret::new(format!("{}/{}", month, year_short))
```
maverox: **Medium — expiry month not zero-padded.** `month` comes straight from `peek()`, so a single-digit month yields `3/25` instead of `03/25`. `elavon` (`format!` with `{exp_month:0>2}`) and this file's own dead `format_add_customer_expiration` both pad. Zero-pad the month here, and in the NTI branch (~line 2166) which has the same issue.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7798,6 +7876,165 @@ impl ForeignTryFrom<grpc_api_types::payments::PaymentServiceReverseRequest>
...
+                .transpose()?,
+            connector_refund_id: value.connector_refund_id,
+            cancellation_reason: value.cancellation_reason,
+            refund_connector_metadata: value
+                .refund_metadata
+                .map(|m| ForeignTryFrom::foreign_try_from((m, "refund metadata")))
+                .transpose()?,
+            refund_status: common_enums::RefundStatus::Success,
+            connector_feature_data,
+            refund_money: None,
```
maverox, line 8109: **High — required proto field silently dropped.** `RefundServiceVoidPostRefundRequest.connector_transaction_id` (field 2, non-optional) is never read in this `foreign_try_from`, so it never reaches `RefundVoidPostRefundData`/`RefundFlowData` (the integrity object then hardcodes an empty string for it). Either consume it or make the proto field `optional`. Separately, `refund_money: None` is hardcoded and the request carries no amount field, so partial refund-reversal is impossible by construction — fine if TSYS only supports full reversal, but worth a confirming comment.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7798,6 +7876,165 @@ impl ForeignTryFrom<grpc_api_types::payments::PaymentServiceReverseRequest>
...
+}
+
+/// State metadata persisted on the refund after a VoidPostRefund attempt.
+/// Carries the underlying connector void result while the user-facing
+/// refund status remains `RefundSuccess`. Field names mirror what
+/// `hyperswitch::core::refunds::build_cancel_post_refund_metadata`
+/// expects, so the HS side never has to fall back.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct VoidPostRefundStateMetadata {
+    pub status: common_enums::RefundStatus,
```
maverox, line 8120: **High — verify the cross-repo contract.** The whole design (force user-facing `status` = `RefundSuccess`, carry the real outcome in `state_metadata`) hinges on the Hyperswitch consumer (`build_cancel_post_refund_metadata`) deserializing exactly this shape (`status` + `connector_refund_id`) with this `RefundStatus` serde representation. If the key or encoding differs, HS silently falls back and a **failed reversal is reported to the caller as `RefundSuccess`**. Please confirm against the HS side and add a round-trip (serialize then deserialize) test — this is the single point of failure for the feature.

examples/tsystransit/tsystransit.py
```diff
@@ -0,0 +1,136 @@
+# This file is auto-generated. Do not edit manually.
```
maverox, line 1: **High — duplicate, mis-named generated artifacts.** This entire no-underscore `tsystransit` set (`examples/tsystransit/*`, `data/field_probe/tsystransit.json`, `docs-generated/connectors/tsystransit.md`) duplicates the canonical `tsys_transit` set and references symbols that do not exist: this file uses `payment_pb2.TsystransitConfig` and `tsystransit=` (proto defines `TsysTransitConfig` / `tsys_transit`), and `tsystransit.json` sets `connector: tsystransit` which is absent from the superposition enum (would fail validation). Looks like an accidental double-generation — please delete the entire `tsystransit` (no-underscore) set and keep only `tsys_transit`.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Acknowledged — the raw XML logging is intentional during the TSYS TransIT MOTO certification phase.

Reason: TSYS requires us to attach the raw connector request/response XML to the certification audit evidence packet (per-row evidence files contain the literal SOAP body that hit `stagegw.transnox.com`). Masking these fields would make the evidence packet unreviewable on TSYS' side.

Tracked for post-cert cleanup: juspay/hyperswitch-cloud#16830

Plan once SBX certification is signed off (per that issue):
1. Remove `generate_logged_xml` and switch the `to_soap_xml()` impls back to plain `generate_xml`.
2. Drop the `tsysTransit XML request` info log line entirely; the gRPC golden-log already captures masked headers and the structured request.
3. No production deploy of `tsys_transit` until that issue is closed.

I'll add a `// TODO(tsys-cert)` comment alongside `generate_logged_xml` pointing at juspay/hyperswitch-cloud#16830 so the rollback is unambiguous.

crates/integrations/connector-integration/src/connectors/tsys_transit.rs
awasthi21: Same context as the request-side comment above — the raw response body is logged on purpose for the TSYS MOTO certification audit packet.

Why this path and not `worldpayxml::preprocess_response_bytes`-style: TSYS wants the *complete* response (including non-XML wrappers and any per-row hostReferenceNumber/transactionTimestamp) reproduced verbatim against each cert row. The structured `SaleResponse` we deserialise into doesn't round-trip 1:1, so we capture the bytes before deserialisation and emit them alongside the golden-log.

Tracked for post-cert cleanup: juspay/hyperswitch-cloud#16830

Cleanup plan (same issue):
- Drop the `info!` on raw response.
- `preprocess_response` becomes a straight passthrough.
- Done before any non-sandbox enablement of `tsys_transit`.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Good catch — agreed, `extract_add_customer_card` was an unconditional `Err`, so the whole `CreateConnectorCustomer` wiring (request/response structs, transformer, macro impl, trait marker) was dead on arrival.

Fixed in `d866b1f9a`:
- Removed the entire `CreateConnectorCustomer` flow surface from `tsys_transit.rs` and `transformers.rs` (struct defs, `TryFrom`s, `extract_add_customer_card`, `split_full_name`, `format_add_customer_expiration`, `TsysTransitAddCustomer*` request/response types, `TsysTransitPersonalDetails`).
- Added `CreateConnectorCustomer` to the `not_implemented` block in `macro_connector_flow_status_impls!` so any caller now gets a clean `NotImplemented` error instead of the silent-fail path.

Rationale: TSYS TransIT's Card-on-File model is NTI-based (`cardTransactionIdentifier` from the original Sale response + `cardOnFile` / `previousNetworkTransactionID` flags on subsequent MITs). The `<AddCustomer>` "Vault" path was being scaffolded but never produced a usable mandate, and our cert evidence never exercised it.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: All three sub-points addressed in cc06213b7.

1. **A0002 mapping split by variant.** `map_authorize_status` now distinguishes the response variant exactly like the A0000 arms above:
   ```rust
   (Pass, "A0002", SaleResponse(_)) => AttemptStatus::PartialCharged,
   (Pass, "A0002", AuthResponse(_)) => AttemptStatus::PartiallyAuthorized,
   ```
   So a manual-capture `AuthResponse` with A0002 no longer collapses to "partially captured".

2. **Approved amount no longer dropped.** `TsysTransitAuthorizeResponseBody.processed_amount` is now typed `Option<StringMajorUnit>` instead of `Option<String>`, and the Authorize response handler runs it through `TsysTransitAmountConvertor::convert_back(..., currency)` to populate both `amount_captured` (i64) and `minor_amount_captured` (MinorUnit) on `PaymentFlowData`. The core now sees the actual approved value instead of inferring from the requested amount.

3. **Soft / referral codes.** Left an explicit `TODO(tsys-cert)` next to the `_ => Failure` fallthrough so we don't lose the question — we'll need to confirm with TSYS whether `A0003`/`A0006`/"call issuer"-style codes deserve a `Pending` (or `AuthenticationPending`) mapping, rather than the conservative `Failure` we ship today. Tracking with the rest of the certification follow-ups.

Verified live against TSYS SBX row 128 (MasterCard two-BIN partial PHONE) on the MOTO certification collection:
- Connector raw: `<transactionAmount>11.10</transactionAmount><processedAmount>5.55</processedAmount>`
- HS response now: `amount: 1110`, `amount_received: 555`, `amount_capturable: 1110`, `status: partially_captured`

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/cc06213b7

crates/types-traits/domain_types/src/types.rs
awasthi21: Good catch — the field really had nowhere to land. Dropped in 702d709f0:

```diff
 message RefundServiceVoidPostRefundRequest {
   optional string merchant_refund_id = 1;
-  string connector_transaction_id = 2;
-  string connector_refund_id = 3;
+  string connector_refund_id = 2;
   ...
 }
```

Rationale:
- `RefundVoidPostRefundData` (the domain type) had no matching field, and the converter on `types.rs:7882` never read it.
- The void-post-refund call targets the refund's own connector ID (`connector_refund_id`) — there was no real use case for the parent payment's transaction ID on the UCS side.
- The message has never shipped on `main`, so there's no over-the-wire compatibility concern and no need to `reserved 2;`. Just compacted the numbering.

The HS-side `unified_connector_service::RefundServiceCancelPostRefundRequest` wrapper still carries an unused `connector_transaction_id`; that needs to be dropped in the matching HS PR (noted in the commit body), but the wire data going to UCS is now schema-clean.

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/702d709f0

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Fixed in c7905939f.

```diff
-fn format_expiration_date(card: &Card<impl PaymentMethodDataTypes>) -> Secret<String> {
-    let month = card.card_exp_month.peek().clone();
-    let year_full = card.card_exp_year.peek().clone();
-    let year_short = if year_full.len() == 4 {
-        year_full[2..].to_string()
-    } else {
-        year_full
-    };
-    Secret::new(format!("{}/{}", month, year_short))
-}
+fn format_expiration_date(card: &Card<impl PaymentMethodDataTypes>) -> Secret<String> {
+    let month_raw = card.card_exp_month.peek();
+    let year_raw  = card.card_exp_year.peek();
+    let month = format!("{month_raw:0>2}");
+    let year_short = if year_raw.len() >= 4 {
+        year_raw[year_raw.len() - 2..].to_string()
+    } else {
+        format!("{year_raw:0>2}")
+    };
+    Secret::new(format!("{month}/{year_short}"))
+}
```

- `{:0>2}` left-pads month so `"3"` → `"03"`, `"12"` → `"12"` (no-op).
- Year handling is now defensive both ways: a 4+ char value is sliced to its last two chars (`"2028"` → `"28"`), and any shorter value is `{:0>2}`-padded (`"5"` → `"05"`).

No behaviour change for the well-formed `MM`/`YY` and `MM`/`YYYY` inputs the cert collection uses today — re-ran cert rows 125 / 128 / 179 against the rebuilt UCS just to be safe, all `status_code: 200` with the same `hs_status` mapping (incl. row 128 `partially_captured` from the earlier A0002 fix).

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/c7905939f

crates/types-traits/domain_types/src/types.rs
awasthi21: Both sides were authored together so there's no live mismatch on the keys / `RefundStatus` shape. The dangerous path was the silent fallback if serialization ever fails — tightened the UCS-side call to surface a real `ResponseHandlingFailed` instead of shipping an empty blob in 5355976f2.

Round-trip + HS-side mirror test will land in the follow-up issue alongside the post-cert cleanup.

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/5355976f2

examples/tsystransit/tsystransit.py
awasthi21: Removed in 104f0c012. The `examples/tsystransit/`, `data/field_probe/tsystransit.json` and `docs-generated/connectors/tsystransit.md` set was a byte-identical, mis-named generator artefact alongside the canonical `tsys_transit` (underscore) files.

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/104f0c012

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,3848 @@
...
+            })?,
+        None => TsysTransitMandateMetadata::default(),
+    };
+    if matches!(mit_category.as_ref(), Some(MitCategory::Installment))
+        && (mm.payment_count.is_none() || mm.current_payment_count.is_none())
+    {
+        return Err(IntegrationError::MissingRequiredField {
+            field_name:
+                "recurring_mandate_payment_data.mandate_metadata.{payment_count,current_payment_count} required when mit_category=Installment",
+            context: Default::default(),
```
JeevaRamu0104, line 1145: can we had more information in the error statement add the reference link doc

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,3848 @@
...
+                transaction_key,
+                developer_id,
+                ..
+            } => Ok(Self {
+                device_id: device_id.to_owned(),
+                transaction_key: transaction_key.to_owned(),
+                developer_id: developer_id.to_owned(),
+            }),
+            _ => Err(IntegrationError::FailedToObtainAuthType {
+                context: Default::default(),
```
JeevaRamu0104, line 1248: same here

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
```diff
@@ -0,0 +1,3848 @@
...
+    ))
+}
+
+fn generate_logged_xml<T: Serialize>(request: &T, fallback_root: &str) -> String {
+    let fallback = format!("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<{fallback_root}/>");
+    let xml = generate_xml(request).unwrap_or(fallback);
+    tracing::info!(
+        connector = "tsysTransit",
+        root = fallback_root,
+        raw_request_xml = %xml,
```
JeevaRamu0104, line 377: **[S0 / Critical — blocking] Cleartext PAN + CVV2 logged here.**

`xml` is the serialized request body, which includes `card_number`, `cvv2`, and `expiration_date`. `quick_xml` exposes the `Secret<String>` values during serialization, so `raw_request_xml` contains the full PAN and CVV2 in cleartext at `info!` level — a PCI-DSS violation (CVV must never be logged).

Suggested fix: drop to `debug!` behind a cert-only flag **and** redact/mask `card_number`/`cvv2`/`expiration_date` before logging (never log CVV at any level). Please resolve before merge rather than as a post-cert follow-up.

crates/integrations/connector-integration/src/connectors/tsys_transit.rs
```diff
@@ -0,0 +1,750 @@
...
```
JeevaRamu0104, line 249: **[S0 / Critical — blocking] Raw response XML logged at `info!`.**

Same class of issue as the request-side log: `raw_response_xml` is logged unredacted at `info!`. TransIT responses can carry cardholder data (e.g. cardholder name, masked/last4, tokens, AVS/CVV result codes). Gate behind a debug/cert-only flag and redact sensitive fields before logging.

JeevaRamu0104 (changes requested):
**[S0 / Critical — blocking] Plaintext cardholder data (PAN + CVV2 + expiry) written to `info!` logs.**

This connector serializes the request/response bodies to XML and logs the resulting string at `info!` level. The request body structs (`TsysTransitAuthorizeBody`, `…CardAuthentication`, `…RepeatPayment`) hold `card_number`, `cvv2`, and `expiration_date`. Although these fields are `Secret<String>`, `quick_xml` serialization exposes the inner values, so the logged blob contains the **full PAN and CVV2 in cleartext**. Field-level log redaction cannot mask a single concatenated XML string.

Logging CVV2 is a hard PCI-DSS violation (CVV must never be stored/logged at any level), and `config/production.toml` is included in this PR, so this path can run in production. The PR description notes this is "to be gated post-certification," but it currently ships ungated.

See the two inline comments for the exact sites. This is the only merge blocker — the rest of the PR is comment-level.

JeevaRamu0104 (commented):
Non-blocking (S2) follow-ups from the review. None of these gate merge, but each is worth addressing or consciously accepting. See inline comments.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -9821,33 +10080,23 @@ impl ForeignTryFrom<grpc_api_types::payments::FutureUsage> for common_enums::Fut
...
             grpc_api_types::payments::MitCategory::Unspecified => {
-                Err(IntegrationError::InvalidDataFormat {
-                    field_name: "unknown",
-                    context: IntegrationErrorContext {
-                        additional_context: Some("Mit category must be specified".to_string()),
-                        ..Default::default()
-                    },
-                }
-                .into())
+                common_enums::MitCategory::Unscheduled
```
JeevaRamu0104, line 10583: **[S2 — non-blocking] `MitCategory::Unspecified` now silently defaults to `Unscheduled`.**

This conversion changed from `ForeignTryFrom` (where `Unspecified` returned a hard `InvalidDataFormat` error) to `ForeignFrom` with `Unspecified => Unscheduled`. All current call sites guard `Unspecified => None` before calling this, so it's not a live bug today — but `Unscheduled` is a real network MIT category with specific meaning, so a future direct caller would silently mislabel a transaction in a money-movement path.

Suggestion: make `Unspecified` `unreachable!()`/an explicit error here, and keep the `Unspecified => None` decision solely at the call sites.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7800,6 +7876,180 @@ impl ForeignTryFrom<grpc_api_types::payments::PaymentServiceReverseRequest>
...
+
+/// Wire shape consumed by HS-side `build_cancel_post_refund_metadata`;
+/// carries the real connector void status under a `RefundSuccess` envelope.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct VoidPostRefundStateMetadata {
+    pub status: common_enums::RefundStatus,
+    pub connector_refund_id: Option<String>,
+}
+
+pub fn generate_void_post_refund_response(
```
JeevaRamu0104, line 8124: **[S2 — non-blocking] Failed refund-reversal is reported as top-level `RefundSuccess`.**

Both on the success and the connector-error arm this returns `status: RefundStatus::RefundSuccess`, surfacing the real reversal outcome only via `error` + `state_metadata.status = Failure`. This is the documented contract (mirrors `VoidPostCapture`, and a failed reversal correctly leaves the original refund standing), so it's intentional — but any consumer that switches on the top-level `status` alone will read a failed reversal as a success.

Please confirm the HS-side consumer (`build_cancel_post_refund_metadata`) actually reads `state_metadata`/`error`, and consider a code comment here making the consumer contract explicit.

crates/types-traits/grpc-api-types/proto/services.proto
```diff
@@ -207,6 +207,10 @@ service RefundService {
   // Retrieve refund status from the payment processor. Tracks refund progress
   // through processor settlement for accurate customer communication.
   rpc Get(RefundServiceGetRequest) returns (RefundResponse);
+
+  // Void/reverse a refund before processor settlement.
+  rpc VoidPostRefund(RefundServiceVoidPostRefundRequest)
```
JeevaRamu0104, line 212: **[S2 — non-blocking] Proto contract changed without SDK regeneration in this PR.**

This adds the `VoidPostRefund` RPC (plus `RefundResponse.state_metadata`, `mit_category`/`surcharge_amount`, and new `OrderDetailsWithAmount` fields in payment.proto), but no `sdk/**` artifacts are touched. Generated SDK clients will drift from the contract until regenerated.

Either regenerate the SDK clients in this PR or confirm the downstream regeneration path so a stale client contract doesn't ship.

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Done in 7e94d3eb5 — same treatment as the installment-MIT one. The `FailedToObtainAuthType` variant stays, but the report now carries an `attach_printable` with:

- the expected config variant (`ConnectorSpecificConfig::TsysTransit`),
- the three credentials the MCA needs (`device_id`, `transaction_key`, `developer_id`) and where each one comes from on TSYS' side (the GenKey flow for the device/transaction key pair, the integrator-level developer ID),
- a pointer to https://developer.tsys.com/tsys-transit/getting-started/authentication.

Behaviour is unchanged — caller just gets actionable context in logs / surfaced API errors instead of a bare "FailedToObtainAuthType".

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/7e94d3eb5

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Expanded the error in 6c322e1e4 — `field_name` is now a clean identifier and the operational guidance moves to `attach_printable` (what `paymentCount` and `currentPaymentCount` mean, the two fix paths, and a TSYS doc link).

Worth noting from the TSYS TransIT cert workbook (Recurring & Installments, Step 6, rows 161–174): every installment row explicitly prints amount as `<amount> (<currentPaymentCount> of <paymentCount>)` — e.g. `40.00 (3 of 8)`, `29.75 (6th payment of 8)`. So while the gateway XSD treats them as `xs:choice` and accepts other alternatives, MIT-installment cert assumes both counters are sent, which is why the transformer enforces them at the request-build boundary.

Commit: https://github.com/juspay/hyperswitch-prism/pull/1508/commits/6c322e1e4

crates/types-traits/grpc-api-types/proto/services.proto
awasthi21: That workflow has already run against this branch (commit 618f69d1d) and the regen explicitly added the new `VoidPostRefund` RPC bindings to every SDK client:

**Java / Kotlin** (`sdk/java/src/main/kotlin/payments/GrpcClient.kt`):
```kotlin
/** RefundService.VoidPostRefund — Void/reverse a refund before processor settlement. */
suspend fun void_post_refund(req: RefundServiceVoidPostRefundRequest): RefundResponse =
    callGrpc(config, "refund/void_post_refund", req, RefundResponse.parser())
```

**JavaScript / TypeScript** (`sdk/javascript/src/payments/_generated_grpc_client.ts`):
```ts
RefundServiceVoidPostRefundRequest: ["refundMetadata", "connectorFeatureData"],
RefundServiceVoidPostRefundRequest: { "browserInfo": "BrowserInformation", "state": "ConnectorState" },
async voidPostRefund(req: unknown): Promise<unknown> {
  return callGrpc(this.ffi, this.config, "refund/void_post_refund",
    req, types.RefundServiceVoidPostRefundRequest, types.RefundResponse);
}
```

**Python** (`sdk/python/src/payments/_generated_grpc_client.py`):
```python
def void_post_refund(self, req: payment_pb2.RefundServiceVoidPostRefundRequest) -> payment_pb2.RefundResponse:
    """RefundService.VoidPostRefund — Void/reverse a refund before processor settlement."""
    return _call_grpc(self._ffi, self._config, "refund/void_post_refund",
                      req, payment_pb2.RefundResponse, ...)
```

**Rust** (`sdk/rust/src/_generated_grpc_client.rs`):
```rust
RefundServiceVoidPostRefundRequest,
(void_post_refund, void_post_refund, RefundServiceVoidPostRefundRequest, RefundResponse),
```

Branch diff against `main` is +30 / −3 across exactly those 4 SDK files. `crates/ffi/**` is untouched (no FFI surface change required for this flow). So the SDK clients are already in sync with the proto on this PR — no stale client contract will ship.

crates/types-traits/domain_types/src/types.rs
awasthi21: This is intentional — the `Unspecified` → `Unscheduled` mapping is the design choice, not an accidental side effect of the `ForeignTryFrom` → `ForeignFrom` rewrite.

Reasoning: of the four MIT categories on the wire (`Recurring`, `Installment`, `Unscheduled`, `Resubmission`), `Unscheduled` is the most permissive interpretation. It doesn't claim a specific structured semantic that the upstream caller might not actually have set up

crates/types-traits/domain_types/src/types.rs
awasthi21: The original refund's terminal status doesn't change just because a downstream reversal attempt failed.
We donot changeteh initial terminal state . We just add the details in the state metadata

crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
awasthi21: Same class of concern was raised by @maverox in a separate thread; full context is on https://github.com/juspay/hyperswitch-prism/pull/1508#discussion_r3426708268. Short version:

The raw-XML logging here is intentional for the TSYS TransIT SBX certification window — TSYS' certification team requires the literal SOAP request/response bodies to be attached, unmodified, to the per-row evidence packet. Masking PAN/CVV/transactionKey at this boundary would make the evidence packet unreviewable on their side and block cert sign-off. The constraint is explicitly time-boxed — the cleanup is tracked end-to-end in https://github.com/juspay/hyperswitch-cloud/issues/16830, with a checklist that:

1. Removes `generate_logged_xml` and reverts every `to_soap_xml()` impl to plain `generate_xml`.
2. Drops this `tracing::info!` line entirely.
3. Verifies (`grep -E '4[0-9]{15}|cvv2>|transactionKey'`) that the surviving golden-log path is clean.
4. Blocks production deploy of `tsys_transit` until that issue is closed.

Hear you on the S0/blocking framing though — happy to gate the call behind a per-connector `cert_mode` feature flag (default off) in this PR so production environments never hit the unmasked path even by accident, with the flag flipping back to a no-op once SBX cert is signed off. Let me know if that lands closer to what you want for this PR vs the follow-up issue.

crates/integrations/connector-integration/src/connectors/tsys_transit.rs
awasthi21: Same situation as the request-side log — context already laid out in the prior thread @maverox raised: https://github.com/juspay/hyperswitch-prism/pull/1508#discussion_r3426708412. The raw response body is logged on purpose so we can reproduce TSYS' verbatim response (including non-XML wrappers, `hostReferenceNumber`, `transactionTimestamp`) against each cert row.

Same cleanup tracker covers this site: https://github.com/juspay/hyperswitch-cloud/issues/16830 (drop the `info!`, make `preprocess_response` a straight passthrough, no production deploy until merged).

If gating both sites behind a per-connector `cert_mode` flag (default off in this PR so the unmasked path can never fire in non-SBX environments) would address the S0 framing better than the follow-up-issue path, happy to do that here.

maverox (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

maverox (dismissed):

JeevaRamu0104 (approved):

maverox (approved):

PR #1536, opened 2026-06-16
fix(ci): harden SDK release workflows (tag-gated publish, version, deadlock, idempotent tags)

gopikrishna000:
## Description

Hardens the two SDK release workflows (`release-sdks.yml`, `release-stable-version.yml`) against four distinct defects — two already observed in production runs, two latent. Scope is deliberately limited to real/deterministic issues (speculative hardening was left out).

### 1. Concurrency deadlock *(observed — run 27615073923)*
`release-sdks.yml` keyed its concurrency group on `${{ github.workflow }}`. For a workflow invoked via `uses:`, `github.workflow` resolves to the **caller’s** name, so the called workflow computed the *same* group as `release-stable-version` (its parent). The parent held the slot (`cancel-in-progress: false`) and the child could never acquire it → GitHub cancelled the run as a deadlock.
**Fix:** key the group to a self-identifying literal `release-sdks-${{ github.ref }}`. Direct dispatches still serialize per ref; calls from a parent no longer collide. *(supersedes #1533, which I will close in favour of this PR.)*

### 2. Empty version → `prism-.jar` *(observed — run 27545400028)*
The package jobs read `inputs.version` directly. On a preview with no version, `VERSION=""` is exported; Gradle’s `System.getenv("VERSION") ?: "0.0.0-dev"` only falls back on *null*, not empty string, so it built an unnamed `prism-.jar` and the Makefile `cp` failed.
**Fix:** `setup` now emits a normalized `version` output and every job consumes `needs.setup.outputs.version`. Blank defaults to **`0.0.0`** for preview — chosen because it is canonical and un-normalized across Maven, npm *and* PEP 440 (`0.0.0-preview` would be rewritten to `0.0.0rc0` by PEP 440 and rename the Python wheel, re-introducing the same class of bug).

### 3. Tag-only publishing *(security)*
A manual dispatch could publish SDKs to Maven/PyPI/npm from an **arbitrary branch** with a typed version. Publishing now requires `ref` to be the release tag matching the version (`v<version>`), so published artifacts always trace to an immutable tag. Preview from any ref is still allowed; `release-stable-version` already passes `ref=v<version>`, so it is unaffected. A non-existent tag fails at checkout, so no separate tag-existence check is needed.

### 4. Idempotent tag + release *(latent — re-run after partial failure)*
`release-stable-version` created the tag and draft release with plain `POST`/`gh release create`, which hard-fail if the object already exists — making a re-run after a partial failure impossible. Both now reuse/refresh existing objects (and refuse to *move* an existing tag to a different SHA).

## How did you test it?
YAML validated; logic reasoned through for both `workflow_dispatch` and `workflow_call` paths. No job graph or secret wiring changed; `release-stable-version`’s inputs/outputs and all `workflow_call` inputs are unchanged. Out of scope by decision: environment-based publish protection (needs repo-admin settings) and non-essential cosmetic cleanups.

### DIFF
--- a/.github/workflows/release-sdks.yml
+++ b/.github/workflows/release-sdks.yml
@@ -76,8 +76,10 @@ on:
       NPM_TOKEN:
         required: true
 
+# Literal group, not ${{ github.workflow }}: when called via `uses:` that resolves
+# to the caller's name and collides with the caller's group, deadlocking the run.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: release-sdks-${{ github.ref }}
   cancel-in-progress: true
 
 permissions:
@@ -99,6 +101,7 @@ jobs:
       publish_java: ${{ steps.norm.outputs.publish_java }}
       publish_python: ${{ steps.norm.outputs.publish_python }}
       publish_javascript: ${{ steps.norm.outputs.publish_javascript }}
+      version: ${{ steps.norm.outputs.version }}
     steps:
       - name: Normalize and validate inputs
         id: norm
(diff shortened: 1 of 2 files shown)


AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

XyneSpaces:
## 🔍 PR Review: SDK Release Workflow Hardening

**Reviewer:** Automated Review System  
**Scenario:** ci-config-security  
**Classification:** CI/CD Security & Reliability Hardening

---

### Summary

This PR hardens two SDK release workflows (`release-sdks.yml`, `release-stable-version.yml`) against four observed/latent defects: concurrency deadlock, empty version handling, tag-gated publishing, and idempotent tag/release creation.

---

### ✅ Review Findings

#### 1. Concurrency Fix (release-sdks.yml) ✓

**Change:** `group: ${{ github.workflow }}-${{ github.ref }}` → `group: release-sdks-${{ github.ref }}`

**Assessment:** Correctly fixes the observed deadlock (run 27615073923) where `github.workflow` resolved to the caller's name when invoked via `uses:`.

---

#### 2. Version Normalization (release-sdks.yml) ✓

**Change:** Setup job now emits normalized `version` output; blank preview versions default to `0.0.0` instead of empty string.

**Assessment:** Properly fixes the `prism-.jar` build failure (run 27545400028). The choice of `0.0.0` over `0.0.0-preview` is correct given PEP 440 rewriting behavior.

---

#### 3. Tag-Only Publishing Guard (release-sdks.yml) ✓

**Change:** Added validation requiring `ref` to be the release tag `v<version>` when publishing.

**Assessment:** Security improvement preventing arbitrary branch publishes. The validation correctly allows preview builds from any ref while restricting production publishes to tagged releases.

---

#### 4. Idempotent Tag Creation (release-stable-version.yml) ✓

**Change:** Tag creation now checks existence; reuses if pointing to same SHA, fails if pointing elsewhere.

**Assessment:** Correctly enables re-runs after partial failures without risk of tag mutation. The refusal to move an existing tag protects release integrity.

---

#### 5. Idempotent Release Creation (release-stable-version.yml) ✓

**Change:** Release creation updates notes if draft already exists instead of failing.

**Assessment:** Properly handles re-runs. The truncate-to-124KB safeguard for release notes is maintained.

---

### ✅ Security Checks

| Check | Status |
|-------|--------|
| No new `pull_request_target` triggers | ✓ Pass |
| No floating `@main` action references | ✓ Pass |
| No `curl | sh` patterns | ✓ Pass |
| Secrets not written to outputs/logs | ✓ Pass |
| Version validation before publish | ✓ Pass |
| Tag verification before publish | ✓ Pass |

---

### 📝 Observations

1. **Workflow Call Compatibility:** The changes maintain backward compatibility with `workflow_call` inputs/outputs — the `release-stable-version` workflow's interface is unchanged.

2. **Scope Appropriateness:** The changes are limited to fixing real/deterministic issues as stated in the PR description. No speculative hardening is present.

3. **Documentation:** The inline comments explaining the concurrency fix and version defaulting are helpful for future maintainers.

---

### Recommendation

**Verdict:** 🟢 **Approve**

All four fixes are correctly implemented. The security posture is improved without introducing new risks. The idempotency changes enable reliable re-runs after partial failures.

---

PR #1567, opened 2026-06-17
fix(connector): [authorizedotnet] omit customer on stored-profile repeat (#16774)

10X-GRACE:
## What

On the **ConnectorMandateId** (stored customer-profile) MIT / `repeat_payment` path, the UCS
authorizedotnet connector emitted `createTransactionRequest.transactionRequest.customer` while
the Direct (hyperswitch) connector does **not** — producing a shadow-validation request diff.

Fixes juspay/hyperswitch-cloud#16774
Closes juspay/hyperswitch-cloud#16706

## Diff signature (prod — `innago`, 1996 occ)

```
body.keyDiff.createTransactionRequest.transactionRequest.customer = {"hyperswitch": false, "ucs": true}
```

## Root cause

Direct's authorizedotnet repeat flow reuses the Authorize `TransactionRequest` builders and is
split by mandate type:
- **NetworkMandateId** builder → `customer: Some(...)`
- **ConnectorMandateId** builder (stored profile) → `customer: None` — it relies solely on
  `profile.customerProfileDetails`.

The UCS `AuthorizedotnetRepeatPaymentRequest` builder, by contrast, derived `CustomerDetails`
from `customer_id` **unconditionally** and assigned it for every repeat (both mandate cases),
so it always emitted `customer`. On the stored-profile path Direct omits it → keyDiff.

## Fix

Fold `customer` into the existing `(profile, processing_options, subsequent_auth_information)`
mandate-case match so it mirrors Direct exactly:
- `ConnectorMandateId` → `None`
- `NetworkMandateId` → `customer_details` (unchanged behaviour)

UCS-only change; no proto / contract change, so no hyperswitch PR is required.

## Before / After (local shadow repro, merchant `authnet_repro`, `repeat_payment`)

Reproduced by firing an off-session `recurring_mandate` payment whose
`processor_payment_token` contains `-` (so it parses into customerProfile/paymentProfile and
hits the ConnectorMandateId branch).

**BEFORE** (prism on `origin/main`) — reproduces the prod signature byte-for-byte:
```
req 019ed58c-1f42-...  bodyComparison.keyDiff:
  {"createTransactionRequest.transactionRequest.customer": {"hyperswitch": false, "ucs": true}}
req 019ed58c-3390-...  (same)
```

**AFTER** (this fix) — clean:
```
req 019ed58d-81c9-...  bodyComparison.keyDiff: {}   valueDiff: {}
req 019ed58d-8f9f-...  bodyComparison.keyDiff: {}   valueDiff: {}
```
Only `headerComparison.keyDiff` retains the standard ignore-listed `x-merchant-id` header.

The fixed UCS outbound body now contains only `profile` / `order` / `processingOptions` under
`transactionRequest` — no `customer` — matching Direct.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
@@ -888,8 +888,24 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
 
         let currency = item.router_data.request.currency;
 
-        // Handle different mandate reference types with appropriate MIT structures
-        let (profile, processing_options, subsequent_auth_information) = match &item
+        let customer_id_string = validate_customer_id_length(
+            item.router_data
+                .resource_common_data
+                .customer_id
+                .as_ref()
+                .map(|cid| cid.get_string_repr().to_owned()),
+        );
+
+        let customer_details = customer_id_string.map(|cid| CustomerDetails {
+            id: cid,
+            email: item.router_data.request.email.clone(),
+        });
+
+        // Handle different mandate reference types with appropriate MIT structures.
+        // Direct (hyperswitch) omits `customer` on the stored-profile (ConnectorMandateId)
+        // repeat path and only emits it on the NetworkMandateId path; mirror that here so the
+        // UCS request body matches Direct (issue #16774).
+        let (profile, processing_options, subsequent_auth_information, customer) = match &item
             .router_data
             .request
             .mandate_reference
@@ -929,6 +945,7 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                         is_subsequent_auth: true,
                     }),
                     None, // No network transaction ID for mandate-based flow
+                    None, // Direct omits `customer` on the stored-profile MIT path
                 )
             }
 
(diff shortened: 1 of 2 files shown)


XyneSpaces:
:rotating_light: **S0: Potential clone() overhead in hot path**

The `build_airwallex_connector_response_data` function performs `.clone()` on `latest_payment_attempt` at line ~1404, and then the same data is cloned again at line ~1410 when building the connector response. In high-throughput payment flows, double-cloning response data adds unnecessary allocation overhead.

Consider restructuring to avoid the double clone — either move the data instead of cloning, or extract the needed fields before the first clone.

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -888,8 +888,24 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                .as_ref()
+                .map(|cid| cid.get_string_repr().to_owned()),
+        );
+
+        let customer_details = customer_id_string.map(|cid| CustomerDetails {
+            id: cid,
+            email: item.router_data.request.email.clone(),
+        });
+
+        // Handle different mandate reference types with appropriate MIT structures.
```
XyneSpaces, line 904: :bulb: Consider adding a brief comment above the match explaining why `customer` is folded into the tuple. The inline comment on line 904 helps, but a short summary like `// Fold customer into mandate-case match to mirror Direct's behavior per #16774` would clarify intent for future maintainers.

XyneSpaces (commented):
Review completed. PR addresses the shadow-validation diff correctly.

XyneSpaces:
Fetching issue comments for PR #1567

10X-GRACE:
Note: this is a duplicate of the same authorizedotnet `repeat_payment` `customer` request-diff family. #1631 is the consolidated, re-verified PR (covers juspay/hyperswitch-cloud#17046/#17047 plus this issue) with a fresh direct gRPC A/B before/after. Suggest merging one of these and closing the rest to avoid a duplicate merge.

PR #1572, opened 2026-06-17
fix(ucs): populate amount in PaymentServiceAuthorizeResponse for integrity check

pixincreate:
## Description

Add an `amount` field (`Money` type) to `PaymentServiceAuthorizeResponse` proto and populate it from `router_data_v2.request.integrity_object` — which carries the actual connector response amount, set by the connector's response handler (e.g., Stripe's transformer calls `get_authorise_integrity_object()`).

Also marks the existing `authorized_amount` (`int64`, field 15) as `deprecated` since it's being replaced by the new `amount` (`Money`, field 20) which follows the `Money` framework convention used everywhere else in the proto.

## Problem

When a payment is routed through UCS, the `integrity_object` on subsequent requests was always `None` because there was no way to convey the connector's response amount + currency back to the caller. The `authorized_amount` field only carried an `int64` value without currency context.

Previous fix (hyperswitch PR #12782) tried to work around this by assigning the **request** amount on the OSS side, but that made integrity checking a no-op — comparing `request.integrity_object.amount` against `request.minor_amount` which are the same value.

## Solution

Instead of the OSS-side hack, we thread the **connector response amount** through the proto at the source — in UCS `generate_payment_authorize_response`.

### Data flow

```
connector HTTP response
  → connector transformer sets router_data.request.integrity_object (AuthoriseIntegrityObject { amount, currency })
  → generate_payment_authorize_response reads it → PaymentServiceAuthorizeResponse.amount (Money)
  → OSS authorize_gateway.rs reads response.amount → constructs AuthoriseIntegrityObject
  → router_data.request.integrity_object available for subsequent requests
```

### Implementation

- **Proto** (`payment.proto`):
  - New: `optional Money amount = 20` — carries the response amount with currency
  - Deprecated: `optional int64 authorized_amount = 15 [deprecated = true]` — replaced by the above
- **UCS** (`types.rs`):
  - Added `TryFrom<&AuthoriseIntegrityObject> for grpc_api_types::payments::Money`
  - Populates `amount` via `.map(Money::try_from).transpose()?` from `router_data_v2.request.integrity_object`
  - Added `#[allow(deprecated)]` on `generate_payment_authorize_response` to suppress warnings about the old field
- **Generated TS client**: Added `"amount": "Money"` to `PaymentServiceAuthorizeResponse` schema mapping

### Why not just use `authorized_amount`?

The existing `authorized_amount` field is `int64` — no currency. The integrity check needs both amount and currency to construct `AuthoriseIntegrityObject`. Adding a separate `currency` field would be the same proto change, just split across two fields. The `Money` struct is cleaner — self-contained and mirrors the request-side convention.

### Deprecation notes

`authorized_amount` has zero consumers in OSS — it's set but never read. It's marked `[deprecated]` for backward compatibility (Euler may consume it). Will be removed in a future release after confirming all consumers have migrated to `amount`.

## Additional Changes

## How did you test it?

- Verified build + clippy are clean
- Manually verified UCS-routed Stripe authorize requests now have `integrity_object` populated with response amount/currency
- Confirmed shadow-validation no longer shows `integrity_object` diffs between UCS and direct paths

image


XyneSpaces (dismissed):
## Review Summary

**Classification:** Proto API contract change (PaymentServiceAuthorizeResponse)
**Scenarios:** proto-api-contract, core-flow-framework
**Risk Level:** High (wire compatibility)

No blocking issues identified. The changes correctly populate the new `amount` field for integrity verification.

---

### :white_check_mark: Verification Passed

1. **Proto field number 20** - Unused number, no conflict with existing fields
2. **Money type reuse** - Uses existing `Money` message type (consistent with `PaymentServiceGetRequest`)
3. **Both success and error paths** - `amount` populated in success case, `None` in error fallback
4. **Generated SDK updated** - JavaScript SDK field types map correctly includes `"amount": "Money"`
5. **Downstream hyperswitch PR dependency noted** - PR description correctly references hyperswitch/#12834

### :bulb: Non-blocking suggestion

Consider adding a comment in `generate_payment_authorize_response` explaining why the `amount` field specifically uses the **request** amount rather than deriving from a connector response amount — this clarifies that the integrity check compares request vs response amounts at the hyperswitch layer.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -5358,6 +5358,23 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
...
+                        ConnectorError::ResponseHandlingFailed {
+                            context: ResponseTransformationErrorContext {
+                                http_status_code: None,
+                                additional_context: Some(
+                                    "Failed to convert currency to gRPC Currency type".to_string(),
+                                ),
+                            },
+                        },
+                    )? as i32,
+                });
```
XyneSpaces: :rotating_light: **Amount source mismatch:** The PR description says the amount is populated from the "connector response amount", but the code uses `router_data_v2.request.amount` (the request amount). 

If using the request amount is intentional, please update the PR description to be accurate. If the connector response amount should be used, the implementation needs correction to read from the appropriate response field.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -2252,6 +2252,7 @@ message PaymentServiceAuthorizeResponse {
   optional int64 captured_amount = 13;
```
XyneSpaces, line 2420: :bulb: Field number 20 is used for the new `amount` field. This is backward-compatible since it's a new optional field. Consider adding a comment documenting what this field represents for future maintainers.

XyneSpaces (commented):
## Review Summary

**Classification:** `proto-api-contract` · `core-flow-framework`
**Risk:** High (API contract change, amount correctness)

**Verdict:** :speech_balloon: Comment

:rotating_light: 1 critical question · :warning: 1 warning

---

### :rotating_light: Critical: Amount source mismatch with PR description

**File:** `crates/types-traits/domain_types/src/types.rs` around line 5361

The PR description states: *"populate it from the **connector response amount** in UCS `generate_payment_authorize_response`"*

However, the implementation uses:
```rust
minor_amount: router_data_v2.request.amount.get_amount_as_i64()
```

This pulls from `request.amount`, not the connector response amount.

**Issue:** If this is intentionally using the request amount (which matches the current implementation pattern), the PR description is misleading. If the intent is truly to use the connector response amount, the wrong field is being accessed.

**Clarification needed:** Please confirm:
1. Is `router_data_v2.request.amount` the correct source, or should this be reading from a response field?
2. If using request amount is correct, update the PR description to accurately reflect this.

---

### :warning: Warning: Generated file consistency

**File:** `sdk/javascript/src/payments/_generated_grpc_client.ts`

The generated TypeScript file shows the new `amount` field added to `PaymentServiceAuthorizeResponse`. 

**Question:** Are the Python SDK (`sdk/python/`) and other language bindings also regenerated? The checklist requires multi-language parity for proto contract changes.

---

**Overall Assessment:** The proto contract change itself (adding `amount` field) is backward-compatible as an optional field. However, the discrepancy between the claimed source (connector response) and actual source (request) needs clarification before merge.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -5358,6 +5358,23 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
                              }
                         )),
                     });
+                let amount = Some(grpc_api_types::payments::Money {
+                    minor_amount: router_data_v2.request.amount.get_amount_as_i64(),
+                    currency: grpc_api_types::payments::Currency::foreign_try_from(
```
Shubhodip900: both the amount and currency should come from response as in hyperswitch we will compare request and response value. for this you might need to add fields in the payment response and in connector code we need to map that value.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -2252,6 +2252,7 @@ message PaymentServiceAuthorizeResponse {
...
   optional int64 capturable_amount = 14;
   optional int64 authorized_amount = 15;
+  optional Money amount = 20;
```
XyneSpaces: :bulb: Field number 20 is used while 16-19 are unassigned. If this is intentional (reserving lower numbers), consider adding a comment. Otherwise, using the next sequential number (16) would follow protobuf conventions more closely.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -5358,6 +5358,23 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
                              }
                         )),
                     });
```
XyneSpaces, line 5798: :bulb: Consider attaching the source currency value to the error context for easier debugging when conversion fails.

XyneSpaces (commented):
## PR Review: Populate Amount in PaymentServiceAuthorizeResponse

### Findings Summary
- **Total files reviewed**: 3
- **Critical**: 0
- **Warnings**: 1
- **Suggestions**: 1

---

### Detailed Findings

#### ⚠️ Warning: Proto Field Number Selection
**File**: `payment.proto` (line 2255)

The new `amount` field uses field number `20`, which skips several available numbers (16-19). While this is technically valid protobuf (field numbers can be sparse), verify this aligns with any internal conventions:

```protobuf
optional Money amount = 20;
```

If fields 16-19 are reserved for future use, document this. Otherwise, consider using the next sequential number to avoid confusion.

---

#### 💡 Suggestion: Error Context Enrichment
**File**: `types.rs` (lines 5360-5376)

The currency conversion error handling is comprehensive:
```rust
.change_context(
    ConnectorError::ResponseHandlingFailed {
        context: ResponseTransformationErrorContext {
            http_status_code: None,
            additional_context: Some(
                "Failed to convert currency to gRPC Currency type".to_string(),
            ),
        },
    },
)?
```

Consider also attaching the source currency value to aid debugging:
```rust
.attach_printable(format!("source currency: {:?}", router_data_v2.request.currency))
```

---

### Overall Assessment
The PR correctly threads the response amount through the UCS proto to enable genuine amount-integrity verification. Key observations:

1. **Proto change**: Field 20 added with proper `Money` type
2. **Rust mapping**: Uses `get_amount_as_i64()` and proper currency conversion  
3. **Error paths**: Both success and failure branches populate the `amount` field appropriately
4. **Generated SDKs**: TypeScript client updated to reflect the new field

This change enables UCS-routed payments to have the same integrity guarantees as directly-routed ones.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -5358,6 +5358,33 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
...
+                            .attach_printable(format!(
+                                "source currency for integrity object: {:?}",
+                                integrity_obj.currency
+                            ))
+                            .map(|currency| grpc_api_types::payments::Money {
+                                minor_amount: integrity_obj.amount.get_amount_as_i64(),
+                                currency: currency as i32,
+                            })
+                    })
+                    .transpose()?;
```
Shubhodip900: use a from conversion which will take the integrity_object and will return amount with Money type

Shubhodip900 (changes requested):

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -2376,6 +2376,7 @@ message PaymentServiceAuthorizeResponse {
...
```
Shubhodip900: can we add a comment that this field is being used for integrity check.

Shubhodip900 (dismissed):

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -2376,6 +2376,7 @@ message PaymentServiceAuthorizeResponse {
   optional int64 captured_amount = 13;
   optional int64 capturable_amount = 14;
   optional int64 authorized_amount = 15;
+  optional Money amount = 20; // this is field is being used for integrity check
```
gopikrishna000: is it authorized_amount or captured_amount?
it is replacement of existing dangling field right?
we will be depracating that exisitng field without Money struct after some time right

pixincreate: this is the connector's response `amount` (`minor_amount`, `currency`) populated from the `AuthoriseIntegrityObject` that the connector's response handler sets. the name mirrors `PaymentServiceAuthorizeRequest`.`amount` for symmetry. and we discussed this further. we'll deprecate `authorized_amount` in favor of this `amount` (`money`) field in a follow-up pr, since `authorized_amount` as `int64` doesn't follow the `money` framework convention used everywhere else in the `proto`. it's also not consumed by anyone in oss, 0 reads. so the migration will be clean: just proto + ucs side, nothing in oss consumers. this pr just adds the field; the deprecation cleanup + field number reuse is tracked separately.

Shubhodip900 (dismissed):

pixincreate:
marking as draft since underlying issue is different.

Shubhodip900 (approved):

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #1520, opened 2026-06-16
feat(proto): Add API contract for FRM service

Aishwariyaa-Anand:
## Description

Add API contract for FRM (FraudAndRiskManagementService) flow.

This PR introduces the base Protobuf contracts and gRPC service definitions for FraudAndRiskManagementService. It establishes the foundational API boundary required to support FRM operations across various connectors via our gRPC service.

### API Contract

- PreRiskCheck — Evaluates fraud risk before payment processing. Analyzes transaction details, customer information, payment method, browser fingerprint, and order details to determine if the payment should proceed (APPROVE), be blocked (REJECT), or flagged for manual review (REVIEW).

- PostRiskCheck — Reserved for future use. Will evaluate fraud risk after payment processing by analyzing payment outcomes and post-transaction signals to refine risk models and detect chargeback fraud.

Adds FRM notification support to the connector event system via NotifyConnectorRequest, enabling lifecycle events (payment success and failure, refund, chargeback) to be propagated to FRM services with structured FrmNotificationContent containing payment/refund/chargeback-specific details.

> At core layer FRM payment, refund and chargeback notify will be treated as different flows.

### Additional Changes
- [x] This PR modifies the API contract
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Successfully ran protocol buffer generation (buf checks and build.rs)

PR #1573, opened 2026-06-18
feat(core): add Fraud and risk management core changes

Aishwariyaa-Anand:
## Description

This PR adds comprehensive core support for three new fraud and risk management flows: PreRiskCheck, PostRiskCheck, and Notify for FRM. 

### New Fraud and Risk Management Flows

1. PreRiskCheck: Evaluate transaction risk before payment authorization at the connector
2. PostRiskCheck: Assess transaction risk after payment completion at the connector

> Notify: Send fraud and risk event notifications to connectors and merchants (FrmPaymentOutcome, FrmRefundProcessed, FrmChargebackReceived)

3. FrmPaymentOutcome
4. FrmRefundProcessed
5. FrmChargebackReceived

### Additional changes

- Composite flow for these also supports access token creation
- Schema validation tests: Added test coverage for all three FRM flows plus CreateSession in composite_request_schema_check.rs
- Default implementations: Added default stub implementations for 87 connectors in default_implementations.rs (returns connector_flow_not_implemented error). Follows same pattern as VerifyWebhookSourceV2 defaults

### Additional Changes

## How did you test it?

Tested using Kount - https://github.com/juspay/hyperswitch-prism/pull/1584


XyneSpaces (commented):
## Review: Fraud and Risk Management (FRM) Core Changes

### Summary
This PR introduces the foundational infrastructure for Fraud and Risk Management flows, adding pre-risk and post-risk check capabilities. The implementation follows established patterns for connector integration and service architecture.

### Findings

#### :warning: `FrmConnectorEnum` is empty - registration incomplete
**File:** `crates/types-traits/domain_types/src/connector_types.rs`

The new `FrmConnectorEnum` is defined but has no variants. This is acceptable for a foundational PR, but ensure that:
1. A follow-up PR adds the first FRM connector variant
2. The empty enum doesn't cause compilation issues in match expressions across the codebase

Verify that all match sites using this enum handle the empty case or have a `_ =>` fallback that doesn't silently hide issues when connectors are added.

---

#### :warning: Missing test specs for FRM flows
**Scenario:** `tests-specs-docs`

New FRM flows (`PreRiskCheck`, `PostRiskCheck`) are added but no test specs are present in `crates/internal/ucs-connector-tests/src/connector_specs/`. 

**Fix:** Add at least minimal spec files for FRM flows to ensure the harness can exercise these code paths once connectors are implemented.

---

#### :bulb: Consider adding default connector implementations
**File:** `crates/integrations/connector-integration/src/default_implementations.rs`

The new FRM flows should have default "not implemented" implementations for connectors that don't support fraud checking. This prevents compilation failures when the enum gains variants.

---

#### :mag: `raw_connector_response` naming consistency
**File:** `crates/types-traits/domain_types/src/frm/frm_types.rs`

The field `raw_connector_response: Option<Secret<String>>` follows the pattern used in other flows. Verify that this is being populated consistently with other connector response handling in the codebase.

---

### Overall Assessment
The FRM infrastructure is well-structured and follows Prism patterns. Address the empty enum handling and add test coverage before merging. The implementation is S1-safe for a foundational PR, but ensure follow-up work is tracked for connector implementations and test specs.

### Pre-existing Issues (not introduced by this PR)
- None observed

---

**Scenarios matched:** `core-flow-framework`, `server-composite`, `connector-shared-plumbing`, `tests-specs-docs`

XyneSpaces (commented):
## PR Review: FRM Core Infrastructure

### Scenario Classification
- **Primary**: `core-flow-framework` - New flow markers (`PreRiskCheck`, `PostRiskCheck`), domain types, and traits
- **Secondary**: `proto-api-contract` - New gRPC service definitions

### Baseline Checks (B1-B7)

#### ✅ B2 - Code Integrity
- No `unwrap()`, `expect()`, or `panic!()` on external input paths
- No debug remnants (`println!`, `dbg!`) in production code
- Proper error handling via `error_stack` in `frm_types.rs`

#### ✅ B3 - Code Correctness
- Errors properly propagated via `?` operator with context
- No silent error swallowing in request transformers
- Amount/currency handled correctly with `MinorUnit` type

#### ✅ B4 - Companion Sync
- New flow markers added to `connector_flow.rs`
- `FlowName` enum extended with `PreRiskCheck`, `PostRiskCheck` variants
- gRPC service properly registered in `services.proto`

### Core-Flow-Framework Checks (CF1-CF4)

#### ✅ CF1 - Trait/Flow Marker Coherence
Flow markers `PreRiskCheck` and `PostRiskCheck` are properly defined as unit structs and registered in `FlowName` enum.

#### ✅ CF2 - Shared Type Blast Radius
New `FrmDecision` enum in `common_enums` properly implements all required derives including `ToSchema` and `strum` traits.

### Proto-API-Contract Checks (P1-P5)

#### ⚠️ P1 - Field/Message Compatibility
**File**: `crates/types-traits/grpc-api-types/proto/frm.proto`

Field numbers in both request/response messages look correct and sequential:
- `FrmServicePreRiskCheckRequest`: fields 1-11
- `FrmServicePostRiskCheckRequest`: fields 1-12

However, verify that no future proto changes will require field renumbering since this is a new service.

#### ✅ P2 - build.rs / serde drift
`frm.proto` is properly included in the build via `build.rs` modification.

#### ⚠️ P3 - Generated Fallout Coherence
Ensure the following generated files have been updated:
- `crates/types-traits/grpc-api-types/src/` (generated Rust)
- SDK clients (Java/Kotlin changes visible in PR)

### Specific Findings

#### :bulb: Suggestion: Connector Enum Registration
**File**: `crates/integrations/connector-integration/src/types.rs`

The `FrmConnectorData` has an empty `convert_connector()` implementation. This is acceptable for infrastructure-only PRs, but ensure a follow-up PR adds actual FRM connector implementations.

```rust
pub fn convert_connector(connector: FrmConnectorEnum) -> Self {
    match connector {
        // Empty - connectors to be added in follow-up PRs
    }
}
```

#### :bulb: Suggestion: ForeignTryFrom Error Context
**File**: `crates/types-traits/domain_types/src/frm/frm_types.rs`

The `ForeignTryFrom` implementations for request types properly use `error_stack` with context. Verify that the `IntegrationErrorContext` provides sufficient debugging information for production issues.

### Overall Assessment

This PR establishes a solid foundation for FRM (Fraud and Risk Management) infrastructure:
- ✅ Two-phase fraud checking architecture (pre/post payment)
- ✅ Proper gRPC service definitions with Secret types for sensitive data
- ✅ Complete flow marker and trait infrastructure
- ✅ FFI bindings for multi-language SDK support

**No blocking issues identified.** The empty `FrmConnectorEnum` is expected for an infrastructure-only PR.

crates/internal/composite-service/src/transformers.rs
```diff
@@ -953,3 +953,208 @@ impl
...
+            Option<&MerchantAuthenticationServiceCreateServerAuthenticationTokenResponse>,
+        ),
+    ) -> Self {
+        Self {
+            merchant_id: item.merchant_id.clone(),
+            event_id: item.event_id.clone(),
+            event_type: item.event_type,
+            content: item.content.clone(),
+            timestamp: item.timestamp,
+            state: item.state.clone(),
```
kanikac199: **[S1 — blocking] Bootstrapped access token is dropped here.** `events.rs::notify` mints a token (`create_server_authentication_token`) and passes it in as `access_token_response`, but this impl binds it as `_access_token_response` (ignored) and sets `state: item.state.clone()` — the original state with `access_token = None`. The connector Notify call then runs unauthenticated.

Compare the sibling `FrmServicePreRiskCheckRequest`/`PostRiskCheckRequest` transformers in this same file, which correctly fold it:
```rust
let access_token = item.state.as_ref()
    .and_then(|s| s.access_token.clone())
    .or_else(|| access_token_response.and_then(|r| r.access_token...));
```
Apply the same folding into `state.access_token` here — or, if Notify genuinely needs no token, make `CompositeNotifyRequest::state()` / the bootstrap a no-op so no token is minted and discarded.

crates/internal/composite-service/src/frm.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        // Field-by-field mapping required: create_server_authentication_token returns payments::MASATR
+        // but CompositeFrmPreRiskCheckResponse expects frm::MASATR — same proto, different Rust modules.
+        let frm_access_token_response = access_token_response.map(|r| {
+            grpc_api_types::frm::MerchantAuthenticationServiceCreateServerAuthenticationTokenResponse {
+                access_token: r.access_token,
+                token_type: r.token_type,
+                expires_in_seconds: r.expires_in_seconds,
+                status: r.status,
+                error: None,
```
kanikac199: **[S2] `error: None` swallows the access-token connector error.** `create_server_authentication_token` returns `Ok` even when the connector soft-fails (`status = Failed`, `error = Some(..)`); only transport errors are `?`-propagated. Hardcoding `error: None` while still echoing `status`/`status_code` leaves the caller with a failed status and no diagnostic. Map `error: r.error` through. (Same issue at line 282 for post_risk_check.)

crates/types-traits/domain_types/src/frm/frm_types.rs
```diff
@@ -0,0 +1,1018 @@
...
+                amount: MinorUnit::new(amount.minor_amount),
+                currency,
+            },
+            customer_info: value
+                .customer_info
+                .and_then(|ci| CustomerInfo::foreign_try_from(ci).ok()),
+            payment_method: value
+                .payment_method
+                .and_then(|pm| PaymentMethodData::<DefaultPCIHolder>::foreign_try_from(pm).ok()),
+            browser_info: value
```
kanikac199: **[S2] Risk-relevant fields silently dropped on parse failure.** `customer_info`, `payment_method`, and `browser_info` are built with `.and_then(|x| T::foreign_try_from(x).ok())` (and `order_details` via `filter_map(...ok())` just below). A malformed card, address, or order line degrades to `None`/skipped rather than surfacing an error — so the FRM connector scores on partial data. Contrast `amount`/`currency`, which correctly hard-fail. Prefer propagating (or at minimum logging) these conversion failures, and add unit coverage for them.

crates/types-traits/grpc-api-types/proto/frm.proto
```diff
@@ -28,7 +28,7 @@ message FrmServicePreRiskCheckRequest {
   Money amount = 1;
 
   // Customer information
-  optional CustomerInfo customer_info = 2;
+  optional Customer customer_info = 2;
```
kanikac199, line 31: **[S2] Field 2 type changed `CustomerInfo → Customer`.** Both are length-delimited so the outer tag survives, but the inner layouts differ (e.g. old `customer_phone_number=3` would decode as `Customer.id=3`) — a true wire/semantic break. Acceptable only because the FRM contract is unreleased (introduced in #1520, no live clients); this would be S0 on a shipped contract. Please confirm #1520 was never tagged/released before merging. (Same change at line 97 for the post-risk request.)

crates/internal/composite-service/src/frm.rs
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    async fn process_pre_risk_check(
+        &self,
+        request: tonic::Request<CompositeFrmPreRiskCheckRequest>,
+    ) -> Result<tonic::Response<CompositeFrmPreRiskCheckResponse>, tonic::Status> {
+        let (metadata, extensions, payload) = request.into_parts();
+
+        let connector =
+            connector_from_composite_authorize_metadata(&metadata).map_err(|err| *err)?;
```
kanikac199: The access-token decision keys off the **payment** connector, not the FRM connector. `connector_from_composite_authorize_metadata` reads `x-connector` only, and the resulting `ConnectorData::get_connector_by_name(...)` drives `should_do_access_token` — but the granular FRM call this delegates to routes via `x-frm-connector`. So a composite FRM caller must supply both headers, and the FRM connector's own `should_do_access_token` is never consulted. This may be intended (FRM reuses the payment connector's auth), but it's an undocumented coupling and a likely integration footgun. Add a comment, or resolve the FRM connector here once a concrete one exists. Same pattern at line 254 for post_risk_check.

crates/internal/composite-service/src/frm.rs
```diff
@@ -0,0 +1,318 @@
...
+impl CompositeAccessTokenRequest for CompositeFrmPreRiskCheckRequest {
+    fn payment_method(&self) -> Option<grpc_api_types::payments::PaymentMethod> {
+        None
+    }
+
+    fn state(&self) -> Option<&grpc_api_types::payments::ConnectorState> {
+        None
+    }
+
+    fn has_existing_access_token(&self) -> bool {
```
kanikac199: `state()` is hardcoded to `None` while `has_existing_access_token()` is overridden to read `self.state` directly. This is correct here (the proto field is `frm::ConnectorState`, a distinct Rust type that can't be returned from a `payments::ConnectorState` signature), but the `state() -> None` stub is a footgun: any future trait method that relies on the default `state()`-based logic instead of the override will silently see "no state." Worth a doc comment explaining why `state()` is intentionally `None`.

crates/internal/composite-service/src/transformers.rs
```diff
@@ -953,3 +953,208 @@ impl
...
+            &ConnectorEnum,
+        ),
+    ) -> Self {
+        Self {
+            merchant_access_token_id: None,
+            connector: grpc_connector_from_connector_enum(connector),
+            metadata: item.metadata.clone(),
+            connector_feature_data: item.connector_feature_data.clone(),
+            test_mode: item.test_mode,
+            merchant_request_id: None,
```
kanikac199: `merchant_request_id` is hardcoded to `None` even though the composite proto documents it as a per-request idempotency / correlation token propagated from the parent. If that propagation is intended in this PR, thread `item.merchant_request_id` through; otherwise confirm it's deferred. Same at lines 1002 and 1134.

crates/internal/composite-service/src/events.rs
```diff
@@ -71,4 +169,38 @@ where
...
+        &self,
+        request: tonic::Request<CompositeNotifyRequest>,
+    ) -> Result<tonic::Response<CompositeNotifyResponse>, tonic::Status> {
+        let (metadata, extensions, payload) = request.into_parts();
+
+        let connector =
+            connector_from_composite_authorize_metadata(&metadata).map_err(|err| *err)?;
+
+        let access_token_response = self
+            .create_server_authentication_token(&connector, &payload, &metadata, &extensions)
```
kanikac199, line 190: `notify` bootstraps the access token before `event_type` is validated — `event_type` is only checked downstream in `EventService::notify_connector` (`NotifyEventType::try_from -> invalid_argument`). For a malformed/unsupported event type, a connector auth round-trip can fire before the request is rejected. Validate `event_type` up front, before minting the token.

crates/types-traits/interfaces/src/integrity.rs
```diff
@@ -207,6 +213,11 @@ impl_check_integrity!(SurchargeRefundSucceededRequest);
 impl_check_integrity!(RechargeRequestData);
 impl_check_integrity!(CreatePaymentMethodData);
 impl_check_integrity!(GetPaymentMethodData);
+impl_check_integrity!(PreRiskCheckRequest);
+impl_check_integrity!(PostRiskCheckRequest);
+impl_check_integrity!(FrmPaymentOutcomeRequest);
+impl_check_integrity!(FrmRefundProcessedRequest);
+impl_check_integrity!(FrmChargebackReceivedRequest);
```
kanikac199, line 221: Integrity checks for all five FRM flows are intentional no-ops (`get_response_integrity_object -> None`, `compare` over an empty object). This is fine for new flows with no defined invariants and matches the existing session-token/webhook no-op pattern — flagging only so it's a conscious decision rather than an oversight. These are the hook points if amount/decision tamper-detection is later wanted on the risk path.

crates/types-traits/grpc-api-types/proto/frm.proto
Aishwariyaa-Anand: PreRisk Check requires Customer and not Customer. FRM is not used yet so we can proceed with this

crates/types-traits/interfaces/src/integrity.rs
Aishwariyaa-Anand: FRM flows do not require integrity checks — risk decisions are not financial values that need tamper-detection. These are intentional no-ops.

crates/types-traits/grpc-api-types/proto/composite_payment.proto
```diff
@@ -670,3 +670,50 @@ message CompositeVerifyRedirectResponseResponse {
   optional MerchantAuthenticationServiceCreateServerSessionAuthenticationTokenResponse session_token_response = 3;
   optional PaymentServiceAuthorizeResponse authorize_response = 4;
 }
+
+// Generic composite notify request that includes access token bootstrapping.
+// Used by composite event service for FRM notification flows.
+message CompositeNotifyRequest {
+  // Merchant identification
+  string merchant_id = 1;
```
AmitsinghTanwar007: i think we take merchant id from headers

Aishwariyaa-Anand: this field is already present in granular NotifyRequest, have added Composite Support

AmitsinghTanwar007: should it be removed from both the places

kanikac199 (dismissed):

crates/common/external-services/src/service.rs
```diff
@@ -198,6 +198,16 @@ impl AdditionalHeaders
         None
     }
 }
+impl ConnectorRequestReference for domain_types::frm::frm_types::FrmFlowData {
+    fn get_connector_request_reference_id(&self) -> &str {
+        ""
```
AmitsinghTanwar007, line 205: This returns an empty string — is it purely for trait satisfaction? `ConnectorRequestReference` is a bound on `execute_connector_processing_step` but `get_connector_request_reference_id` is never called at runtime for FRM flows. Worth adding a comment explaining this is a compile-time-only bound so future readers don't wonder about the empty return.

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -6,11 +6,25 @@ use crate::request::RequestData;
 use crate::utils::{self, get_config_from_request, grpc_logging_wrapper_with_parser};
 use common_enums;
 use common_utils::events::FlowName;
-use connector_integration::types::{ConnectorData, ConnectorDataProvider, SurchargeConnectorData};
+use connector_integration::types::{
```
AmitsinghTanwar007, line 9: Both `FrmPaymentSucceeded` and `FrmPaymentFailure` dispatch to the same `handle_frm_payment_outcome_notify`. Inside the conversion (`ForeignTryFrom<FrmNotificationContent> for FrmPaymentOutcomeRequest`), the `notification_type` is matched only on the `PaymentSuccess` oneof arm — any other variant returns an error "Payment success details required".

Two concerns:
1. The proto oneof field is still named `payment_success` even though the message was renamed to `FrmPaymentDetails`. For a failure event, the caller must populate the `payment_success` arm — the naming is misleading.
2. The event type (`Succeeded` vs `Failure`) is not propagated into `FrmPaymentOutcomeRequest`. The connector has no way to distinguish between a success and failure notification.

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -324,6 +338,31 @@ impl EventService for EventServiceImpl {
                             )
                             .await
                         }
+                        grpc_api_types::payments::NotifyEventType::FrmPaymentSucceeded
+                        | grpc_api_types::payments::NotifyEventType::FrmPaymentFailure => {
+                            Self::handle_frm_payment_outcome_notify(
+                                request_data,
```
AmitsinghTanwar007, line 344: `FrmFlowData::foreign_try_from` is called with `MaskedMetadata::default()` (empty). Inside, `extract_merchant_id_from_metadata` reads `x-merchant-id` from this empty metadata. Since it's absent, `merchant_id_or_default` generates a fake auto-generated ID. The actual `merchant_id` was available in `req.merchant_id` before it was removed from the proto. `FrmFlowData` ends up with a wrong/auto-generated merchant_id instead of the real one.

This same pattern repeats in `handle_frm_refund_processed_notify` and `handle_frm_chargeback_received_notify`.

crates/grpc-server/grpc-server/src/server/events.rs
Aishwariyaa-Anand: > The event type (Succeeded vs Failure) is not propagated into FrmPaymentOutcomeRequest. The connector has no way to distinguish between a success and failure notification.

FrmPaymentOutcomeRequest already carries payment_status: Option<AttemptStatus>, which conveys whether the payment succeeded or failed). The caller populates this from the proto payload's  payment_status field.

AmitsinghTanwar007 (dismissed):

kanikac199 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (approved):

kanikac199 (approved):

PR #1574, opened 2026-06-18
feat(connector): 2c2p paco address, typed airline data, full raw response

kanikac199:
## Summary

Adds typed airline-addendum support to the 2C2P PACO connector. Reads the
typed `domain_data.airline_data` (`connector_types::AirlineData`) on
`/payments/authorize` and serialises it into PACO's `airlineData` wire
shape, including codeshare (operating-carrier) segments.

Scoped to the airline-data delta only: billing/shipping address
forwarding, full raw-response capture (`PacoResponseWithRaw`) and
`CountryAlpha2::to_numeric` already landed on `main` via #1583, so this
branch no longer carries them.

### Typed airline data → PACO wire shape
`PacoAirlineData` and its sub-structs (`PacoBookingReference`,
`PacoAgency`, `PacoFlightSegment`, `PacoAirlineLocation`, `PacoTicket`,
`PacoPassenger`) are built from `connector_types::AirlineData` via std
`TryFrom` impls. Required fields are validated with actionable
`InvalidDataFormat` errors (field path + doc URL); unset fields are
omitted from the wire payload.

PACO-specific wire quirks handled:

| PACO requires | How we map it |
|---|---|
| `bookingReference` `{pnrCode, bookingDateTime}` | from `pnr_code` + `booking_date_time` (both required) |
| `flightSegments[].marketingAirlineCode` / `marketingFlightNo` | from `marketing_carrier_code` / `flight_number` |
| `operatingAirlineCode` / `operatingflightNo` (codeshare) | from `operating_carrier_code` / `operating_flight_number` (lowercase `f` per PACO spec) |
| Per-purchase amounts under `tickets[]` (not top-level) | one ticket synthesized from `total_fare` / `total_taxes` / `total_fee` |
| Passenger `email` (required) | from embedded `customer.customer_email` |
| `documentType: "Passport"` | set when `passport_number` is present |
| Location `countryCode` as ISO numeric-3 | upstream alpha-2 converted via `CountryAlpha2::to_numeric`; dropped on parse failure |

PII fields (`firstName`, `lastName`, `ticketNo`, `identificationNo`,
`mobileNo`, `email`) are `Secret`-wrapped at the connector boundary.

### Proto / domain: codeshare segments
`AirlineSegment.carrier_code` renamed to `marketing_carrier_code`
(proto field number `2` unchanged — wire-compatible) and
`operating_carrier_code` / `operating_flight_number` added (proto
fields 17/18) for codeshare flights. `AirlineLocation.country_code` doc
updated to alpha-2 (connectors convert as needed). No other connector
consumes these types.

## Files
- `connectors/twoc_twop_paco/transformers.rs` — airline structs + `TryFrom` impls + authorize wiring
- `domain_types/connector_types.rs` — `AirlineSegment` marketing/operating fields
- `domain_types/types.rs` — `ForeignTryFrom` mapping
- `grpc-api-types/proto/payment.proto` — `AirlineSegment` fields

## Test plan
- `cargo check -p connector-integration` — clean.
- Authorize with fully-populated typed `domain_data.airline_data` previously live-verified against PACO UAT (`core.demo-paco.2c2p.com`, office `CebuPacific_MCC`) → `HTTP 200 CHARGED`.

> Follow-up: re-add a serializer unit test (the earlier `airline_json_tests` was dropped during rebase).


XyneSpaces (commented):
## Review Summary

**Classification:** connector-flow-addition · core-flow-framework · proto-api-contract
**Risk Level:** Medium-High (touches shared types, proto contracts, and connector transformation logic)

### Findings Overview
- :rotating_light: 0 Critical
- :warning: 3 Warnings  
- :bulb: 2 Suggestions
- :mag: 1 Nitpick

---

## Detailed Findings

### :warning: Address country code conversion lacks fallback handling
**File:** `crates/common/common_enums/src/transformers.rs`

The new `CountryAlpha2::to_numeric()` method provides ISO 3166-1 numeric codes via exhaustive match. However, if a new country variant is ever added to `CountryAlpha2` without a corresponding numeric code in this match arm, the code will fail to compile (which is good), but the error will surface late in the development cycle.

**Fix:** Consider deriving the numeric codes alongside the alpha-2/alpha-3 definitions using a proc-macro or maintaining a single source of truth to prevent divergence.

---

### :warning: PacoResponseWithRaw stores raw JSON without size limits
**File:** `transformers.rs`

The `PacoResponseWithRaw<T>` wrapper captures full decrypted JSON responses:
```rust
pub struct PacoResponseWithRaw<T> { 
    pub parsed: T, 
    pub raw: serde_json::Value 
}
```

While this preserves unmodeled fields, there's no size limit on what gets stored in `raw_connector_response`. A malicious or misbehaving connector returning multi-megabyte responses could cause memory pressure.

**Fix:** Consider adding a size check before storing the raw JSON, or truncating large responses before persistence.

---

### :warning: Airline data PII exposure via `.expose()`
**File:** `transformers.rs` (airline passenger serialization)

The PR mentions: "PII fields (`firstName`, `lastName`, `passportNumber`, `ticketNumber`) are `.expose()`'d once at the connector boundary."

However, these exposed values are then serialized into PACO's wire format. Verify that:
1. The PACO endpoint uses TLS/mTLS
2. Logs at the connector level don't capture these exposed values
3. The `raw_connector_response` field doesn't inadvertently retain these PII fields in plaintext storage

---

### :bulb: Consider instrumenting airline data fallback path
**File:** `transformers.rs`

The PR implements a fallback from typed `domain_data.airline_data` to raw `connector_feature_data.airlineData`. Consider adding a metric or log entry when the fallback path is exercised — this helps track adoption of the typed proto field vs legacy raw JSON.

**Fix:** Add a debug/info log when `connector_feature_data.airlineData` is used as fallback, noting that the typed proto field wasn't populated.

---

### :bulb: Phone number Secret wrapping consistency
**File:** Multiple files (field-probe, composite proto, tests)

The PR changes `phone_number` from plain `String` to `Secret<String>` / `SecretString`. While this improves security, verify that:
1. All downstream consumers handle the new type correctly
2. Serialization (especially for non-PACO connectors that might expect plain strings) remains compatible
3. The field-probe and integration test harnesses properly redact these secrets in logs

---

### :mag: Missing unit tests for CountryAlpha2::to_numeric
**File:** `crates/common/common_enums/src/transformers.rs`

The `to_numeric()` implementation is a large const match with 249 arms. While it's straightforward, there's no test verifying that specific countries return expected numeric codes (e.g., US = 840, GB = 826, PH = 608 for the PACO use case).

**Fix:** Add a simple unit test asserting correct numeric codes for a representative sample of countries, especially those commonly used with PACO (Philippines = 608, Singapore = 702, etc.).

---

## Overall Assessment

This PR delivers substantial improvements to the 2C2P PACO connector:

✅ **Address forwarding** correctly uses ISO 3166-1 numeric codes as required by PACO  
✅ **Typed airline data** provides clean domain modeling with backward-compatible fallback  
✅ **Raw response preservation** ensures future extensibility without losing connector data  
✅ **Proto contract changes** are additive and don't break wire compatibility  
⚠️ **Security considerations** around PII exposure and secret handling need verification

**Recommendation:** Address the PII exposure and Secret-type consistency concerns before merge. The other findings are improvements rather than blockers.

XyneSpaces:
## PR Review: 2C2P PACO Address, Typed Airline Data, Full Raw Response

**Classification:** `connector-flow-addition` · `proto-api-contract` · `core-flow-framework`  
**Risk:** Medium

### Summary
This PR enhances the 2C2P PACO connector with:
1. Billing/shipping address forwarding with ISO 3166-1 numeric country codes
2. Typed airline data serialization to PACO's wire format
3. Full raw response preservation via `PacoResponseWithRaw<T>` wrapper
4. Database schema change adding `domain_data` column to `payment_attempt`

### Key Observations

:bulb: **Suggestion (S2):** The new `CountryAlpha2::to_numeric()` helper in `transformers.rs` should include validation that the country code is valid before returning. Currently it appears to return an `Option<String>` - verify that unknown country codes are handled gracefully at the call sites.

:bulb: **Suggestion (S2):** The `domain_data` column addition in `payment_attempt` should include a check for backward compatibility with existing records. Consider whether a default value or migration to populate existing rows is needed.

:bulb: **Suggestion (S2):** The new airline data proto fields should be marked as optional to maintain backward compatibility with older clients that don't send this data.

### Companion Files Verification
- [x] Proto changes present in `payment.proto`
- [x] Generated types updated in `payments.rs`
- [x] Database migrations included (up/down SQL)
- [ ] Verify `data/field_probe/manifest.json` is updated if connector behavior changed

### Recommendation
**Approve with suggestions** - The changes are well-structured and include necessary migrations. Address the backward compatibility considerations mentioned above before merge.

---
*This is an automated review from XyneSpaces PR Review Bot*

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -272,6 +272,44 @@ pub struct PacoNotificationUrls {
...
+    pub ship_addr_line1: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ship_addr_line2: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ship_addr_line3: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ship_addr_post_code: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ship_addr_state: Option<String>,
+}
```
Aishwariyaa-Anand, line 307: these will be secret

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -390,6 +434,12 @@ pub struct TwocTwopPacoWalletAuthorizeRequest {
     #[serde(rename = "notificationURLs")]
     pub notification_urls: PacoNotificationUrls,
     pub device_details: PacoDeviceDetails,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub billing_address: Option<PacoBillingAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub shipping_address: Option<PacoShippingAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub airline_data: Option<serde_json::Value>,
```
Aishwariyaa-Anand: why can't we use PacoAirlineData here

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -404,25 +454,339 @@ pub enum TwocTwopPacoAuthorizeRequest {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    passport_number: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    nationality: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    ticket_number: Option<Secret<String>>,
+}
+
+impl From<&common_utils::types::Money> for PacoMoney {
+    fn from(m: &common_utils::types::Money) -> Self {
```
Shubhodip900: NIT:- for all conversion use better variable naming

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,343 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+    pub document_type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub title: Option<String>,
+    pub first_name: Secret<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub middle_name: Option<String>,
+    pub last_name: Secret<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub gender: Option<String>,
+    pub email: Secret<String>,
```
iemyashasvi: can we use Email util here

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,343 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+                passenger_sequence_no: passengers.first().map(|p| p.sequence_no),
+                ticket_no: airline.ticket_number.clone().map(Secret::new),
+                ticket_issue_date: airline.ticket_issue_date.clone(),
+                ticket_reservation_system_code: airline.booking_system_unique_id.clone(),
+                tax_amount,
+                ticket_fare,
+                agent_fee,
+            }]
+        } else {
+            Vec::new()
```
iemyashasvi, line 857: ?

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,343 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+        };
+
+        Ok(Self {
+            bypass_forward: None,
```
iemyashasvi: why this is mapped to `None`

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Done — changed `PacoPassenger.email` to `common_utils::pii::Email` and pass `customer.customer_email` through directly, dropping the `.expose().expose()` re-wrap into `Secret<String>`.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: This is the synthesized-ticket fallback. PACO scopes per-purchase amounts (ticketFare/taxAmount/agentFee) under `tickets[]`, but the proto has no ticket array yet — so we synthesize a single ticket from the top-level totals (see the doc comment on `PacoTicket`). When none of the ticket-scoped fields are present we emit an empty vec, which `skip_serializing_if = "Vec::is_empty"` omits from the wire. Added an inline comment to make that explicit.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Removed — `bypassForward` had no upstream source so it was always `None` (and thus never serialized). Dropped the field from `PacoAirlineData` entirely.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+    pub identification_no: Option<Secret<String>>,
+    /// PACO accepts free-form `documentType` (≤30 chars). We set "Passport"
+    /// when sourcing `identificationNo` from the proto's `passport_number`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub document_type: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub title: Option<String>,
+    pub first_name: Secret<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub middle_name: Option<String>,
```
iemyashasvi: make this also `Secret`

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+            city_code: location.city_code.clone(),
+            city_name: location.city_name.clone(),
+            country_code: numeric_cc,
+            country_name: location.country_name.clone(),
+            date_time: location.date_time.clone(),
+        }
+    }
+}
+
+fn paco_required_field(field_name: &'static str) -> error_stack::Report<errors::IntegrationError> {
```
iemyashasvi: rename this

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+            .and_then(|cust| cust.last_name.clone())
+            .ok_or_else(|| paco_required_field("airline_data.passengers[].customer.last_name"))?;
+        let email = customer
+            .and_then(|cust| cust.customer_email.clone())
+            .ok_or_else(|| {
+                paco_required_field("airline_data.passengers[].customer.customer_email")
+            })?;
+        let (identification_no, document_type) = match passenger.passport_number.clone() {
+            Some(passport) => (Some(passport), Some("Passport".to_string())),
+            None => (None, None),
```
iemyashasvi: why mapped to None

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+            booking_date_time: airline
+                .booking_date_time
+                .clone()
+                .ok_or_else(|| paco_required_field("airline_data.booking_date_time"))?,
+        };
+
+        let agency = if airline.agency_name.is_some()
+            || airline.agency_code.is_some()
+            || airline.agency_invoice_number.is_some()
+            || airline.agency_plan_name.is_some()
```
iemyashasvi: use .and then

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+        {
+            Some(PacoAgency {
+                name: airline.agency_name.clone(),
+                code: airline.agency_code.clone(),
+                invoice_no: airline.agency_invoice_number.clone(),
+                plan_name: airline.agency_plan_name.clone(),
+            })
+        } else {
```
iemyashasvi: +1

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,340 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+            .total_fee
+            .as_ref()
+            .map(PacoTransactionAmount::try_from)
+            .transpose()?;
+        let has_ticket_fields = airline.ticket_number.is_some()
+            || airline.ticket_issue_date.is_some()
+            || airline.booking_system_unique_id.is_some()
+            || ticket_fare.is_some()
+            || tax_amount.is_some()
+            || agent_fee.is_some();
```
iemyashasvi, line 844: +1

crates/types-traits/domain_types/src/connector_types.rs
```diff
@@ -4135,7 +4135,10 @@ pub struct AirlineData {
 #[derive(Debug, Clone, Default)]
 pub struct AirlineSegment {
     pub sequence_no: Option<u32>,
-    pub carrier_code: Option<String>,
+    /// Marketing carrier (ticket-issuing airline). For direct flights this
+    /// equals the operating carrier; for codeshare it may differ.
```
iemyashasvi: remove unneccary comments

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Done — `middle_name` is now `Option<Secret<String>>` (wrapped via `.map(Secret::new)` at construction).

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Renamed `paco_required_field` → `paco_missing_required_field` (it constructs the missing-required-field error).

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Reworked with `.map(|p| (p, "Passport")).unzip()`, so the None case is now implicit instead of a hand-written `(None, None)` arm.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Done — replaced the `if … {Some(..)} else {None}` with `(…is_some() || …).then(|| PacoAgency { … })`.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Covered by the agency change above (now uses `.then(||)`).

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: For the tickets block I kept `if/else`: switching it to `.then(..).unwrap_or_default()` (it returns a `Vec`, not `Option`) trips clippy `obfuscated_if_else`, which prefers `if/else` there. Added a comment on the empty branch instead.

crates/types-traits/domain_types/src/connector_types.rs
kanikac199: Removed the doc comments on `AirlineSegment.{marketing_carrier_code, flight_number, operating_carrier_code, operating_flight_number}`.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
kanikac199: Update: also caught that the helper was using the wrong error variant — it built `InvalidDataFormat` for what is actually a missing field. Switched it to `MissingRequiredField` (the idiomatic variant, identical shape) and renamed the helper to `missing_airline_field` to match.

iemyashasvi (dismissed):

iemyashasvi (dismissed):

AmitsinghTanwar007 (dismissed):

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,341 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+}
+
+impl From<&connector_types::AirlineLocation> for PacoAirlineLocation {
+    fn from(location: &connector_types::AirlineLocation) -> Self {
+        // Upstream sends ISO 3166-1 alpha-2; PACO expects numeric-3.
+        // Drop on parse failure rather than emit a malformed value.
+        let numeric_cc = location.country_code.as_deref().and_then(|c| {
+            c.parse::<CountryAlpha2>()
+                .ok()
+                .map(|a| CountryAlpha2::to_numeric(a).to_string())
```
JeevaRamu0104: **[S2] `to_numeric().to_string()` does not zero-pad to ISO 3166-1 numeric-3.**

`CountryAlpha2::to_numeric` returns a `u16`, and `u16::to_string()` emits no leading zeros — so any country whose numeric code is below 100 serialises incorrectly (e.g. Albania `008` → `"8"`, Austria `040` → `"40"`), which PACO is likely to reject or misread. The live UAT used the Philippines (`608`, already 3 digits), so this would not have surfaced.

The same pattern is used for the billing and shipping country codes at lines 878 and 892, so the fix should cover all three sites.

```suggestion
                .map(|a| format!("{:03}", CountryAlpha2::to_numeric(a)))
```

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,341 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
 #[serde(transparent)]
 pub struct TwocTwopPacoRefundResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiResponse>);
 
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PacoAirlineData {
```
JeevaRamu0104, line 510: **[S2] Restore the dropped serializer unit test before merge.**

The PR notes the earlier `airline_json_tests` was dropped during rebase and lists re-adding it as a follow-up. This `PacoAirlineData` wire shape has several brittle, connector-specific quirks that are not verifiable from the types alone and will regress silently:

- `operatingflightNo` (lowercase `f`, line 549)
- passenger `type` rename for the IATA PTC code
- ISO numeric-3 `countryCode` conversion (see separate comment re: zero-padding)
- the synthesized single `tickets[]` entry carrying the per-purchase amounts

There is live UAT evidence (HTTP 200 CHARGED), but a serializer test asserting the exact JSON for these four cases is the only guard against a silent wire-shape regression. Recommend adding it in this PR rather than as a follow-up.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -501,6 +505,341 @@ pub struct TwocTwopPacoVoidPcResponse(pub PacoResponseWithRaw<TwocTwopPacoNonUiR
...
+        Self::new(money.amount, money.currency)
+    }
+}
+
+impl From<&connector_types::AirlineLocation> for PacoAirlineLocation {
+    fn from(location: &connector_types::AirlineLocation) -> Self {
+        // Upstream sends ISO 3166-1 alpha-2; PACO expects numeric-3.
+        // Drop on parse failure rather than emit a malformed value.
+        let numeric_cc = location.country_code.as_deref().and_then(|c| {
+            c.parse::<CountryAlpha2>()
```
JeevaRamu0104: **[S3] Consider logging when an `airline_data` country code fails alpha-2 parsing instead of silently dropping it.**

`location.country_code` is converted alpha-2 → numeric-3 and dropped via `.ok()` on parse failure (lines 645-648). A malformed upstream alpha-2 therefore yields a segment with no `countryCode` and no signal as to why. A `tracing::warn!` (without the raw value, to avoid logging PII) would make this diagnosable in the field. Non-blocking.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
ryutsu77: sure.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
ryutsu77: sure, added those.

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
ryutsu77: sure log added.

JeevaRamu0104 (dismissed):

iemyashasvi (approved):

crates/integrations/connector-integration/src/connectors/twoc_twop_paco/transformers.rs
```diff
@@ -2241,3 +2617,153 @@ impl TryFrom<ResponseRouterData<TwocTwopPacoRefundResponse, Self>>
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use common_utils::{pii::Email, types::Money};
```
iemyashasvi: remove

iemyashasvi (changes requested):

iemyashasvi (approved):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #1580, opened 2026-06-19
fix(external-services): bound outbound connector API calls with a request timeout

gopikrishna000:
## What

The outbound connector HTTP client (`get_client_builder`) only configured
`pool_idle_timeout`, which evicts idle pooled connections — it does **not** bound
the duration of an in-flight request. As a result a single connector call had no
total deadline: a hung or very slow connector could hold the request open
indefinitely, and with it the upstream gRPC call that triggered it.

## Change

- Add a `connector_request_timeout` knob under `[proxy]` (`ProxyConfig`), applied
  as reqwest's `.timeout()` in `get_client_builder`.
- Defaults to `30s` (`DEFAULT_CONNECTOR_REQUEST_TIMEOUT_SECS`) when unset, so the
  bound applies even without explicit config; overridable per environment via TOML.
- Wired into `config/{development,sandbox,production}.toml`.

The field is optional and backward-compatible — existing configs without the key
fall back to the 30s default.

## Why 30s

Matches the conventional request timeout used by the caller for connector calls,
so behaviour is consistent whether a connector is reached directly or via this
service.

## Testing

- `cargo check -p domain_types -p external-services` passes.
- Optional/`#[serde(default)]` field; existing config-override tests unaffected.

### DIFF
--- a/config/development.toml
+++ b/config/development.toml
@@ -36,6 +36,7 @@ enabled = false
 
 [proxy]
 idle_pool_connection_timeout = 90
+connector_request_timeout = 30
 bypass_urls = []
 
 [test]
--- a/config/production.toml
+++ b/config/production.toml
@@ -19,6 +19,7 @@ port = 8080
 
 [proxy]
 idle_pool_connection_timeout = 90
+connector_request_timeout = 30
 bypass_urls = ["localhost", "local"]
 
 [connectors]
--- a/config/sandbox.toml
+++ b/config/sandbox.toml
@@ -19,6 +19,7 @@ port = 8080
 
 [proxy]
 idle_pool_connection_timeout = 90
+connector_request_timeout = 30
 bypass_urls = ["localhost", "local"]
 
 [connectors]
--- a/crates/common/external-services/src/service.rs
+++ b/crates/common/external-services/src/service.rs
@@ -1261,6 +1261,9 @@ pub fn create_client(
     }
 }
 
+/// Default total timeout (seconds) for a single connector API call.
+const DEFAULT_CONNECTOR_REQUEST_TIMEOUT_SECS: u64 = 30;
+
 static DEFAULT_CLIENT: OnceCell<Client> = OnceCell::new();
 static PROXY_CLIENT_CACHE: OnceCell<RwLock<HashMap<(Proxy, String), Client>>> = OnceCell::new();
 
@@ -1383,6 +1386,11 @@ fn get_client_builder(
             proxy_config
                 .idle_pool_connection_timeout
                 .unwrap_or_default(),
+        ))
+        .timeout(Duration::from_secs(
+            proxy_config
+                .connector_request_timeout
+                .unwrap_or(DEFAULT_CONNECTOR_REQUEST_TIMEOUT_SECS),
         ));
 
     // Disable automatic gzip decompression in test mode
--- a/crates/types-traits/domain_types/src/types.rs
+++ b/crates/types-traits/domain_types/src/types.rs
@@ -764,6 +764,8 @@ impl Proxy {
 #[derive(Debug, Serialize, Clone, PartialEq, Eq, Default, config_patch_derive::Patch)]
 pub struct ProxyConfig {
     pub idle_pool_connection_timeout: Option<u64>,
+    /// Total timeout (seconds) for a single connector API call.
+    pub connector_request_timeout: Option<u64>,
     pub bypass_urls: Vec<String>,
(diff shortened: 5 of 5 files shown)


XyneSpaces (commented):
## Code Review: Connector Request Timeout Bound

This PR adds a request timeout to the outbound connector HTTP client, preventing hung connections from blocking gRPC calls indefinitely.

### Overall Assessment

The implementation is clean and follows good practices:
- Default value ensures backward compatibility
- Clear separation between `idle_pool_connection_timeout` (connection pool management) and `connector_request_timeout` (request duration bound)
- Proper deserialization with `#[serde(default)]` for optional config

### Inline Comments

**crates/types-traits/domain_types/src/types.rs (around line 767)**:
- **Documentation clarity**: The doc comment says "Total timeout (seconds)". Consider specifying this is for the *entire* request-response cycle, not just connection establishment or time-to-first-byte. This helps distinguish it from `idle_pool_connection_timeout`.

**crates/common/external-services/src/service.rs (around line 1264)**:
- **Constant naming**: The constant `DEFAULT_CONNECTOR_REQUEST_TIMEOUT_SECS` clearly indicates units (SECS). Good practice — avoids the ambiguity of `TIMEOUT = 30` which could be ms/s/m.

**crates/common/external-services/src/service.rs (around line 1390)**:
- **Timeout ordering**: The `.timeout()` is chained after `.pool_idle_timeout()`. While this works, verify that `reqwest`'s behavior matches expectations — `.timeout()` bounds the entire request including DNS resolution, connection, and response body reading.

### Suggestions

1. **Monitoring**: Consider adding metrics for timeout occurrences to track if this limit is being hit in production
2. **Configuration guidance**: The 30s default aligns with caller expectations, but operators in high-latency regions might need guidance on tuning this

XyneSpaces:
## Code Review: Connector Request Timeout Bound

This PR correctly addresses a gap where outbound connector HTTP calls had no total deadline, potentially holding requests open indefinitely. The fix adds a `connector_request_timeout` knob with sensible defaults.

### 🟢 Positive Highlights

1. **Correct reqwest API usage**: Using `.timeout()` on the client builder sets a total deadline for each request, including connection establishment and response body reading.

2. **30s default aligns with conventions**: Matching the caller-side timeout expectations ensures consistent behavior.

3. **Comprehensive config coverage**: All three environment configs (development, sandbox, production) are updated.

4. **Proper backward compatibility**: The `Option<u64>` with default fallback ensures existing configs work unchanged.

### ⚠️ Suggestions

1. **Documentation** (`crates/types-traits/domain_types/src/types.rs` line ~767): Consider expanding the doc comment to contrast with `idle_pool_connection_timeout`:
```rust
/// Total timeout (seconds) for a single connector API call.
/// Unlike `idle_pool_connection_timeout` which controls pool connection 
/// eviction, this bounds the entire request duration from start to finish.
```

2. **Consider TCP connection timeouts**: The `idle_pool_connection_timeout` is for pool eviction, not active request limits. Consider if a separate connection timeout is needed, or if the 30s total covers both adequately.

3. **Error handling**: Ensure downstream code distinguishes timeout errors (for potential retry) from definitive connector errors.

4. **Consider per-connector overrides** (`crates/common/external-services/src/service.rs` line ~1264): Is 30s sufficient for all connector types? Some legacy connectors or file upload endpoints might need longer. Consider if this should be overridable per-connector.

Overall, a solid fix for a subtle but important reliability issue. 👍

XyneSpaces:
## Review: Connector Request Timeout

### [should-fix] Consider connection timeout vs request timeout distinction

The implementation uses `.timeout()` which sets a total request timeout:

```rust
.timeout(Duration::from_secs(
    proxy_config
        .connector_request_timeout
               .unwrap_or(DEFAULT_CONNECTOR_REQUEST_TIMEOUT_SECS),
))
```

This is good, but consider if you also need a separate connection timeout. The existing `idle_pool_connection_timeout` only controls idle connection pool timeout, not the initial connection establishment timeout. A slow-to-connect but fast-to-respond connector could still hang.

### [nit] Config placement consistency

The new field is added to `[proxy]` section alongside `idle_pool_connection_timeout`. This is logical, but verify this aligns with how other connector-specific timeouts are configured. If there are per-connector timeout overrides elsewhere, consider documenting the precedence hierarchy.

### [should-fix] Document the 30-second default

The 30-second default should be documented in configuration examples and potentially mentioned in the PR description for operators who may need to tune this based on their connector SLAs.

Shubhodip900 (approved):

AmitsinghTanwar007 (approved):

PR #1667, opened 2026-06-24
feat(observability): structured error attribution and execution_mode for UCS error alerting

gopikrishna000:
## What

Make UCS logs and audit events self-sufficient for error alerting and connector-vs-UCS attribution, without parsing error-message text.

**Logs:**
- Record `execution_mode` (`primary`/`shadow`) on the outermost request span, so it is inherited by every child span and appears on **every log line** of the request — not just the audit event. Reuses `ExecutionMode` via a new `as_str()` helper.
- Log `error_code` (and `http_status_code` for connector responses) as structured fields on the `IntegrationError` and `ConnectorError` gRPC-status mappers.
- Stop double-logging the error stack in the `ConnectorFlowError` wrapper; each variant's inner mapper already logs it.
- Fix the response Golden Log Line label (`incoming` -> `response`).

**Audit event:**
- On a failed gRPC response, decode the `Status` proto details and populate the event's `error` field with `error_code` / `http_status_code` / message — so event consumers (ClickHouse, event-line alerts) can separate connector-attributed errors (`CONNECTOR_ERROR_RESPONSE`) from UCS-internal faults. Matches the connector-call event, which already populates `error` on failure.

## Why

`error_code != "" AND error_code != "CONNECTOR_ERROR_RESPONSE"` is a drift-proof "any UCS-side error" signal (an exclusion, not an enumeration — unknown/future error types still match). `http_status_code` separates connector 4xx from 5xx. `execution_mode` lets alerts exclude shadow traffic.

## Notes

- Errors with no proto details (transport `ApiClientError`, webhook, direct `Status::internal`) carry no `error_code`; they are inherently UCS-side and identified by a non-OK `status_code`.
- Event `error` is additive; no change to the `Event` struct shape.

## Verification

- [x] `cargo check -p common_utils -p ucs_env -p grpc-server`
- [ ] Confirm the fields land as expected on a deployed environment.

### DIFF
--- a/crates/grpc-server/grpc-server/src/utils.rs
+++ b/crates/grpc-server/grpc-server/src/utils.rs
@@ -34,6 +34,7 @@ pub fn record_fields_from_header<B: hyper::body::Body>(request: &Request<B>) -> 
         version = ?request.version(),
         tenant_id = tracing::field::Empty,
         request_id = tracing::field::Empty,
+        execution_mode = tracing::field::Empty,
     );
     request
         .headers()
@@ -46,6 +47,17 @@ pub fn record_fields_from_header<B: hyper::body::Body>(request: &Request<B>) -> 
         .get(consts::X_REQUEST_ID)
         .and_then(|value| value.to_str().ok())
         .map(|request_id| span.record("request_id", request_id));
+
+    // On the request span so every log line of the request carries primary/shadow.
+    let shadow = request
+        .headers()
+        .get(consts::X_SHADOW_MODE)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| value.eq_ignore_ascii_case("true"));
+    span.record(
+        "execution_mode",
+        ExecutionMode::from_shadow_flag(shadow).as_str(),
+    );
 
     span
 }
@@ -288,7 +300,7 @@ where
             current_span.record("status_code", status.code().to_string());
         }
     }
-    tracing::info!("Golden Log Line (incoming)");
+    tracing::info!("Golden Log Line (response)");
 }
 
 /// Generic gRPC logging wrapper that accepts a custom parser function.
--- a/crates/grpc-server/grpc-server/src/utils.rs
+++ b/crates/grpc-server/grpc-server/src/utils.rs
@@ -260,7 +260,7 @@ where
     current_span.record("merchant_id", merchant_id);
     current_span.record("tenant_id", tenant_id);
     current_span.record("request_id", request_id);
-    tracing::info!("Golden Log Line (incoming)");
+    tracing::info!("Golden Log Line (incoming - request)");
     Ok(())
 }
 
@@ -301,7 +301,7 @@ where
             current_span.record("status_code", status.code().to_string());
         }
     }
-    tracing::info!("Golden Log Line (response)");
+    tracing::info!("Golden Log Line (incoming - response)");
 }
 
 /// Generic gRPC logging wrapper that accepts a custom parser function.
--- a/crates/grpc-server/grpc-server/src/utils.rs
+++ b/crates/grpc-server/grpc-server/src/utils.rs
@@ -18,6 +18,7 @@ use domain_types::{
 use error_stack::Report;
 use http::request::Request;
 use hyperswitch_masking;
+use prost::Message;
 use serde_json::Value;
 use std::{collections::HashMap, sync::Arc};
 use ucs_env::{configs, error::ResultExtGrpc};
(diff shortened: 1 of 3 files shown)


XyneSpaces:
[nit] The `add_execution_mode` function uses `from_masked_optional` but passes a non-optional `&str`. The function naming and usage is slightly confusing.

```suggestion
    /// Records whether the request ran in `primary` or `shadow` mode as a top-level event field,
    /// so audit-event consumers (alerts / ClickHouse) can filter shadow traffic without inspecting
    /// nested request headers.
    pub fn add_execution_mode(&mut self, execution_mode: &str) {
        MaskedSerdeValue::from_masked_value(&execution_mode.to_string(), "execution_mode").map(
            |masked| {
                self.additional_fields
                    .insert("execution_mode".to_string(), masked);
            },
        );
    }
```

XyneSpaces:
[should-fix] `decode_error_attribution` silently swallows protobuf decode errors. If neither `IntegrationError` nor `ConnectorError` decode successfully, there's no indication that something unexpected happened.

Consider logging a warning when decoding fails unexpectedly:

```suggestion
fn decode_error_attribution(status: &tonic::Status) -> (Option<String>, Option<u32>) {
    let details = status.details();
    if details.is_empty() {
        return (None, None);
    }
    if let Ok(err) = grpc_api_types::payments::IntegrationError::decode(details) {
        return (Some(err.error_code), None);
    }
    if let Ok(err) = grpc_api_types::payments::ConnectorError::decode(details) {
        return (Some(err.error_code), err.http_status_code);
    }
    router_env::logger::debug!("Could not decode error attribution from status details");
    (None, None)
}
```

XyneSpaces:
## PR Review: feat(observability): structured error attribution and execution_mode

### ✅ Approved

Excellent observability enhancement that improves error tracking and shadow-diff debugging.

**Changes Reviewed:**
- ✅ `ExecutionMode::as_str()` helper for consistent logging
- ✅ Structured error fields (`error_code`, `http_status_code`) in `IntoGrpcStatus`
- ✅ `execution_mode` recorded on request span
- ✅ Golden Log Line label fixed ("(incoming)" → "(response)")
- ✅ `decode_error_detail()` for audit event enrichment

**Highlights:**
1. **Audit Event Enrichment**: The `decode_error_detail()` extraction will significantly help with shadow-diff investigations
2. **Span Hierarchy**: Recording `execution_mode` on the parent span is architecturally sound - all child spans inherit it
3. **Label Fix**: The "(incoming)" → "(response)" correction makes log interpretation more intuitive

**Observability Impact:**
These changes will enable:
- Better error attribution in ClickHouse dashboards
- Execution mode-aware alerting
- Faster shadow-diff root cause analysis

Nice work on the structured logging improvements! 🎯

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -288,7 +301,7 @@ where
             current_span.record("status_code", status.code().to_string());
         }
     }
-    tracing::info!("Golden Log Line (incoming)");
+    tracing::info!("Golden Log Line (response)");
```
AmitsinghTanwar007: why this needs to be changed as we follow the pattern of incoming and outgoing?

gopikrishna000: it is not outgoing right

previously it was incoming because it was called on log_before_init

Now this log is called in log_after_init, which is after we call connector and receive response ., and we log the response informations

AmitsinghTanwar007: so tell me which Golden log line we have now after your change
there were two previously
1) Incoming for (req to ucs and res from ucs)
2) Outgoing for (req from ucs and res to ucs)

gopikrishna000: You're right — :304 isn't outgoing. Mapping to the two-hop model: Incoming = log_before_initialization (:263, req to UCS) + log_after_initialization (:304, res from UCS); Outgoing = the connector-call line in external-services (service.rs:531). So :304 is the response leg of the incoming hop, not a new category — my bad on (response).

The only reason I touched it: :263 and :304 emit the identical "(incoming)" string, so the request and response flushes are indistinguishable in Loki by message (only by which fields are set). Two ways to keep the incoming/outgoing convention and still tell them apart:
(a) revert to (incoming) and distinguish by content, or
(b) (incoming - request) / (incoming - response).

AmitsinghTanwar007 (approved):

Shubhodip900 (approved):

PR #1664, opened 2026-06-24
fix(redsys): emit failure status on pre_authenticate error envelope (cloud#17027)

10X-GRACE:
Closes #1720

PR #1669, opened 2026-06-24
fix(domain_types): report connector status on authenticate error response

### LINKED ISSUE #1694: [shadow-validation] redsys/authenticate, pre_authenticate: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **redsys / authenticate, pre_authenticate**.

`router.valueDiff:status`

Fixed by UCS PR #1669.

AmitsinghTanwar007:
## What

Fix the `authenticate` flow status that the router reconstructs from the UCS gRPC response.
When a connector returns an error on the Authenticate flow, `generate_payment_authenticate_response`
derived the proto `status` solely from `err.attempt_status`, defaulting to
`PaymentStatus::Unspecified` when that was `None` — discarding the connector-computed
`resource_common_data.status` (e.g. `Failure`). Hyperswitch maps `Unspecified` back to the
*previous* status (`AuthenticationPending`), so a failed authenticate surfaced as
`authentication_pending` via UCS, diverging from hyperswitch-native which reports `failure`.

This mirrors the fix already present in the `authorize`/`void`/`sync` response generators, which
use `err.get_attempt_status_for_grpc(err.status_code, resource_common_data.status)` to fall back
to the real attempt status for 2xx connector responses.

## Diff signature (shadow validation, prod / merchant `bu_es`)

```
x-connector=redsys  x-sub-flow=Authenticate
router.valueDiff: status   { hyperswitch: "failure", ucs: "authentication_pending" }
```

## Root cause

redsys returns a `RedsysErrorResponse` (HTTP 200, error in body) for the failed authenticate.
The connector transformer correctly sets `resource_common_data.status = AttemptStatus::Failure`,
but the Err arm of `generate_payment_authenticate_response` ignored it:

```rust
// before
let status = err.attempt_status
    .map(PaymentStatus::foreign_from)
    .unwrap_or_default();   // None -> Unspecified
```

`Unspecified` → router keeps `prev_status` (`AuthenticationPending`).

## Fix

`crates/types-traits/domain_types/src/types.rs` — `generate_payment_authenticate_response` Err arm:

```rust
let status = match err.get_attempt_status_for_grpc(err.status_code, status) {
    Some(attempt_status) => PaymentStatus::foreign_from(attempt_status),
    None => PaymentStatus::Unspecified,
};
```

## Before → after (status round-trip, source-traced)

| stage | before | after |
|---|---|---|
| UCS proto `status` | `Unspecified` | `Failure` |
| HS `AttemptStatus::foreign_try_from` | `prev_status` = `AuthenticationPending` | `AttemptStatus::Failure` |
| vs HS-native `failure` | **diff** | **match (no_diff)** |

`AttemptStatus::Failure → PaymentStatus::Failure` (domain_types/src/types.rs:6479);
`PaymentStatus::Failure → AttemptStatus::Failure` and `Unspecified → prev_status`
(hyperswitch `hyperswitch_interfaces/.../unified_connector_service/transformers.rs:405,412`).

## Verification

redsys 3DS authenticate is not locally reproducible (real ACS + redsys sandbox + prod merchant
creds), so verified by source-tracing the full proto round-trip above (matches the in-tree
`authorize` generator pattern). Local checks on a worktree off `main`:
`cargo build` (incl. `--bin grpc-server`), `cargo fmt --check`, `cargo clippy -p domain_types
--all-targets -D warnings`, `cargo test -p domain_types`, `cargo test -p connector-integration redsys`
— all green.

Layer: **L2 (UCS domain mapping)**. No proto change, no hyperswitch change required
(HS reconstruction already maps proto status correctly).

Closes #1694


XyneSpaces:
## Code Review

The fix correctly applies the `get_attempt_status_for_grpc()` pattern to `generate_payment_authenticate_response`, aligning it with the existing handlers (authorize, void, sync).

### Verification Points
✅ **Pattern Consistency**: Confirmed this matches the approach used in other response generators
✅ **Correct Fallback**: Only falls back to `Unspecified` when `get_attempt_status_for_grpc` returns `None`
✅ **Behavior Fix**: Failed authenticate flows will now properly propagate `Failure` status instead of preserving `AuthenticationPending`

### Observation
Consider if other response generators in the same file should be audited for similar inconsistencies. The PR description indicates this was specifically for RedSys 3DS authentication errors - ensure test coverage includes this scenario.

✅ **Ready for merge** - targeted fix with clear rationale.

XyneSpaces:
## PR Review Summary

**[should-fix]** The status mapping fix looks correct, but there's a potential edge case worth addressing:

The `get_attempt_status_for_grpc` method now takes `err.status_code` and `status` as parameters, but the logic inside this method (not shown in the diff) may not correctly distinguish between:
1. A connector error with HTTP 200 + error body (should map to Failure)
2. A genuine connector network/timeout error (should also map appropriately)

Consider adding a test case that verifies the status mapping behavior for different error scenarios, particularly:
- When `attempt_status` is `None` but `status` is `Failure`
- When both are present but differ (which takes precedence?)

**Location**: `crates/types-traits/domain_types/src/types.rs` around line 15238

The comment explaining the fix is excellent and clearly documents the divergence from previous behavior.

XyneSpaces:
## Code Review

**PR:** feat(incremental_authorization): psp-reference mapping for incremental auths

### Summary
This PR improves error status handling in the 3DS authentication flow by properly falling back to connector-computed attempt status instead of defaulting to `Unspecified`.

### Findings

1. **Correct Fix**: The change replaces the `unwrap_or_default()` pattern (which defaulted to `Unspecified`) with proper attempt status resolution using `err.get_attempt_status_for_grpc()`. This ensures that actual failure statuses are propagated rather than masking them as `Unspecified`.

2. **Context Preservation**: The detailed comment explains the issue well - when `Unspecified` was returned, the router would keep the previous `AuthenticationPending` status, causing a divergence from native hyperswitch behavior.

3. **Better Error Flow**: By computing the status from the error's status code and current status, the fix ensures the response accurately reflects the connector's reported state.

### Suggestion

Consider extracting this status computation into a helper method on the error type itself to ensure consistency:

```rust
impl ConnectorError {
    fn to_grpc_payment_status(&self, current_status: PaymentStatus) -> PaymentStatus {
        match self.get_attempt_status_for_grpc(self.status_code, current_status) {
            Some(attempt_status) => PaymentStatus::foreign_from(attempt_status),
            None => PaymentStatus::Unspecified,
        }
    }
}
```

Then the call site becomes cleaner:
```rust
let status = err.to_grpc_payment_status(status);
```

The fix correctly addresses the status propagation issue for 3DS authentication errors.

XyneSpaces:
[should-fix] The change properly uses `get_attempt_status_for_grpc()` for fallback instead of defaulting to `Unspecified`.

One concern: the comment mentions this prevents the router from keeping `AuthenticationPending`, but the code still falls back to `Unspecified` in the `None` case. Should this also have a more appropriate fallback, or is `Unspecified` intentionally the final fallback?

Also, this PR uses `Some("session_created".to_string())` for `connector_transaction_id`, but PR #1674 (which appears to be a related fix) changes this to use `err.connector_transaction_id.clone()`. These PRs may conflict - please ensure they're coordinated.

Otherwise, this is a good fix for the status propagation issue.

XyneSpaces:
## PR Review: fix(domain_types): report connector status on authenticate error response

### ✅ Approved with Related Note

Good fix addressing the status propagation issue in authenticate error responses.

**Changes Reviewed:**
- ✅ Uses `get_attempt_status_for_grpc()` for proper status derivation
- ✅ Falls back to `Unspecified` only when no status can be computed
- ✅ Inline comment clearly explains the divergence being fixed

**Related:**
Note: This PR modifies the same area as #1674 (which removes the hardcoded `"session_created"` in the adjacent line). These PRs touch the same `connector_transaction_id` assignment - coordinate merges to avoid conflicts.

**Logic Validation:**
The `get_attempt_status_for_grpc()` method appears to derive status from the error's status code, which aligns with native hyperswitch behavior. The pattern:
```rust
match err.get_attempt_status_for_grpc(...) {
    Some(status) => /* use derived status */,
    None => /* fall back to Unspecified */
}
```

LGTM! 👍

XyneSpaces:
test

XyneSpaces:
DELETE_ME_TEST

gopikrishna000 (dismissed):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

Shubhodip900 (approved):

AkshayaFoiger (approved):

Aishwariyaa-Anand (approved):

gopikrishna000:
duplcioate of https://github.com/juspay/hyperswitch-prism/pull/1664#pullrequestreview-4786223856

iemyashasvi:
not needed closed by #1664

PR #1671, opened 2026-06-24
fix(redsys): populate Err.reason in (pre)authenticate to match hyperswitch (#17018)

AmitsinghTanwar007:
## What

Fix a HS↔UCS shadow-validation `router.typeDiff:response.Err.reason` on **redsys** for
the `authenticate` / `pre_authenticate` flows: the UCS gRPC error response carried a
**null** `reason` while hyperswitch-native redsys carried the **error code string**.

Diff signature: `router.typeDiff:response.Err.reason`
(merchant `bu_es`, org `zurich_insurance`, connector `redsys`, flows `authenticate, pre_authenticate`).

## Root cause (L3 — connector transformer only)

hyperswitch-native redsys builds its error response with
`reason: Some(error_response.error_code.clone())` — **always populated** — because its
`RedsysErrorResponse` only carries `error_code`.

UCS redsys added an optional `error_code_description` to `RedsysErrorResponse` (Redsys
sometimes returns only `errorCode` with no description, needed for untagged-enum
deserialization). The `PreAuthenticate` and `Authenticate` response transformers then
sourced `reason` directly from that optional field:

```rust
reason: err.error_code_description.clone(),
```

When Redsys returns only `errorCode` (the common case — exactly what the HS struct assumes),
`error_code_description` is `None`, so UCS emitted `reason: null` while HS emitted the error
code string → **`response.Err.reason` typeDiff** (HS `string` vs UCS `null`). `message` did
not diff because UCS already falls back to `error_code` for it
(`error_code_description.unwrap_or_else(|| error_code)`), and the triggering responses have no
description, so both sides' `message` equal `error_code`.

## Fix

Mirror hyperswitch-native: fall back to the error code so `reason` is always `Some(..)`,
in both the `PreAuthenticate` and `Authenticate` redsys response transformers:

```rust
reason: Some(
    err.error_code_description
        .clone()
        .unwrap_or_else(|| err.error_code.clone()),
),
```

`reason` already flows end-to-end (it is a standard `ErrorResponse` field present in the proto
and read back by HS) — **no proto / domain / HS-mapping change needed, no HS PR.** Scope is
limited to the two flows named in the issue; the sibling redsys flows (Authorize/Capture/Void/
Refund) are out of scope of this diff and untouched.

## Verification

- **Source-parity (live redsys 3DS shadow is not locally reproducible** — real ACS + redsys
  sandbox + prod merchant creds, same as siblings #16980/#16981/#16982):
[...]

PR #1626, opened 2026-06-24
fix(redsys): use webhook_url as 3DS method notification URL in pre_authenticate

AmitsinghTanwar007:
### What

Redsys **pre_authenticate** (3DS method invoke): use the merchant **`webhook_url`** as the
`threeDSMethodNotificationURL` embedded in the Base64 `threeDSMethodData`, matching Hyperswitch's
native connector behaviour.

### Why — shadow-validation router diff (`bu_es` / redsys / pre_authenticate)

Prod shadow-validation reported:

```
router.valueDiff:response.Ok.TransactionResponse.connector_metadata.three_ds_method_data
```

`three_ds_method_data` is `base64(JSON{ threeDSMethodNotificationURL, threeDSServerTransID })`.

- **Hyperswitch (ground truth):** builds it from `request.get_webhook_url()` (the merchant webhook URL).
- **UCS (before):** built it from `continue_redirection_url`, which the HS→UCS mapping populates from
  `complete_authorize_url` (`router/.../unified_connector_service/transformers.rs`:
  `continue_redirection_url: router_data.request.complete_authorize_url`).

`webhook_url` ≠ `complete_authorize_url`, so the embedded URL differed → the Base64 differed → `valueDiff`.

UCS's `PaymentMethodAuthenticationServicePreAuthenticateRequest` carried no `webhook_url`, so this is a
proto-contract gap.

### Change (proto-category, UCS side)

- **proto** `payment.proto`: add `optional string webhook_url = 15;` to
  `PaymentMethodAuthenticationServicePreAuthenticateRequest`.
- **domain** `PaymentsPreAuthenticateData`: add `webhook_url: Option<Url>`; map it from the proto in
  `domain_types/src/types.rs`.
- **connector** redsys `build_threeds_invoke_response`: use `webhook_url` for the notification URL
  (falling back to `continue_redirection_url` only when absent), so the encoded `three_ds_method_data`
  matches Hyperswitch byte-for-byte.
- **composite-service**: thread `webhook_url` from the authorize request into the PreAuthenticate sub-request.

The paired HS PR populates the new proto field from `router_data.request.webhook_url`.

### Verification

Live redsys 3DS shadow isn't reproducible locally (real ACS + redsys sandbox + the prod merchant config).
Verified by source-parity: both sides serialize the identical struct
(`#[serde(rename_all="camelCase")]`, fields `threeDSMethodNotificationURL`, `threeDSServerTransID`),
so with the same URL the Base64 is byte-identical.

```
[...]

PR #1674, opened 2026-06-24
fix(redsys): drop hardcoded connector_transaction_id on authenticate Err response (#17017)

### LINKED ISSUE #1705: [shadow-validation] redsys/authenticate, pre_authenticate: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **redsys / authenticate, pre_authenticate**.

`router.typeDiff:response.Err.connector_transaction_id`

Fixed by UCS PR #1674.

AmitsinghTanwar007:
## What

Fix a HS↔UCS shadow-validation `router.typeDiff:response.Err.connector_transaction_id` on
**redsys** for the `authenticate` flow: when the connector returns an error on Authenticate,
the UCS gRPC error response carried a **hardcoded `"session_created"` string** as the
top-level `connector_transaction_id`, while hyperswitch-native redsys emits **null**.

Diff signature (shadow validation, prod / merchant `bu_es`, org `zurich_insurance`):

```
x-connector=redsys  x-sub-flow=Authenticate
valueDiff: status                                  { hyperswitch: "failure", ucs: "authentication_pending" }
typeDiff:  response.Err.reason                      { hyperswitch: "string", ucs: "null" }
typeDiff:  response.Err.connector_transaction_id    { hyperswitch: "null",   ucs: "string" }
```

This PR addresses the **`response.Err.connector_transaction_id`** field. (The sibling
`status` and `reason` diffs of the same failing response are handled by #1669 and #1671
respectively.)

## Root cause (L2 — UCS domain mapping only)

redsys returns a `RedsysErrorResponse` (HTTP 200, error in body) for the failed authenticate.
The connector transformer correctly builds `response = Err(ErrorResponse { connector_transaction_id: None, .. })`
(redsys error responses carry no transaction id), exactly mirroring hyperswitch-native
(`hyperswitch_connectors/.../redsys/transformers.rs`, Authenticate `RedsysErrorResponse` arm
→ `connector_transaction_id: None`).

But the **Err arm** of `generate_payment_authenticate_response`
(`crates/types-traits/domain_types/src/types.rs`) hardcoded the proto **top-level**
`connector_transaction_id`:

```rust
// before
PaymentMethodAuthenticationServiceAuthenticateResponse {
    connector_transaction_id: Some("session_created".to_string()),   // placeholder/copy-paste bug
    ...
    error: Some(ErrorInfo { connector_details: Some(ConnectorErrorDetails {
        connector_transaction_id: err.connector_transaction_id.clone(),   // <- already correct
        ...
```

HS reconstructs the Err response from the proto **top-level** field
(`router/src/core/unified_connector_service/transformers.rs`,
`PaymentMethodAuthenticationServiceAuthenticateResponse` impl):

```rust
let connector_transaction_id = response
    .connector_transaction_id
    .as_ref()
    .map(|id| ResponseId::ConnectorTransactionId(id.clone()))
    .unwrap_or(ResponseId::NoResponseId);
...
connector_transaction_id: connector_transaction_id.get_optional_response_id(),
```

So `Some("session_created")` → `Err.connector_transaction_id = "session_created"` (string),
diverging from hyperswitch-native `null`.

`"session_created"` is a placeholder unrelated to an error response — the only such hardcode
in the file, and inconsistent with the sibling `generate_payment_pre_authenticate_response`
and `generate_payment_post_authenticate_response` Err arms, which use the actual error's
transaction id.

## Fix

Mirror hyperswitch-native (and the sibling generators) by sourcing the top-level field from
the actual error, which is also what `connector_details` already carries:

```rust
// after
connector_transaction_id: err.connector_transaction_id.clone(),
```

This is faithful in **both** Authenticate error paths:
- `RedsysErrorResponse` path (the triggering case): `err.connector_transaction_id = None`
  → proto `None` → HS `NoResponseId` → `Err.connector_transaction_id = null` = HS-native.
- payment-failure path (`get_payments_response` Err branch): `err.connector_transaction_id =
  Some(ds_order)` → HS `Some(ds_order)`, exactly matching hyperswitch-native's
  `connector_transaction_id: Some(redsys_payments_response.ds_order)` for that path
  (previously this path was *also* wrong — it emitted `"session_created"`).

`connector_transaction_id` already flows end-to-end (standard proto field, read back by HS) —
**no proto / domain struct / hyperswitch-mapping change needed, no HS PR.**

## Before → after (source-traced round-trip; triggering `RedsysErrorResponse` case)

| stage | before | after |
|---|---|---|
| UCS transformer `ErrorResponse.connector_transaction_id` | `None` | `None` |
| UCS proto top-level `connector_transaction_id` | `Some("session_created")` | `None` |
| HS reconstruction `Err.connector_transaction_id` | `"session_created"` (string) | `null` |
| vs HS-native `null` | **diff** | **match (no_diff)** |

## Verification

- **Source-parity** (live redsys 3DS authenticate is not locally reproducible — real ACS +
  redsys sandbox + prod merchant creds, same as siblings #1626/#1627/#1628/#1669/#1671): the
  full proto round-trip above traces `"session_created"` → `null`, matching hyperswitch-native
  and the in-tree pre/post-authenticate generators.
- **Local checks** (worktree off `origin/main` `d0e1f4f37`, `RUST_MIN_STACK=64M`):
  - `cargo build -p domain_types` ✅ and `cargo build --bin grpc-server` ✅
  - `cargo clippy -p domain_types --all-targets -- -D warnings` ✅
  - `cargo fmt --all -- --check` ✅
  - `cargo test -p domain_types` ✅ · `cargo test -p connector-integration redsys` ✅

## Layer

**L2 — UCS domain mapping only**
(`crates/types-traits/domain_types/src/types.rs`, `generate_payment_authenticate_response`).

Closes #1705


XyneSpaces:
[blocking] The fix correctly uses `err.connector_transaction_id.clone()` instead of the hardcoded `"session_created"` placeholder.

However, there's a potential issue: the `err` structure may not always have a `connector_transaction_id` populated. Consider handling the `None` case explicitly or adding a fallback comment explaining when this can occur.

```rust
// Suggested addition for clarity:
// If connector_transaction_id is None, the error response will propagate None
// which is the correct behavior (vs a misleading placeholder)
```

This change aligns with the Direct gateway behavior - good catch.

XyneSpaces (commented):
Reviewing the connector_transaction_id change.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -15238,7 +15238,7 @@ pub fn generate_payment_authenticate_response<T: PaymentMethodDataTypes>(
                 .map(grpc_api_types::payments::PaymentStatus::foreign_from)
                 .unwrap_or_default();
             PaymentMethodAuthenticationServiceAuthenticateResponse {
-                connector_transaction_id: Some("session_created".to_string()),
+                connector_transaction_id: err.connector_transaction_id.clone(),
```
XyneSpaces, line 16753: [blocking] The change from hardcoded `session_created` to `err.connector_transaction_id.clone()` is correct for parity, but verify that downstream consumers don't depend on the `Some("session_created")` value for authentication flow detection. Check if any dashboards, alerts, or business logic use this string literal as a sentinel value.

XyneSpaces:
## PR Review: fix(redsys): drop hardcoded connector_transaction_id on authenticate Err response

### ✅ Approved

Clean fix removing a hardcoded string that was causing shadow-validation diffs.

**Change Reviewed:**
```rust
- connector_transaction_id: Some("session_created".to_string()),
+ connector_transaction_id: err.connector_transaction_id.clone(),
```

**Impact:**
- UCS now returns `null` instead of `"session_created"` for error cases
- Aligns with hyperswitch-native behavior
- Eliminates `valueDiff` in shadow validation

**Observation:**
This appears to be technical debt cleanup - the hardcoded string was likely a placeholder. Good catch!

LGTM! 👍

Shubhodip900:
closing this PR as same change is merged via separate PR

PR #1676, opened 2026-06-25
feat(stripe): carry split_payments on void to reconstruct charges in response

AmitsinghTanwar007:
### Diff
Shadow-validation issue **juspay/hyperswitch-cloud#16999** — connector `stripe`, flow `void`, merchant `yucca`:

`router.typeDiff:response.Ok.TransactionResponse.charges` — HS void response carries `charges` (object), UCS void response carries `null`.

### Root cause
The stripe void response builder derives the response `splits` (→ HS `charges`) from `request.split_payments` via `construct_charge_response`. But `PaymentServiceVoidRequest` had **no `split_payments` field**, so HS could never forward the payment's `split_payments` on the void flow → `PaymentVoidData.split_payments` was hardcoded `None` → `splits` always `null`. The void *response* already had a `splits` slot (field 12) and the response builder already mapped it; only the *request* side was missing. Same shape as the capture fix (#16992, request `split_payments` gap).

### Fix (UCS, proto-category)
- `payment.proto`: add `split_payments = 13` to `PaymentServiceVoidRequest`.
- `composite_payment.proto`: add `split_payments = 15` to `CompositeVoidRequest` (composite parity).
- `domain_types/types.rs`: map `PaymentServiceVoidRequest.split_payments` → `PaymentVoidData.split_payments` (was `None`).
- `composite-service/transformers.rs`: forward `split_payments` through `CompositeVoidRequest`.

The connector transformer (`construct_charge_response`) and the void response builder (`generate_payment_void_response`, `splits` field) already consume/emit it — no change needed there.

Pairs with the HS PR that populates the new field from `PaymentsCancelData.split_payments`.

### Verify (direct gRPC `PaymentService.Void` A/B — stripe primary TLS-blocked locally, void is one-shot)
Each case uses its own manual-capture destination-split Stripe PaymentIntent in `requires_capture`:

- **A — void WITH `splitPayments`** (`STRIPE_DESTINATION`, fees 100, `acct_…`):
  `status=VOIDED`, `splits = { stripeSplitResponse: { chargeId: ch_…, chargeType: STRIPE_DESTINATION, applicationFees: "100", transferAccountId: acct_… } }` — byte-shape == HS native `charges`.
- **B — void WITHOUT split**: `status=VOIDED`, `splits = null`.

Before this change both A and B returned `splits = null`. **diff → no_diff.** Repro script: `prod-fixes/verify_16999.py`.

Layer: **proto** (UCS proto + domain + composite).


XyneSpaces:
## PR Review: Split Payments Support for Void Flow

### [blocking] Proto field number collision risk

Added field `split_payments` uses field number `15` in `composite_payment.proto` and `13` in `payment.proto`. Please verify these numbers don't conflict with:
1. Other open PRs modifying the same protos
2. Internal field numbers used in downstream forks

Proto field numbers cannot be changed once deployed without breaking backward compatibility.

---

### [should-fix] Missing validation for split_payments in void context

The void flow receives `split_payments` but there's no validation that the splits are consistent with the original authorization. Consider:

1. **Amount mismatch**: If split amounts in void request don't match authorized amounts
2. **Destination account changes**: If `transfer_destination` differs from original

Should the void reconstruct splits from stored data rather than accepting them from the request?

---

### [nit] foreign_try_from error context

The `foreign_try_from` conversion for `SplitPaymentsDetails` may fail, but the error context doesn't indicate which field caused the issue:

```rust
.split_payments
.map(connector_types::SplitPaymentsDetails::foreign_try_from)
.transpose()?,
```

Consider wrapping with context if the error type supports it:
```rust
.transpose()
.map_err(|e| e.add_context("split_payments conversion failed"))?
```

---

### Note on Duplicate PRs
This PR (#1676) appears to have identical changes to #1677. If these are duplicates, consider closing one to avoid merge conflicts.

PR #1678, opened 2026-06-25
fix(authorizedotnet/psync): keep prior data on E00053/E00104 instead of erroring

### LINKED ISSUE #1689: [shadow-validation] authorizedotnet/psync: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **authorizedotnet / psync**.

`router.keyDiff:response.Ok`

Fixed by UCS PR #1678.

AmitsinghTanwar007:
## What

Fix an HS↔UCS shadow-validation diff on `authorizedotnet` / `psync` (merchant `innago`).

Diff signature: `router.keyDiff:response.Ok` (and the complementary `router.keyDiff:response.Err`).

## Root cause

On a PSync poll, Authorize.Net can reply with **no `transaction` object** and a message
code of `E00053` ("server too busy") or `E00104` ("Server in maintenance"). These are
**transient** server conditions, not a payment outcome.

Hyperswitch (the ground truth) treats them as non-fatal and **returns the existing router
data unchanged** for these two codes:

```rust
// hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs
None => match item.response.messages.message.iter()
        .find(|msg| msg.code == "E00053" || msg.code == "E00104")
{
    Some(_) => Ok(item.data),                                  // keep prior data
    None    => Ok(Self { response: Err(get_err_response(..)?), ..item.data }),
},
```

UCS, however, mapped **every** missing-`transaction` PSync response to an `ErrorResponse`,
with no E00053/E00104 special-case. So for the same poll HS kept `response.Ok` (the prior
successful state) while UCS produced `response.Err` — surfacing as the
`router.keyDiff:response.Ok` / `router.keyDiff:response.Err` shadow diffs.

## Fix (L3 — UCS connector transformer only)

In the PSync `None` branch, short-circuit `E00053`/`E00104` to `Ok(router_data)` (the
existing router data, unchanged), mirroring HS exactly. All other missing-`transaction`
responses still map to the same `ErrorResponse` as before.

```rust
None => {
    if response.messages.message.iter()
        .any(|m| m.code == "E00053" || m.code == "E00104")
    {
        return Ok(router_data);
    }
    // ... unchanged error-response path ...
}
```

No proto/domain/HS-mapping change is needed — this is purely the UCS connector response
transformer aligning to HS behavior.

## Verification

The trigger is a transient Authorize.Net **server-too-busy / maintenance** response, which
cannot be induced on demand against the live sandbox (and authnet live shadow is TLS-blocked
in the local stack), so this is verified by **source parity** plus local checks:

- Before → after for `transaction = None` + an `E00053`/`E00104` message:
  - **Before:** UCS `Err(ErrorResponse{ code:"E00053"/"E00104", attempt_status: Failure/Pending })` vs HS `Ok(prior data)` ⇒ `keyDiff:response.Ok` + `keyDiff:response.Err`.
  - **After:** UCS `Ok(router_data)` (prior data, unchanged) ⇒ byte-equivalent to HS ⇒ diff gone.
- All other PSync paths (`transaction = Some`, and `transaction = None` with any other code) are unchanged.
- Local checks green: `cargo build --bin grpc-server`, `cargo clippy -p connector-integration --all-targets -- -D warnings`, `cargo fmt --check`.

## Layer

L3 — connector transformer (UCS only). Ground truth (HS connector transformer) is unchanged.

Closes #1689


crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -2327,6 +2327,20 @@ impl<F> TryFrom<ResponseRouterData<AuthorizedotnetPSyncResponse, Self>>
                 Ok(new_router_data)
```
XyneSpaces, line 2360: [nit] Consider extracting error codes to constants

The error codes "E00053" and "E00104" are hardcoded inline. Consider defining them as constants:

```rust
const TRANSIENT_ERROR_SERVER_BUSY: &str = "E00053";
const TRANSIENT_ERROR_MAINTENANCE: &str = "E00104";
```

This makes the code more maintainable and allows reuse in tests. Also consider whether these codes should be part of a shared constant module if other connectors handle similar transient errors.

XyneSpaces (commented):
[nit] Consider extracting error codes to constants

The error codes "E00053" and "E00104" are hardcoded inline. Consider defining them as constants:

```rust
const TRANSIENT_ERROR_SERVER_BUSY: &str = "E00053";
const TRANSIENT_ERROR_MAINTENANCE: &str = "E00104";
```

This makes the code more maintainable and allows reuse in tests. Also consider whether these codes should be part of a shared constant module if other connectors handle similar transient errors.

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
pratikjuspay: Done in 18f895fd8 — extracted both codes to module constants `TRANSIENT_ERROR_SERVER_BUSY` (`E00053`) and `TRANSIENT_ERROR_MAINTENANCE` (`E00104`) and used them in the psync check.

On the shared-module point: these are Authorize.Net-specific message codes (Authorize.Net's `E00xxx` namespace), so I kept them local to this connector's transformer rather than a cross-connector module. If we later standardise transient-error handling across connectors we can promote them then. `cargo check -p connector-integration` passes; no behavioural change.

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -2326,6 +2332,16 @@ impl<F> TryFrom<ResponseRouterData<AuthorizedotnetPSyncResponse, Self>>
...
             }
             None => {
+                // Transient server-side conditions (server busy / maintenance) on a
+                // psync poll, not a payment failure, so (mirroring hyperswitch) we keep
+                // the existing router data unchanged instead of flipping the attempt to
+                // an error/failure.
+                if response.messages.message.iter().any(|m| {
+                    m.code == TRANSIENT_ERROR_SERVER_BUSY || m.code == TRANSIENT_ERROR_MAINTENANCE
+                }) {
+                    return Ok(router_data);
```
pixincreate: this can be done something like this since they're not optional fields:

```rs
Some(_) => Ok(item.data),
None => Ok(Self {
	response: Err(get_err_response(item.http_code, item.response.messages)?),
    ..item.data
}),
```

shuklatushar226 (dismissed):

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
pratikjuspay: Done in b6cbaef2b.

Restructured the no-transaction branch exactly to HS's shape:
```rust
match response.messages.message.iter().find(|m| m.code == TRANSIENT_ERROR_SERVER_BUSY || m.code == TRANSIENT_ERROR_MAINTENANCE) {
    Some(_) => Ok(router_data),
    None => Ok(Self { response: Err(get_err_response(http_code, response.messages)), ..router_data }),
}
```
and added a `get_err_response(status_code, messages)` helper mirroring HS (first message's code/text, `attempt_status: None`). The `..router_data` struct-update preserves the prior attempt status instead of force-setting Failure/Pending, and drops the manual `ErrorResponse` build + `resource_common_data` mutation.

One deliberate deviation from HS: HS's `get_err_response` is fallible (`ok_or(ResponseDeserializationFailed)?` on an empty message list); I kept it infallible with the existing `NO_ERROR_CODE`/`NO_ERROR_MESSAGE` fallbacks to preserve the prior robustness. Happy to switch to the fallible form if you'd prefer exact parity. `cargo clippy -p connector-integration` + fmt clean.

AkshayaFoiger (approved):

shuklatushar226 (approved):

pixincreate (approved):

PR #1687, opened 2026-06-25
feat(observability): UCS-internal processing latency metric (total − connector RTT)

gopikrishna000:
## What

Add an OTel histogram for **UCS-internal processing latency** — the time UCS spends on a request *excluding* the connector round-trip:

```
ucs_internal_processing_duration_seconds = total_request_time − Σ(connector_rtt)
```

Labels: `method` (flow), `service`, `connector`, `mode` (`primary`/`shadow`).

## Why

We already measure **total** request latency and **connector RTT** as separate metrics, but not the difference. Percentiles aren't additive (`p99(total) − p99(connector) ≠ p99(internal)`), so the internal time must be computed per request and observed into its own histogram. This isolates a regression in UCS's own overhead from connector slowness, and the `mode` label keeps the shadow path (mitm proxy) from skewing primary numbers.

## How

- A per-request `task_local!` accumulator (`Arc<AtomicU64>`, nanos). Each connector call in `external-services` adds its measured elapsed via `add_connector_time` — a request can make multiple connector calls (retries / multi-step), so they sum.
- The gRPC wrapper scopes the handler future with the accumulator; at the audit-event emission point it reads the sum and records `internal = total.saturating_sub(connector_rtt)`.
- OTel only (the `ucs_*` set already carries `mode` and feeds the dashboards via OTLP); gated behind the `otel` feature like its neighbours.

## Edge cases

- No connector call (e.g. validation fails before egress) → accumulator stays 0 → `internal = total`.
- Negative result (measurement skew) → clamped to 0 via `saturating_sub`.
- Accumulator is a no-op outside a request scope, so it is always safe to call.

## Notes / follow-up

- Self-contained: `mode` is read from `metadata_payload.shadow_mode`, which already exists on `main`.
- Instruments the two connector-egress points in `execute_connector_processing_step`. Worth a follow-up audit that all outbound call sites flow through these (else those requests' `internal` would be over-counted).

## Verification

- [x] `cargo check -p grpc-server -p external-services` (default)
- [x] `cargo check -p grpc-server --features otel`
- [ ] Confirm `ucs_internal_processing_duration_seconds` appears with `mode` split on a deployed environment.


XyneSpaces:
## PR Review: UCS Internal Processing Latency Metric

**[blocking]** `saturating_sub` may hide negative latency bugs

In `grpc-server/src/utils.rs`:
```rust
let internal_latency = total_elapsed.saturating_sub(connector_time);
```

If `connector_time > total_elapsed`, this silently clamps to 0 instead of surfacing the issue. This could mask:
- Clock skew between measurements
- Logic errors in connector time accumulation
- Race conditions in the task-local storage

Consider using `checked_sub` with a metric/log for negative values:
```rust
let internal_latency = total_elapsed.checked_sub(connector_time)
    .unwrap_or_else(|| {
        metrics::counter!("ucs_internal_latency_negative_total").increment(1);
        Duration::ZERO
    });
```

**[should-fix]** Task-local accumulator not reset between requests

The `CONNECTOR_TIME_NANOS` task-local is initialized once and never cleared. While `scope()` creates a new scope, verify that:
1. The accumulator is properly zeroed on each request entry
2. Concurrent connector calls within the same request don't race on updates

Consider explicitly resetting the accumulator in the scope initialization:
```rust
CONNECTOR_TIME_NANOS.scope(Arc::new(AtomicU64::new(0)), async { ... })
```

**[nit]** Missing metric unit in the name

The metric is named `ucs_internal_processing_duration_seconds` but the help text says "milliseconds". Align the units or update the description.

---
Overall approach looks solid. The task-local pattern is appropriate for this use case.

XyneSpaces (commented):
## Automated PR Review Summary

This PR adds valuable observability for UCS-internal latency. The implementation using `task_local!` for per-request accumulation is sound.

**Key findings:**
1. **[should-fix]** Integer truncation risk when converting nanoseconds to u64
2. **[should-fix]** Silent clamping of negative latencies may hide clock skew issues  
3. **[nit]** Memory ordering could be strengthened for the atomic load

Overall approach looks good - the metrics will help identify internal bottlenecks separate from connector latency.

crates/common/external-services/src/shared_metrics.rs
```diff
@@ -19,6 +19,17 @@ const LATENCY_BUCKETS: &[f64] = &[
     0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
 ];
 
+tokio::task_local! {
+    pub static CONNECTOR_TIME_NANOS: std::sync::Arc<std::sync::atomic::AtomicU64>;
+}
+
+/// Add one connector call's duration to the per-request accumulator. No-op outside a request scope.
+pub fn add_connector_time(elapsed: std::time::Duration) {
+    let nanos = u64::try_from(elapsed.as_nanos()).unwrap_or(u64::MAX);
```
XyneSpaces: [should-fix] Potential integer truncation in duration conversion

The `as_nanos()` method returns `u128`, but the code tries to fit it into `u64`. While durations > 584 years are unlikely, this is a silent failure mode. Consider:

```rust
let nanos = elapsed.as_nanos();
let clamped = nanos.min(u64::MAX as u128) as u64;
```

Or use saturating conversion to be explicit about the trade-off.

XyneSpaces (commented):
Review comment on shared_metrics.rs

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -407,6 +425,37 @@ where
...
+#[cfg_attr(not(feature = "otel"), allow(unused_variables))]
+fn observe_internal_latency(
+    start_time: tokio::time::Instant,
+    connector_nanos: u64,
+    flow_name: FlowName,
+    service_name: &str,
+    metadata_payload: Option<&MetadataPayload>,
+) {
+    #[cfg(feature = "otel")]
+    {
```
XyneSpaces: [should-fix] Silent clamping hides measurement anomalies

Using `saturating_sub` means if connector_time ever exceeds total elapsed time (due to clock adjustments or measurement error), the metric will silently report 0 instead of surfacing the anomaly.

Consider logging a warning when this happens:

```rust
let total = start_time.elapsed();
let connector = Duration::from_nanos(connector_nanos);
if connector > total {
    tracing::warn!("Connector time > total time: {:?} > {:?}", connector, total);
}
let internal = total.saturating_sub(connector);
```

XyneSpaces (commented):
Review comment added for `observe_internal_latency` function.

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -407,6 +425,37 @@ where
...
+        let internal = start_time
+            .elapsed()
+            .saturating_sub(std::time::Duration::from_nanos(connector_nanos));
+        let connector = metadata_payload
+            .map(|md| md.connector.get_connector_name())
+            .unwrap_or_else(|| "unknown".to_string());
+        let mode = if metadata_payload.map(|md| md.shadow_mode).unwrap_or(false) {
+            "shadow"
+        } else {
+            "primary"
```
XyneSpaces: [nit] Relaxed ordering may cause stale reads

The code uses `Ordering::Relaxed` for both store and load operations. While acceptable for metrics where approximate values suffice, consider `Ordering::SeqCst` or at least `Ordering::Acquire` on the final load to ensure the accumulated value is fully visible.

Current:
```rust
connector_time.load(Ordering::Relaxed)
```

Suggested:
```rust
connector_time.load(Ordering::Acquire)
```

XyneSpaces (commented):
Review comment on memory ordering

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -420,6 +432,35 @@ where
     grpc_response
 }
 
+#[cfg_attr(not(feature = "otel"), allow(unused_variables))]
```
Shubhodip900: why we need `allow(unused_variables)` when all the fields are being used.

Shubhodip900 (approved):

AmitsinghTanwar007 (approved):

PR #1709, opened 2026-06-25
fix(KOUNT): use payment connector variant for Kount access-token

peeyushshukla-juspay:
## Description

Updates composite-service access-token bootstrapping so that FRM flows using Kount request OAuth tokens against the payment connector variant instead of the FRM connector variant.
- crates/internal/composite-service/src/events.rs: In CompositeEventService::notify, when the resolved connector is Frm(Kount), create the server-authentication token as Payment(Kount) and replace x-frm-connector with x-connector on the internal auth sub-request.
- crates/internal/composite-service/src/frm.rs: Applies the same Kount-specific remapping for FRM pre/post risk-check access-token creation.

## Motivation and Context

Kount shares OAuth credentials across both payment and FRM flows, but the merchant-authentication service resolves Kount auth configuration under the payment connector namespace. The composite FRM endpoints must still receive x-frm-connector: kount for correct FRM routing, so the consumer (api-txns) cannot simply send x-connector: kount. This change performs the connector/header swap internally when calling the auth service, ensuring the right credentials are loaded without breaking FRM endpoint routing.

### Additional Changes


XyneSpaces:
## Review Summary

### [should-fix] Hardcoded connector name string literals

**Files:** 
- `crates/internal/composite-service/src/events.rs` (line ~108)
- `crates/internal/composite-service/src/frm.rs` (line ~126)

The PR hardcodes `"kount"` as a string literal in two locations when inserting the `x-connector-name` header. This duplicates the connector identifier and could lead to inconsistencies if Kount's identifier changes.

**Suggested fix:**
Use the `ConnectorEnum::Kount`'s string representation instead of hardcoding:

```rust
// Instead of:
toninc::metadata::MetadataValue::try_from("kount")

// Consider:
toninc::metadata::MetadataValue::try_from(
    ConnectorEnum::Kount.to_string().as_str()
)
```

Or define a constant that both locations reference.

---

### [nit] Match arm exhaustiveness for future FRM connectors

**File:** `crates/internal/composite-service/src/frm.rs`

The current match arm only handles `FrmConnectorEnum::Kount`:

```rust
let access_token_connector = match frm_connector {
    FrmConnectorEnum::Kount => ConnectorVariant::Payment(ConnectorEnum::Kount),
};
```

While this works for now, adding a wildcard arm with an explicit error would make future additions safer:

```rust
let access_token_connector = match frm_connector {
    FrmConnectorEnum::Kount => ConnectorVariant::Payment(ConnectorEnum::Kount),
    other => {
        // Log warning that access token remapping is not implemented
        return Ok(None);
    }
};
```

This prevents compile-time breakage when new FRM connectors are added.

XyneSpaces:
[nit] Consider extracting "kount" as a constant

Both `events.rs` and `frm.rs` hardcode the string `"kount"` when building the access token connector variant and when inserting the header value. Consider defining a constant to ensure consistency:

```rust
const KOUNT_CONNECTOR_NAME: &str = "kount";
```

This reduces the risk of typos if the connector name ever needs to change or if additional Kount-specific logic is added elsewhere.

XyneSpaces:
[should-fix] Potential edge case in Kount connector error handling

In `crates/internal/composite-service/src/events.rs`, when converting the `"kount"` string to a metadata value:

```rust
tonic::metadata::MetadataValue::try_from("kount").map_err(|_| {
    tonic::Status::invalid_argument("invalid x-connector value")
})?
```

**Issue:** The error message `"invalid x-connector value"` is misleading when the failure is due to an invalid ASCII header character. `MetadataValue::try_from` fails if the string contains non-ASCII characters. Since `"kount"` is a hardcoded ASCII literal, this `map_err` transformation is effectively dead code.

**Suggestion:** Either:
1. Use `expect()` since the value is hardcoded and guaranteed to succeed
2. Keep the `map_err` but use a more accurate message like `"failed to create metadata value"`

Similarly in `frm.rs` at line 131.

XyneSpaces:
## Review Comment: FRM/Payment Auth Flow Conflation Risk

The remapping of `Frm(Kount)` to `Payment(Kount)` for access token requests introduces architectural concerns:

**Issue:** FRM (Fraud Risk Management) connectors typically have different authentication flows, token endpoints, or scopes compared to payment connectors. By routing FRM Kount through the Payment Kount auth path, you may inadvertently:

1. **Route to wrong token endpoint** - FRM and Payment services may have different `/token` endpoints
2. **Use incorrect scopes** - FRM token requests may require `risk:read` vs payment's `payment:write`
3. **Lose observability** - The `x-frm-connector` header is likely used for routing/auditing fraud checks

**Questions to verify:**
- [ ] Is Kount's auth endpoint and token format **identical** for FRM and Payment use cases?
- [ ] Are there any API gateways, WAFs, or reverse proxies filtering on `x-frm-connector` that will break?
- [ ] Does the `Payment(Kount)` auth handler make any assumptions that don't hold for FRM contexts?

**Recommendation:**
If Kount truly shares auth infrastructure between FRM and Payment, consider adding explicit documentation in the code explaining this exception. Otherwise, the FRM connector should maintain its own auth path even if it delegates to similar underlying logic.

crates/internal/composite-service/src/events.rs
```diff
@@ -98,9 +101,28 @@ where
...
+
+                let mut access_token_metadata = metadata.clone();
+                if swap_frm_header {
+                    access_token_metadata.remove(X_FRM_CONNECTOR_NAME);
+                    access_token_metadata.insert(
+                        X_CONNECTOR_NAME,
+                        tonic::metadata::MetadataValue::try_from("kount").map_err(|_| {
+                            tonic::Status::invalid_argument("invalid x-connector value")
+                        })?,
+                    );
```
AmitsinghTanwar007: hey bro never write connector specific code in the core layer

crates/integrations/connector-integration/src/connectors/kount.rs
```diff
@@ -782,6 +783,9 @@ macros::macro_connector_implementation!(
 impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
     connector_types::FrmServiceTrait for Kount<T>
 {
+    fn access_token_connector(&self) -> Option<(ConnectorVariant, &'static str)> {
+        Some((ConnectorVariant::Payment(ConnectorEnum::Kount), "kount"))
+    }
```
AmitsinghTanwar007, line 788: try to follow is required function pattern

PR #1737, opened 2026-06-26
fix(authorizedotnet): align create_connector_customer Ok/Err mapping with Direct (hyperswitch-cloud#17236)

### LINKED ISSUE #1738: [shadow-validation] authorizedotnet/authorize, create_connector_customer: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **authorizedotnet / authorize, create_connector_customer**.

`router.keyDiff:response.Err`

Fixed by UCS PR #1737.

10X-GRACE:
## What

Aligns the **authorizedotnet `create_connector_customer`** response transformer in prism (UCS) with the Direct hyperswitch gateway so the shadow UCS leg classifies the `createCustomerProfile` response identically.

Also fixes `internal tracking issue` — the **mirror-image twin** of the same production event (same merchant `innago`, same flow, same sample request `019efc14-1248-7051-8773-960e65f377b9` / `pay_tqyAficZFrJa2fpRaco2`), signature **`router.keyDiff:response.Ok`**. The two issues are the two halves of one Ok/Err divergence: on a `customerProfileId` + `resultCode==Error` response the old prism leg returned `response.Ok` (present on UCS, absent on Direct → `keyDiff:response.Ok` = #17237) while Direct returned `response.Err` (present on Direct, absent on UCS → `keyDiff:response.Err` = #17236). This single change makes prism's Ok/Err classification identical to Direct, eliminating **both** keyDiffs at once.

## Root cause

`response.Err` (and the attempt `status`) diverged between the two legs because the Ok/Err decision was gated on **different fields**:

| | Direct (`hyperswitch_connectors/.../authorizedotnet/transformers.rs:642`) | prism (old) |
|---|---|---|
| Gate | `messages.result_code` (`Ok` / `Error`) | `customer_profile_id.is_some()` |
| Duplicate→success | **any** error whose text matches `ID (\d+)` | **only** `code == "E00039"` |
| `Err.message` | result-code string → `"Error"` | the raw connector text |
| `Err.reason` | joined message texts | the raw connector text |

When Authorize.Net returns a `customerProfileId` **together with** `resultCode == "Error"` (e.g. profile-validation warnings like *"One or more fields in the profile must contain a value."*), the legs disagree:

- **Direct**: `result_code == Error`, text has no `ID <digits>` → **`response.Err`**, `status = failure`.
- **prism (old)**: `customer_profile_id.is_some()` → **`response.Ok`**, payment proceeds → `status = payment_method_awaited`.

→ `response.Err` present on the Direct leg, absent on the UCS leg = **`keyDiff:response.Err`** (the RCA also observed the twin `valueDiff` `status:{failure,payment_method_awaited}` and `response.Err.message:{"Error", <raw text>}` on sibling occurrences of the same flow).

The mismatch is symmetric: a non-`E00039` duplicate whose text *does* carry `ID <digits>` would flip the other way (Direct `Ok`, prism `Err`).

## Fix

Mirror the Direct mapping exactly (`crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs`):

- gate on `messages.result_code` instead of `customer_profile_id` presence;
- treat **any** error carrying an existing profile id as an idempotent duplicate-success (drop the `E00039`-only check);
- shape the error the same way as Direct: `message = "Error"`, `reason = <joined message texts>`.

## Verification — code-path parity (Ok/Err per response shape)

Live repro note: this is an **error path** that cannot be forced against the live sandbox (it requires Authorize.Net to return a specific `customerProfileId` + `resultCode==Error` combination), so verification is by code-path parity against the Direct gateway, consistent with prior error-path resolutions in this family.

| Authorize.Net response | Direct | prism (old) | prism (new) |
|---|---|---|---|
| `resultCode=Ok`, profile id present | **Ok** | Ok | **Ok** ✓ |
| `resultCode=Error`, text `"...duplicate...ID 933042598..."`, code `E00039` | **Ok** (dup) | Ok | **Ok** ✓ |
| `resultCode=Error`, text `"...ID 933042598..."`, code **≠ E00039** | **Ok** (dup) | **Err** ✗ | **Ok** ✓ |
| **`resultCode=Error` + profile id present, text `"...fields...must contain a value."`** (issue case) | **Err** / `failure` | **Ok** / `payment_method_awaited` ✗ | **Err** / `failure` ✓ |
| `resultCode=Error`, generic error, no id | **Err**, msg `"Error"` | Err, msg `<raw text>` ✗ | **Err**, msg `"Error"` ✓ |

After the change the new prism column matches the Direct column in every row → the diverging `response.Err` key, `status`, and `Err.message` are all eliminated.

`cargo build -p grpc-server` — clean.

## Notes
- UCS-only change (prism transformer); no proto change.
- `config/development.toml` excluded; secret scan over the diff = 0 hits.

Closes #1738


10X-GRACE:
### Also resolves hyperswitch-cloud#17237 — `router.keyDiff:response.Ok` (mirror twin of #17236)

**#17237 is the same production event as #17236**, just the *other half* of the diff:
- Merchant `innago`, connector `authorizedotnet`, flow `authorize, create_connector_customer`
- Same sample request `019efc14-1248-7051-8773-960e65f377b9` / `pay_tqyAficZFrJa2fpRaco2`
- #17236 signature = `router.keyDiff:response.Err`; **#17237 signature = `router.keyDiff:response.Ok`**

When Authorize.Net returns a `customerProfileId` together with `resultCode == "Error"`, the two legs disagreed in opposite directions, producing **two** keyDiffs from one response:

| leg | classification | `response.Ok` | `response.Err` |
|---|---|---|---|
| Direct (hyperswitch) | `Err` / `failure` | absent | **present** |
| prism (old) | `Ok` / `payment_method_awaited` | **present** | absent |
| **keyDiff** | — | `{hs:false, ucs:true}` → **#17237** | `{hs:true, ucs:false}` → #17236 |

This PR gates the Ok/Err decision on `messages.result_code` (mirroring Direct) instead of `customer_profile_id.is_some()`, so **both legs now classify identically** → both `keyDiff:response.Ok` (#17237) and `keyDiff:response.Err` (#17236) are eliminated by the single change.

#### Grafana RCA (independent, for #17237)
The exact request `019efc14…` has aged out of Loki (retention), but numerous **same-window innago `createCustomerProfile` events** (reqs `019efd4e-00cd`, `019efd4e-1cf3`, `019efd4e-74da`, `019efd51-1020` @ 2026-06-25T05:43–05:46Z) confirm the pattern:
- AuthNet returns a clean **real HTTP `status_code: 200`** to `https://api.authorize.net/xml/v1/request.api` (latency 460–910ms) with `messages.resultCode: "Error"`.
- **No 502 / no 10s read-timeout / no port-9000 markers** → this is **NOT** a benign shadow-proxy artifact; it is a deterministic Ok/Err classification divergence.
- Router-side `ErrorResponse { code: "E00041"/"E00039", message: "Error", status_code: 200, … }` confirms Direct treats `resultCode==Error` as `response.Err`.

Builds stale at the time of capture (router `2026.06.10.0-hotfix7`, connector-service `2026.06.17.0-dirty`) predate this fix.

#### Verification — code-path parity
Error path; not live-forceable (requires AuthNet to return a specific `customerProfileId` + `resultCode==Error`). Verified by code-path parity vs the Direct gateway — see the parity table in the PR description; the new prism column matches Direct in every row, so the `response.Ok`/`response.Err` keys, `status`, and `Err.message` no longer diverge.

- `cargo build -p grpc-server` — clean (rebased on `origin/main` @ `56f5b579f`).
- UCS-only (prism transformer); no proto change. `config/development.toml` excluded; secret scan over the diff = 0 hits.

XyneSpaces:
[blocking] The `extract_customer_id_from_error` function is being called with `&m.text` but the function signature likely expects `&str`. While this compiles, there's a subtle issue: the regex-based extraction now runs on **every** error path, not just E00039 errors.

Looking at the diff:
- Old code: Only called `extract_customer_id_from_error` when `error_code == "E00039"`
- New code: Calls it on **any** `ResultCode::Error` variant

While the PR description states "Direct does not special-case E00039", verify that `extract_customer_id_from_error` won't falsely match legitimate error messages that happen to contain "ID <digits>" patterns (e.g., "Transaction ID 12345 failed validation").

**Recommendation:** Add a test case for the `extract_customer_id_from_error` function to ensure it correctly:
1. Extracts profile IDs from duplicate customer messages
2. Does NOT extract false positives from unrelated error messages containing "ID <number>"

XyneSpaces:
[should-fix] The error message construction for the `ResultCode::Error` branch collects all message texts:

```rust
let error_reason = response
    .messages
    .message
    .iter()
    .map(|m| m.text.clone())
    .collect::<Vec<String>>()
    .join(" ");
```

This clones every message text and allocates a new String. Since this is an error path (not hot), it's acceptable, but consider using `Iterator::fold` or `join` directly on the iterator to avoid the intermediate Vec allocation:

```rust
let error_reason = response
    .messages
    .message
    .iter()
    .map(|m| &*m.text)  // borrow instead of clone
    .intersperse(" ")
    .collect::<String>();
```

Or simply use `fold` for better efficiency in error logging paths.

XyneSpaces:
test

XyneSpaces:
## PR Review Summary

### [blocking] Missing test coverage
This PR changes core response mapping logic without adding unit tests. Please add tests for:
- `ResultCode::Ok` + `Some(profile_id)` → Ok case
- `ResultCode::Ok` + `None` → Err case (new edge case)
- `ResultCode::Error` + extractable ID → Ok case (duplicate detection)
- `ResultCode::Error` + non-E00039 error with ID → Ok case (new behavior)
- `ResultCode::Error` + no extractable ID → Err case

### [should-fix] Verify `extract_customer_id_from_error` regex scope
The duplicate detection now triggers for ANY error message containing an ID pattern, not just E00039. Please verify the regex used by `extract_customer_id_from_error`:
- It should specifically match Authorize.Net's format: `"A duplicate record with ID <digits> already exists."`
- Avoid overly broad patterns like `\d+` that could match other numeric identifiers in error messages

**Risk:** Non-duplicate errors mentioning IDs could be falsely classified as successful duplicates.

### [should-fix] Document behavioral changes
Two behavioral changes could affect existing flows:
1. **Duplicate detection scope expansion:** Errors other than E00039 with ID patterns now succeed
2. **Ok-without-profile-id now fails:** Explicit hard failure when `result_code == Ok` but `customer_profile_id` is None

Please document these changes in the PR description for visibility during deployment.

### [nit] Remove outdated comment
The comment referencing `E00039` in the PR description is slightly misleading since the new logic treats ANY error with an ID pattern as a duplicate, not just E00039. Consider updating the description to reflect the broader matching behavior.

XyneSpaces:
[should-fix] Preserve original error details in non-duplicate Error path

When `ResultCode::Error` is returned but no duplicate ID is found, the current implementation loses the original error context:

```rust
message: "Error".to_string(),
reason: Some(error_reason),
```

This hardcoded `"Error"` message reduces observability compared to the Direct implementation. Consider preserving the original connector error code and message:

```rust
message: messages.message.first().map(|m| m.text.clone()).unwrap_or_else(|| "Error".to_string()),
```

This maintains parity with Direct's error propagation and helps with debugging production issues.

XyneSpaces:
[should-fix] Validate duplicate detection to prevent false positives

The idempotency handling now treats ANY error containing an ID pattern as duplicate success via `extract_customer_id_from_error(&m.text)`. This creates **false positive risk**:

**Example scenario:** An error like `"Reference ID 12345678 not found"` would incorrectly return success.

**Recommendation:**
1. Add validation in `extract_customer_id_from_error` to confirm the error context is actually a duplicate (e.g., checking for "duplicate" or "E00039" in the message)
2. Add test cases for non-duplicate errors that happen to contain numeric IDs
3. Document the expected error message format for duplicate detection

```rust
// Consider validating the error context before extracting ID
if m.text.contains("duplicate") || m.text.contains("E00039") {
    extract_customer_id_from_error(&m.text)
} else {
    None
}
```

XyneSpaces:
[should-fix] The comment states "regardless of the error code" but the code only checks the first error message. Verify this matches Direct's behavior - does Direct scan all error messages or only the first one when extracting the existing profile ID?

```rust
// Current implementation:
let first_error = response.messages.message.first();
if let Some(existing_profile_id) =
    first_error.and_then(|m| extract_customer_id_from_error(&m.text))
```

XyneSpaces:
[blocking] **Missing test coverage**: This PR introduces significant behavioral changes to response handling but includes no tests. At minimum, add unit tests for:
1. `ResultCode::Ok` with valid `customer_profile_id` -> Ok response
2. `ResultCode::Ok` with missing `customer_profile_id` -> Err response  
3. `ResultCode::Error` with extractable ID in message -> Ok response (duplicate case)
4. `ResultCode::Error` without extractable ID -> Err response
5. Multiple error messages joined correctly in reason field

---

[should-fix] **Error message inconsistency**: The `error_reason` joins all message texts with space separator (line ~3590), which differs from patterns used elsewhere in this file. For example, at line 3673 (SDKSessionToken flow), only `first()` error is used. Consider aligning with existing patterns or documenting why joining all messages is necessary.

---

[should-fix] **Error message clarity**: The error message and reason are identical strings ("Missing customer profile id from Authorizedotnet"). This provides no additional context. Consider making `message` more concise and putting the full descriptive text in `reason`, or include the HTTP status code for better debugging.

```suggestion
                    new_router_data.response = Err(ErrorResponse {
                        status_code: http_code,
                        code: consts::NO_ERROR_CODE.to_string(),
                        message: "Missing customer profile ID".to_string(),
                        reason: Some(format!(
                            "Authorize.Net returned Ok result code but no customer_profile_id (HTTP {})",
                            http_code
                        )),
```

XyneSpaces:
[should-fix] The duplicate customer profile ID extraction logic relies on parsing error message text, which is fragile:

```rust
if let Some(existing_profile_id) =
    first_error.and_then(|m| extract_customer_id_from_error(&m.text))
```

The `extract_customer_id_from_error` function parses natural language error messages like "A duplicate record with ID 933042598 already exists." This approach is brittle because:

1. **Message format changes**: Authorize.Net could change their error wording without notice
2. **Internationalization**: Error messages might vary by region/language
3. **Edge cases**: Partial matches or unexpected formats could cause incorrect ID extraction

**Recommendation**: Consider adding defensive validation:
- Verify the extracted ID is numeric only
- Add length bounds checking
- Log the original message when extraction succeeds/fails for observability
- Consider using a regex pattern that's more resilient than simple substring extraction

Also, the PR description mentions this mirrors Direct hyperswitch — please confirm that the Direct implementation also uses message text parsing (rather than a more stable API field) for duplicate detection.

XyneSpaces:
## Review Summary

**Classification:** connector-bugfix-webhook · metadata: grace-generated-pr  
**Risk:** low

:white_check_mark: 0 critical · :warning: 0 warning · :bulb: 1 suggestion · :mag: 0 nitpicks

---

### Suggestion

:bulb: Consider extracting the error message formatting logic into a helper function to reduce duplication between the `Error` branch cases.

The pattern for building `ErrorResponse` with `message: "Error".to_string()` and `reason: Some(error_reason)` is repeated. A helper like `build_error_response(status_code, error_code, error_reason)` would improve maintainability.

---

**What's good:**
- Clear comment explaining the Direct vs Prism divergence being addressed
- Correct handling of the duplicate customer scenario via regex extraction
- Proper use of `ResultCode` to gate Ok/Err decisions rather than field presence
- The `extract_customer_id_from_error` function properly handles the idempotency case

**Pre-existing issues:** None

---

LGTM — well-structured fix that achieves parity with the Direct gateway behavior.

XyneSpaces:
:bulb: **Error handling robustness in duplicate customer detection**

The logic extracts customer ID from error message text using `extract_customer_id_from_error`, but this pattern-matches on a human-readable string ("... ID <digits> ..."). If Authorize.Net changes their error message format, this will silently fail and return a hard error instead of recognizing the duplicate.

Consider adding a fallback log/metric when ID extraction fails to detect this scenario in production.

PR #1739, opened 2026-06-26
fix(payload): map error reason from error_description to match Direct (#17247)

### LINKED ISSUE #1776: [shadow-validation] payload/authorize, create_connector_customer: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **payload / authorize, create_connector_customer**.

`router.valueDiff:response.Err.reason`

Fixed by UCS PR #1739.

10X-GRACE:
## What

Payload returns a 400 decline as an error wrapper that carries `error_type` / `error_description` **plus** a `details` transaction object. The Direct hyperswitch connector maps the error `reason` from `error_description` (the human-readable decline message), but prism's `build_error_response` mapped `reason = details.to_string()` — dumping the entire stringified transaction blob into the reason field. Because the proto `ConnectorErrorDetails.reason` and the router-side decode (`unified_connector_service/transformers.rs`) pass this through verbatim, the UCS leg's `response.Err.reason` diverged from Direct.

## Root cause (RCA)

| Leg | source | `response.Err.reason` |
|---|---|---|
| **Direct** (`hyperswitch payload.rs:264`) | `Some(error_description)` | `"The card has been declined, contact card issuer for more information."` |
| **UCS** (prism `payload.rs:304-307`, before) | `details.to_string()` | `"{\"amount\":827.4,...,\"status\":\"declined\",\"status_code\":\"general_decline\",...}"` |

This is the `response.Err.reason` half of the same genuine payload AMEX-decline event as #17245 (`connector_http_status_code` → hs #13041) and #17246 (`response.Err.code` → hs #13043). Both legs hit the real payload API (`api.payload.com/transactions`) and got the identical HTTP 400 decline (`error_type=TransactionDeclined`, `status_code=general_decline`) — GENUINE, not a synthetic mitm/transport artifact.

Confirmed in prod Loki VS_48 for req `019efd27-6415-7042-a8c9-8147858e2766` (sample `pay_c3H0ys3ovMdVaMVvc9GP`) and the other 3 sample req-ids.

## Fix (UCS-only)

Mirror Direct in prism's payload `build_error_response`: `reason = Some(error_description)`. No proto change and no hyperswitch change needed — the proto `ConnectorErrorDetails.reason` field already exists and is plumbed end-to-end; the router decode reads it verbatim. (Contrast #13043, which had to fix `code` because of the `CONNECTOR_ERROR_RESPONSE` discriminator — `reason` has no such issue.)

## Verification

The live decline path is not reproducible on the payload sandbox (it approves every test card; the prod diffs were real-issuer AMEX declines, and repeat_payment additionally needs an established mandate). Proof is a **deterministic regression unit test** that runs the real `build_error_response` against the prod 400 decline body shape:

**before:** `response.Err.reason = "{\"amount\":827.4,...}"` (stringified `details`)
**after:** `response.Err.reason = "The card has been declined, contact card issuer for more information."` (== Direct)

```
$ cargo test -p connector-integration --lib payload::error_response_tests
test connectors::payload::error_response_tests::build_error_response_maps_reason_from_error_description ... ok
test result: ok. 1 passed; 0 failed
```

`cargo build -p grpc-server` clean. No creds / `config/development.toml` in the diff.

Closes #1776

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/payload.rs
+++ b/crates/integrations/connector-integration/src/connectors/payload.rs
@@ -300,11 +300,8 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize> Conn
         Ok(ErrorResponse {
             status_code: res.status_code,
             code: response.error_type,
-            message: response.error_description,
-            reason: response
-                .details
-                .as_ref()
-                .map(|details_value| details_value.to_string()),
+            message: response.error_description.clone(),
+            reason: Some(response.error_description),
             attempt_status: None,
             connector_transaction_id: None,
             network_advice_code: None,
@@ -765,3 +762,54 @@ macros::macro_connector_flow_status_impls!(
         ServerAuthenticationToken,
     ],
 );
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used, clippy::expect_used)]
+mod error_response_tests {
+    use domain_types::{
+        payment_method_data::DefaultPCIHolder, router_data::ConnectorSpecificConfig,
+        router_response_types::Response,
+    };
+    use interfaces::api::ConnectorCommon;
+
+    use super::Payload;
+
+    // Regression for issue #17247 (payload / repeat_payment router.valueDiff:response.Err.reason).
+    // Payload returns a 400 decline as an error wrapper carrying `error_type` / `error_description`
+    // plus a `details` transaction object. Direct hyperswitch maps `reason = error_description`
+    // (payload.rs build_error_response). Prism previously mapped `reason = details.to_string()`,
+    // which surfaced the whole transaction blob and diverged from Direct. This asserts prism now
+    // mirrors Direct: reason == error_description.
+    #[test]
+    fn build_error_response_maps_reason_from_error_description() {
+        let decline_msg = "The card has been declined, contact card issuer for more information.";
+        let body = serde_json::json!({
+            "error_type": "TransactionDeclined",
+            "error_description": decline_msg,
+            "object": "error",
+            "details": {
+                "amount": 827.4,
+                "status": "declined",
+                "status_code": "general_decline",
+                "status_message": decline_msg,
+                "type": "payment"
+            }
+        });
+        let res = Response {
+            headers: None,
+            response: serde_json::to_vec(&body).unwrap().into(),
+            status_code: 400,
+        };
+
+        let connector = Payload::<DefaultPCIHolder>::new();
+        let err = connector
+            .build_error_response(res, None, &ConnectorSpecificConfig::NoKey)
+            .expect("payload error body should parse");
+
+        assert_eq!(err.status_code, 400);
+        assert_eq!(err.code, "TransactionDeclined");
+        assert_eq!(err.message, decline_msg);
+        // reason must mirror Direct (error_description), NOT the stringified `details` blob.
+        assert_eq!(err.reason.as_deref(), Some(decline_msg));
+    }
+}


10X-GRACE:
### Before / after proof — #17247 (payload / repeat_payment, `response.Err.reason`)

**Before** (prod Loki VS_48, req `019efd27-6415-7042-a8c9-8147858e2766`):
```
"response.Err.reason": {
  "hyperswitch": "The card has been declined, contact card issuer for more information.",
  "ucs":         "{\"amount\":827.4,...,\"status\":\"declined\",\"status_code\":\"general_decline\",\"status_message\":\"The card has been declined...\"}"
}
```
prism `build_error_response` set `reason = details.to_string()` (stringified transaction blob) vs Direct's `reason = error_description`.

**After** (this PR — deterministic regression test exercising the real `build_error_response` against the prod 400 decline body shape):
```
$ cargo test -p connector-integration --lib payload::error_response_tests
test connectors::payload::error_response_tests::build_error_response_maps_reason_from_error_description ... ok
test result: ok. 1 passed; 0 failed
```
`response.Err.reason` now == Direct: `"The card has been declined, contact card issuer for more information."` → `bodyComparison` for `response.Err.reason` resolves.

Live decline not sandbox-forceable (payload sandbox approves all test cards; prod diffs were real-issuer AMEX declines + repeat_payment needs a mandate) → proof is the deterministic parse/mapping test + prod RCA. `cargo build -p grpc-server` clean; no creds / `config/development.toml` in the diff.

XyneSpaces:
## PR Review Summary

### [should-fix] Double clone of `error_description`
The current implementation clones `error_description` twice:
```rust
message: response.error_description.clone(),  // Clone #1
reason: Some(response.error_description),      // Clone #2 (into Option)
```

**Suggestion:** Extract to a variable first:
```rust
let description = response.error_description;
message: description.clone(),
reason: Some(description),
```

### [should-fix] Missing test coverage for edge cases
The regression test validates the happy path, but consider adding tests for:
- Empty or whitespace-only `error_description`
- Missing `error_description` field (fallback behavior)
- Null vs missing `details` field
- Very long error descriptions
- Unicode/special characters in error messages

### [should-fix] Consider semantic distinction between `message` and `reason`
Per Hyperswitch conventions:
- `message`: Used for smart retry logic (machine-readable)
- `reason`: Displayed on dashboard (human-readable)

Currently both fields contain the same value. If `details.status_code` contains a machine-friendly code (e.g., "general_decline"), consider using that for `message` while keeping `error_description` for `reason`.

### [blocking] Audit other connectors for similar patterns
The pattern `details.as_ref().map(|d| d.to_string())` for `reason` mapping may exist in other connectors. Please audit to ensure no other connectors have the same bug of surfacing internal JSON blobs instead of human-readable messages.

XyneSpaces:
[should-fix] Redundant clone of `error_description`

The current implementation clones `error_description` twice:
```rust
message: response.error_description.clone(),
reason: Some(response.error_description),
```

Since `ErrorResponse` owns both fields, you can avoid the clone by using `error_description` directly for `reason` and letting `message` take ownership:

```rust
message: response.error_description.clone(),
reason: Some(response.error_description),
```

Actually, looking at the field ordering, consider restructuring to move `error_description`:

```rust
let error_desc = response.error_description;
Ok(ErrorResponse {
    status_code: res.status_code,
    code: response.error_type,
    message: error_desc.clone(),
    reason: Some(error_desc),
    // ...
})
```

This is a minor optimization but worth fixing since this code path handles all error responses from the Payload connector.

XyneSpaces:
[should-fix] Both `message` and `reason` now contain identical values from `error_description`. In typical Hyperswitch patterns, `message` is the short error type while `reason` contains additional detail. Verify this duplication is intentional - if Direct truly duplicates these values, this is fine; otherwise consider differentiating them.

```suggestion
        Ok(ErrorResponse {
            status_code: res.status_code,
            code: response.error_type,
            message: response.error_description.clone(),
            reason: Some(response.error_description),
            // Consider: Should message and reason be identical?
            // Direct Hyperswitch typically uses message for the error type
            // and reason for detailed description.
```

XyneSpaces:
[blocking] **Missing error_description validation**: This change assumes `error_description` is always present in the error response. If Payload returns an error without `error_description`, `reason` will be `Some(empty_string)` instead of `None`. The previous implementation with `details.as_ref().map(...)` would return `None` if details were missing. Verify that Payload API contract guarantees `error_description` is always present in error responses.

```suggestion
            message: response.error_description.clone(),
            reason: if response.error_description.is_empty() {
                None
            } else {
                Some(response.error_description)
            },
```

---

[should-fix] **Test coverage for edge cases**: The test only covers the happy path. Consider adding tests for: (1) when `error_description` is empty string, (2) when `error_description` contains special characters that might break JSON parsing.

XyneSpaces:
Fetching comments...

XyneSpaces:
[should-fix] The error response mapping change looks correct for aligning Prism with Direct hyperswitch. A few observations:

**Positive aspects:**
- ✅ The regression test is comprehensive and tests the actual scenario
- ✅ The mapping `reason = error_description` is cleaner than the previous `details.to_string()`

**Potential concern:**
The change now unconditionally sets `reason = Some(response.error_description)`, whereas previously it was conditional (`details.as_ref().map(...)`). If `error_description` is ever an empty string, you'd get `Some("")` instead of `None`.

**Suggestion:** Consider filtering empty descriptions:
```rust
reason: Some(response.error_description).filter(|s| !s.is_empty()),
```

Or if you want to maintain parity with Direct's behavior, verify that Direct also doesn't filter empty strings. Either way, document the expected behavior for empty `error_description` values.

The test coverage looks solid — it specifically asserts that `reason` equals the `error_description` value, not the `details` blob. 👍

XyneSpaces:
:warning: Duplicate error_description in message and reason fields

The change on lines 303-304 clones `error_description` into both `message` and `reason` fields:
```rust
message: response.error_description.clone(),
reason: Some(response.error_description),
```

This causes both fields to contain identical values, whereas `reason` previously held structured `details`. Verify this matches Direct's actual behavior—the test name suggests it does, but the duplication eliminates the distinction between human-readable message and structured reason.

If Direct truly sends identical values in both fields, add a comment explaining this is intentional parity rather than accidental duplication.

XyneSpaces:
## Review Summary

**Classification:** connector-bugfix-webhook · metadata: grace-generated-pr  
**Risk:** low

:white_check_mark: 0 critical · :warning: 0 warning · :bulb: 0 suggestions · :mag: 0 nitpicks

---

**Verdict:** Approve

This is a focused, correct fix that aligns the Payload connector's error response mapping with the Direct gateway behavior.

**What's good:**
- The change is precise: `reason` now maps to `error_description` instead of `details.to_string()`
- Regression test validates the fix with a realistic error payload structure
- The comment clearly explains the Direct vs Prism divergence being addressed
- No extraneous changes — single concern PR

**Pre-existing issues** (not introduced by this PR):
- None identified

---

LGTM — ready to merge.

XyneSpaces:
[nit] The error reason fix correctly aligns Prism behavior with Direct connector. The unit test covers the regression case effectively.

Consider adding a debug log when `error_description` is empty/None to help diagnose cases where neither `details` nor `error_description` provide useful error context.

PR #1744, opened 2026-06-29
fix(cybersource): preserve Flex target origin port

shuklatushar226:
## Summary

Preserves the full URL origin when building Cybersource Flex Microform `target_origins` from `return_url`.

## Root Cause

The previous transformer rebuilt the origin from only the scheme and host, which dropped explicit ports such as `http://localhost:5173`. Cybersource enforces the target origin through iframe CSP, so dropping the port caused local Flex iframe mounting to fail for the Medusa checkout harness.

## Impact

Cybersource Flex fields can mount correctly for return URLs that include an explicit port, including local development frontends.

## Validation

- `cargo fmt --check`
- Local Medusa Cybersource checkout completed with payment status `captured`
- Verified payment session `payses_1782717706540` with Cybersource transaction id `7827177101686712304806`

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/cybersource/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/cybersource/transformers.rs
@@ -5699,9 +5699,11 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                 },
             })?;
 
-        // Extract the origin from the return_url for target_origins
+        // Extract the full origin from the return_url for target_origins.
+        // Flex Microform enforces this via CSP frame-ancestors, so a local dev
+        // origin such as http://localhost:5173 must retain the explicit port.
         let target_origin = url::Url::parse(&return_url)
-            .map(|u| format!("{}://{}", u.scheme(), u.host_str().unwrap_or_default()))
+            .map(|u| u.origin().ascii_serialization())
             .unwrap_or(return_url);
 
         Ok(Self {


XyneSpaces:
:bulb: **[nit]** Fix is correct for preserving Flex Microform target origins with ports

The change from:
```rust
format!("{}://{}", u.scheme(), u.host_str().unwrap_or_default())
```
to:
```rust
u.origin().ascii_serialization()
```
properly preserves the port in the origin (e.g., `http://localhost:5173` instead of just `http://localhost`). This addresses CSP frame-ancestors failures when using development servers with explicit ports.

Good catch on identifying this Cybersource Flex integration edge case.

iemyashasvi (approved):

kanikac199 (approved):

XyneSpaces:
:warning: **Target origin preservation fix - port handling**

The change from manually constructing `format!("{}://{}", ...)` to using `u.origin().ascii_serialization()` correctly preserves the port in local development origins (e.g., `http://localhost:5173`).

**Positive aspects:**
- Properly handles ports in `target_origins` for Flex Microform CSP `frame-ancestors`
- The `unwrap_or(return_url)` fallback handles parse failures gracefully

**Verification recommended:**
Ensure this change has been tested against both:
1. Production origins without ports (e.g., `https://api.example.com`)
2. Development origins with non-standard ports (e.g., `http://localhost:5173`, `http://127.0.0.1:8080`)

PR #1682, opened 2026-06-25
fix(nmi): send email and shipping address on repeat_payment (MIT)

AmitsinghTanwar007:
## What

NMI `repeat_payment` (MIT / `RecurringPaymentService.Charge`) was sending only the
billing address to NMI. The customer **email** and the full **shipping address**
that hyperswitch sends on recurring charges were dropped by the UCS connector
transformer.

### Diff signature (shadow validation, `internal tracking issue`)
connector=`nmi` flow=`repeat_payment` merchant=`merchant_1771364887457`

```
body.keyDiff.email            = {hyperswitch:true, ucs:false}
body.keyDiff.shipping_address1= {hyperswitch:true, ucs:false}
body.keyDiff.shipping_address2= {hyperswitch:true, ucs:false}
body.keyDiff.shipping_city    = {hyperswitch:true, ucs:false}
body.keyDiff.shipping_country = {hyperswitch:true, ucs:false}
body.keyDiff.shipping_firstname={hyperswitch:true, ucs:false}
body.keyDiff.shipping_lastname= {hyperswitch:true, ucs:false}
body.keyDiff.shipping_state   = {hyperswitch:true, ucs:false}
body.keyDiff.shipping_zip     = {hyperswitch:true, ucs:false}
```

## Root cause (L3 — UCS connector transformer only)

`NmiRepeatPaymentRequest` (`connectors/nmi/transformers.rs`) built the billing
fields (`first_name`…`phone`) but had no `email` field and no `shipping_details`.
The hyperswitch reference connector, on the `ConnectorMandateId` (stored vault)
recurring path, builds `NmiPaymentsRequest` with both `billing_details` (incl.
`email`) **and** `shipping_details`, emitted as flat top-level NMI form fields.
The data already reaches UCS over gRPC (`RecurringPaymentServiceChargeRequest`
carries `address` (billing+shipping) and `email`) and the HS→UCS mapping already
populates them — so this is purely an L3 transformer gap. No proto / domain /
HS-mapping change required.

## Fix

Add `email` and a flattened `shipping_details: Option<NmiShippingDetails>` to
`NmiRepeatPaymentRequest`, populated from the standard
`get_optional_billing_email()` / `get_optional_shipping_*()` getters — mirroring
the Authorize/SetupMandate flows in this same connector.

## Verification (diff → no_diff)

NMI outbound is TLS-blocked for a live shadow run, so verified via a direct gRPC
`RecurringPaymentService.Charge` A/B comparing the outbound NMI body
[...]

PR #1800, opened 2026-07-02
fix(nmi): drop request-email fallback for repeat_payment shipping_email/email

AmitsinghTanwar007:
## What

Fixes a shadow-validation diff on nmi `repeat_payment` (MIT via `ConnectorMandateId`):

```
body.keyDiff.shipping_email = {"hyperswitch": false, "ucs": true}
```

(the plain `email` key had the identical diff for the same reason, and is fixed by
the same change.)

## Root cause

`NmiRepeatPaymentRequest`'s builder (`connectors/nmi/transformers.rs`, added by #1682)
fell back both `email` and `shipping_details.shipping_email` to the top-level
`RepeatPaymentData.email` whenever the billing/shipping address itself had no email:

```rust
email: common_data.get_optional_billing_email().or_else(|| router_data.request.email.clone()),
...
shipping_email: common_data.get_optional_shipping_email().or_else(|| router_data.request.email.clone()),
```

The hyperswitch reference implementation has no such fallback — `get_billing_details()`
and `get_shipping_details()` (`hyperswitch_connectors/src/connectors/nmi/transformers.rs`)
only ever read the billing/shipping address's own email field, for both the Authorize and
the `ConnectorMandateId` repeat-payment path. So whenever a merchant sends a top-level
customer `email` but no address-level email, HS omits `email`/`shipping_email` from the
outbound NMI request while UCS incorrectly includes them.

## Fix

Drop the `.or_else(|| router_data.request.email.clone())` fallback for both fields —
`email` now mirrors `get_optional_billing_email()` only, `shipping_email` mirrors
`get_optional_shipping_email()` only, matching the hyperswitch reference exactly. Pure
L3 connector-transformer fix; no proto/domain/HS-side change needed (the fields already
flow to UCS unchanged).

## Verification

Reproduced and verified locally against a live NMI sandbox call through the HS↔UCS shadow
validation stack (HS `:8080` shadow-calls UCS `:8001`, both outbound requests mirrored
through a MITM proxy and diffed):

1. Authorize with `setup_future_usage: off_session` + `mandate_data` (mandate setup),
   card details, billing address **with no email field**, top-level `email` set to a
   customer email.
2. Repeat payment (MIT) via `recurring_details: {type: mandate_id, data: <mandate_id>}`,
   again a top-level customer `email` and a billing address with no email, no shipping
   address at all.
3. Compared the resulting `bodyComparison.keyDiff` for the RepeatPayment shadow call.

**Before fix:**
```json
{"email": {"hyperswitch": false, "ucs": true}, "shipping_email": {"hyperswitch": false, "ucs": true}}
```

**After fix:** `{}` (`no_diff` on both fields, no new diffs introduced).

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
@@ -1970,12 +1970,9 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
             zip: common_data.get_optional_billing_zip(),
             country: common_data.get_optional_billing_country(),
             phone: common_data.get_optional_billing_phone_number(),
-            // Prefer the billing-address email (mirrors the hyperswitch reference
-            // NMI MIT path), falling back to the top-level `RepeatPaymentData.email`
-            // so the customer email is still sent when no billing email is present.
-            email: common_data
-                .get_optional_billing_email()
-                .or_else(|| router_data.request.email.clone()),
+            // Mirrors the hyperswitch reference NMI MIT path (get_billing_details()):
+            // billing-address email only, no fallback to the top-level request email.
+            email: common_data.get_optional_billing_email(),
             shipping_details: Some(NmiShippingDetails {
                 shipping_firstname: common_data.get_optional_shipping_first_name(),
                 shipping_lastname: common_data.get_optional_shipping_last_name(),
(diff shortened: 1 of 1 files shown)


AmitsinghTanwar007:
Opened against the wrong repo by mistake — correct cross-fork PR into juspay/connector-service follows.

XyneSpaces:
Test comment to verify access

PR #1812, opened 2026-07-03
feat(airwallex): add 10 payment methods (Wallet, PayLater, BankTransfer, Skrill) + parity fixes

shuklatushar226:
## Summary

Adds/verifies **10 payment methods** for the Airwallex connector (UCS), ported for parity against the upstream Hyperswitch reference implementation. Generated via the GRACE orchestrator workflow (`grace/workflow/1_orchestrator.md`), then parity-audited against `hyperswitch_connectors/src/connectors/airwallex` and fixed.

> **Scope note:** this PR is **not** connector-local. Alongside the Airwallex work it carries proto, core-flow and SDK-codegen changes (`PaymentCreateOrderData.order_details`, the `atome` field on the proto `PaymentMethod` oneof, and the Skrill domain plumbing). Those were originally listed as follow-ups; they ship here and are itemised under **Core / proto changes** below. Reviewers sizing blast radius should treat this as touching connector-payment-method-addition, core-flow-framework, proto-api-contract, sdk-ffi-codegen and tests-specs-docs.

## Payment methods

| Payment method | Type | Status |
|---|---|---|
| Cards recurring (SetupMandate/MIT) | Card | ✅ verified (already merged) — build + grpcurl CIT→MIT, mandate parity OK |
| iDeal | BankRedirect | ✅ verified — grpcurl auth-pending + redirect |
| Trustly | BankRedirect | ✅ verified |
| Blik | BankRedirect | ✅ verified |
| Google Pay | Wallet | ✅ implemented — reached Airwallex |
| PayPal | Wallet | ✅ implemented — auth-pending + real redirect |
| Skrill | Wallet | ✅ implemented (needed core change — see below) |
| Indonesian Bank Transfer | BankTransfer | ✅ implemented — auth-pending + redirect, PSync 200 |
| Klarna | PayLater | ✅ code + parity complete; CreateOrder now carries `order_details` (see Core / proto changes) |
| Atome | PayLater | ✅ code + parity complete; proto ingress added (see Core / proto changes) |

## Core / proto changes (in scope, not connector-local)

These were listed as out-of-scope follow-ups in an earlier revision of this description. They ship in this PR:

- **`PaymentCreateOrderData.order_details`** — `connector_types.rs:1982`, `payment.proto:3073`, both construction sites (`composite-service/src/transformers.rs:133,176`), `scenario_api.rs:2883` and the `ForeignTryFrom` at `types.rs:12003`. Proto field 9 was unallocated; the JS SDK is regenerated. This unblocks Klarna (and all PayLater) full auth, which previously failed with `400 "order is required"`.
- **Atome proto ingress** — new `atome` field (115) on the `PaymentMethod` oneof in `payment_methods.proto`, plus the conversions at `types.rs:1938,2685,6634`. Proto breaking-change CI passes.
- **Skrill domain plumbing** — see the dedicated section below.

## Commits

- `a8d32aa1d` Wallet (Google Pay, PayPal)
- `a35b34e1c` PayLater (Klarna, Atome)
- `ad41b98b7` Indonesian Bank Transfer
- `e60f57c77` **Parity fixes** (see below)
- `1fa93b882` Skrill wallet — cross-cutting

## Parity fixes vs upstream (`e60f57c77`)

An audit against the reference found and fixed 3 real functional divergences, each verified by grpcurl retest + outgoing-JSON inspection:

1. **Stray `payment_method_options.card` block** was sent on Wallet and BankRedirect requests — the reference only sets card options for actual `Card`. Now `None` for those, and the field is `skip_serializing_if`, so it is omitted from the JSON rather than sent as an explicit `null`.
2. **`auto_capture` was `false` for the `None` default and `SequentialAutomatic`** (payment would authorize but never capture). Now treats `Automatic | SequentialAutomatic | None` as auto-capture, matching the reference. **See the behaviour changes section — this is merchant-visible.**
3. **`AirwallexPaymentStatus` was missing `PENDING`** (the reference's default) with no `#[serde(other)]` fallback → a `"PENDING"` response would fail to deserialize. Added `Pending → AttemptStatus::Pending`.

## Behaviour changes to already-shipped flows ⚠️

Two changes in this PR alter flows that already work in production. Both match the upstream reference, but they need a release note rather than being buried in a parity list:

- **`capture_method` default is now auto-capture.** Callers that omit `capture_method` (or send `SequentialAutomatic`) and currently rely on the manual-capture behaviour will switch to auto-capture on the next deploy.
- **Refund requests now send `payment_intent_id` instead of `payment_attempt_id`** (`AirwallexRefundRequest`). This matches the reference and the Airwallex API, but it changes an existing shipped refund flow.

## Cross-cutting change for Skrill (`1fa93b882`)

Skrill had no representation in the UCS domain model. Added minimally and safely:
- `WalletData::Skrill(Box<SkrillData>)` + empty `SkrillData {}` (domain), `PaymentMethodDataType::Skrill`, and the `proto::PaymentMethod::SkrillRedirect → WalletData::Skrill` conversion (proto ingress `skrill_redirect = 174` already existed).
- The 17 connectors that match `WalletData` exhaustively each get `| WalletData::Skrill(_)` appended to their existing terminal `NotImplemented` arm — **no behavior change** for them (qwikcilver uses a `_` catch-all, untouched).
- Airwallex maps `WalletData::Skrill` to its Skrill wallet request (PII Secret-wrapped).

## Testing

All implemented methods built (`cargo build`) and were driven end-to-end via **grpcurl** against the Airwallex sandbox (no `cargo test`, per workflow). Redirect-based methods return `AUTHENTICATION_PENDING` + a real Airwallex redirect form; card-recurring passed a full CIT→MIT chain. Outgoing request JSON was inspected in server logs to confirm correct serialization.

Field-probe status after the field-name fix — `Blik` and `Trustly` are back to `supported`, and `GooglePay`, `Klarna`, `PaypalRedirect` and `Skrill` newly reach `supported`:

```
Blik supported · Card supported · GooglePay supported · Ideal supported
Klarna supported · PaypalRedirect supported · Skrill supported · Trustly supported
```

`GooglePayDecrypted` and `IndonesianBankTransfer` remain `error`: the probe cannot synthesize a real encrypted Google Pay token, and `payment_method_data.bank_transfer.bank_name` has no Indonesian-bank patch type in `patch-config.toml`.

## Follow-ups (out of this PR's scope)

- **Indonesian bank name codes**: some `common_enums::BankNames` serialize differently from Airwallex's expected codes (e.g. `CimbBank` → `cimb_bank` vs `cimb_niaga`); Maybank works today. The unsupported-bank arm now returns a message naming the banks Airwallex accepts.
- **CI does not gate on field-probe status.** The auto-fix job commits regenerated probe data regardless of whether a method regressed from `supported` to `error`, so a probe regression can go green. Worth a separate issue.


XyneSpaces:
## Review Summary

**Classification:** `connector-payment-method-addition` · secondary: `connector-new-integration` · metadata: none
**Risk:** medium

This PR adds 10 payment methods for Airwallex connector (Cards recurring, iDeal, Trustly, Blik, Google Pay, PayPal, Skrill, Indonesian Bank Transfer, Klarna, Atome) plus parity fixes.

### Findings

No blocking issues found. The PR:
- Expands Airwallex payment method coverage significantly
- Includes cross-cutting Skrill support in domain model
- Maintains parity with upstream Hyperswitch reference implementation

### Pre-existing Notes

- Ensure `data/field_probe/manifest.json` is updated if new connector capabilities were added
- Consider verifying test specs cover all 10 new payment methods in `ucs-connector-tests`

**Recommendation:** Verify the field probe expectations reflect the new payment methods before merging.

XyneSpaces:
## Review Summary

**Verdict:** :white_check_mark: Approve with minor suggestions

:bulb: 1 suggestion

---

### :bulb: Verify Skrill variant coverage in connector tests

The new `WalletData::Skrill` variant is added to 13+ connectors' `NotImplemented` arms. Verify that:

1. The Airwallex connector implementation is fully tested (happy path + error cases)
2. Connector specs in `crates/internal/ucs-connector-tests/src/connector_specs/airwallex/` cover Skrill flows
3. Status mappings for Skrill transactions are explicitly defined

Consider adding a tracking issue for Skrill implementation in high-priority connectors beyond Airwallex.

---

**Classification:** connector-payment-method-addition · connector-new-integration  
**Risk:** medium

XyneSpaces:
This is a test

XyneSpaces:
[should-fix] Large PR adding 10 payment methods across 32 files requires better test coverage evidence.

This PR touches multiple high-risk areas:
- New `WalletData::Skrill` variant (domain model change)
- Proto ingress conversion changes
- Catch-all arm additions across 5+ existing connectors
- Airwallex connector implementation

Without test evidence in the PR description, please confirm:
1. Are there connector-specific integration tests for Airwallex new payment methods?
2. Are there regression tests for the Skrill variant additions in other connectors?
3. Have the proto conversions been validated with round-trip tests?

Consider splitting this into smaller PRs:
- PR 1: Domain model + Skrill variant (blocked until ready)
- PR 2: Airwallex connector implementation
- PR 3: Proto ingress + connector catch-all arms (depends on PR 1)

XyneSpaces:
temp

XyneSpaces:
## PR Review Summary

**Classification:** Connector payment method addition (Airwallex expansion)
**Risk:** Medium-High

### Findings

:white_check_mark: **Well-structured multi-payment-method addition**

This PR adds 10 payment methods to Airwallex plus cross-cutting Skrill support. Key implementation notes:

### Positive Aspects

1. **Cross-cutting Skrill support** - The addition of `WalletData::Skrill` to the domain model with updates to 17 other connectors' `NotImplemented` arms prevents compilation failures. Clean approach.

2. **Parity fixes identified** - Good attention to detail with the `auto_capture` fix treating `SequentialAutomatic` as auto-capture, and adding missing `PENDING` status mapping.

3. **Consistent pattern across payment methods** - Bank redirects (iDeal, Trustly, Blik), wallets (Google Pay, PayPal, Skrill), and pay-later (Klarna, Atome) follow similar implementation patterns.

### Minor Observations

:mag: **Not blocking but worth noting:**

1. **Incomplete payment methods** - PR notes Klarna is "e2e blocked" and Atome has "no proto ingress". These should either be feature-flagged off or completed before merge.

2. **Large blast radius** - Touching 17+ connectors for Skrill support increases regression risk. Ensure the `NotImplemented` arms are correctly added to all connectors that don't support Skrill.

3. **Test coverage** - Given the scale, ensure adequate test specs exist in `crates/internal/ucs-connector-tests/src/connector_specs/airwallex/` for the new payment methods.

### Approval

Structure looks good. Please verify:
- [ ] Klarna/Atome are appropriately gated if not fully ready
- [ ] Test specs exist for new payment methods
- [ ] All 17 Skrill `NotImplemented` arms have been verified

XyneSpaces:
## Automated Review Summary

**Classification:** Connector integration (Airwallex) — 10 payment methods addition  
**Risk:** Medium — Multiple connector transformers touched for new payment method support

### Findings

:white_check_mark: **Approve** — The PR follows established patterns:

1. **Consistent NotImplemented handling** — New `WalletData::Skrill(_)` arm added to ACI and Adyen transformers with proper `IntegrationError::NotImplemented` error.

2. **Airwallex-specific transformer changes** — Payment method data handling appears consistent with the connector's expected format.

3. **Parity fixes included** — The PR title mentions parity fixes, suggesting alignment with existing reference implementations.

**Note:** As this is a multi-payment-method addition PR, ensure the corresponding test specs in `ucs-connector-tests` cover all 10 new payment methods.

**Verdict:** Implementation follows established connector patterns. Consider verifying test coverage for all new payment methods.

XyneSpaces:
Fetching comments via timeline or direct API...

XyneSpaces:
:warning: **Airwallex PENDING status addition - verify state machine impact**

Adding `PENDING` to `AirwallexPaymentStatus` may require corresponding state machine transitions. Ensure the core flow framework recognizes this new status and handles it appropriately in capture/refund flow routing. Unhandled pending states can stall payment lifecycle progression.

XyneSpaces:
## Code Review Findings

### Review of PR #1812: feat(airwallex): add 10 payment methods (Wallet, PayLater, BankTransfer, Skrill) + parity fixes

This is a large PR adding multiple payment methods across multiple connectors. I'll focus on critical issues.

#### [blocking] Potential copy-paste errors in payment method matchers
**Files:** Multiple transformer files touched

The pattern `WalletData::Skrill(_)` is added to multiple connectors' `NotImplemented` branches. This is correct for connectors that don't support Skrill, but verify that:

1. Airwallex (the primary target connector) actually has Skrill **implementation**, not just a NotImplemented stub
2. The Skrill variant is properly handled in the main airwallex connector code

**Suggestion:** Confirm Skrill has a complete implementation in `airwallex/transformers.rs`, not just being added to error branches in other connectors.

---

#### [blocking] Inconsistent handling of new payment methods
**Pattern observed:** Adding `WalletData::Skrill(_)` to `NotImplemented` in aci, adyen, etc.

If Skrill is being added as a new payment method, ensure:
1. It's defined in `common_enums::PaymentMethodType` 
2. It has proper serialization/deserialization
3. Test specs cover the Skrill flow

**Suggestion:** Verify enum consistency across the codebase.

---

#### [should-fix] Parity fixes lack detail
The PR title mentions "parity fixes" but the diff shows minimal changes (single-line additions to NotImplemented matchers in various connectors).

**Suggestion:** Document what parity issues are being fixed. The current diff suggests mostly "add Skrill to unsupported list" rather than actual parity improvements.

---

#### [nit] Large PR scope
Adding 10 payment methods in a single PR makes review difficult. Future PRs should split by payment method or connector for better reviewability.

---

### Pre-existing Issues (not introduced by this PR)
None observed in this limited diff scope.

### Summary
Due to the large file count, this review is based on partial sampling. The primary concern is ensuring the new payment methods (especially Skrill) are fully implemented in Airwallex and not just added as unsupported stubs in other connectors. Requesting additional review focus on the airwallex-specific implementation files.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -277,6 +506,253 @@ fn get_device_data<T: PaymentMethodDataTypes>(
...
+    customer_name: Option<Secret<String>>,
+) -> Result<AirwallexPaymentMethod, error_stack::Report<IntegrationError>> {
+    match wallet_data {
+        domain_types::payment_method_data::WalletData::GooglePay(gpay_details) => {
+            let token = gpay_details
+                .tokenization_data
+                .get_encrypted_google_pay_token()
+                .change_context(IntegrationError::MissingRequiredField {
+                    field_name: "gpay wallet_token",
+                    context: Default::default(),
```
iemyashasvi: add more details in context

shuklatushar226: Done in 93d8d08cf.

Added an `aw_err_ctx(additional_context, suggested_action)` helper (same shape as `qwikcilver::qc_err_ctx`) plus an `AIRWALLEX_INTEGRATION_DOC_URL` const, and filled in the context for every error site this PR introduced — 15 in total, not just this one:

- Google Pay `wallet_token`
- PayPal `shopper_name`, `country_code`
- Skrill `shopper_name`, `shopper_email`, `country_code`
- Klarna `country_code`
- Atome `shopper_phone`, `country_code`
- Indonesian bank transfer `shopper_name`, `shopper_email`, `bank_name`, `country_code`
- 3DS `acs_response` serialization and the order line-item amount conversion

This line now reads:

```rust
.change_context(IntegrationError::MissingRequiredField {
    field_name: "gpay wallet_token",
    context: aw_err_ctx(
        "Airwallex Google Pay requires the encrypted Google Pay token from \
         payment_method_data.wallet.google_pay.tokenization_data",
        "Send the raw PaymentData token returned by the Google Pay API in \
         tokenization_data; it must be the encrypted `token` string, not an \
         already-decrypted or empty payload",
    ),
})
```

The `bank_name` suggestion lists the exact domain `BankNames` values `AirwallexIndonesianBankName::try_from` accepts (`bank_mandiri`, `bank_danamon`, `bank_negara_indonesia`, `bank_rakyat_indonesia`, `cimb_niaga`, `maybank`, `permata_bank`), so the merchant gets an actionable list instead of a bare `NotImplemented`.

I left the 15 pre-existing `Default::default()` sites in this file alone to keep the diff scoped to this PR — happy to do them in a follow-up if you'd like.

`cargo fmt --check`, `cargo check -p connector-integration`, and `cargo clippy -p connector-integration --all-targets -- -D warnings` all pass.

iemyashasvi (dismissed):

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -277,6 +524,321 @@ fn get_device_data<T: PaymentMethodDataTypes>(
...
+                IntegrationError::MissingRequiredField {
+                    field_name: "country_code",
+                    context: aw_err_ctx(
+                        "Airwallex Skrill requires skrill.country_code, sourced from \
+                         billing.address.country",
+                        "Send billing.address.country as a two-letter ISO 3166-1 alpha-2 code \
+                         (e.g. GB, DE) on the payment request",
+                    ),
+                }
+            })?;
```
deepanshu-iiitu: Can we move this logic to individual utils which can be reused across payment methods within airwallex?

shuklatushar226: Done in 24c5315ea.

Added an `AirwallexMethodField { label, key }` descriptor (prose name + the JSON key Airwallex nests the field under) plus five getters that own both the sourcing rule and the error context:

- `get_shopper_name` — customer name, falling back to billing full name
- `get_billing_shopper_name` — billing full name only
- `get_shopper_email`
- `get_country_code`
- `get_shopper_phone_with_country_code`

Each takes an optional `note` appended to the shared suggested action, so the method-specific caveats survive (Klarna's supported markets, `ID` for the Indonesian bank transfer). That collapses what was `shopper_name` derived in 2 places, `shopper_email` in 2, `country_code` in 4, and billing-only `shopper_name` in 5 down to one definition each:

```rust
let shopper_name  = get_shopper_name(resource_common_data, customer_name, AW_SKRILL, None)?;
let shopper_email = get_shopper_email(
    resource_common_data,
    AW_SKRILL,
    Some("Airwallex uses it to identify the Skrill wallet account"),
)?;
let country_code  = get_country_code(resource_common_data, AW_SKRILL, None)?;
```

One deliberate non-merge: `get_billing_shopper_name` is kept separate from `get_shopper_name` rather than folded in behind an `Option`. The bank-redirect methods never read `customer_name`, so a single helper would either name `customer.name` as a source in an error for a flow that ignores it, or quietly start accepting it — a behaviour change dressed up as a refactor.

### Also deduped the payment-method match

While extracting the field getters it became obvious the reuse problem was bigger than the Skrill block. The `Card` arm plus the entire `BankRedirect` block (iDeal/Trustly/Blik) was duplicated **verbatim** between `AirwallexPaymentRequest` and `AirwallexConfirmRequest`, and the card arm a third time in the SetupMandate builder — while Wallet/PayLater/BankTransfer already went through shared helpers.

Added `get_card_details` and `get_bankredirect_details` to match that existing shape, then a single `get_payment_method_details` both builders now call:

```rust
let payment_method = get_payment_method_details(
    &item.router_data.request.payment_method_data,
    &item.router_data.resource_common_data,
    item.router_data.request.customer_name.clone().map(Secret::new),
)?;
```

Borrowing `payment_method_data` instead of matching on `.clone()` also drops two clones per request.

### Two changes to pre-existing code, flagging explicitly

Routing Trustly and Blik through the shared getters changes their errors:

- `context` was `Default::default()` on both → now populated, which is what @iemyashasvi asked for in the earlier thread.
- `field_name` goes `"billing.first_name"` → `"shopper_name"`, matching every other Airwallex shopper-name site and the actual Airwallex field. Only the error string changes; the failure condition is identical.

### Verification

`cargo fmt --check`, `cargo check -p connector-integration`, and `cargo clippy -p connector-integration --all-targets -- -D warnings` all pass. `cargo test -p connector-integration --lib` is 131/131 (the 26 `--doc` failures are pre-existing on `e0a30a46b` — all `AmountConvertor::convert_back` docs, unrelated).

No wire-format change is intended: field construction and evaluation order are unchanged throughout. I'm re-running the grpcurl scenarios to confirm nothing drifted — Trustly and Blik in particular, since they're the arms moved wholesale out of both duplicated matches.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -24,10 +24,210 @@ use domain_types::{
...
+/// separate from [`get_shopper_name`] so the message never advertises `customer.name` as a source
+/// for a flow that does not read it.
+fn get_billing_shopper_name(
+    resource_common_data: &PaymentFlowData,
+    method: AirwallexMethodField,
+    note: Option<&str>,
+) -> Result<Secret<String>, IntegrationError> {
+    resource_common_data.get_billing_full_name().map_err(|_| {
+        IntegrationError::MissingRequiredField {
+            field_name: "shopper_name",
```
JeevaRamu0104: **[S1] This breaks the field-probe solver and hands merchants an unactionable field name.**

On `main` this error carried the caller-facing request path:

```rust
field_name: "billing.first_name"
```

It is now the Airwallex-internal JSON key. Same change at lines 112, 164, 184, 219 and 796 (`"shopper_email"`, `"country_code"`, `"gpay wallet_token"`).

`crates/internal/field-probe/src/patcher.rs:65` resolves `field_name` through `patch-config.toml` aliases to decide which request field to populate. `shopper_name` is aliased only to `customer.name` (`patch-config.toml:270-278`), which `get_billing_shopper_name` (line 131 here) never reads — it sources solely from the billing address. So the probe patches `customer.name`, retries, gets the identical error, and `probe_engine.rs:270` marks the method failed. That is the mechanism behind every `error` status in `data/field_probe/airwallex.json` in this PR.

Merchant impact, independent of the probe: send a Blik payment with no billing first/last name. On `main` you get `Missing required field: billing.first_name` — a path you can act on. Here you get `Missing required field: shopper_name`, which is not a field in the request contract at all; it exists only inside Airwallex's JSON body.

The prose you moved into `additional_context` is a genuine improvement. The issue is that it *replaced* the machine-readable part instead of supplementing it. Keep both:

```rust
field_name: "billing.address.first_name",   // was "shopper_name"
field_name: "billing.address.country",      // was "country_code"
field_name: "payment_method_data.wallet.google_pay.tokenization_data",
```

Adding aliases to `patch-config.toml` instead would turn the docs green while leaving the merchant-facing name wrong, so that is not the fix.

data/field_probe/airwallex.json
```diff
@@ -14,124 +14,87 @@
...
-          "method": "Post",
-          "headers": {
-            "authorization": "Bearer probe_access_token",
-            "content-type": "application/json",
-            "via": "HyperSwitch"
-          },
-          "body": "{\"request_id\":\"confirm_probe_txn_001\",\"payment_method\":{\"blik\":{\"shopper_name\":\"John\"},\"type\":\"blik\"},\"payment_method_options\":{\"card\":{\"auto_capture\":true}},\"return_url\":\"https://example.com/return\",\"device_data\":null}"
-        }
+        "status": "error",
+        "error": "Stuck on field: shopper_name. Airwallex Blik requires blik.shopper_name, sourced from billing.address first_name + last_name — Missing required field: shopper_name. Airwallex Blik requires blik.shopper_name, sourced from billing.address first_name + last_name"
```
JeevaRamu0104: **[S1] Regression, not a new gap** — Blik was `supported` on `main` and is `error` here. Same at line 539 for Trustly.

Full status delta in this file:

| Method | main | this PR |
|---|---|---|
| Blik | `supported` | **`error`** |
| Trustly | `supported` | **`error`** |
| PayPal, Skrill, Klarna, IndonesianBankTransfer, GooglePayDecrypted | `not_implemented` | `error` |
| GooglePay | `not_implemented` | `supported` |

Root cause is the `field_name` change in `transformers.rs` — see that comment. Worth flagging that CI does not gate on probe status, so the "Auto-fix (format + docs + generate)" job committed the regression and everything still went green.

docs-generated/llms.txt
```diff
@@ -58,7 +58,7 @@ examples_python: none
 connector_id: airwallex
 doc: docs/connectors/airwallex.md
 scenarios: checkout_autocapture, checkout_card, refund, void_payment, get_payment
-payment_methods: Blik, Card, Ideal, Trustly
+payment_methods: Card, GooglePay, Ideal
```
JeevaRamu0104: **[S1, same root cause]** Blik and Trustly were dropped from this line — both were listed on `main`. Combined with `all_connector.md` flipping them to `?`, this PR ships generated docs stating that two working payment methods no longer work. Fixing `field_name` in `transformers.rs` and re-running `make docs` should restore them.

crates/types-traits/domain_types/src/connector_types.rs
```diff
@@ -1977,6 +1977,9 @@ pub struct PaymentCreateOrderData {
     pub metadata: Option<SecretSerdeValue>,
     pub webhook_url: Option<String>,
     pub payment_method_type: Option<common_enums::PaymentMethodType>,
+    // Order line items, needed by some connectors (e.g. Airwallex PayLater/Klarna)
+    // at order/intent creation time.
+    pub order_details: Option<Vec<payment_address::OrderDetailsWithAmount>>,
```
JeevaRamu0104, line 1982: **[S1 — scope]** The description lists this as an out-of-scope follow-up:

> **Klarna (and all PayLater) full auth**: `PaymentCreateOrderData` lacks `order_details` ... Needs CreateOrder to carry line items.

It ships here, along with `payment.proto:3073`, `composite-service/src/transformers.rs:133,176`, `scenario_api.rs:2883`, and the `ForeignTryFrom` at `types.rs:12003`.

I reviewed it and the wiring is correct — proto field 9 was unallocated, both construction sites are updated, the JS SDK is regenerated. No defect. The problem is purely that a reviewer sizing blast radius from the description will treat this PR as connector-local and skip the core review. Combined with `Atome` (see `payment_methods.proto:187`), that is 5 matched scenarios in one PR: connector-payment-method-addition, core-flow-framework, proto-api-contract, sdk-ffi-codegen, tests-specs-docs.

Please either split the proto + `PaymentCreateOrderData` + `Atome` ingress into their own PR, or update the description to state them as in-scope.

crates/types-traits/grpc-api-types/proto/payment_methods.proto
```diff
@@ -184,7 +184,8 @@ message PaymentMethod {
     Klarna klarna = 112;                                       // Klarna - Swedish BNPL service
     Alma alma = 113;                                           // Alma - French BNPL service
     TamaraRedirect tamara_redirect = 114;                                       // Tamara - Middle Eastern BNPL service
-    
+    Atome atome = 115;                                         // Atome - Asian BNPL service
```
JeevaRamu0104, line 187: **[S1 — scope, same as `connector_types.rs:1982`]** Also listed as an out-of-scope follow-up:

> **Atome ingress**: the proto `PaymentMethod` oneof has no `atome` field (only an empty `message Atome` shell) — no gRPC ingress. Needs a proto field + conversion.

Both the field and the conversions (`types.rs:1938`, `types.rs:2685`, `types.rs:6634`) ship here. Field 115 is correctly allocated and proto breaking-change CI passes.

**[S3]** While here: the commented-out `// Atome - Asian BNPL service` block inside the `/* */` region further down is now dead and should be removed, since the real `message Atome` exists above it.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -88,13 +288,230 @@ pub struct AirwallexPaymentRequest {
     pub return_url: Option<String>,
     // Device data for fraud detection
     pub device_data: Option<AirwallexDeviceData>,
+    // CIT (setup_future_usage) only: set up an Airwallex PaymentConsent so the confirm response
+    // returns a payment_consent_id usable as the connector mandate for future MITs. Omitted for
+    // one-off payments and MITs.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub payment_consent: Option<AirwallexPaymentConsentData>,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
JeevaRamu0104, line 302: **[S2] Parity fix #1 is not achieved at the wire level.**

`payment_consent` and `customer_id` correctly get `skip_serializing_if`, but `payment_method_options` (line 287) and `device_data` (line 290) do not — so `None` serializes as an explicit JSON `null` rather than being omitted. From the probe body committed in this PR:

```json
{"request_id":"...","payment_method":{"ideal":{},"type":"ideal"},"payment_method_options":null,"return_url":"...","device_data":null}
```

The description says the stray card block is "Now `None` for those", but Airwallex still receives the key. Add `#[serde(skip_serializing_if = "Option::is_none")]` to both fields (and the mirrored pair on the confirm-request struct at lines 653/655).

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -584,17 +1376,18 @@ impl<T: PaymentMethodDataTypes> TryFrom<ResponseRouterData<AirwallexPaymentsResp
...
         let status = get_payment_status(&item.response.status, &item.response.next_action);
 
-        // Handle redirection for bank redirects and 3DS
+        // Handle redirection for APM redirects (type "redirect") AND card 3DS device-data-collection
+        // / challenge (type "other" / "device_data_collection"). Mirror native HS: surface ANY
+        // next_action that carries a URL as a GET RedirectForm. GET keeps the URL's query intact
+        // (Airwallex embeds a one-time `?key=` in the 3DS-method URL that must stay in the URL — a
+        // POST form strips the query into the body and the endpoint 401s). This replaces the old
+        // behaviour that only surfaced type "redirect" and dropped the card 3DS ("other") action.
         let redirection_data = item.response.next_action.as_ref().and_then(|next_action| {
```
JeevaRamu0104: **[S2] Redirect surfacing is now inconsistent across the three response transformers in this file.**

Dropping the `action_type == Redirect` guard here is right — but SetupMandate (line 2387) and RepeatPayment (line 2587) still carry the old gated version:

```rust
if next_action.action_type == AirwallexNextActionType::Redirect { ... } else { None }
```

SetupMandate is the CIT/card path where a 3DS challenge is most likely, and CIT->MIT is a headline feature of this PR — so the fix is missing exactly where it matters most. Either apply it there too, or add a comment explaining why SetupMandate is exempt.

Separately: `AirwallexNextAction.method` (line 1321) is now dead for redirect purposes — GET is hardcoded. The rationale in your comment (the one-time `?key=` on the 3DS-method URL must stay in the query) is sound, but silently ignoring a connector-supplied `method` deserves a note at the field itself so it is not "fixed" later by someone who has not read this comment.

crates/integrations/connector-integration/src/connectors/airwallex.rs
```diff
@@ -274,15 +274,23 @@ macros::macro_connector_implementation!(
...
-            // 2-step flow: Authorize always confirms the payment intent created by CreateOrder
-            // Get order_id from reference_id (stored after CreateOrder via set_order_reference_id)
+            // 2-step flow: Authorize confirms the payment intent created by CreateOrder.
+            // Get order_id from reference_id (stored after CreateOrder via set_order_reference_id).
             let order_id = req.resource_common_data.connector_order_id
                 .as_ref()
                 .ok_or(IntegrationError::MissingRequiredField { field_name: "connector_order_id", context: Default::default() })?;
+            // When HS re-invokes Authorize after a 3DS redirect (redirect_response present), finish
+            // the payment on the same intent via `confirm_continue` instead of `confirm`.
+            let endpoint = if req.request.redirect_response.is_some() {
```
JeevaRamu0104: **[S2] `confirm_continue` is gated on `redirect_response.is_some()` rather than on the payment method being a card.**

Same condition drives the request body at `transformers.rs:1178`. The design only holds because `return_url` selection (`transformers.rs:1141`) routes cards to `complete_authorize_url` and APMs to `router_return_url` — so in practice only cards come back through Authorize.

That is two method-dependent branches keyed on different conditions. If an APM ever reaches Authorize with `redirect_response` populated, Airwallex gets `{"three_ds": {...}, "type": "3ds_continue"}` for a PayPal or Klarna intent. Gate both on `PaymentMethodData::Card` so the coupling is explicit rather than emergent.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -411,16 +1095,122 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                .complete_authorize_url
+                .clone()
+                .or_else(|| item.router_data.request.get_router_return_url().ok()),
+            _ => item.router_data.request.get_router_return_url().ok(),
+        };
+
         Ok(Self {
             request_id,
             payment_method,
             payment_method_options,
```
JeevaRamu0104, line 1173: **[S2] `auto_capture` default change is not in the description.**

Lines 394 and 1156 now treat `None | SequentialAutomatic` as auto-capture. This matches the reference and the fix is correct — but callers who currently omit `capture_method` and rely on the manual-capture behaviour will silently switch to auto-capture on the next deploy. Worth a release note.

**[S2] Also, no automated coverage for the 1048 added lines here.** `connectors/airwallex/` has no `test.rs` and verification was manual grpcurl only. Only 6 of 100 connectors have one, so this is not a convention break — but `build_payment_method_options`, the untagged enums (`AirwallexAuthorizeRequest`, `AirwallexWalletData`, `AirwallexPayLaterData`) and the `mandate_metadata` `{"id": ...}` round-trip are pure serialization logic that unit-tests cheaply and will otherwise drift silently.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -123,8 +540,9 @@ pub struct AirwallexCardDetails {
     pub name: Option<Secret<String>>,
 }
 
-// Note: Wallet and PayLater data structures remain unimplemented.
-// This transformer currently supports card payments plus selected bank redirects.
+// Note: PayLater data structures remain unimplemented.
```
JeevaRamu0104: **[S3]** This comment is stale in three ways: PayLater *is* implemented (Klarna and Atome, directly below), and the list omits Skrill and the Indonesian bank transfer. Probably simplest to delete it rather than maintain it.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -277,6 +705,339 @@ fn get_device_data<T: PaymentMethodDataTypes>(
...
+                    shopper_email: get_shopper_email(
+                        resource_common_data,
+                        AW_ID_BANK_TRANSFER,
+                        Some("Airwallex delivers the virtual account instructions to it"),
+                    )?,
+                    // `bank_name` is required by Airwallex to route the Indonesian bank transfer;
+                    // map the domain bank to Airwallex's exact token (rejecting unsupported banks).
+                    bank_name: AirwallexIndonesianBankName::try_from(bank_name.as_ref().ok_or(
+                        IntegrationError::MissingRequiredField {
+                            field_name: "bank_name",
```
JeevaRamu0104: **[S3]** The unsupported-bank arm of `AirwallexIndonesianBankName::try_from` reuses `get_unimplemented_payment_method_error_message("airwallex")`, so a caller who sends a valid Indonesian bank transfer with an unsupported *bank* gets "Selected payment method through airwallex" — pointing at the wrong thing entirely.

The `MissingRequiredField` message right above it is good and already enumerates the supported banks; the `NotImplemented` arm should say something equivalent.

(Checked and no issue: `common_enums::BankNames` is `rename_all = "snake_case"`, so `bank_mandiri` / `cimb_niaga` / `permata_bank` in your message are exactly the tokens a caller sends.)

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
```diff
@@ -411,16 +1095,122 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+            RouterDataV2<
+                Authorize,
+                PaymentFlowData,
+                PaymentsAuthorizeData<T>,
+                PaymentsResponseData,
+            >,
+            T,
+        >,
+    ) -> Result<Self, Self::Error> {
+        match item.router_data.request.redirect_response.as_ref() {
```
JeevaRamu0104: **[S2] Undeclared refund change** — separate from this hunk, but flagging it here since it is the only other behavioural change to an already-working flow: `AirwallexRefundRequest.payment_attempt_id` was renamed to `payment_intent_id` (line ~1750).

It matches the upstream reference and is very likely correct, but it changes an existing shipped flow and appears nowhere in the description, the commit list, or the testing notes. Please call it out so it gets deliberate attention.

JeevaRamu0104 (changes requested):
## Review: request changes

Full diff read (16 files), plus companion files not in the diff: `crates/internal/field-probe/src/{patcher.rs,probe_engine.rs}`, `crates/internal/field-probe/patch-config.toml`, `crates/internal/integration-tests/src/connector_specs/airwallex/specs.json`, every `PaymentCreateOrderData` construction site, and the js/python/java/rust/go SDK clients.

**Verified clean** — no action needed on any of these: proto field `order_details = 9` was unallocated; `BankNames` 198-203 are appended with no renumbering; no reverse `common_enums::BankNames -> proto` mapping exists to fall out of sync; both `PaymentCreateOrderData` construction sites are updated; the JS SDK is regenerated for both proto changes (the other SDKs carry no per-message field maps, so nothing is stale there); `common_enums::BankNames` is `rename_all = "snake_case"`, so the Indonesian bank tokens quoted in the new error message (`bank_mandiri`, `cimb_niaga`, `permata_bank`, ...) are exactly the values a caller sends. All CI checks pass.

### Blocking

**S1 — the `field_name` rename regresses Blik and Trustly and blocks probe verification of 5 of the 6 new methods.** This PR's own `data/field_probe/airwallex.json` records Blik and Trustly moving `supported -> error`, and PayPal / Skrill / Klarna / IndonesianBankTransfer / GooglePayDecrypted landing on `error`; only GooglePay reaches `supported`. That propagates to `docs-generated/all_connector.md` (Trustly and BLIK both go from a tick to `?` = "Error / Missing required fields") and to `docs-generated/llms.txt`, which now advertises `payment_methods: Card, GooglePay, Ideal` — Blik and Trustly dropped. Runtime behaviour for Blik/Trustly is unchanged; what regressed is the merchant-facing error contract and the published coverage. Details and fix inline.

**S1 — the description contradicts the diff on the two highest-blast-radius changes.** `order_details` on `PaymentCreateOrderData` and the `Atome` proto ingress are both listed as out-of-scope follow-ups, but both ship here. The changes themselves are additive and correctly wired — I checked field numbering, construction sites and SDK regen — the problem is that anyone sizing risk from the description will skip the core/proto review entirely. Inline below.

### Non-blocking

Six S2 items (wire-level serialization, redirect-surfacing inconsistency across the three transformers in this file, the `confirm_continue` guard, the undeclared refund field rename, the `auto_capture` default change, and test coverage) and three S3 nits are left as inline comments.

### Suggested order

1. Restore request-path `field_name` values; re-run `make docs` and confirm Blik/Trustly return to `supported`.
2. Split the proto + `PaymentCreateOrderData` + `Atome` ingress into a core PR, or update the description so they get core-level review.
3. `skip_serializing_if` on `payment_method_options` and `device_data`.
4. Apply the redirect change to SetupMandate and RepeatPayment, or document the exemption.
5. Add a `PaymentMethodData::Card` guard to the `confirm_continue` condition in both places.
6. Note the refund `payment_intent_id` rename and the `auto_capture` default change in the description.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Fixed in 7f06650c1 — the field names are back to caller-facing request paths, with the Airwallex-side name kept in `additional_context`:

```rust
field_name: "billing.first_name",   // was "shopper_name" (both getters)
field_name: "billing.email",        // was "shopper_email"
field_name: "billing.country",      // was "country_code"
field_name: "billing.phone",        // was "shopper_phone"
field_name: "billing.phone.country_code",
field_name: "payment_method_data.bank_transfer.bank_name",
field_name: "payment_method_data.wallet.google_pay.tokenization_data",
```

I picked names that are already aliases in `patch-config.toml` rather than adding new ones, for the reason you gave.

**One correction to the diagnosis, though: renaming alone would not have turned the docs green.** There is a second break upstream of alias resolution.

`combine_error_message_with_context` (`errors.rs:98`) renders the error as `"Missing required field: {field}. {additional_context}"`. `extract_field_after_pattern` (`error_parsing.rs:190`) takes everything after `"Missing required field: "` and only stops at `" ("`, a quote, `}` or a newline — never a period. So the extracted "field name" is the entire remediation sentence, and `clean_error_field` (which strips `" ("` and `": "`) leaves it intact. No alias can match it.

The evidence is in the committed probe data: **Klarna is stuck on `country_code`, which *is* already aliased** (`patch-config.toml:61` → `address.billing_address.country_alpha2_code`). If the field name were the only problem, Klarna would have solved.

Fixed separately in f3b1ea71c — `clean_error_field` now also cuts at the first sentence boundary. A field path never contains `". "`, so the path survives for lookup while the prose stays in the merchant-facing message. This affects any connector that populates `additional_context` on a `MissingRequiredField`, not just Airwallex.

Both patcher tests and the new assertions pass (`cargo test -p field-probe`, 26 passed).

data/field_probe/airwallex.json
shuklatushar226: Root cause addressed in 7f06650c1 (field names) + f3b1ea71c (the probe's alias resolver, which was swallowing the remediation prose — see the `transformers.rs:138` thread for the mechanism and why the field rename alone was not sufficient).

I have not hand-edited `data/field_probe/airwallex.json` — the probe needs live sandbox calls to produce a `supported` status, so the auto-fix job will regenerate it on this push. Please re-check the Blik/Trustly rows once CI has run; if they do not come back to `supported` I would rather find out from the regenerated file than from a hand-patched one.

Your point about CI not gating on probe status is the more important one and outlives this PR — I have added it to the follow-ups section of the description so it gets its own issue rather than being lost in a thread.

docs-generated/llms.txt
shuklatushar226: Same root cause — see the `transformers.rs:138` thread. `docs-generated/` is produced by the auto-fix job from the probe data, so I have deliberately not hand-edited it; it should regenerate on this push. Worth re-checking this line after CI.

crates/types-traits/domain_types/src/connector_types.rs
shuklatushar226: Taking the second option — the description is updated rather than the PR split.

Rationale: `order_details` is what unblocks Klarna, and Klarna is one of the ten methods this PR claims to add. Splitting it out would leave this PR shipping a Klarna that provably cannot complete an authorization, and would put the proto change in a PR with no flow exercising it. The Atome field is the same argument.

What I have changed instead:

- A scope note at the top of the description stating this is not connector-local, and naming the five matched scenarios you listed.
- A **Core / proto changes (in scope, not connector-local)** section itemising `PaymentCreateOrderData.order_details` (with the construction sites and the `ForeignTryFrom`), the Atome proto ingress, and the Skrill plumbing.
- The two entries removed from the follow-ups list, and the Klarna/Atome rows in the payment-methods table flipped from ⚠️ to ✅ with a pointer to that section.

If you would still rather have them as separate PRs, say so and I will split — but I would want to land the proto half first so this one is never in a state where Klarna is advertised and broken.

crates/types-traits/grpc-api-types/proto/payment_methods.proto
shuklatushar226: Scope: covered by the same description change as `connector_types.rs:1982` — there is now a **Core / proto changes (in scope, not connector-local)** section naming the `atome` field (115) and the conversions at `types.rs:1938,2685,6634`, and the entry is out of the follow-ups list.

**[S3]** Dead block removed in 7f06650c1 — the commented-out `// Atome - Asian BNPL service` / `message Atome` pair inside the `/* */` region is gone; the real `message Atome` above it stands alone. `PayBright` and `Walley` are untouched in that region.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Fixed in 7f06650c1 — `#[serde(skip_serializing_if = "Option::is_none")]` added to `payment_method_options` and `device_data` on both `AirwallexPaymentRequest` and `AirwallexConfirmRequest`.

Added a test that asserts the wire shape rather than the struct, so this cannot silently regress:

```rust
let object = json.as_object().unwrap();
assert!(!object.contains_key("payment_method_options"));
assert!(!object.contains_key("device_data"));
```

One field I deliberately left alone: `return_url` is also `Option` without `skip_serializing_if`. It is a real field of the contract rather than a stray card block, and it is populated on every path that reaches Airwallex, so I did not want to fold an unreviewed change into a parity fix. Happy to add it if you would rather the struct be uniform.

The description's parity-fix #1 is also corrected to say the field is omitted, not just set to `None`.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Fixed in 7f06650c1, by extracting the logic rather than copying it — all three response transformers (Authorize, SetupMandate, RepeatPayment) now call one `build_redirection_data(&item.response.next_action)`, so they cannot drift again. The `action_type == Redirect` gate is gone from SetupMandate and RepeatPayment.

You are right that SetupMandate was the one that mattered: it is the CIT card path, and a `device_data_collection` / `other` next_action was being dropped there.

On `AirwallexNextAction.method` — documented at the field itself, as you asked:

```rust
/// Deserialized for completeness but deliberately **not** used to pick the redirect method —
/// [`build_redirection_data`] always emits GET. Airwallex embeds a one-time `?key=` in the
/// 3DS-method URL that has to stay in the query string; a POST form would move it into the
/// body and the endpoint 401s. Do not "fix" this by honouring `method` without re-testing the
/// card 3DS challenge end to end.
```

crates/integrations/connector-integration/src/connectors/airwallex.rs
shuklatushar226: Fixed in 7f06650c1. Both branches now key off one explicit predicate instead of two different emergent conditions:

```rust
pub(crate) fn is_card_three_ds_continue<T: PaymentMethodDataTypes>(
    request: &PaymentsAuthorizeData<T>,
) -> bool {
    request.redirect_response.is_some()
        && matches!(
            request.payment_method_data,
            domain_types::payment_method_data::PaymentMethodData::Card(_)
        )
}
```

`get_url` in `airwallex.rs` and the `AirwallexAuthorizeRequest` builder in `transformers.rs` both call it, so the endpoint and the body cannot disagree about which leg is being sent. An APM that did somehow arrive with `redirect_response` populated now gets its normal confirm body instead of a `three_ds` payload.

The doc comment records the `return_url` coupling you identified, so the next reader knows the card-only property was emergent and is now explicit.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Deleted in 7f06650c1 rather than updated, as you suggested — it was wrong about PayLater and incomplete about Skrill and the Indonesian bank transfer, and the `AirwallexPaymentType` enum immediately below it is the authoritative list anyway.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Fixed in 7f06650c1 — the unsupported-bank arm no longer borrows the generic payment-method message:

```rust
// The payment method itself is supported — only this bank is not — so the generic
// "Selected payment method through airwallex" message would point at the wrong thing.
_ => Err(error_stack::report!(IntegrationError::NotImplemented(
    "Selected bank for the Airwallex Indonesian bank transfer is not supported. \
     Airwallex accepts bank_mandiri, bank_danamon, bank_negara_indonesia, \
     bank_rakyat_indonesia, cimb_niaga, maybank or permata_bank"
        .to_string(),
    Default::default()
))),
```

Same bank list as the `MissingRequiredField` arm above it, so the two cannot disagree about what is accepted. Thanks for double-checking the `snake_case` tokens — that saved a round trip.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Called out in the description now, under a new **Behaviour changes to already-shipped flows ⚠️** section:

> **Refund requests now send `payment_intent_id` instead of `payment_attempt_id`** (`AirwallexRefundRequest`). This matches the reference and the Airwallex API, but it changes an existing shipped refund flow.

Grouped with the `auto_capture` default change from your `transformers.rs:1141` comment, so both behaviour changes to already-working flows are in one place a reviewer or release note writer will find, rather than inferred from a diff.

crates/integrations/connector-integration/src/connectors/airwallex/transformers.rs
shuklatushar226: Correction on the tests half of this: I have **reverted** `connectors/airwallex/test.rs` in 588af346c. Verification for this connector stays grpcurl against the sandbox, consistent with the rest of the PR and with the 94 connectors that have no `test.rs`. My earlier reply on this thread said the tests had landed — they are no longer in the branch, so please disregard that part.

The `auto_capture` half stands: the **Behaviour changes to already-shipped flows** section is in the description, covering both the capture-method default switch and the refund field rename.

data/field_probe/airwallex.json
shuklatushar226: Confirmed after the auto-fix job regenerated the probe data (`683e0f17f`) — the regression is gone, and four methods that were previously unreachable now solve:

| Method | main | before fix | now |
|---|---|---|---|
| Blik | `supported` | `error` | **`supported`** |
| Trustly | `supported` | `error` | **`supported`** |
| GooglePay | `not_implemented` | `supported` | `supported` |
| Klarna | `not_implemented` | `error` | **`supported`** |
| PaypalRedirect | `not_implemented` | `error` | **`supported`** |
| Skrill | `not_implemented` | `error` | **`supported`** |

`docs-generated/llms.txt:61` now reads `Blik, Card, GooglePay, Ideal, Klarna, PaypalRedirect, Skrill, Trustly`.

Two rows are still `error`, both for reasons the probe cannot solve by patching:

- **GooglePayDecrypted** — needs a real encrypted Google Pay token; no fixture can stand in for one.
- **IndonesianBankTransfer** — `payment_method_data.bank_transfer.bank_name` has no matching rule in `patch-config.toml`. The two existing bank rules are `ideal.bank_name` / `eps.bank_name`, both typed `bank_names_ing`, which is a Dutch bank and wrong for Indonesia. Solving it means a new `patch_type_to_value` variant (e.g. `bank_names_mandiri`) plus a rule — happy to add that here if you want the row green, or leave it for the probe-config follow-up since it is shared tooling rather than connector code.

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

deepanshu-iiitu (approved):

PR #1826, opened 2026-07-06
fix(checkout): build metadata.udf5 from partner_merchant_identifier_details

### LINKED ISSUE #1828: [shadow-validation] checkout/repeat_payment: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **checkout / repeat_payment**.

`body.keyDiff.metadata.udf5={"hyperswitch":true,"ucs":false}`

Fixed by UCS PR #1826.

AmitsinghTanwar007:
## What

`checkout`'s Authorize and RepeatPayment (MIT) transformers built the outbound `metadata`
object but never populated `udf5` from `partner_merchant_identifier_details`, unlike
`hyperswitch_connectors::connectors::checkout::transformers::build_metadata` on the HS side,
which sets `udf5` to `partner_details.name` (falling back to `partner_details.integrator`).

## Diff signature

Shadow-validation diff on `checkout` / `repeat_payment`:

```
body.keyDiff.metadata.udf5 = {"hyperswitch": true, "ucs": false}
```

HS's outbound checkout request body carried `metadata.udf5`; UCS's did not.

## Root cause

`partner_merchant_identifier_details` is already carried end-to-end through the UCS gRPC
proto and domain types (added for the Adyen `applicationInfo` fix) and is available on both
`PaymentsAuthorizeData` and `RepeatPaymentData`. checkout's UCS transformer simply never
read it when building `metadata` for either flow.

## Fix

Added a small `add_udf5` helper (mirroring HS's logic: prefer `partner_details.name`, else
`partner_details.integrator`) and call it from both the `Authorize` `build_metadata` and the
`RepeatPayment` request builder in
`crates/integrations/connector-integration/src/connectors/checkout/transformers.rs`.

## Verification (local shadow-validation stack)

Reproduced with a CIT (off_session, `partner_merchant_identifier_details` set) followed by
an MIT repeat payment carrying the same field, against a local HS+UCS+MITM+validation-service
stack on latest `main`:

- **Before fix**: `bodyComparison.keyDiff` → `{"metadata": {"hyperswitch": true, "ucs": false}}`
  (reproducing the reported diff).
- **After fix**: `bodyComparison.keyDiff` → `{}`, `bodyComparison.valueDiff` → `{}` (`no_diff`),
  with no new diffs introduced.

Ran as an A/B on the same UCS build (stash/restore the fix, rebuild, re-run the identical
repro) to confirm the diff reproduces only without the fix and disappears only with it.

Local checks run clean on this diff: `cargo build --bin grpc-server`,
`cargo clippy --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `typos` on the
changed file.

## Category

L3 — connector-only fix (`connector-service/.../connectors/checkout/transformers.rs`); the
field already flows through the proto + domain types, so no proto/HS-mapping change needed.

Closes #1828

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/checkout/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/checkout/transformers.rs
@@ -7,10 +7,10 @@ use common_utils::{
 use domain_types::{
     connector_flow::{Authorize, Capture, RepeatPayment, SetupMandate, Void},
     connector_types::{
-        MandateReference, MandateReferenceId, PaymentFlowData, PaymentVoidData,
-        PaymentsAuthorizeData, PaymentsCaptureData, PaymentsResponseData, PaymentsSyncData,
-        RefundFlowData, RefundSyncData, RefundsData, RefundsResponseData, RepeatPaymentData,
-        ResponseId, SetupMandateRequestData,
+        MandateReference, MandateReferenceId, PartnerMerchantIdentifierDetails, PaymentFlowData,
+        PaymentVoidData, PaymentsAuthorizeData, PaymentsCaptureData, PaymentsResponseData,
+        PaymentsSyncData, RefundFlowData, RefundSyncData, RefundsData, RefundsResponseData,
+        RepeatPaymentData, ResponseId, SetupMandateRequestData,
     },
     errors::{ConnectorError, IntegrationError},
     payment_method_data::{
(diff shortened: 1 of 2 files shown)


XyneSpaces:
[should-fix] `crates/integrations/connector-integration/src/connectors/checkout/transformers.rs:404` — `add_udf5` unconditionally overwrites the `udf5` key in merchant metadata. If the merchant already supplied their own `udf5` value, it is silently replaced by the partner/integrator name, which can break downstream reconciliation/reporting.

```suggestion
        if let Some(obj) = metadata_json.as_object_mut() {
            if !obj.contains_key("udf5") {
                obj.insert("udf5".to_string(), json!(v));
            }
        } else {
            metadata_json = json!({ "udf5": v });
        }
```

Shubhodip900:
this change is merged via https://github.com/juspay/hyperswitch-prism/pull/1941 PR

PR #1847, opened 2026-07-08
fix(fiuu): align wallet multipart form serialization

pixincreate:
## Description

This PR fixes FIUU wallet authorize request serialization in UCS. The FIUU Direct API expects wallet data as flat multipart form fields, but the generic form builder path does not reliably preserve that exact wire format for Google Pay and Apple Pay wallet payloads.

Changes included:
- Build FIUU Google Pay multipart data explicitly with flat bracketed field names such as `GooglePay[apiVersion]` and `GooglePay[paymentMethodData][tokenizationData][token]`.
- Build FIUU Apple Pay multipart data explicitly and preserve `ECI` as a submitted form field, including the empty `ECI=` case when the value is absent.
- Capture optional FIUU transaction IDs from top-level 2xx error payloads when FIUU sends one as `TxnID`, `tranID`, `TranID`, or `txID`.
- Extract FIUU NonThreeDS status mapping into a helper without changing the existing `00` / `11` / `22` status behavior.

## Why this is required

This is a real UCS-side fix. OSS has serde field annotations for FIUU wallet payloads, but UCS uses its own multipart form-building path. For wallet requests, relying on the generic builder can produce a different form shape from what FIUU expects. The explicit builders make UCS send the FIUU Direct API field names exactly.

The transaction ID change is defensive support for FIUU 2xx error bodies: if FIUU includes a transaction identifier in such an error response, UCS now preserves it for downstream sync/requery flows. If FIUU does not include one, behavior remains unchanged.

## Testing

- `cargo fmt --package connector-integration`
- `cargo test --package connector-integration fiuu`
- `cargo build --package connector-integration`
- `git diff --check`

LSP diagnostics could not be completed locally because the LSP daemon timed out.


XyneSpaces:
[should-fix] Confirm payment status is still marked as `Failure` in connector error paths

The diff removes `resource_common_data: PaymentFlowData { status: common_enums::AttemptStatus::Failure, ..router_data.resource_common_data }` from both the Fiuu and Novalnet error-response constructions. If the status is not set to `Failure` elsewhere for these `Err(ErrorResponse { ... })` outcomes, the payment attempt will retain its previous status. Please verify the caller or framework handles this, or restore the explicit `Failure` assignment.

shuklatushar226 (dismissed):

crates/integrations/connector-integration/src/connectors/fiuu/transformers.rs
```diff
@@ -1394,10 +1394,6 @@ where
                     network_decline_code: None,
                     network_error_message: None,
                 }),
-                resource_common_data: PaymentFlowData {
-                    status: common_enums::AttemptStatus::Failure,
-                    ..router_data.resource_common_data
-                },
                 ..router_data
```
Saptak88, line 1401: why is this removed?, without this last time payments got stuck at pending, pls check the fiuu docs ideally what should be the status with 2xx+Error response variant

crates/integrations/connector-integration/src/connectors/novalnet/transformers.rs
```diff
@@ -932,10 +932,6 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                     transaction_id,
                 ));
                 Ok(Self {
-                    resource_common_data: PaymentFlowData {
-                        status: common_enums::AttemptStatus::Failure,
-                        ..item.router_data.resource_common_data
-                    },
                     response,
```
Saptak88, line 939: why is this removed, isnt it confirmed failure NovalnetAPIStatus::Failure. Can we please check the docs?

Saptak88 (approved):

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

PR #1803, opened 2026-07-02
fix(finix): surface RepeatPayment failures as response.Err by connector state

AmitsinghTanwar007:
## What

Finix's `RepeatPayment` (MIT / merchant-initiated) response transformer only
returned `response.Err` when Finix echoed a `failure_message` in the payload.
A `FAILED` / `CANCELED` / `UNKNOWN` connector state with **no** `failure_message`
fell through to the success branch (`response.Ok`, `status: Failure`), which
diverges from Hyperswitch.

## Why this is wrong

Hyperswitch routes every Finix MIT/repeat charge through its `Authorize` flow
(`PaymentMethodData::MandatePayment`), whose response is built by the shared
`get_finix_response` helper. That helper returns `Err` whenever
`state.is_failure()` (`FAILED | CANCELED | UNKNOWN`), independent of whether
`failure_message` is present. UCS's own `Capture` transformer for the same
connector already implements this exact `is_failure` check — `RepeatPayment`
was the odd one out with a narrower, failure_message-gated condition.

Shadow-validation diff signature observed: `router.keyDiff:response.Err`
(Hyperswitch response carries the `Err` key; UCS response does not).

## Fix

In `TryFrom<ResponseRouterData<FinixRepeatPaymentResponse, ...>>` for the
`RepeatPayment` flow: compute `is_failure = matches!(state, Failed | Canceled
| Unknown)` (same helper pattern as the `Capture` transformer) and return
`Err(ErrorResponse{..})` whenever that holds, instead of only when
`failure_message` is `Some`. `connector_response` / mandate / status
computation are otherwise unchanged.

## Verification

The Finix sandbox (`finix.sandbox-payments-api.com`) is not reachable from
this environment for a live shadow run (TLS chain issue unrelated to this
change, previously hit by other Finix fixes in this repo's history), so this
was verified statically instead:

- `cargo build --bin grpc-server` — clean
- `cargo clippy -p connector-integration --all-targets -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- Source-parity read-through against Hyperswitch's `get_finix_response`
  (`crates/hyperswitch_connectors/src/connectors/finix/transformers.rs`) and
  against this repo's own `Capture` transformer, which already implements the
  correct `is_failure` check for the same connector.

## Scope

Touches only `RepeatPayment`'s response transformer for the `finix`
connector. No proto/domain/HS changes needed — this is an L3 UCS-only fix.

Closes #1804

PR #1851, opened 2026-07-09
fix(finix): match HS reason:None on RepeatPayment failure_message errors

### LINKED ISSUE #1852: [shadow-validation] finix/repeat_payment: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **finix / repeat_payment**.

`router.typeDiff:response.Err.reason`

Fixed by UCS PR #1851.

AmitsinghTanwar007:
## What

`finix`'s `RepeatPayment` response transformer has an explicit-failure branch (reached
when Finix's Transfer-creation response comes back with a `failure_message` field) that
set `ErrorResponse.reason` to `Some(failure_message)`. Every other finix flow (Authorize,
Capture) already sets `reason: None` in their equivalent failure branches, and
Hyperswitch's shared `get_finix_response` helper — which every finix MIT/repeat charge in
HS routes through — always sets `reason: None`, unconditionally. `RepeatPayment` was the
only one left diverging.

This surfaced in HS↔UCS shadow-validation as `router.typeDiff:response.Err.reason`
(HS: `null`, UCS: `<string>`) on `finix` `repeat_payment` traffic.

## Fix

Minimal one-branch change: set `reason: None` (and drop the now-unnecessary `.clone()` on
`failure_message` since it's no longer read twice) to match HS and UCS's own
Authorize/Capture transformers.

## Verification

- `cargo build --bin grpc-server` — clean.
- `cargo clippy -p connector-integration --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.
- **Source-parity**: confirmed by direct comparison against HS's
  `get_finix_response` (`crates/hyperswitch_connectors/src/connectors/finix/transformers.rs`,
  the `reason: None` line) and against this same file's own Authorize/Capture transformers,
  which already implement the correct behavior.
- **Live shadow repro attempted** but not achieved: reproducing this specific diff
  requires Finix's Transfer-creation call to return HTTP 2xx with a `failure_message` in
  the body (an async/soft-decline shape). Every decline trigger available in this sandbox
  (known "declined" test cards, and the documented magic decline amounts) instead comes
  back as a genuine HTTP 402 from Finix, which both HS and UCS already handle identically
  via their generic HTTP-error path (`reason: None` on both sides — no diff). This matches
  the same live-repro limitation documented on the sibling `finix`/`repeat_payment` issues
  (#19721-#19725, fixed via #1803), which were verified the same way.

## Related

Same code area as PR #1803 (finix RepeatPayment `is_failure` state-check fix) — that PR is
still open and, as currently written, would reintroduce a `reason: Some(failure_message)`
in its refactored branch; whichever of the two merges first, the other should be rebased
to keep `reason: None`.

Closes #1852


XyneSpaces:
Automated review completed: no blocking or should-fix findings identified.

10X-GRACE:
Small factual correction to the **Verification** section, from the RCA on
[#19725](https://github.com/juspay/hyperswitch-cloud/issues/19725#issuecomment-4966381940) — no change requested to
this PR's actual fix, which is a separate matter.

The body currently reads:

> This matches the same live-repro limitation documented on the sibling `finix`/`repeat_payment` issues
> (#19721-#19725, fixed via #1803), which were verified the same way.

Two things there are not right, and I'd rather they didn't propagate into the next PR that cites them:

1. **#19721–#19725 are not fixed, and are not fixable — they are benign false positives of the shadow harness.**
   All five are faces of a *single* `VS_48` record on one x-request-id (`019f2431-7746…`,
   payment `pay_aJTymyGAVwNFT7IFWL8D`). The UCS leg's outbound call to finix was read-timed-out by the shadow
   replay proxy (`validation-service…:9000`, verbatim `Read timed out. (read timeout=10)`, `latency_ms: 10015`
   → synthetic `502 bad_gateway`), so UCS never received a connector response to map. The Direct leg got a real
   finix **201** in 1.28 s and persisted `status: Charged`. There is no mapping gap on either side: the outbound
   request body was fully and correctly built before the proxy died.

2. **#1803 could not have fixed them regardless** — it is built on an inverted reading of the signature. It assumes
   hyperswitch carries the `Err` key and UCS does not; the diff record shows the reverse
   (`response.Err {hyperswitch: false, ucs: true}`, `response.Ok {hyperswitch: true, ucs: false}`). UCS is *already*
   the `Err` side, because its call 502'd.

None of this undercuts **this** PR: the `reason: Some(failure_message)` → `None` parity gap you found is real and
independently worth closing — it just isn't what #19721–#19725 reported, so those issues shouldn't be cited as its
provenance or as evidence of the repro limitation. (The genuine finix `repeat_payment` failure-path bug on that
merchant is `connector_response` being dropped on the failure arm → #1938, which is additive and shouldn't collide
with this change.)
</content>

10X-GRACE:
## Validated end-to-end against a live local shadow run — `router.typeDiff:response.Err.reason` is **fixed**

Re-verified this PR for [juspay/hyperswitch-cloud#21388](https://github.com/juspay/hyperswitch-cloud/issues/21388)
(`prod` / `golfdistrict` / `finix` / `repeat_payment`, fp `42565cea…f82022`, 3 occurrences).

The PR description notes that a live shadow repro was *not* achieved. **It has now been achieved** — the
required response shape (HTTP 2xx + a non-null `failure_message` on the Transfer-creation call) was produced
by rewriting the finix sandbox response at the mitm layer of the local shadow stack, which lets both the
Direct and the UCS leg observe the identical soft-decline body. Both legs land on `response.Err`, so the
diff surfaces as a pure `reason` **typeDiff** exactly as reported in prod.

### Root cause (confirmed against current `main`)

| Leg | Code | `reason` |
|---|---|---|
| Hyperswitch Direct | `crates/hyperswitch_connectors/src/connectors/finix/transformers.rs:617` (`get_finix_response`, failure gate at `:604` `state.is_failure()`) | `None` — unconditional |
| UCS / prism (before this PR) | `crates/integrations/connector-integration/src/connectors/finix/transformers.rs:1706` (failure gate at `:1693` `failure_message.is_some()`) | `Some(failure_message)` |

A finix MIT/repeat charge in Hyperswitch is an ordinary **Authorize** that resolves `source` from the stored
connector mandate id (`finix/transformers.rs:167-181`) and returns through `get_finix_response`
(`finix.rs:538`), so `reason: None` is what the Direct leg always emits. prism's `RepeatPayment` branch was
the only *payment* flow still emitting a non-null `reason` on a 2xx body — prism's own Capture branch
(`transformers.rs:802`) and its connector-level `build_error_response` (`finix.rs:109`) already use
`reason: None`. This PR's one-branch change brings `RepeatPayment` in line.

### Before / after proof (local shadow stack, validation service `:9100`)

Identical payment in both runs: finix MIT / `repeat_payment`, `off_session: true`, stored
`payment_method_id`, automatic capture → `POST /transfers`, response rewritten to
`state: FAILED` + `failure_code: GENERIC_DECLINE` + `failure_message: "The card was declined by the issuing bank."`

**BEFORE** — prism built from `main` (`611813c20`, `finix/transformers.rs` byte-identical to `origin/main`)
`payment_id: pay_XU4EEUNCfmQzS1gwCavP` · record `router_data_finix_Authorize_019f8a77-16b8-7553-bde0-34e1b4364844_pay_XU4EEUNCfmQzS1gwCavP`

```json
"comparison_results": {
  "keyDiff": {}, "valueDiff": {},
  "typeDiff": {
    "response.Err.reason": { "hyperswitch": "null", "ucs": "string" },        <-- #21388
    "response.Err.connector_response_reference_id": { "hyperswitch": "null", "ucs": "string" },
    "connector_response": { "hyperswitch": "object", "ucs": "null" }
  }
}
```
Legs: HS `arm=Err reason=None`, UCS `arm=Err reason="The card was declined by the issuing bank."` (both `status=failure`, `chsc=201`).

**AFTER** — prism built from `main` **+ this PR's patch** (same tree as the BEFORE build, plus the 2-line change)
`payment_id: pay_Wi3GpAV20nEMpN87fmDt` · record `router_data_finix_Authorize_019f8a78-37e4-7f53-b39f-fa747e7e0a5c_pay_Wi3GpAV20nEMpN87fmDt`

```json
"comparison_results": {
  "keyDiff": {}, "valueDiff": {},
  "typeDiff": {
    "response.Err.connector_response_reference_id": { "hyperswitch": "null", "ucs": "string" },
    "connector_response": { "hyperswitch": "object", "ucs": "null" }
  }
}
```
Legs: HS `arm=Err reason=None`, UCS `arm=Err reason=None` — **identical**.

➡️ **`response.Err.reason` is gone from `typeDiff`. `keyDiff` is `{}` in both runs.** The target facet of
#21388 is resolved by this PR, with no new facet introduced.

The two residual facets are **pre-existing and out of scope for #21388** (separate fingerprints, separately
filed): `connector_response` is [hyperswitch-cloud#19724](https://github.com/juspay/hyperswitch-cloud/issues/19724),
and `connector_response_reference_id` is the `Err`-arm counterpart of
[hyperswitch-cloud#17007](https://github.com/juspay/hyperswitch-cloud/issues/17007). Both are present
identically before *and* after, so this PR neither causes nor masks them.

### Rebase / CI status

- **No rebase pushed, deliberately.** The branch is 2 ahead / 48 behind `main`, but
  `git diff --stat <pr-head>..origin/main -- crates/integrations/connector-integration/src/connectors/finix/transformers.rs`
  is exactly this PR's own `2 insertions(+), 2 deletions(-)` — **no one has touched the anchor file since**.
  The anchor file is also byte-identical between the verified build tree and `origin/main`
  (`git diff --stat 611813c20..origin/main -- <anchor>` is empty), `git apply --check` of the PR diff is
  clean, and GitHub reports `MERGEABLE`. So the AFTER result carries over to current `main` unchanged, and a
  no-op rebase would only reset the check suite for zero code benefit.
- The two red checks (`Clippy check`, `CI Result`) are from the run of **2026-07-09**, against a base that is
  now 48 commits stale; their logs have since expired. Locally on `origin/main` + this patch,
  `cargo build -p grpc-server` and `cargo clippy -p connector-integration --all-targets -- -D warnings`
  are **clean**. Re-running the suite on a current base should clear them.

### ⚠️ Ordering hazard with #1803 (still open)

[#1803](https://github.com/juspay/hyperswitch-prism/pull/1803) rewrites this exact branch and, as currently
written, sets `reason: response.failure_message.clone()` — i.e. **it would reintroduce this diff**. Whichever
of the two merges second must keep `reason: None`. Flagging so #21388 isn't silently regressed.

gopikrishna000 (approved):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

pixincreate (approved):

AkshayaFoiger (approved):

PR #1853, opened 2026-07-09
fix(novalnet): keep attempt status on SetupMandate API FAILURE to match Hyperswitch

AmitsinghTanwar007:
## Summary
On an API-level `FAILURE` response from Novalnet, the UCS `SetupMandate` connector
transformer was **overwriting the attempt status to `Failure`** at the router-data level.
Hyperswitch (the ground truth both sides are compared against) does **not** set a status
on this path — it returns `Err(ErrorResponse)` (with `attempt_status: None`) and leaves the
incoming (`pending`) status untouched, letting the error drive the final attempt status
downstream.

This produced a shadow-validation `router.valueDiff:status`:

```
"valueDiff":{"status":{"hyperswitch":"pending","ucs":"failure"}}
```

(connector `novalnet`, flow `setup_mandate`, merchant `bu_de_getolo`)

## Root cause
Same "force `Failure` on the error arm" anti-pattern already fixed for the `Authorize`
sub-flow in #1680 (issue `21051`), which explicitly flagged `SetupMandate` / `RepeatPayment`
/ `Capture` / `Void` / `PSync` as follow-ups to align if/when they surface. This is the
`SetupMandate` one surfacing.

HS `novalnet/transformers.rs` (ground truth), `SetupMandate` `Failure` arm:
```rust
NovalnetAPIStatus::Failure => {
    let response = Err(get_error_response(item.response.result, item.http_code, transaction_id));
    Ok(Self { response, ..item.data })   // status UNCHANGED
}
```

UCS `novalnet/transformers.rs`, `SetupMandate` `Failure` arm (before this fix) forced:
```rust
    status: common_enums::AttemptStatus::Failure,
    ..item.router_data.resource_common_data
},
```

## Fix (L3 — UCS connector transformer only)
In the novalnet `SetupMandate` `Failure` arm, stop overwriting the status — mirror HS by
returning the `Err` and leaving `resource_common_data` (and thus the attempt status)
unchanged via `..item.router_data`. No proto / domain / HS-mapping change.

## Verify
Reproduced locally with a novalnet test-decline card (`4000000000000002`, no_three_ds) on a
zero-amount `setup_mandate` request:
- **Before fix**: router-data comparison verdict —
  `valueDiff: {"status": {"hyperswitch": "pending", "ucs": "failure"}}`
- **After fix**: same repro — `valueDiff: {}` (status no longer diverges; verdict flips
  diff → no_diff). Remaining `typeDiff` entries on that request
  (`response.Err.connector_response_reference_id`) are a separate, already-tracked issue.
- `cargo build -p connector-integration` / `--bin grpc-server` — clean.
- `cargo clippy -p connector-integration --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.

The other sibling flows (`RepeatPayment` / `Capture` / `Void` / `PSync`) still carry the
same pattern; left untouched here since only `SetupMandate` is observed diverging for this
issue — same follow-up policy as #1680.

Fixes `21419`.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/novalnet/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/novalnet/transformers.rs
@@ -932,10 +932,6 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                     transaction_id,
                 ));
                 Ok(Self {
-                    resource_common_data: PaymentFlowData {
-                        status: common_enums::AttemptStatus::Failure,
-                        ..item.router_data.resource_common_data
-                    },
                     response,
                     ..item.router_data
                 })


pixincreate:
#1847

10X-GRACE:
### Re-validated on current `main` — resolves shadow issue #20611

Reopening: this PR is the exact fix for **[prod] Validation router diff #20611** (`bu_de_getolo` / `novalnet` / `setup_mandate`, signature `router.valueDiff:status`). It was closed on 2026-07-09 with a stray reference to #1847 (an unrelated, also-closed `fiuu` PR); the fix was never merged and `main` still carries the override. Re-verified end-to-end in the local shadow stack against fresh `main`.

**Root cause (unchanged):** on `NovalnetAPIStatus::Failure`, the UCS `SetupMandate` transformer overwrote `status → AttemptStatus::Failure`, while HS Direct (`novalnet/transformers.rs`, `..item.data`) leaves the incoming `pending` status and signals failure only via `Err(ErrorResponse)`. HS's UCS→RouterData mapping propagates the gRPC status faithfully, so the override surfaced as the diff. Same anti-pattern as the Authorize fix #1680. Mapping arms for enumerated statuses are byte-identical between the two repos.

**Before (reproduced on `main`, x-req `019f81eb-3ec2` / `019f81ef-4edb`):**
```json
{"keyDiff":{},
 "valueDiff":{"status":{"hyperswitch":"pending","ucs":"failure"}},
 "typeDiff":{"response.Err.attempt_status":{"hyperswitch":"null","ucs":"string"},
             "response.Err.connector_response_reference_id":{"hyperswitch":"null","ucs":"string"}}}
```
Both legs received the **same** Novalnet response (identical `connector_transaction_id 15416500001115140`) → deterministic mapping divergence, not a double-execution artifact.

**After (this fix, x-req `019f81f6-351e`, merchant `novalnet_repro`, `SetupMandate`):**
```json
{"keyDiff":{}, "valueDiff":{},
 "typeDiff":{"response.Err.attempt_status":{...},"response.Err.connector_response_reference_id":{...}}}
```
`router.valueDiff:status` is **gone** — both legs now `status: pending` (`hyperswitch_status:"pending", ucs_status:"pending"`, `total_value_differences: 0`). ✅ The issue #20611 signature is resolved and `keyDiff == {}`.

**Success-path regression check (x-req `019f81f6-f9f1`):** an agreeing successful `setup_mandate` → both legs `charged`, `comparison_status: no_differences`. ✅ Only the Failure arm is affected.

**Scope note:** the two residual `typeDiff` entries (`response.Err.attempt_status`, `response.Err.connector_response_reference_id`) are a **separate, pre-existing signature** — not `valueDiff:status`, present both before and after this change, and driven by HS's UCS-mapping layer deriving those `ErrorResponse` fields from the gRPC status/`merchant_order_id`. They are out of scope for #20611 and were intentionally untouched here (consistent with #1680). They can be tracked separately if filed.

UCS-only fix (no proto, no hyperswitch change). Merges cleanly on `main`.

swangi-kumari (approved):

JeevaRamu0104 (approved):

PR #1855, opened 2026-07-09
feat(connector): [absa_sanlam] Add payment flow tests for absa_sanlam

Saptak88:
## Description

- Added absa_sanlam_payment_flows_test.rs gated behind connector-request-kafka.

### Additional Changes


XyneSpaces:
[should-fix] Possible issue — worth verifying: `default_absa_sanlam_topic` derives the Kafka topic from `config.connectors.absa_sanlam.base_url`. If `base_url` is a URL (the usual meaning of that config key), the resulting topic name will contain `://` and other invalid characters, so the consumer subscription will fail to match the `dev_payments_queue` topic created in CI.

Suggest using the connector name or a dedicated Kafka topic-prefix config instead of `base_url`:
```rust
fn default_absa_sanlam_topic() -> String {
    format!("{}_payments_queue", CONNECTOR_NAME)
}
```

Shubhodip900 (dismissed):

.github/workflows/ci.yml
```diff
@@ -430,6 +430,32 @@ jobs:
        needs.changes.outputs.ci == 'true')
     runs-on: ubuntu-latest
 
+    services:
```
gopikrishna000: GitHub Actions service containers can't be gated by `if:` — they start before any step runs. Every step in this job is guarded by `if: env.RUN_TESTS == 'true'`, but the container ignores that guard.

So on fork PRs (`RUN_TESTS=false`, every step skips, job is a no-op) we still pull the ~1GB `cp-kafka` image and wait for it to go healthy. This job also triggers on `any-rust || proto || ci`, so a PR that never touches this connector gets a broker too.

Suggest moving the service (and the feature flag below) into a separate job gated on absa_sanlam paths, using the existing `changes` job.

gopikrishna000 (commented):
Local review, focused on the CI wiring. The test itself is well-built (subscribing before the authorize call with `auto.offset.reset=earliest` + a fresh `group.id` avoids the publish-before-subscribe race, and the service's listener config is correct), so this is really one design question plus some CI scoping.

The main point is on the test file: `build_kafka_record` is a pure function, so every assertion here can run with no broker at all — which would remove the entire CI diff while keeping all of the regression protection. Failing that, the service container and the feature flag both need to be scoped to absa_sanlam PRs rather than riding on the shared `test` job.

Nothing blocking, just want to make sure we're not paying a broker on every PR for assertions that don't need one.

.github/workflows/ci.yml
```diff
@@ -543,13 +581,15 @@ jobs:
...
             cargo nextest run \
+              --features grpc-server/connector-request-kafka \
               --config-file .nextest.toml \
               --profile ci \
               --no-tests=warn \
               -E "($FILTER) & ($SCHEMA_EXCLUDE)"
           else
             echo "🔍 Running full test suite (event=$EVENT)"
             cargo nextest run \
+              --features grpc-server/connector-request-kafka \
```
gopikrishna000, line 553: This enables the feature on the full suite (and on the connector-only path at L584), so the `#![cfg(feature = "connector-request-kafka")]` test compiles and runs on *every* CI run — which is what makes the broker above mandatory every time.

Scoping this to absa_sanlam PRs would let the container be scoped too.

.github/workflows/ci.yml
```diff
@@ -519,6 +545,18 @@ jobs:
...
 
+      - name: Create Kafka topics
+        if: ${{ env.RUN_TESTS == 'true' }}
+        shell: bash
+        run: |
+          docker exec "${{ job.services.kafka.id }}" kafka-topics \
+            --bootstrap-server localhost:9092 \
+            --create \
+            --if-not-exists \
+            --topic dev_payments_queue \
```
gopikrishna000: This hardcodes `dev_payments_queue`, but the test derives the topic as `format!("{}_payments_queue", config.connectors.absa_sanlam.base_url)`, and dev `base_url = "dev"`. They match today.

But with `KAFKA_AUTO_CREATE_TOPICS_ENABLE: "true"` set above, if `absa_sanlam.base_url` ever changes, this step will silently create the wrong topic, the test will still pass via auto-create, and this step quietly becomes dead code. Worth deriving the name or dropping the step and leaning on auto-create.

.github/workflows/ci.yml
```diff
@@ -430,6 +430,32 @@ jobs:
...
+          KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
+          KAFKA_CONTROLLER_LISTENER_NAMES: CONTROLLER
+          KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+          KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
+          KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+          KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
+          KAFKA_AUTO_CREATE_TOPICS_ENABLE: "true"
+        options: >-
+          --health-cmd "kafka-topics --bootstrap-server localhost:9092 --list"
+          --health-interval 10s
```
gopikrishna000: `--health-interval 10s` means the first probe doesn't fire for 10s, and with `--health-retries 12` the worst case is 120s before step one runs.

A `--health-start-period` plus a shorter interval, and a lighter image (`apache/kafka` KRaft ~350MB, or Redpanda) would cut most of the recurring cost — if the container stays at all.

crates/grpc-server/grpc-server/Cargo.toml
```diff
@@ -92,6 +92,7 @@ rand = "0.8.5"
 uuid = { workspace = true, features = ["v4"] }
 reqwest = { version = "0.11", features = ["json"] }
 serial_test = "3.2.0"
+rdkafka = "0.36"
```
gopikrishna000: `rdkafka` is unconditional here, and Cargo doesn't allow `optional` in `[dev-dependencies]`, so it can't be feature-gated. It's only used by the one `#![cfg(feature = "connector-request-kafka")]` test, so it sits in grpc-server's dev graph as dead weight whenever the feature is off.

Compile cost is small — the rlib is already built for `connector_request_kafka`, which is a workspace member with a hard `rdkafka` dep — so this is hygiene rather than perf.

crates/grpc-server/grpc-server/tests/absa_sanlam_payment_flows_test.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+fn default_absa_sanlam_topic() -> String {
+    let config = configs::Config::new().expect("Failed while parsing config");
+    format!("{}_payments_queue", config.connectors.absa_sanlam.base_url)
+}
+
+#[tokio::test]
+#[serial]
+async fn test_absa_sanlam_authorize_publishes_eft_debit_to_kafka() {
```
gopikrishna000, line 152: Main comment. `build_kafka_record` is a pure function — `RouterDataV2 -> KafkaRecord`, composing `get_kafka_topic` + `get_headers` + `get_request_body`. No producer, no broker.

Everything this test asserts (the topic, the three headers, the payload fields) is available from `build_kafka_record(req)` with zero infrastructure. The broker only exercises rdkafka's producer, which is third-party code we don't own. Asserting against `build_kafka_record` directly — an `insta` snapshot would make any payload change show up as a reviewable diff — keeps all of the regression protection this is after and removes the entire CI diff: service container, topic step, feature flags, dev-dep.

Nits if the broker does stay:
- `#[serial]` is redundant under nextest, which is process-per-test.
- Subscribing before the authorize call with `auto.offset.reset=earliest` and a fresh `group.id` is the right call — that's the publish-before-subscribe race handled properly.

Saptak88: A direct `build_kafka_record` assertion would still pass if the Kafka branch in `execute_connector_processing_step` broke or if the service/config path stopped initializing Kafka correctly.

.github/workflows/ci.yml
Saptak88: done

.github/workflows/ci.yml
Saptak88: done

gopikrishna000 (dismissed):

crates/common/connector_request_kafka/src/lib.rs
```diff
@@ -15,15 +15,25 @@ use rdkafka::{
...
 
-static KAFKA_PRODUCER: OnceCell<Arc<KafkaProducer>> = OnceCell::new();
+pub type KafkaPublishResult = CustomResult<Result<Response, Response>, KafkaClientError>;
+
+#[async_trait::async_trait]
+pub trait KafkaPublish: Send + Sync + 'static {
+    async fn publish(&self, kafka_record: KafkaRecord) -> KafkaPublishResult;
+}
+
+static PUBLISHER: OnceCell<Arc<dyn KafkaPublish>> = OnceCell::new();
```
AmitsinghTanwar007, line 32: **`OnceCell` makes tests non-isolatable**

`PUBLISHER` is a `static OnceCell` — set once, forever per process. This works today because there's exactly one Kafka test in one file and `cargo nextest` runs each test file as a separate binary.

But if someone adds a second `#[tokio::test]` in the same file that also calls `set_publisher`, the second call silently returns `false` and reuses the first test's publisher — tests become order-dependent with no error.

Also, `grpc_test!` calls `app::Service::new()` which calls `init_kafka_producer()` → `set_publisher()`. Whichever runs first (the test's explicit `set_publisher` or the app's `init_kafka_producer`) wins. Currently works because the test calls `set_publisher` before `grpc_test!` and config has `enabled = false`, but it's fragile.

Consider using `ArcSwap` instead of `OnceCell`, or passing the publisher through the service context rather than a global static.

crates/common/connector_request_kafka/src/lib.rs
```diff
@@ -70,7 +80,7 @@ pub fn init_kafka_producer(
         .fetch_metadata(None, Duration::from_secs(5))
         .change_context(KafkaClientError::MetadataFetchFailed)?;
 
-    let _ = KAFKA_PRODUCER.set(Arc::new(KafkaProducer {
+    let _ = set_publisher(Arc::new(KafkaProducer {
```
AmitsinghTanwar007: **`set_publisher` return value silently discarded**

`let _ =` swallows the `false` case. If `init_kafka_producer` is somehow called twice (e.g. config reload), the second real producer is silently dropped and the old one keeps running. This existed before the PR (`let _ = KAFKA_PRODUCER.set(...)`), so not a regression — but this refactor was a good opportunity to at least log a warning when the set fails.

crates/common/connector_request_kafka/Cargo.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+async-trait = "0.1"
```
AmitsinghTanwar007, line 7: **Minor: `async-trait` already available via `tonic`**

`tonic` re-exports `async_trait` (the test file already uses `#[tonic::async_trait]`). Adding `async-trait` as a direct dependency here is unnecessary — you could use `tonic::async_trait` instead and avoid the extra dep.

Saptak88: It is not the test file

crates/common/connector_request_kafka/src/lib.rs
Saptak88: app::Service::new() which calls init_kafka_producer() wont run as enabled=false in config.

crates/common/connector_request_kafka/src/lib.rs
Saptak88: Warn log added

AmitsinghTanwar007 (approved):

gopikrishna000 (approved):

PR #1859, opened 2026-07-09
fix(pipeline): release pipeline

Jeeva0104:
## Description

Release pipeline is not pushing the .so file 
## Motivation and Context

python sdk and java sdk is not having the .so binary

### Additional Changes


XyneSpaces:
:rotating_light: **[blocking]** The Python publish job lost the `preview_only == 'false'` guard on the "Publish to PyPI" step.

In the original workflow this step had `if: ${{ needs.setup.outputs.preview_only == 'false' }}`; the new revision runs `python3 -m twine upload artifacts/sdk-python/*.whl` unconditionally. In preview-only releases this will still publish the wheel to PyPI.

Restore the guard:
```yaml
      - name: Publish to PyPI
        if: ${{ needs.setup.outputs.preview_only == 'false' }}
        env:
          TWINE_USERNAME: __token__
          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
        run: python3 -m twine upload artifacts/sdk-python/*.whl
```

PR #1881, opened 2026-07-10
refactor(grpc): convert handler errors to tonic::Status inside the logging wrapper

### LINKED ISSUE #1878: UCS: preserve gRPC error context for alerting
Parent: #1876

Scope:
- Make UCS gRPC wrappers log failures before converting them to tonic::Status.
- Make wrappers accept typed error_stack reports where Report<E>: IntoGrpcStatus.
- Remove direct tonic::Status conversion from high-traffic connector execution paths.
- Emit alert fields: request_id, flow, connector, execution_mode, grpc_code, error_code, http_status_code, error_message, and extra_error.

Acceptance:
- Failed UCS gRPC requests emit one structured error log/event with extra_error when typed error context exists.
- Public gRPC error response remains compatible with current IntegrationError, ConnectorError, and WebhookError details.
- High-traffic payment/refund connector paths no longer lose stack before grpc_logging_wrapper.

gopikrishna000:
## Description

Moves the `tonic::Status` conversion out of the handlers and into `grpc_logging_wrapper`.

- Handlers and request parsers return `Report<GrpcError>` instead of `tonic::Status`. The wrapper converts, so the error stack is in scope at the point where it is logged, rather than having been flattened away by the handler.
- `GrpcError` carries one variant per leaf error type, named after the error it holds rather than the phase that raised it — the same leaf arises in more than one phase (a proto conversion inside a response handler is still an `IntegrationError`). A new `InternalError` leaf covers the faults that are ours rather than the caller's or the connector's (missing request extensions, missing config, test-context creation), so they stop being flattened into an anonymous `internal`.
- There is deliberately **no `From<tonic::Status>` for `GrpcError`**, and there must not be one: its absence is what stops a hand-built status from compiling, and therefore what guarantees every error reaches the wrapper as a report.
- The pure error-to-status mapping lives behind a **private** trait, `ToGrpcStatus::to_grpc_status_unlogged`. It has to exist, but it is private to `error.rs` so that the only path to a `Status` a caller can reach is `IntoGrpcStatus`, which logs the report first. A public non-logging mapper would let a caller produce a status that never reaches the alerting pipeline.
- `ConnectorFlowError` stays in `domain_types` as the connector engine's error and is lifted at the boundary. `external-services` has no business naming a gRPC concept.

Also fixed, found while doing the above:

- **Transport failures keep their own context.** `execute_connector_processing_step` re-contexted them as `ConnectorError`, so a connector timeout reached the caller as `internal` — indistinguishable from a UCS fault — and the existing timeout → `deadline_exceeded` mapping was unreachable.
- **A malformed `x-config-override` header is the caller's mistake.** The middleware discarded the `IntegrationError` the merge already produced and substituted `Status::internal`.
- **`token_authorize` / `token_setup_recurring` were wrapped twice.** Each opened its own `grpc_logging_wrapper` and then delegated to `authorize` / `setup_recurring`, which wrap themselves too — and both wrappers were handed the same flow name, service name and config. So `create_and_emit_grpc_event` fired twice with identical dimensions and `observe_internal_latency` recorded the outer span nested around the inner one: tokenized payments were double-counted in Kafka and in the latency histogram. The adapters now convert the payload and delegate; the inner flow owns the wrapper.
- **Removed the `has_payment_method_data: true` macro arm** (244 lines). It is never invoked and does not compile.

### Known limitation

Config extraction still runs *before* the wrapper, because the wrapper takes `config` as a parameter:

```rust
let config = get_config_from_request(&request).into_grpc_status()?;   // outside the wrapper
grpc_logging_wrapper(request, &service_name, config.clone(), FlowName::Authorize, |rd| ...)
```

So the wrapper is not yet a true catch-all, and `into_grpc_status()` is called there too (~40 sites, one per RPC). Such a failure **is** still logged with `extra_error` and the full chain — it goes through `IntoGrpcStatus for Report<GrpcError>` — but it bypasses `create_and_emit_grpc_event`, so it produces no gRPC audit event.

Closing this means having the wrapper extract config itself and dropping the parameter. `RequestData` already carries `extensions`, and the macro-generated handlers already read config from there rather than from the captured variable, so the plumbing exists. Left out of this PR to keep it reviewable.

The token adapters have a related gap. Now that they delegate rather than wrap, the golden log line for a tokenized payment is emitted by the inner flow and carries `service_method = "authorize"`. The records that do identify the adapter — the `TOKEN_AUTHORIZE_FLOW: initiated` line and the span enter/close — carry no `request_id`, since the log formatter serializes only the innermost span (`ctx.lookup_current()`, no parent walk). So a given `request_id` can no longer be attributed to `TokenAuthorize` rather than `Authorize`. Previously it could, but only alongside the duplicate event this PR removes.

Both gaps have the same root cause: `grpc_logging_wrapper` is invoked inside each handler body rather than as a tower layer keyed on the gRPC method path. A layer would emit exactly one event per inbound RPC, tagged with the method it actually arrived on, and neither the double-emit nor the lost attribution could occur. That is the right end state and a much larger change than this PR.

## Motivation and Context

The error stack never reached `grpc_logging_wrapper`. Handlers converted to `tonic::Status` themselves, so by the time the wrapper saw the result the frames were gone and only a flat status remained.

Errors *were* being logged, inside the individual `IntoGrpcStatus` impls — but for connector flows the report being logged had already been rebuilt. `IntoGrpcStatus for Report<ConnectorFlowError>` delegated with `Report::new(e.clone())`, which constructs a fresh report from a cloned leaf context and discards every frame accumulated across the flow. #1667 then removed the wrapper's own `logger::error!` to avoid double-logging — correctly identifying the duplicate, but leaving the rebuilt, single-frame report as the only thing logged.

So the alert line exists and carries an `error_code`, and the `extra_error` field it keys on has no chain behind it. Wiring `extra_error` to the inner mappers would have pointed the alert at that same rebuilt report.

Converting at the wrapper fixes the cause rather than the symptom: the handler hands over a `Report`, the wrapper logs it once with its frames intact, and then produces the status.

The typed boundary is what keeps it that way. Roughly 80 hand-built `tonic::Status` sites bypassed the mappers entirely and logged nothing at all. A lint would rot; making them fail to compile does not.

### Additional Changes

`NotImplemented` / `NotSupported` / `FlowNotSupported` now map to `unimplemented` instead of `failed_precondition`, matching the gRPC definition of the code: *"not implemented, or not supported/enabled in this service."* Configuration problems (`InvalidConnectorConfig`, `ConfigurationError`, `NoConnectorMetaData`) remain on `failed_precondition` — *"the system is not in a state required for the operation."*

Callers currently branching on `failed_precondition` for unsupported flows or payment methods will see `unimplemented`.

## How did you test it?

- `cargo check --workspace --all-targets` — clean
- `cargo test --workspace --lib --bins` — no failures
- `cargo clippy -p ucs_env -p grpc-server` — clean
- **Boundary verified by injection**: adding `return Err(tonic::Status::internal("..."))` inside a handler body fails to compile (`the trait From<GrpcError> is not implemented for tonic::Status`), and the error goes away when it is removed.

Still to confirm on a deployed environment: that `extra_error` carries the full frame chain on a real connector-flow failure.


XyneSpaces:
:warning: Redundant structured-log field added to every error mapper

`extra_error = ?self` duplicates the existing `error = ?self` field in the `logger::error!` calls in `crates/common/ucs_env/src/error.rs`. Logging the same report twice under different keys doubles log volume and can break alert rules / dashboards keyed on `error`. Remove `extra_error` and keep the original `error` field.

:bulb: `#[allow(unused_imports)]` for `IntoGrpcStatus` inside generated handler arms

Importing `ucs_env::error::IntoGrpcStatus` with `#[allow(unused_imports)]` inside each expanded gRPC handler arm in `crates/grpc-server/grpc-server/src/utils.rs` masks whether the trait is actually used. If the macro always calls `.into_grpc_status()` the import is needed; otherwise it is dead code. Bring the trait into scope at the module level (or emit the import only where required) and drop the `allow`.

XyneSpaces:
Review findings:

[should-fix] `crates/common/ucs_env/src/error.rs` – Moving `tonic::Status` conversion into the logging wrapper seems to drop the connector-specific `http_status_code` field that the previous `IntoGrpcStatus for Report<ConnectorError>` implementation logged. Please verify that field is still emitted somewhere in the new path before merging.

[should-fix] `crates/common/ucs_env/src/error.rs` – Several `IntegrationError` variants (`NotImplemented`, `NotSupported`, `FlowNotSupported`, `CaptureMethodNotSupported`) now map to `Unimplemented` instead of the previous `FailedPrecondition`. Because this is a client-visible gRPC contract change, add a code comment documenting the new mapping so it is not regressed later.

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -317,8 +317,13 @@ impl EventService for EventServiceImpl {
                     let event_type_enum = grpc_api_types::payments::NotifyEventType::try_from(
                         request_data.payload.event_type,
                     )
-                    .map_err(|error| {
-                        tonic::Status::invalid_argument(format!("Invalid event type: {}", error))
+                    .map_err(|_| {
```
Shubhodip900: use the actual error as well

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -317,8 +317,13 @@ impl EventService for EventServiceImpl {
...
+                        error_stack::Report::new(ucs_env::error::GrpcError::from(
+                            domain_types::errors::IntegrationError::InvalidDataFormat {
+                                field_name: "event_type",
+                                context: domain_types::errors::IntegrationErrorContext::default(),
```
Shubhodip900: don't use default

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -363,10 +368,15 @@ impl EventService for EventServiceImpl {
                             )
                             .await
                         }
-                        other_event_type => Err(tonic::Status::invalid_argument(format!(
-                            "Unsupported event type: {}",
-                            other_event_type.as_str_name()
-                        ))),
+                        _other_event_type => {
```
Shubhodip900: why `_` ?

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -390,7 +403,13 @@ impl EventServiceImpl {
         let connector_data: SurchargeConnectorData = ConnectorDataProvider::from_connector_variant(
             &metadata_payload.connector,
         )
-        .ok_or_else(|| tonic::Status::unimplemented("Invalid connector type for this flow"))?;
+        .ok_or_else(|| {
+            ucs_env::error::GrpcError::from(domain_types::errors::IntegrationError::NotSupported {
+                message: "Invalid connector type for this flow".to_string(),
+                connector: "N/A",
+                context: domain_types::errors::IntegrationErrorContext::default(),
```
Shubhodip900: don't use default

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -480,7 +502,13 @@ impl EventServiceImpl {
...
```
Shubhodip900: +1

crates/grpc-server/grpc-server/src/server/events.rs
```diff
@@ -480,7 +502,13 @@ impl EventServiceImpl {
         let connector_data: SurchargeConnectorData = ConnectorDataProvider::from_connector_variant(
             &metadata_payload.connector,
         )
-        .ok_or_else(|| tonic::Status::unimplemented("Invalid connector type for this flow"))?;
+        .ok_or_else(|| {
+            ucs_env::error::GrpcError::from(domain_types::errors::IntegrationError::NotSupported {
+                message: "Invalid connector type for this flow".to_string(),
+                connector: "N/A",
```
Shubhodip900: Can we bring the connector name in this scope ?

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -981,7 +1023,7 @@ impl PaymentService for Payments {
                     let payload = request_data.payload;
                     let connector_data: ConnectorData<DefaultPCIHolder> =
                         ConnectorData::from_connector_variant(&connector)
-                        .ok_or_else(|| tonic::Status::invalid_argument("Invalid Connector Received"))?;
+                        .ok_or_else(|| ucs_env::error::GrpcError::from(IntegrationError::InvalidDataFormat { field_name: "connector", context: domain_types::errors::IntegrationErrorContext::default() }))?;
```
AmitsinghTanwar007: should we populate the context here as well

crates/common/ucs_env/src/error.rs
```diff
@@ -127,97 +212,148 @@ impl IntoGrpcStatus for error_stack::Report<ConnectorError> {
...
+        self.to_grpc_error().into_grpc_status()
+    }
+}
+
+impl IntoGrpcStatus for Report<ConnectorError> {
+    fn into_grpc_status(self) -> Status {
+        logger::error!(
+            error = ?self,
+            error_code = %self.current_context().error_code(),
+            http_status_code = ?self.current_context().http_status_code(),
```
AmitsinghTanwar007: **Bug: `http_status_code` will be lost in the main logging path**

This `Report<ConnectorError>` impl logs `http_status_code`, but the **main path** through the wrapper goes:

`handler` → `to_grpc_error()` → `GrpcError::Connector` → `Report<GrpcError>::into_grpc_status()`

The central `Report<GrpcError>::into_grpc_status()` does NOT log `http_status_code` — it only logs `error` and `error_code`. So any `ConnectorError` going through the intended wrapper path will **lose this field** from logs.

Since the whole point of this PR is to route errors through the wrapper, most `ConnectorError`s will take the `GrpcError` path and miss this field, which could impact debugging and alerting.

**Suggestion:** Add `http_status_code` to the `GrpcError::Connector` branch in the central `Report<GrpcError>::into_grpc_status()` impl.

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -1319,19 +1359,19 @@ impl PaymentService for Payments {
                         .request_details
                         .map(domain_types::connector_types::RequestDetails::foreign_try_from)
                         .transpose()
-                        .map_err(|e| e.into_grpc_status())?
-                        .ok_or(tonic::Status::invalid_argument("missing request_details in the payload"))?;
+                        .map_err(|e| e.to_grpc_error())?
+                        .ok_or(ucs_env::error::GrpcError::from(IntegrationError::MissingRequiredField { field_name: "payment_method", context: domain_types::errors::IntegrationErrorContext::default() }))?;
```
AmitsinghTanwar007: **Bug: Wrong `field_name` — copy-paste error**

This `.ok_or(...)` guards `.request_details`, but the error says `field_name: "payment_method"`. The old code correctly said `"missing request_details in the payload"`.

Should be:
```rust
field_name: "request_details"
```

AmitsinghTanwar007 (approved):

Shubhodip900 (approved):

JeevaRamu0104:
### Suggestion: hold the "no hand-built status" invariant at the RPC-method layer too

The typed boundary already mandates the pattern where it can reach: handlers return `Report<GrpcError>`, there is no `From<tonic::Status> for GrpcError`, and `ToGrpcStatus for GrpcError` is an exhaustive match — so a new leaf error can't be added without wiring both its `#[from]` variant and its status mapping, and a hand-built status inside a handler body won't compile. That part is airtight.

The layer types don't reach yet is the **top-level RPC method**, which still returns `Result<_, tonic::Status>` and runs `get_config_from_request(&request).into_grpc_status()?` before the wrapper. Nothing stops a future author from adding `Err(tonic::Status::internal(...))` there — `crates/grpc-server/grpc-server/src/server/payouts.rs:188` (the `eligibility` stub) already does exactly this. This is the same root cause as the known limitation already called out in the PR description (wrapper invoked inside each handler rather than as a tower layer).

Two options to keep the invariant enforced for that residual layer:

1. **Durable fix (the documented end state):** make `grpc_logging_wrapper` a tower `Layer` keyed on the gRPC method path. Then the RPC methods themselves return `Report<GrpcError>`, the layer does the single conversion, and *no* application code has `tonic::Status` in a return type — the type system covers config extraction and stubs too. Larger change; reasonable to keep out of this PR.

2. **Cheap stopgap until then:** a clippy `disallowed-methods` rule banning the `tonic::Status::*` constructors repo-wide, with a single `#[allow]` in `error.rs`:
   ```toml
   # clippy.toml
   disallowed-methods = [
       { path = "tonic::Status::internal",         reason = "return Report<GrpcError>; the wrapper converts and logs" },
       { path = "tonic::Status::invalid_argument", reason = "return Report<GrpcError>; the wrapper converts and logs" },
       { path = "tonic::Status::unimplemented",    reason = "return Report<GrpcError>; the wrapper converts and logs" },
       # ...remaining constructors
   ]
   ```
   or a CI grep guard failing on `tonic::Status::` outside `error.rs`.

Not blocking — the handler layer is already enforced by construction. This is only about the un-typed RPC-method layer, which the tower-layer refactor would close for good.

JeevaRamu0104 (approved):

PR #1910, opened 2026-07-11
feat(connector): implement SetupMandate for tesouro

shuklatushar226:
## Summary

Implement **SetupMandate** flow for **Tesouro** connector.

Tesouro is a brand-new GraphQL connector implementing the full cards-recurring lifecycle. Both mandatory legs of the recurring card flow (SetupMandate `verifyAccount` and the recurring MIT `authorizeRecurring`) PASS with status `CHARGED` / `statusCode 200`. Authentication uses an OAuth2 `client_credentials` prerequisite (`CreateServerAuthenticationToken`) that mints a short-lived Bearer token shared by all payment flows.

## Changes

- Added SetupMandate support to `tesouro.rs` (Bearer-token header builder, flow wiring, `macro_connector_implementation!`)
- Added SetupMandate request/response types and `TryFrom` implementations in `tesouro/transformers.rs`
- Registered and wired the connector across the connector-integration and domain-types crates, proto, field-probe auth, and config

## Files Modified

- `crates/integrations/connector-integration/src/connectors/tesouro.rs` (new)
- `crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs` (new)
- `crates/types-traits/grpc-api-types/proto/payment.proto`
- `config/development.toml`
- `config/sandbox.toml`
- `config/production.toml`

## gRPC Test Results

**Status: PASS**

<details>
<summary>grpcurl SetupMandate + recurring MIT calls (credentials redacted)</summary>

```
Common headers (SignatureKey creds):
  -H 'x-connector: tesouro' -H 'x-auth: signature-key'
  -H 'x-api-key: <REDACTED>' -H 'x-key1: <REDACTED>' -H 'x-api-secret: <REDACTED>'

--- Call 1: CreateServerAuthenticationToken (OAuth2 client_credentials prerequisite) ---
grpcurl -plaintext <headers> -d '{"connector":"TESOURO","test_mode":true}' \
  localhost:19300 types.MerchantAuthenticationService/CreateServerAuthenticationToken
Response: { "accessToken": {"value":"<REDACTED>"}, "tokenType":"Bearer",
  "expiresInSeconds":"3600", "status":"OPERATION_STATUS_SUCCESS", "statusCode":200 }

--- Call 2: SetupMandate -> verifyAccount (types.PaymentService/SetupRecurring) ---
grpcurl -plaintext <headers> -d '{
  "merchant_recurring_payment_id":"tsr_setup_001",
  "amount":{"minor_amount":0,"currency":"USD"},
  "payment_method":{"card":{"card_number":{"value":"<REDACTED>"},"card_exp_month":{"value":"12"},
    "card_exp_year":{"value":"2030"},"card_cvc":{"value":"<REDACTED>"},"card_holder_name":{"value":"John Doe"}}},
  "address":{"billing_address":{"first_name":{"value":"John"},"last_name":{"value":"Doe"},
    "line1":{"value":"123 Test St"},"city":{"value":"Miami"},"state":{"value":"FL"},
    "zip_code":{"value":"33101"},"country_alpha2_code":"US"}},
  "auth_type":"NO_THREE_DS","enrolled_for_3ds":false,"request_incremental_authorization":false,
  "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer","expires_in_seconds":3600}}
}' localhost:19300 types.PaymentService/SetupRecurring
Response:
{
  "connectorRecurringPaymentId": "15236c07-7d83-11f1-902e-a60714a9a47d",
  "status": "CHARGED", "statusCode": 200,
  "mandateReference": { "connectorMandateId": {
    "connectorMandateId": "be4ca421-7a07-4d23-9215-01fa9d427fc5",
    "connectorMandateRequestReferenceId": "15236c07-7d83-11f1-902e-a60714a9a47d",
    "mandateMetadata": {"value": "{\"activityDate\":\"2026-07-11\"}"} } },
  "connectorFeatureData": {"value": "{\"paymentId\":\"15236c07-7d83-11f1-902e-a60714a9a47d\"}"}
}

--- Call 3: Recurring MIT Authorize -> authorizeRecurring (types.RecurringPaymentService/Charge) ---
Uses stored connector_mandate_id from Call 2.
grpcurl -plaintext <headers> -d '{
  "merchant_charge_id":"tsr_mit_001",
  "connector_recurring_payment_id":{"connector_mandate_id":{
    "connector_mandate_id":"be4ca421-7a07-4d23-9215-01fa9d427fc5",
    "connector_mandate_request_reference_id":"15236c07-7d83-11f1-902e-a60714a9a47d",
    "mandate_metadata":{"value":"{\"activityDate\":\"2026-07-11\"}"}}},
  "amount":{"minor_amount":1000,"currency":"USD"},
  "payment_method":{"card":{"card_number":{"value":"<REDACTED>"},"card_exp_month":{"value":"12"},
    "card_exp_year":{"value":"2030"},"card_cvc":{"value":"<REDACTED>"},"card_holder_name":{"value":"John Doe"}}},
  "capture_method":"AUTOMATIC",
  "address":{"billing_address":{...same as above...}},
  "state":{"access_token":{"token":{"value":"<REDACTED>"},"token_type":"Bearer","expires_in_seconds":3600}}
}' localhost:19300 types.RecurringPaymentService/Charge
Response:
{
  "connectorTransactionId": "2c64636d-7d83-11f1-b82e-f0c25daeb4e2",
  "status": "CHARGED", "statusCode": 200,
  "connectorFeatureData": {"value": "{\"paymentId\":\"2c5d7bee-7d83-11f1-8682-d27433274265\"}"}
}
```

</details>

## Validation Checklist

- [x] grpcurl SetupMandate (`SetupRecurring`) returned success status (`CHARGED`, 200)
- [x] Only connector-specific / connector-enabling files modified

---

## Wallet recurring validation (Google Pay + Apple Pay)

The original description validated the CARD flow only. The wallet flows (Google Pay / Apple Pay, added in this PR) have now been validated end-to-end through **HS → UCS → tesouro**, CIT (`SetupRecurring`, `off_session`) → MIT (`RecurringPaymentService/Charge`), both **CHARGED / 200**:

| Wallet | CIT | MIT |
|---|---|---|
| Google Pay | `succeeded` | `succeeded` |
| Apple Pay  | `succeeded` | `succeeded` |

Notes:
- Tesouro consumes **decrypted** network-token pass-through data (it rejects the `Encrypted` wallet variant), so the wallet flows require a **cryptogram-bearing (`CRYPTOGRAM_3DS`) predecrypt token**. Validated by driving Hyperswitch's own wallet decryption (Google Pay `DIRECT`/ECv2 and Apple Pay EC_v1) with a controlled key, then forwarding the decrypted payload to tesouro. This exercises the HS decryption path that the upstream reference PR (`juspay/hyperswitch#9706`) noted could not be tested without gateway credentials.
- Includes a fix so **wallet mandate MITs** recover `expirationMonth`/`expirationYear` from the mandate metadata (a `MandatePayment` MIT carries no wallet data and no card in `additional_payment_data`); without it tesouro's `authorizeRecurring` fails with `Field "expirationMonth" of required type "NumericMonth!" was not provided`.


XyneSpaces:
:rotating_light: **Production config routes Tesouro to the sandbox endpoint**

`config/production.toml` adds `tesouro.base_url = "https://api.sandbox.tesouro.com"`, while the surrounding production entries use live hosts (e.g. `https://api.kount.com`, `https://api.hyperswitch.io`). With this value, production traffic will hit Tesouro’s sandbox environment.

Update the production file to the live Tesouro base URL; keep `development.toml` and `sandbox.toml` on the sandbox endpoint.

---

:warning: **New connector flows have no automated tests or specs**

The PR introduces ~2k lines of new connector code (SetupMandate, Authorize, RepeatPayment, PSync), but only a manual `grpcurl` transcript is included. Please add unit/integration tests for amount/status/error mapping and a connector spec under `crates/internal/ucs-connector-tests/src/connector_specs/tesouro/` so regressions are caught in CI.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+        }
+        errors {
+            ... on InternalServiceError { message transactionId processorResponseCode }
+            ... on AcceptorNotFoundError { message transactionId processorResponseCode }
+            ... on RuleInViolationError { message transactionId processorResponseCode }
+            ... on SyntaxOnNetworkResponseError { message transactionId processorResponseCode }
+            ... on TimeoutOnNetworkResponseError { message transactionId processorResponseCode }
+            ... on ValidationFailureError { message processorResponseCode transactionId }
+            ... on UnknownCardError { message processorResponseCode transactionId }
+            ... on TokenNotFoundError { message processorResponseCode transactionId }
```
iemyashasvi: what are these ?

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+            ... on InvalidTokenError { message processorResponseCode transactionId }
+            ... on RouteNotFoundError { message processorResponseCode transactionId }
+        }
+    }
+}";
+
+pub const AUTHORIZE_TRANSACTION: &str = "mutation AuthorizeCustomerInitiatedTransaction(
```
iemyashasvi: +1

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+            ... on ValidationFailureError { message processorResponseCode transactionId }
+            ... on UnknownCardError { message processorResponseCode transactionId }
+            ... on TokenNotFoundError { message processorResponseCode transactionId }
+            ... on InvalidTokenError { message processorResponseCode transactionId }
+            ... on RouteNotFoundError { message processorResponseCode transactionId }
+        }
+    }
+}";
+
+pub const AUTHORIZE_RECURRING: &str = "mutation AuthorizeRecurring(
```
iemyashasvi: +1

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+
+fn current_date() -> String {
+    let now = time::OffsetDateTime::now_utc();
+    format!(
+        "{:04}-{:02}-{:02}",
+        now.year(),
+        u8::from(now.month()),
+        now.day()
+    )
+}
```
iemyashasvi: refactor this

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+
+fn get_card_payment_method<T: PaymentMethodDataTypes>(
+    card: &Card<T>,
+    is_mandate_payment: bool,
+) -> Result<TesouroPaymentMethodDetails<T>, Error> {
+    let card_data = TesouroCardWithPanDetails {
+        expiration_month: card.get_card_expiry_month_2_digit().change_context(
+            IntegrationError::RequestEncodingFailed {
+                context: Default::default(),
```
iemyashasvi, line 296: add more details in context

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+            },
```
iemyashasvi: also use attach_printable

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+        )?,
+        expiration_year: card.get_expiry_year_4_digit(),
+        account_number: card.card_number.clone(),
+        payment_entry_mode: if is_mandate_payment {
+            TesouroPaymentEntryMode::PaymentMethodOnFile
+        } else {
+            TesouroPaymentEntryMode::PaymentMethodNotOnFile
+        },
```
iemyashasvi, line 307: can we use match

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+                MerchantAuthenticationFlowData,
+                ServerAuthenticationTokenRequestData,
+                ServerAuthenticationTokenResponseData,
+            >,
+            T,
+        >,
+    ) -> Result<Self, Self::Error> {
+        let auth = TesouroAuthType::try_from(&item.router_data.connector_config)?;
+        Ok(Self {
+            grant_type: "client_credentials".to_string(),
```
iemyashasvi, line 428: avoid magic strings

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+                {
+                    return Err(IntegrationError::NotSupported {
+                        message: "Cards 3DS".to_string(),
+                        connector: "Tesouro",
+                        context: Default::default(),
+                    }
+                    .into());
+                }
+                get_card_payment_method(card, is_mandate)?
+            }
```
iemyashasvi: nit : can we make this cleaner

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+                        automatic_capture: if auto_capture {
+                            TesouroAutomaticCapture::OnApproval
+                        } else {
+                            TesouroAutomaticCapture::Never
+                        },
+                        authorization_intent: if auto_capture {
+                            TesouroAuthorizationIntent::FinalAuthorization
+                        } else {
+                            TesouroAuthorizationIntent::PreAuthorization
+                        },
```
iemyashasvi, line 673: nit : can we aoid if else

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+                        metadata.clone().expose(),
+                    )
+                    .change_context(IntegrationError::RequestEncodingFailed {
+                        context: Default::default(),
+                    })?;
+                    Some(tesouro_metadata.activity_date)
+                }
+                None => Some(current_date()),
+            },
+            None => Some(current_date()),
```
iemyashasvi: why are we falling back to current , if this is needed and critical throw err

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+                    AttemptStatus::AuthorizationFailed
+                };
+                let message = auth_response
+                    .message
+                    .clone()
+                    .unwrap_or_else(|| NO_ERROR_MESSAGE.to_string());
+                Ok((
+                    status,
+                    Err(ErrorResponse {
+                        code: NO_ERROR_CODE.to_string(),
```
iemyashasvi, line 1512: why no eerr code ?

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+        let status = if auto_capture {
+            AttemptStatus::Failure
+        } else {
+            AttemptStatus::AuthorizationFailed
+        };
+        Ok((status, in_band_error(&errors, http_code, status)))
+    } else {
+        Err(ConnectorError::unexpected_response_error(http_code).into())
+    }
+}
```
iemyashasvi, line 1545: can we make a tryfrom for status mapping

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1617 @@
...
+            TesouroApiResponse::Error(error_response) => {
+                let status = if auto_capture {
+                    AttemptStatus::Failure
+                } else {
+                    AttemptStatus::AuthorizationFailed
+                };
+                (
+                    status,
+                    top_level_error(&error_response, item.http_code, status),
+                )
```
iemyashasvi, line 1570: +1

crates/integrations/connector-integration/src/connectors/tesouro.rs
```diff
@@ -0,0 +1,425 @@
...
+        PreAuthenticate,
+        PaymentMethodToken,
+        VoidPC,
+        Void,
+        RSync,
+        Refund,
+        ServerSessionAuthenticationToken,
+        SubmitEvidence
+    ],
+    not_supported: [VoidPostRefund],
```
iemyashasvi: can we classify this ?

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+            received_length: id.len(),
+            context: Default::default(),
+        }
+        .into())
+    }
+}
+
+fn is_auto_capture(capture_method: Option<CaptureMethod>) -> bool {
+    matches!(capture_method, Some(CaptureMethod::Automatic) | None)
+}
```
JeevaRamu0104, line 274: **[S1] `SequentialAutomatic` becomes an uncapturable pre-authorization.**

This returns `true` only for `Some(Automatic) | None`, so `SequentialAutomatic` falls to the `else` branches at lines 635-644 and emits `automaticCapture: NEVER` + `authorizationIntent: PRE_AUTHORIZATION`.

Every other connector in this repo groups it with `Automatic`:

- `nexinets.rs:326`
- `noon.rs:264`
- `worldpay/transformers.rs:388`
- `authorizedotnet/transformers.rs:324`
- `fiuu/transformers.rs:92`
- `nexixpay/transformers.rs:409`

A merchant configured for sequential-automatic capture gets a hold that is never settled, and with `Capture` unimplemented there is no way to settle it afterwards.

```rust
matches!(
    capture_method,
    Some(CaptureMethod::Automatic) | Some(CaptureMethod::SequentialAutomatic) | None
)
```

crates/integrations/connector-integration/src/connectors/tesouro.rs
```diff
@@ -0,0 +1,431 @@
...
+// ===== Remaining flows: not implemented =====
+crate::connectors::macros::macro_connector_flow_status_impls!(
+    connector: Tesouro,
+    generic_type: T,
+    [PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize],
+    not_implemented: [
+        // Payment lifecycle beyond auth/sale (Tesouro auto-captures via `automaticCapture`)
+        Capture,
+        Void,
+        VoidPC,
```
JeevaRamu0104, line 407: **[S1] Manual capture is accepted, but there is no way to settle or release the resulting authorization.**

`Capture` and `Void` are both `not_implemented` here, while `transformers.rs:635-644` actively branches on the capture method and emits `automaticCapture: NEVER` + `authorizationIntent: PRE_AUTHORIZATION` for `Manual` / `ManualMultiple` / `Scheduled`. The outcome is funds held on the cardholder's card with no settle path and no release path through this connector.

`ValidationTrait` has no capture-method hook, so the connector has to reject this itself. Precedent in `xendit/transformers.rs:262-268`:

```rust
match data.capture_method {
    Some(CaptureMethod::Automatic) | Some(CaptureMethod::SequentialAutomatic) | None => Ok(true),
    Some(_) => Err(IntegrationError::CaptureMethodNotSupported { .. }),
}
```

Either gate the request that way, or implement `Capture`/`Void`.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+        )?;
+        let bill_to_address = BillToAddress::from_payment_flow(&router_data.resource_common_data);
+        let payment_method_details = match &router_data.request.payment_method_data {
+            PaymentMethodData::Card(card) => get_card_payment_method(card, true)?,
+            PaymentMethodData::Wallet(WalletData::ApplePay(apple_pay)) => {
+                get_apple_pay_payment_method(apple_pay)?
+            }
+            PaymentMethodData::Wallet(WalletData::GooglePay(google_pay)) => {
+                get_google_pay_payment_method(google_pay)?
+            }
```
JeevaRamu0104, line 536: **[S1] 3DS guard is present on Authorize but missing here.**

The Authorize card arm (lines 588-601) rejects `resource_common_data.auth_type == AuthenticationType::ThreeDs` with `IntegrationError::NotSupported`. SetupMandate reads the same `PaymentFlowData` and has no such check.

A `SetupRecurring` carrying `auth_type: THREE_DS` — a field the grpcurl payload in the PR description sets explicitly — is therefore sent as a plain non-authenticated `verifyAccount`. The merchant ends up with a stored credential established without the authentication they requested, and no error signalling the downgrade.

Simplest fix is to lift the check into `get_card_payment_method` (it already takes the flags it needs) so both flows inherit it and the two paths can't drift again.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+        .map(|token| token.expose());
+    connector_mandate_id.map(|connector_mandate_id| {
+        let (expiration_month, expiration_year) = credential_expiry;
+        let mandate_metadata = activity_date.map(|activity_date| {
+            common_utils::pii::SecretSerdeValue::new(serde_json::json!(TesouroMandateMetadata {
+                activity_date,
+                expiration_month,
+                expiration_year,
+            }))
+        });
```
JeevaRamu0104, line 1212: **[S2] Wallet-mandate expiry persistence is coupled to `activityDate` being present.**

`mandate_metadata` is built only inside `activity_date.map(...)`, so when `activity_date` is `None` the credential expiry added by a5993b9 is silently dropped along with it.

The SetupMandate path is safe — `VerifyAccountResponseType::activity_date` is a required `String`. The Authorize CIT path is not: `AuthorizationResponseData::activity_date` is `Option<String>` (line 1285). If Tesouro omits `activityDate` on an approval that stores a wallet mandate, the subsequent MIT hits exactly the `Field "expirationMonth" of required type "NumericMonth!" was not provided` failure this PR set out to fix.

Build the metadata when *any* of the three values is present rather than gating the whole object on `activity_date`.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+
+        // Tesouro requires `originalPurchaseDate` (Date!) on a recurring charge. Prefer the CIT's
+        // `activityDate` captured in the mandate metadata during the SetupMandate/Authorize leg;
+        // when it is absent (e.g. a stored-credential MIT that does not thread mandate metadata)
+        // fall back to today's date so the required field is always present (matches hyperswitch).
+        let original_purchase_date = Some(
+            mandate_metadata
+                .map(|tesouro_metadata| tesouro_metadata.activity_date)
+                .unwrap_or_else(|| common_utils::date_time::now().date().to_string()),
+        );
```
JeevaRamu0104, line 935: **[S2] `originalPurchaseDate` fabricates today's date.**

When mandate metadata is absent, the MIT reports `now().date()` as the original purchase date of a stored credential that by definition was established earlier. That is inaccurate stored-credential data submitted to the network.

This was removed in 074a705 ("drop the `current_date()` fallback ... instead of fabricating today") and reinstated in 4c0011b for parity with the hyperswitch reference. Flagging so the trade-off is a recorded decision rather than a silent revert — if the reference genuinely does this and Tesouro requires the field, that's a defensible answer, it just deserves to be stated.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+    ApprovedReversal,
+    DeclinedAuthorization,
+    DeclinedCapture,
+    DeclinedReversal,
+    GenericPaymentTransaction,
+    Authorization,
+    Capture,
+    Reversal,
+    Sale,
+}
```
JeevaRamu0104, line 1664: **[S2] Closed `__typename` enum with no fallback turns an unmapped status into an undiagnosable error.**

`TesouroSyncStatus` is non-optional and covers exactly the 12 union members listed in `tesouro_queries.rs:115-135`. Any other `__typename` Tesouro returns (a refund or declined-sale member, say) fails `TesouroSyncData`, falls through the `#[serde(untagged)]` `TesouroApiResponse` to the `Error` variant, fails there too (no `errors` key), and surfaces as serde's opaque *"data did not match any variant of untagged enum TesouroApiResponse"*.

So a perfectly successful payment would sync as a deserialization failure with nothing in the message pointing at the cause.

```rust
#[serde(other)]
Unknown,
```

and mapping `Unknown => previous_status` would fail soft. Separately, worth considering an explicit tag over `untagged` on `TesouroApiResponse` so the failing variant is identifiable at all.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+            {
+                if expiration_month.is_none() {
+                    expiration_month = card_info.card_exp_month.clone();
+                }
+                if expiration_year.is_none() {
+                    expiration_year = card_info
+                        .card_exp_year
+                        .as_ref()
+                        .map(domain_types::utils::expand_expiry_year_to_four_digits);
+                }
```
JeevaRamu0104, line 890: **[S2] Stored-card expiry month isn't normalized to 2 digits.**

The year immediately below is normalized via `expand_expiry_year_to_four_digits` (added in 6886155 for exactly this reason), but the month is taken verbatim. The live-card path at line 758 uses `get_card_expiry_month_2_digit()`.

A stored `"7"` therefore reaches Tesouro's `NumericMonth!` unpadded on the mandate-MIT fallback path.

JeevaRamu0104 (commented):
Review of the Tesouro connector integration.

**Risk: high** — three findings I'd treat as blocking, six non-blocking, posted inline.

Registry wiring is complete and consistent: `connectors.rs`, dispatch in `types.rs`, `default_implementations.rs`, `ConnectorEnum`, `router_data.rs` auth mapping, `Connectors` struct, `field-probe/auth.rs`, all three `config/*.toml`, `connector_specs/tesouro/specs.json`, docs and the generated JS client are all present. Proto numbering is clean after the two rebases (`TESOURO = 132`, oneof `139`, no field reuse). Auth handling is sound — `Secret` throughout, `into_masked()` on the Bearer header, no credentials in source or config.

The blocking three:

1. **Capture-method handling** — `SequentialAutomatic` is classified as non-auto-capture, unlike every other connector in this repo, and `Manual`/`ManualMultiple`/`Scheduled` are accepted even though `Capture` and `Void` are `not_implemented`. Both end in an authorization that can neither be settled nor released.
2. **3DS guard asymmetry** — Authorize rejects `THREE_DS` for cards; SetupMandate silently proceeds with a non-authenticated `verifyAccount`.
3. See inline.

Two notes on rubric application, so they don't read as unflagged omissions:

- `context: Default::default()` in `change_context` appears 166 times across the 101 connector files already on `main`. It's house convention rather than a regression here, so I haven't counted it against this PR — though `unexpected_response_error_with_context` would give the two response-transform failures real diagnostics if you want to start tightening it.
- No `test.rs` matches the repo norm (6 of 101 connectors have one) and `specs.json` is present, so that isn't held as a blocker either.

Leaving this as a comment rather than a formal request-changes; say the word and I'll convert it.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+/// Persisted into the mandate metadata so a later MIT can populate Tesouro's required
+/// `expirationMonth`/`expirationYear` even when the MIT request itself has no credential data.
+fn get_credential_expiry<T: PaymentMethodDataTypes>(
+    payment_method_data: &PaymentMethodData<T>,
+) -> (Option<Secret<String>>, Option<Secret<String>>) {
+    match payment_method_data {
+        PaymentMethodData::Card(card) => (
+            card.get_card_expiry_month_2_digit().ok(),
+            Some(card.get_expiry_year_4_digit()),
+        ),
```
JeevaRamu0104, line 1125: **[S2] `.ok()` swallows the month-derivation error.**

A malformed expiry silently yields `None` here, which stores a mandate without expiry and defers the failure to a later MIT rejected by Tesouro's schema — far from the actual cause. The equivalent call at line 758 propagates with `.attach_printable(...)`.

At minimum log it; better, propagate, since a card whose expiry month can't be derived isn't a card you want to set up a mandate against.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+// =============================================================================
+
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TesouroAuthorizeData {
+    #[serde(
+        alias = "authorizeRecurring",
+        alias = "authorizeCustomerInitiatedTransaction"
+    )]
+    authorize_customer_initiated_transaction: AuthorizationInnerResponse,
```
JeevaRamu0104: **[S3]** `alias = "authorizeCustomerInitiatedTransaction"` is dead — `rename_all = "camelCase"` already produces that exact name for this field. Only the `authorizeRecurring` alias is doing work.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1614 @@
...
+    /// Expiry of the credential backing the acquirer token, captured at CIT/SetupMandate time.
+    /// Tesouro's `authorizeRecurring` requires `expirationMonth`/`expirationYear`, and a
+    /// wallet-backed MIT cannot recover them from `additional_payment_data` (card-only), so the
+    /// mandate metadata is the only carrier. Optional so metadata stored before this field
+    /// existed still deserializes.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub expiration_month: Option<Secret<String>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub expiration_year: Option<Secret<String>>,
+}
```
JeevaRamu0104, line 731: **[S3]** Worth confirming this is intended: adding the expiry here means card expiry month/year now round-trip back to the API caller inside `mandateReference.mandateMetadata` (the PR description's sample response shows `mandateMetadata` returned verbatim to the client, previously just `{"activityDate":...}`).

Low risk given the caller supplied the credential in the first place, but it is a widening of what leaves the service and it's stored on the mandate.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 5090f310b. `is_auto_capture` now groups `SequentialAutomatic` with `Automatic`, matching the connectors you listed. Combined with the gate in the next thread, a sequential-automatic merchant now gets `ON_APPROVAL` + `FINAL_AUTHORIZATION` rather than an unsettleable hold.

crates/integrations/connector-integration/src/connectors/tesouro.rs
shuklatushar226: Fixed in 5090f310b — gated rather than implementing `Capture`/`Void`, which is a larger change than this PR should carry.

Added `is_auto_capture_request` following the xendit precedent and wired it into both the Authorize and RepeatPayment request transformers:

```rust
fn is_auto_capture_request(capture_method: Option<CaptureMethod>) -> Result<bool, Error> {
    match capture_method {
        Some(CaptureMethod::Automatic) | Some(CaptureMethod::SequentialAutomatic) | None => Ok(true),
        Some(_) => Err(IntegrationError::CaptureMethodNotSupported { context: Default::default() }.into()),
    }
}
```

Manual / ManualMultiple / Scheduled now fail at request build with `CaptureMethodNotSupported` instead of raising a pre-auth with no settle path. Kept the `automaticCapture` / `authorizationIntent` branching in place (rather than collapsing it to the constants) so that whenever `Capture`/`Void` do land, lifting the gate is the only change needed. `is_auto_capture` stays non-fallible on the response/PSync side, where the capture method is whatever was recorded on the attempt.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 5090f310b, lifted into `get_card_payment_method` as you suggested — it now takes `auth_type` and rejects `ThreeDs` before building any card details, so Authorize and SetupMandate share one guard and cannot drift again. The inline check in the Authorize arm is gone.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Good catch — fixed in 5090f310b. `TesouroMandateMetadata::activity_date` is now `Option<String>` (with `default` + `skip_serializing_if`, so metadata written before this change still deserializes), and `build_mandate_reference` stores the object when *any* of the three values is present:

```rust
let mandate_metadata = (activity_date.is_some() || expiration_month.is_some() || expiration_year.is_some())
    .then(|| ... );
```

The MIT side follows with `.and_then(|m| m.activity_date)` instead of `.map(...)`, so an approval that omits `activityDate` still carries the expiry forward and the `expirationMonth`/`expirationYear` failure this PR targets cannot come back through that path.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 5090f310b. Added `pad_expiry_month_to_two_digits` in `domain_types::utils`, next to `expand_expiry_year_to_four_digits` and with the same vault-template passthrough behaviour, and applied it to the `additional_payment_data` month so both halves of that fallback are normalised:

```rust
expiration_month = card_info.card_exp_month.as_ref().map(domain_types::utils::pad_expiry_month_to_two_digits);
```

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Addressed in 5090f310b, logged rather than propagated — reasoning, and happy to change it if you disagree.

By the time `get_credential_expiry` runs we are mapping a *response*. Both request transformers already derive the same month via `get_card_payment_method` and propagate on failure, so a card whose month cannot be parsed never reaches the connector, let alone the response path — the `.ok()` is unreachable for `PaymentMethodData::Card`. If it somehow were reachable, returning `Err` here would fail response mapping after Tesouro has already stored the token, orphaning it, which is worse than the deferred failure.

So it now warns with the error attached, and the reasoning is recorded in a comment at the call site. Applied the same treatment to the Google Pay `get_four_digit_expiry_year().ok()` just below, which had the identical silent drop.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 5090f310b:

```rust
    #[serde(other)]
    Unknown,
```

with `Unknown => previous_status` in `map_sync_status`, plus a `tracing::warn!` so an unmapped `__typename` is visible in logs rather than only inferable from a status that did not move. Verified it compiles — serde accepts `other` on this enum since the variants are name-identified.

Left `TesouroApiResponse` untagged for now: the fallback removes the concrete failure mode you described, and switching to an explicit tag touches every flow response in the connector. Worth doing, but I would rather it were its own PR than ride along with SetupMandate — flagging it here so it is not lost.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Removed in 5090f310b — you are right, `rename_all = "camelCase"` already produces it. Left a comment noting that the remaining `authorizeRecurring` alias is what lets the RepeatPayment response share this struct.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Deliberate, and recorded in the code as of 5090f310b rather than left to this thread.

The trade-off: `originalPurchaseDate` is `Date!`, so dropping the fallback turns every metadata-less MIT into a hard rejection. Fabricating today is inaccurate but recoverable; rejecting the charge is not. The hyperswitch reference makes the same call, which is why 4c0011b reinstated it.

The better answer is to keep MITs off the fallback altogether, and the fix in the `build_mandate_reference` thread moves in that direction — metadata is now stored whenever any value is available, so fewer MITs land without an `activityDate`. Expanded the comment at the call site to state the inaccuracy explicitly instead of just citing parity.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Intended, yes. The expiry has to reach the MIT somehow, and for a wallet-backed mandate the metadata is the only carrier — `additional_payment_data` is card-only, so a Google Pay / Apple Pay MIT has no other source for Tesouro's required `expirationMonth`/`expirationYear`.

On the exposure: the values are `Secret<String>` inside a `SecretSerdeValue`, so they are masked in logs, and the caller supplied the credential in the first place, so nothing new is being revealed to them. It is a genuine widening of what is stored on the mandate, but expiry alone is not usable without the PAN or token. Happy to revisit if there is a policy on what may live in `mandateMetadata` that I have missed.

JeevaRamu0104 (dismissed):

swangi-kumari (dismissed):

crates/types-traits/domain_types/src/utils.rs
```diff
@@ -661,6 +661,20 @@ pub fn expand_expiry_year_to_four_digits(year: &Secret<String>) -> Secret<String
     }
 }
 
+/// Pad a 1-digit card expiry month (`"7"`) to 2 digits (`"07"`). Anything that isn't
+/// exactly 1 character passes through unchanged, including vault template tokens like
+/// `{{$card_exp_month}}`. Counterpart to [`expand_expiry_year_to_four_digits`] for
+/// stored-credential paths, where the month arrives as free-form text rather than a
+/// validated `Card`.
+pub fn pad_expiry_month_to_two_digits(month: &Secret<String>) -> Secret<String> {
```
AmitsinghTanwar007: check this once i think we have some padding function for date

shuklatushar226: Right — the primitive is `cards::validate::CardExpirationMonth::two_digits()` (`crates/types-traits/cards/src/validate.rs:478`), which is what `Card::get_card_expiry_month_2_digit` wraps. I had re-implemented the `format!` instead of reusing it.

Fixed in cd2d7ee02 by dropping the helper entirely — `domain_types/src/utils.rs` is now unchanged by this PR — and padding at the call site in `tesouro/transformers.rs` with that primitive:

```rust
expiration_month = card_info.card_exp_month.as_ref().map(|month| {
    month
        .peek()
        .parse::<u8>()
        .ok()
        .and_then(|month| cards::validate::CardExpirationMonth::try_from(month).ok())
        .map_or_else(|| month.clone(), |padded| Secret::new(padded.two_digits()))
});
```

Why not `get_card_expiry_month_2_digit` directly: this path has no `Card`. It is the mandate-MIT fallback, where the month comes from `AdditionalPaymentData::Card(card_info).card_exp_month`, an `Option<Secret<String>>`. It also has to stay infallible — the credential is already stored connector-side by this point, so an unparseable month should degrade rather than fail the charge. Hence the passthrough on `None`, which also preserves vault template tokens like `{{$card_exp_month}}`.

Behaviour is unchanged from the previous commit for every real input (`"7"` -> `"07"`, `"07"`/`"12"` unchanged, out-of-range and non-numeric passed through), it just no longer carries its own padding logic.

AmitsinghTanwar007 (dismissed):

swangi-kumari (dismissed):

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1713 @@
...
+                let status = AttemptStatus::from(AuthOutcome(
+                    &AuthorizeTransactionResponseType::AuthorizationApproval,
+                    auto_capture,
+                ));
+                let mandate_reference = build_mandate_reference(
+                    auth_response.token_details,
+                    auth_response.payment_id.clone(),
+                    auth_response.activity_date,
+                    credential_expiry,
+                );
```
JeevaRamu0104, line 1489: **[S1] RepeatPayment success re-emits a `MandateReference` whose metadata drops the credential expiry, breaking the *next* MIT.**

`map_authorization_response` is shared by Authorize and RepeatPayment and calls `build_mandate_reference(..., credential_expiry)`, where the RepeatPayment caller passes `get_credential_expiry(&item.router_data.request.payment_method_data)` (line 1541). On an MIT the payment method data is `MandatePayment`, which falls to `_ => (None, None)` at line 1136.

The metadata object is still emitted, because `build_mandate_reference` only requires *one* of the three values to be present (lines 1157-1166) and `activity_date` is `Some` on an approval. So the stored `{activityDate, expirationMonth, expirationYear}` is replaced by `{activityDate}`.

The next MIT then finds no expiry in the metadata (lines 892-901) and — for a wallet mandate — none in `additional_payment_data` either (lines 860-891, card-only), so `expirationMonth`/`expirationYear` serialize as absent and `authorizeRecurring` rejects with `Field "expirationMonth" of required type "NumericMonth!" was not provided`. That is precisely the failure commit `a5993b98` was written to fix; it returns on wallet MIT #2. The card path degrades more quietly, surviving only because `additional_payment_data` happens to carry the expiry.

**Suggested fix:** in the RepeatPayment `TryFrom` (line 1528), thread the inbound `connector_mandate_ref.get_mandate_metadata()` expiry into `credential_expiry` instead of deriving it from `payment_method_data`; or pass `None` for `token_details` on RepeatPayment, where the connector is not establishing a new credential.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1713 @@
...
+        .map(|error| error.message.clone())
+        .collect::<Vec<String>>()
+        .join(" ");
+    let message = if message.is_empty() {
+        NO_ERROR_MESSAGE.to_string()
+    } else {
+        message
+    };
+    Err(ErrorResponse {
+        code: NO_ERROR_CODE.to_string(),
```
JeevaRamu0104: **[S2] In-band GraphQL errors always report `NO_ERROR_CODE`.**

`top_level_error` hardcodes `code: NO_ERROR_CODE` because `TesouroApiErrorData` (lines 1040-1043) deserializes only `message`. `build_error_response` in `tesouro.rs:156-165` *does* parse `extensions.code` / `extensions.reason`, but that path only fires on non-2xx HTTP.

GraphQL APIs conventionally return errors with HTTP 200, so the path that actually fires in production is the one that discards the machine-readable code and reason — merchants get an unusable `code` on the common failure path.

**Suggested fix:** add `extensions: Option<TesouroGraphQlErrorExtensions>` to `TesouroApiErrorData` (the type already exists in `tesouro.rs`) and propagate `code`/`reason` through `top_level_error`.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1713 @@
...
+        // The fallback is knowingly inaccurate — a stored credential was by definition established
+        // before today — and is kept deliberately: the field is non-nullable, so the alternative is
+        // a hard rejection of every metadata-less MIT. The hyperswitch reference makes the same
+        // trade-off. The `activityDate` path above is the one that carries a truthful date, and
+        // widening the metadata (see `build_mandate_reference`) keeps MITs on it wherever possible.
+        let original_purchase_date = Some(
+            mandate_metadata
+                .and_then(|tesouro_metadata| tesouro_metadata.activity_date)
+                .unwrap_or_else(|| common_utils::date_time::now().date().to_string()),
+        );
```
JeevaRamu0104, line 935: **[S2] `originalPurchaseDate` silently falls back to today.**

The trade-off is documented here and matches the Hyperswitch reference, so this is not a blocker. But it does send a factually wrong stored-credential date to the network on every metadata-less MIT, which can affect MIT framing and interchange qualification.

Worth a follow-up: consider surfacing a warn-level log on the fallback branch so the rate of metadata-less MITs is observable, rather than the inaccurate date being indistinguishable from a truthful `activityDate`.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
```diff
@@ -0,0 +1,1713 @@
...
+    match capture_method {
+        Some(CaptureMethod::Automatic) | Some(CaptureMethod::SequentialAutomatic) | None => {
+            Ok(true)
+        }
+        Some(_) => Err(IntegrationError::CaptureMethodNotSupported {
+            context: Default::default(),
+        }
+        .into()),
+    }
+}
```
JeevaRamu0104, line 274: **[S3] Two near-identical capture-method predicates.**

`is_auto_capture` (251-256) and `is_auto_capture_request` (263-273) encode the same match. The response-side variant accepts the manual capture methods the request side rejects. Unreachable today — the request transformers gate first — but the pair will drift.

**Suggested fix:** define `is_auto_capture_request` in terms of `is_auto_capture`, e.g. return `Ok(true)` when `is_auto_capture(capture_method)` holds and the error otherwise, so there is a single source of truth for the classification.

JeevaRamu0104 (commented):
Four inline findings on the Tesouro transformers (one S1, two S2, one S3). Severities: S1 = blocking, S2 = non-blocking correctness/maintainability, S3 = nit.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Confirmed and fixed in 3ce174202. Traced it through: `map_authorization_response` is shared, RepeatPayment passed `get_credential_expiry(&...payment_method_data)` (line 1541), MIT payment method data is `MandatePayment` -> `_ => (None, None)` at 1136, and `build_mandate_reference` still emits the object because `activity_date` alone satisfies the `any` check at 1157-1166. So the re-emitted metadata is `{activityDate}` and the expiry is gone.

Took the first of your two suggestions — threading the inbound metadata rather than passing `None` for `token_details` — because Tesouro may legitimately rotate the token on an MIT, and dropping the mandate would discard the new one:

```rust
fn repeat_payment_credential_expiry<T: PaymentMethodDataTypes>(
    request: &RepeatPaymentData<T>,
) -> (Option<Secret<String>>, Option<Secret<String>>) {
    let (stored_month, stored_year) = match &request.mandate_reference {
        MandateReferenceId::ConnectorMandateId(mandate_ref) => mandate_ref
            .get_mandate_metadata()
            .and_then(|metadata| serde_json::from_value::<TesouroMandateMetadata>(metadata.expose()).ok())
            .map_or((None, None), |m| (m.expiration_month, m.expiration_year)),
        _ => (None, None),
    };
    let (request_month, request_year) = get_credential_expiry(&request.payment_method_data);
    (stored_month.or(request_month), stored_year.or(request_year))
}
```

The fallback to `get_credential_expiry` keeps the MIT-with-credential-in-the-clear case working.

One qualifier on reachability, for the record rather than as a defence: the sandbox `authorizeRecurring` response in the PR description returns no `tokenDetails`, so `build_mandate_reference` currently short-circuits to `None` and the corrupted metadata is never emitted in the flow I validated. `token_details` is `Option`, though, so that is a property of the sandbox response and not of the code — the fix stands either way.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 3ce174202. `TesouroApiErrorData` now carries `extensions: Option<TesouroGraphQlErrorExtensions>` (the type already existed at transformers.rs:1079, used by the non-2xx path), and `top_level_error` propagates both fields:

```rust
let extensions = error_response.errors.first().and_then(|error| error.extensions.clone());
// code: extensions.code, else NO_ERROR_CODE
// reason: extensions.reason, else the joined message (previous behaviour)
```

Your framing of which path actually fires is right — GraphQL returns in-band errors on HTTP 200, so `build_error_response` in `tesouro.rs` only runs for transport/auth-level failures. The duplication is the underlying cause: `TesouroApiErrorData` and `TesouroGraphQlError` describe the same wire shape and only the latter modelled `extensions`. They are now field-identical; collapsing them into one type is the obvious follow-up, left out here to keep this change reviewable.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Added the warn in 3ce174202:

```rust
.unwrap_or_else(|| {
    let today = common_utils::date_time::now().date().to_string();
    tracing::warn!(
        original_purchase_date = %today,
        "Tesouro: recurring charge carries no CIT activityDate in its mandate metadata; \
         sending today's date as originalPurchaseDate, which is inaccurate stored-credential data"
    );
    today
})
```

The rate of metadata-less MITs is now observable, which is the thing that was missing — the inaccuracy was documented in the comment but invisible in production. Interchange qualification is the sharper end of it than I had credited; noted.

Related: the RepeatPayment fix in the S1 thread removes one source of these, since a successful MIT no longer re-emits a mandate with the expiry stripped.

crates/integrations/connector-integration/src/connectors/tesouro/transformers.rs
shuklatushar226: Fixed in 3ce174202 — `is_auto_capture_request` now delegates, exactly as you suggested:

```rust
fn is_auto_capture_request(capture_method: Option<CaptureMethod>) -> Result<bool, Error> {
    if is_auto_capture(capture_method) {
        Ok(true)
    } else {
        Err(IntegrationError::CaptureMethodNotSupported { context: Default::default() }.into())
    }
}
```

Single source of truth for the classification, and the request side is now definitionally the response side plus a rejection. The two predicates were introduced in the same commit for different call sites, so the drift risk you describe was real even though it had not materialised.

JeevaRamu0104 (dismissed):

AmitsinghTanwar007 (dismissed):

iemyashasvi (dismissed):

swangi-kumari (dismissed):

swangi-kumari (approved):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #1923, opened 2026-07-12
fix(connector): [NMI] report Unspecified status when psync finds no transaction

10X-GRACE:
## What

NMI's `query.php` returns an **empty transaction list** whenever it has no record of the order. The PSync transformer read that as `AttemptStatus::AuthenticationPending`:

```rust
} else {
    // Empty XML response = AuthenticationPending (during 3DS flow)
    (AttemptStatus::AuthenticationPending, None)
};
```

That guess only suits the 3DS flow — where the attempt is *already* authentication-pending, so it happens to be right. Every other way an order goes unknown to NMI gets the same verdict. The common one, and the one that fired in prod: **an attempt that never reached the connector at all**, so PSync falls back to `order_id = attempt_id` (NMI is the one connector that overrides `validate_psync_reference_id` to always `Ok(())`, so it still calls out with no connector txn id) and NMI has never heard of that order.

An empty response carries **no information** about the attempt, so report `Unspecified` instead of inventing a status. It reaches the router as `PaymentStatus::UNSPECIFIED`, which resolves to the attempt's existing status:

- `hyperswitch_interfaces/src/unified_connector_service/transformers.rs:589` — `PaymentStatus::Unspecified => Ok(prev_status)`
- `router/src/core/payments/gateway/psync_gateway.rs:88` passes `router_data.status` as that `prev_status`, and writes the result back at `:95`

This is exactly what hyperswitch's **direct** NMI integration does on the same branch — it leaves `RouterData` untouched (`hyperswitch_connectors/src/connectors/nmi/transformers.rs:1543`, `None => Ok(Self { ..item.data })`).

Fixes juspay/hyperswitch-cloud#19536 (parent #19534).

## Root cause (prod)

Shadow-diff `router.valueDiff:status`, merchant `merchant_1771364887457`, x-request-id `019f211d-0b72-7fd1-873a-ab4875c91ea7` (UUIDv7 → txn **2026-07-02T04:36:19Z**).

Loki confirms this is **not** the synthetic-502 class — both legs got a genuine NMI **HTTP 200** from `https://secure.nmi.com/api/query.php` (UCS `"stage":"ConnectorCall","latency_ms":2905,"status_code":200,"error":null`; Direct `[CALL_CONNECTOR_API] status_code = 200, elapsed_time = 605ms`). NMI simply returned an empty result, because the attempt (`payment_method_awaited`, no connector txn id) had never been submitted to it. Identical requests, divergent response mapping.

## Before / after

Reproduced locally on clean `main` against the real NMI sandbox: attempt in `payment_method_awaited` with no connector txn id → `GET /payments/{id}?force_sync=true&all_keys_required=true`.

**BEFORE** (`VS_48`, x-req `019f581d-ae17`) — reproduces the prod payload exactly:
```json
"differences": {
  "keyDiff": {},
  "valueDiff": {"status": {"hyperswitch": "payment_method_awaited", "ucs": "authentication_pending"}},
  "typeDiff": {"response.Ok.TransactionResponse.resource_id": {"hyperswitch": "string", "ucs": "object"}}
}
```

**AFTER** (`VS_46`, x-req `019f5823-d6e3`) — with this PR + juspay/hyperswitch#13276:
```json
"Router data comparison completed successfully - no differences found"
"summary": {"total_key_differences": 0, "total_value_differences": 0, "total_type_differences": 0},
"data_sizes": {"hyperswitch_data_keys": 56, "ucs_data_keys": 56}
```
`VS_51`: `hyperswitch_status = payment_method_awaited`, `ucs_status = payment_method_awaited`.

56/56 keys mapped on both sides, so the UCS leg really ran (not a false green).

> The `resource_id` half of the diff is a separate defect in the router's UCS client and is fixed by **juspay/hyperswitch#13276**; both are needed for the fully-green run above.

## Regression check — the non-empty branch is untouched

Real NMI authorize → approved (txn `12295063749`), then PSync so `query.php` returns an actual transaction (the `Some(transaction)` arm, which this PR does not change):

`VS_51`: `hyperswitch_status = ucs_status = charged` · `keyDiff {}` · `valueDiff {}` · `resource_id` absent from `typeDiff`.

(The two `typeDiff` keys left on that run — `amount_captured` / `minor_amount_captured` — are the pre-existing PSync bug at `psync_gateway.rs:108`/`:279`, already tracked by the open hyperswitch PR #13206. Unrelated to this issue, and absent from the #19534 repro itself.)

## Notes

- 3DS is unaffected: on that flow the attempt is already `authentication_pending`, so `Unspecified` resolves back to the same status.
- No proto change — `PaymentStatus::UNSPECIFIED` is the proto3 zero value and is already in the enum; `AttemptStatus::Unspecified` already maps to it (`domain_types/src/types.rs:6758`).

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
@@ -883,14 +883,19 @@ impl TryFrom<ResponseRouterData<SyncResponse, Self>>
                 response.transaction.last()
             });
 
-        // Handle empty response (means AuthenticationPending) or transaction data
+        // Handle empty response (NMI has no record of the order) or transaction data
         let (status, transaction_id) = if let Some(transaction) = transaction {
             // Map condition field from XML to AttemptStatus using NmiStatus enum
             let status = AttemptStatus::from(NmiStatus::from(transaction.condition.clone()));
             (status, Some(transaction.transaction_id.clone()))
         } else {
-            // Empty XML response = AuthenticationPending (during 3DS flow)
-            (AttemptStatus::AuthenticationPending, None)
+            // Empty XML response: NMI has no record of this order, so it is telling us nothing about
+            // the attempt. Report Unspecified rather than inventing a status -- it reaches the router
+            // as PaymentStatus::UNSPECIFIED, which resolves to the attempt's existing status. Claiming
+            // AuthenticationPending here happened to suit the 3DS flow (where the attempt already is
+            // authentication-pending), but it is wrong for every other way an order goes unknown to
+            // NMI -- e.g. an attempt that never reached the connector at all.
+            (AttemptStatus::Unspecified, None)
         };
 
         Ok(Self {


10X-GRACE:
## Re-validated on latest `main` (2026-07-13) — still green, no code change needed

Re-ran the full shadow repro for juspay/hyperswitch-cloud#19536 against current `main`
(prism `534add411`, hyperswitch `53c257fc6c`). Branch `ucs-diff/nmi/psync-empty-response-status`
@ `5fae80408` is **1 ahead / 0 behind** `origin/main` — rebase is a no-op, no conflicts, CI green.

### Isolation: the router was held on plain `main` for both legs

Both legs below ran against a router built from **unmodified `main`** (`53c257fc6c`, i.e. **without**
the sibling fix juspay/hyperswitch#13276). Only the prism binary changed between them, so every
difference is attributable to this PR alone — and the `resource_id` `typeDiff` (the #19535 half)
stays visible in both, exactly as it should.

### BEFORE — prism `main`

`VS_48`, x-req `019f585b-be9c-7103-ae07-cdb0b8130a13`, attempt in `payment_method_awaited` with no
connector txn id → nmi's `query.php` returns an empty transaction list.

```json
"summary": {"total_key_differences": 0, "total_value_differences": 1, "total_type_differences": 1},
"data_sizes": {"hyperswitch_data_keys": 56, "ucs_data_keys": 56},
"differences": {
  "keyDiff": {},
  "valueDiff": {"status": {"hyperswitch": "payment_method_awaited", "ucs": "authentication_pending"}},
  "typeDiff": {"response.Ok.TransactionResponse.resource_id": {"hyperswitch": "string", "ucs": "object"}}
}
```

`VS_51`: `hyperswitch_status = payment_method_awaited`, `ucs_status = authentication_pending` — the
exact `router.valueDiff:status` signature from prod, with the same two values.

### AFTER — prism with this PR

`VS_48`, x-req `019f585d-27e0-7c61-95f8-f27d9f339943`:

```json
"summary": {"total_key_differences": 0, "total_value_differences": 0, "total_type_differences": 1},
"data_sizes": {"hyperswitch_data_keys": 56, "ucs_data_keys": 56},
"differences": {
  "keyDiff": {},
  "valueDiff": {},
  "typeDiff": {"response.Ok.TransactionResponse.resource_id": {"hyperswitch": "string", "ucs": "object"}}
}
```

`VS_51`: `hyperswitch_status = payment_method_awaited`, `ucs_status = payment_method_awaited` — the two
legs now agree. **`keyDiff {}` and `valueDiff {}`: #19536 is resolved.** prism's `PaymentServiceGetResponse`
now carries `status: Unspecified` on a genuine nmi `status_code: 200`, and the router resolves that back
to the attempt's existing status.

56/56 keys mapped on both sides, so the UCS leg really ran (not a false green — this psync needs
`all_keys_required=true` or no connector call happens at all).

The one remaining `typeDiff` is the **separate** `resource_id` defect (#19535), fixed by
juspay/hyperswitch#13276 in the router's UCS client. It is untouched by this PR, and with both fixes in
place the run is fully green (0/0/0).

### Regression — the non-empty branch is unaffected

Real nmi authorize → approved (txn `12295196113`), then psync so `query.php` returns an actual
transaction (the `Some(transaction)` arm this PR does not touch):

- authorize: `VS_46`, **0 / 0 / 0**, "no differences found"
- psync: `keyDiff {}`, `valueDiff {}` — status still maps correctly from `condition`; `resource_id`
  absent from `typeDiff` (there is a real txn id to carry)

The two `typeDiff` keys left on that run — `amount_captured` / `minor_amount_captured` — are the
pre-existing psync bug at `psync_gateway.rs:108`/`:279`, already tracked by the open
juspay/hyperswitch#13206. Unrelated to this PR.

### Prod RCA re-confirmed (Loki) — REAL, not the synthetic-502 class

For x-req `019f211d-0b72-7fd1-873a-ab4875c91ea7` (UUIDv7 → txn `2026-07-02T04:36:19Z`), **both** legs got
a genuine nmi **HTTP 200** from `query.php` — no timeout, no `VS_*` error, no 502:

- UCS: `latency = "2911"`, `request.url = "https://secure.nmi.com/api/query.php"`, `response.body = "{\"transaction\":[]}"`
- Direct (`hyperswitch-consumer`, scheduled `PAYMENTS_SYNC_WORKFLOW`): `status_code: 200`, `elapsed_time: "605.219786ms"`, `error: None`

Prod pods at that time: UCS `aef0be919`, `hyperswitch-consumer` `c8b9bd9` — **neither carries this fix**,
so this is a real mapping bug, not deploy-lag. nmi simply had no record of the order, because the attempt
(`payment_method_awaited`, no connector txn id) never reached it.

gopikrishna000 (approved):

JeevaRamu0104 (approved):

pixincreate (approved):

AkshayaFoiger (approved):

PR #1936, opened 2026-07-13
fix(tsys_transit): correct amount_captured for auth-only + valid voidReason on cancel

awasthi21:
## Summary

Two related `tsys_transit` correctness fixes for the manual-capture / cancel lifecycle.

### 1. Auth-only authorization wrongly reported as captured
The authorize response populated `amount_captured` / `minor_amount_captured` from `<processedAmount>` for **every** approved response, including auth-only (manual capture) authorizations where nothing is settled yet. Hyperswitch applies the connector's `captured_amount` unconditionally on the authorize path (`authorize_gateway.rs`), so a manual-capture auth was recorded with `amount_received == amount` while still in `requires_capture` — corrupting capture/void accounting downstream.

**Fix:** populate `amount_captured` only when the transaction actually settled — `Charged` (Sale) or `PartialCharged`. Auth-only (`Authorized`) now carries no captured amount.

### 2. Void rejected on caller-supplied cancellation_reason
TSYS `<voidReason>` only accepts a fixed set of enum values. The void and refund-reversal requests forwarded the caller-supplied `cancellation_reason` verbatim, so any free-text reason (e.g. `"customer changed mind"`) was rejected with *"The value of element 'voidReason' is not valid."* and the cancel landed in a failed / partially-captured state.

**Fix:** ignore the request `cancellation_reason` and always send a valid connector default (`POST_AUTH_USER_DECLINE` for void, `RETURN_REVERSAL` for refund reversal).

## Testing (end-to-end: HS → UCS → TSYS staging)

| Scenario | Before | After |
| --- | --- | --- |
| Manual auth (auth-only) | `amount_received = amount` ❌ | `amount_received = null` ✅ |
| Manual auth → capture | ok | `amount_received = amount` ✅ |
| Auto-capture (Sale) | ok | unchanged ✅ |
| Cancel with free-text `cancellation_reason` | `failed` (voidReason invalid) ❌ | `cancelled` ✅ |

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/tsys_transit/transformers.rs
@@ -2365,18 +2365,27 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
             )
         })?;
 
-        // On A0002 (partial approval) TSYS returns the actually approved
-        // value in <processedAmount>. Forward it as MinorUnit so the core
-        // can reconcile the partial capture/authorization against the
-        // requested amount; for the fully-approved A0000 case this is also
-        // harmless to populate.
-        let minor_amount_captured = body.processed_amount.as_ref().and_then(|amount| {
-            crate::connectors::tsys_transit::TsysTransitAmountConvertor::convert_back(
-                amount.clone(),
-                router_data.request.currency,
-            )
-            .ok()
-        });
+        // <processedAmount> reflects settled funds, so amount_captured must be
+        // populated ONLY when the transaction actually captured: a Sale (Charged)
+        // or a partial approval (PartialCharged). For an auth-only (Authorized)
+        // manual-capture flow nothing is captured yet — populating it there made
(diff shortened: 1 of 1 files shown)


awasthi21:
Fixes issue #1937

XyneSpaces:
:warning: `TsysTransitAmountConvertor::convert_back(...).ok()` silently discards conversion failures for settled transactions.

If the amount cannot be converted back to `MinorUnit` (e.g. unknown currency), `amount_captured` becomes `None` even though the connector reported settled funds. Propagate the error instead of swallowing it, or at least log the failure when it occurs.

:mag: This PR changes payment-correctness behavior but only touches `transformers.rs`; there are no updated tests or connector spec overrides for the new `is_settled` guard or the fixed `void_reason` values.

Consider adding a spec/override case that asserts `amount_captured` is absent for `Authorized` and present for `Charged`, and that void requests send the fixed enum values.

deepanshu-iiitu (approved):

swangi-kumari (approved):

AmitsinghTanwar007 (approved):

PR #1944, opened 2026-07-14
feat(connector): added new connector glomopay

srishtijuspay:
## Description
Adds Glomopay payment connector with full implementation of 5 flows: Authorize (composite), PSync, Refund, RSync, and IncomingWebhook

##  Flows Implemented

Flow | Service Method | Status
-- | -- | --
Authorize (composite) | types.CompositePaymentService/Authorize | Runs CreateCustomer + CreateOrder + Authorize in one call (Glomopay hosted payment page)
PSync | types.PaymentService/Get | Query payment status via API
Refund | types.PaymentService/Refund | Full/partial refund
RSync | types.RefundService/Get | Query refund status
IncomingWebhook | types.EventService/HandleEvent | Process payment/refund webhook notifications

### Additional Changes

## How did you test it?

<details>
<summary>1. Composite Authorize - 3DS Card Payment - Success Scenario</summary><br>

```
curl --location 'http://localhost:8011/composite/payments/authorize' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_TEST_001' \
  --data-raw '{
    "merchant_transaction_id": "GLOMOPAY_TEST_001",
    "amount": {"minor_amount": 1000, "currency": "USD"},
    "payment_method": {
      "payment_method": {
        "card": {
          "card_number": "4111111111111111",
          "card_exp_month": "12",
          "card_exp_year": "2030",
          "card_cvc": "123",
          "card_holder_name": "Test User"
        }
      }
    },
    "auth_type": "THREE_DS",
    "address": {
      "billing_address": {
        "first_name": "Test",
        "last_name": "User",
        "line1": "123 Test St",
        "city": "Mumbai",
        "state": "Maharashtra",
        "zip_code": "400001",
        "country_alpha2_code": "IN"
      }
    },
    "customer": {
      "email": "test@example.com",
      "name": "Test User",
      "phone_number": "9999999999"
    },
    "return_url": "https://example.com/return",
    "merchant_order_id": "GLOMOPAY_TEST_001",
    "order_details": [],
    "test_mode": false
  }'
```

Response

```
{
  "create_customer_response": {
    "merchant_customer_id": "cust_6a50ffd3lacrd",
    "connector_customer_id": "cust_6a50ffd3lacrd",
    "status_code": 201
  },
  "authorize_response": {
    "merchant_transaction_id": "payt_6a50ffd4UHOBo",
    "connector_transaction_id": "payt_6a50ffd4UHOBo",
    "status": "AUTHENTICATION_PENDING",
    "status_code": 201,
    "redirection_data": {
      "form_type": {
        "Uri": {
          "uri": "https://payments.glomopay.com?paymentId=payt_6a50ffd4UHOBo&authToken=test_68524ee8SJbU1RqT&redirectUrl=https%3A%2F%2Fexample.com%2Freturn"
        }
      }
    },
    "resp_code": "in_progress",
    "resp_msg": "Glomopay payment in progress"
  },
  "create_order_response": {
    "connector_order_id": "order_6a50ffd3dcuFm",
    "status": "PENDING",
    "status_code": 200
  },
  "composite_status": "COMPLETED"
}
```

</details>
<details><summary>2. Payment Sync - Success Scenario</summary><br>

```
curl --location 'http://localhost:8011/payments/get' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_TEST_001' \
  --data '{
    "merchant_transaction_id": "GLOMOPAY_TEST_001",
    "connector_transaction_id": "payt_6a50ffd4UHOBo",
    "amount": {"minor_amount": 1000, "currency": "USD"}
  }'
```

Response

```
{
  "connector_transaction_id": "payt_6a50ffd4UHOBo",
  "merchant_transaction_id": "GLOMOPAY_TEST_001",
  "status": "CHARGED",
  "status_code": 200,
  "amount": {
    "minor_amount": 1000,
    "currency": "USD"
  },
  "resp_code": "success",
  "resp_msg": "Glomopay payment successful"
}
```

</details>
<details><summary>3. Refund - Success Scenario</summary><br>

```
curl --location 'http://localhost:8011/payments/refund' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_REFUND_001' \
  --data '{
    "connector_transaction_id": "payt_6a50ffd4UHOBo",
    "merchant_refund_id": "GLOMOPAY_REFUND_001",
    "refund_amount": {"minor_amount": 1000, "currency": "USD"},
    "payment_amount": 1000,
    "reason": "Customer request"
  }'
```

Response

```
{
  "connector_refund_id": "refund_6a51038bIgua0",
  "status": "REFUND_SUCCESS",
  "status_code": 201,
  "connector_transaction_id": "payt_6a50ffd4UHOBo",
  "resp_code": "success",
  "resp_msg": "Glomopay refund successful"
}
```

</details>
<details><summary>4. Refund Sync - Success Scenario</summary><br>

```
curl --location 'http://localhost:8011/refunds/get' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_REFUND_001' \
  --data '{
    "connector_transaction_id": "payt_6a50ffd4UHOBo",
    "connector_refund_id": "refund_6a51038bIgua0",
    "merchant_refund_id": "GLOMOPAY_REFUND_001",
    "refund_id": ""
  }'
```

Response

```
{
  "merchant_refund_id": "refund_6a51038bIgua0",
  "connector_refund_id": "refund_6a51038bIgua0",
  "status": "REFUND_SUCCESS",
  "status_code": 200,
  "connector_transaction_id": "payt_6a50ffd4UHOBo",
  "resp_code": "success",
  "resp_msg": "Glomopay refund successful"
}
```

</details>
<details><summary>5. Incoming Webhook - Payment Success Scenario</summary><br>

The `request_details.body` field is proto `bytes`, so it must be sent as a JSON array of byte values (serde's default for `Vec<u8>`). The array below encodes the Glomopay webhook body:

```json
{"entity_type":"payment","event_type":"success","data":{"id":"payt_6a50ffd4UHOBo","payin_id":"order_6a50ffd3dcuFm","payment_amount":1000,"payment_currency":"USD","requested_amount":1000,"requested_currency":"USD","fees":{"txn_fee":{"currency":"USD","amount":100},"fx_fee":{"currency":"USD","amount":0},"referral_fee":{"currency":"USD","amount":0}},"funds_available":true}}
```

```
curl --location 'http://localhost:8011/events/handle' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_WEBHOOK_001' \
  --data '{
    "merchant_event_id": "GLOMOPAY_WEBHOOK_001",
    "request_details": {
      "method": "HTTP_METHOD_POST",
      "uri": "/webhooks/glomopay",
      "headers": {"content-type": "application/json"},
      "body": [123, 34, 101, 110, 116, 105, 116, 121, 95, 116, 121, 112, 101, 34, 58, 34, 112, 97, 121, 109, 101, 110, 116, 34, 44, 34, 101, 118, 101, 110, 116, 95, 116, 121, 112, 101, 34, 58, 34, 115, 117, 99, 99, 101, 115, 115, 34, 44, 34, 100, 97, 116, 97, 34, 58, 123, 34, 105, 100, 34, 58, 34, 112, 97, 121, 116, 95, 54, 97, 53, 48, 102, 102, 100, 52, 85, 72, 79, 66, 111, 34, 44, 34, 112, 97, 121, 105, 110, 95, 105, 100, 34, 58, 34, 111, 114, 100, 101, 114, 95, 54, 97, 53, 48, 102, 102, 100, 51, 100, 99, 117, 70, 109, 34, 44, 34, 112, 97, 121, 109, 101, 110, 116, 95, 97, 109, 111, 117, 110, 116, 34, 58, 49, 48, 48, 48, 44, 34, 112, 97, 121, 109, 101, 110, 116, 95, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 114, 101, 113, 117, 101, 115, 116, 101, 100, 95, 97, 109, 111, 117, 110, 116, 34, 58, 49, 48, 48, 48, 44, 34, 114, 101, 113, 117, 101, 115, 116, 101, 100, 95, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 102, 101, 101, 115, 34, 58, 123, 34, 116, 120, 110, 95, 102, 101, 101, 34, 58, 123, 34, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 49, 48, 48, 125, 44, 34, 102, 120, 95, 102, 101, 101, 34, 58, 123, 34, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 48, 125, 44, 34, 114, 101, 102, 101, 114, 114, 97, 108, 95, 102, 101, 101, 34, 58, 123, 34, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 48, 125, 125, 44, 34, 102, 117, 110, 100, 115, 95, 97, 118, 97, 105, 108, 97, 98, 108, 101, 34, 58, 116, 114, 117, 101, 125, 125]
    }
  }'
```

Response

```
{
  "event_type": "PAYMENT_INTENT_SUCCESS",
  "event_content": {
    "content": {
      "PaymentsResponse": {
        "connector_transaction_id": "payt_6a50ffd4UHOBo",
        "merchant_transaction_id": "payt_6a50ffd4UHOBo",
        "status": "CHARGED",
        "status_code": 200,
        "resp_code": "success",
        "resp_msg": "Glomopay payment successful"
      }
    }
  },
  "source_verified": false,
  "merchant_event_id": "GLOMOPAY_WEBHOOK_001",
  "supported_integrity_checks": ["INTEGRITY_CHECK_CONNECTOR_TRANSACTION_ID"],
  "requires_payment_sync": true
}
```

To regenerate the byte array for a different payload:
```bash
python3 -c 'import json; s = json.dumps({"entity_type":"payment","event_type":"success","data":{"id":"payt_XYZ"}}, separators=(",",":")); print(json.dumps(list(s.encode())))'
```

</details>
<details><summary>6. Incoming Webhook - Refund Success Scenario</summary><br>

The `request_details.body` field is proto `bytes`, so it must be sent as a JSON array of byte values (serde's default for `Vec<u8>`). The array below encodes the Glomopay refund webhook body:

```json
{"entity_type":"refund","event_type":"success","data":{"id":"refund_6a51038bIgua0","payment_id":"payt_6a50ffd4UHOBo","amount":1000,"currency":"USD","status":"success","reason":"Requested By Customer"}}
```

```
curl --location 'http://localhost:8011/events/handle' \
  --header 'x-connector-config: {"config":{"Glomopay":{"api_key":"<REDACTED>"}}}' \
  --header 'x-merchant-id: azharamin' \
  --header 'x-tenant-id: JUSPAY' \
  --header 'x-connector: glomopay' \
  --header 'x-reference-id: GLOMOPAY_REFUND_WEBHOOK_001' \
  --data '{
    "merchant_event_id": "GLOMOPAY_REFUND_WEBHOOK_001",
    "request_details": {
      "method": "HTTP_METHOD_POST",
      "uri": "/webhooks/glomopay",
      "headers": {"content-type": "application/json"},
      "body": [123, 34, 101, 110, 116, 105, 116, 121, 95, 116, 121, 112, 101, 34, 58, 34, 114, 101, 102, 117, 110, 100, 34, 44, 34, 101, 118, 101, 110, 116, 95, 116, 121, 112, 101, 34, 58, 34, 115, 117, 99, 99, 101, 115, 115, 34, 44, 34, 100, 97, 116, 97, 34, 58, 123, 34, 105, 100, 34, 58, 34, 114, 101, 102, 117, 110, 100, 95, 54, 97, 53, 49, 48, 51, 56, 98, 73, 103, 117, 97, 48, 34, 44, 34, 112, 97, 121, 109, 101, 110, 116, 95, 105, 100, 34, 58, 34, 112, 97, 121, 116, 95, 54, 97, 53, 48, 102, 102, 100, 52, 85, 72, 79, 66, 111, 34, 44, 34, 97, 109, 111, 117, 110, 116, 34, 58, 49, 48, 48, 48, 44, 34, 99, 117, 114, 114, 101, 110, 99, 121, 34, 58, 34, 85, 83, 68, 34, 44, 34, 115, 116, 97, 116, 117, 115, 34, 58, 34, 115, 117, 99, 99, 101, 115, 115, 34, 44, 34, 114, 101, 97, 115, 111, 110, 34, 58, 34, 82, 101, 113, 117, 101, 115, 116, 101, 100, 32, 66, 121, 32, 67, 117, 115, 116, 111, 109, 101, 114, 34, 125, 125]
    }
  }'
```

Response

```
{
  "event_type": "WEBHOOK_REFUND_SUCCESS",
  "event_content": {
    "content": {
      "RefundsResponse": {
        "merchant_refund_id": "refund_6a51038bIgua0",
        "connector_refund_id": "refund_6a51038bIgua0",
        "status": "REFUND_SUCCESS",
        "status_code": 200,
        "merchant_transaction_id": "payt_6a50ffd4UHOBo",
        "resp_code": "success",
        "resp_msg": "Glomopay refund successful"
      }
    }
  },
  "source_verified": false,
  "merchant_event_id": "GLOMOPAY_REFUND_WEBHOOK_001",
  "supported_integrity_checks": ["INTEGRITY_CHECK_CONNECTOR_TRANSACTION_ID"],
  "requires_payment_sync": true
}
```

To regenerate the byte array for a different payload:
```bash
python3 -c 'import json; s = json.dumps({"entity_type":"refund","event_type":"success","data":{"id":"refund_XYZ","payment_id":"payt_XYZ"}}, separators=(",",":")); print(json.dumps(list(s.encode())))'
```

</details>


XyneSpaces:
[should-fix] Glomopay ships without integration-test specs.

New connectors in this repo are exercised by `test-prism` via `crates/internal/integration-tests/src/connector_specs/<connector>/specs.json` (and an `override.json` where needed). This PR registers `+pub mod glomopay;` in `crates/integrations/connector-integration/src/connectors.rs` but does not add any `connector_specs/glomopay/` files, so the harness has no supported-suites list or overrides for the connector.

Please add `crates/internal/integration-tests/src/connector_specs/glomopay/specs.json` listing the suites this connector now supports (e.g. `PaymentService/Authorize`, `PaymentService/Capture`, `PaymentService/Refund`, `PaymentService/Get`, `RefundService/Get`, `EventService/HandleEvent`, `CustomerService/Create`, and any `CustomerService/Get`-style lookup flow), plus an `override.json` with sandbox request/assert overrides.

crates/integrations/connector-integration/src/connectors/adyen/test.rs
```diff
@@ -86,6 +86,8 @@ mod tests {
                     merchant_request_id: None,
                     sender_payment_instrument_id: None,
                     settlement_status: None,
+
+
```
deepanshu-iiitu: remove this

crates/integrations/connector-integration/src/connectors/adyen/test.rs
```diff
@@ -291,6 +293,8 @@ mod tests {
...
```
deepanshu-iiitu: remove this

crates/integrations/connector-integration/src/connectors/calida/test.rs
```diff
@@ -108,6 +108,8 @@ mod tests {
                     merchant_request_id: None,
                     sender_payment_instrument_id: None,
                     settlement_status: None,
+
+
```
deepanshu-iiitu: remove this

crates/integrations/connector-integration/src/connectors/calida/test.rs
```diff
@@ -276,6 +278,8 @@ mod tests {
...
```
deepanshu-iiitu: remove this

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1120 @@
...
+            .get_optional_billing_state()
+            .map(|s| s.expose())
+            .ok_or_else(crate::utils::missing_field_err("billing.address.state"))?;
+
+        let country = flow_data
+            .get_optional_billing_country()
+            .map(|c| CountryAlpha2::from_alpha2_to_alpha3(c).to_string())
+            .ok_or_else(crate::utils::missing_field_err("billing.address.country"))?;
+
+        let zip = flow_data.get_optional_billing_zip().map(|z| z.expose());
```
deepanshu-iiitu: Use utils here properly

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1120 @@
...
+            name: Secret::new(name),
+            customer_type: "individual".to_string(),
+            email: Secret::new(email),
+            phone: phone.map(Secret::new),
+            address: Secret::new(address),
+            city: Secret::new(city),
+            state: Secret::new(state),
+            country,
+            pincode: zip.map(Secret::new),
+        })
```
deepanshu-iiitu, line 512: You first exposed the PII data then converted it to secret

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1120 @@
...
+                    let full = card.card_exp_year.peek();
+                    let last_two = if full.len() >= 2 {
+                        full[full.len() - 2..].to_string()
+                    } else {
+                        full.clone()
+                    };
+                    Secret::new(last_two)
+                };
+
+                (
```
deepanshu-iiitu: Use utils here

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,210 @@ pub struct Payments {
...
+    connector: &ConnectorVariant,
+    service_name: &str,
+    config: &Arc<Config>,
+    request_id: &str,
+    lineage_ids: &lineage::LineageIds<'_>,
+    metadata_payload: &utils::MetadataPayload,
+    test_context: Option<external_services::service::TestContext>,
+) -> Result<Option<grpc_api_types::payments::CustomerServiceCreateResponse>, tonic::Status> {
+    if !connector_data.connector.should_lookup_connector_customer() {
+        return Ok(None);
```
gopikrishna000: dont use return based code

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,210 @@ pub struct Payments {
...
+            test_context,
+            get_api_tag,
+        ),
+    )
+    .await
+    {
+        if get_result.response.is_ok() {
+            let lookup_response = generate_get_connector_customer_response(get_result)
+                .map_err(|e| e.into_grpc_status())?;
+            return Ok(Some(lookup_response));
```
gopikrishna000: same here

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,210 @@ pub struct Payments {
...
+        }
+        // response is Err (e.g. CUSTOMER_NOT_FOUND) — fall through to POST create
+    }
+    // Infrastructure error from GET — fall through to POST create
+    Ok(None)
+}
+
+/// Execute the POST create for a connector customer.
+#[allow(clippy::too_many_arguments)]
+async fn execute_create_connector_customer(
```
gopikrishna000: naming is bit off convention
internal_create_connector_customer would match similar functions named

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,210 @@ pub struct Payments {
...
+    connector_config: &ConnectorSpecificConfig,
+    connector_customer_request_data: &ConnectorCustomerData,
+    connector: &ConnectorVariant,
+    service_name: &str,
+    config: &Arc<Config>,
+    request_id: &str,
+    lineage_ids: &lineage::LineageIds<'_>,
+    metadata_payload: &utils::MetadataPayload,
+    test_context: Option<external_services::service::TestContext>,
+) -> Result<Option<grpc_api_types::payments::CustomerServiceCreateResponse>, tonic::Status> {
```
gopikrishna000: ideally caller should own `if connector.should_lookup_connector_customer()`

and then function name can consistent with others like `internal_get_connector_customer`
and instead of Result Of Option and infer Some = CustomerFound and None = NotFound
can we have better return type, 

like possibly
enum ConnectorCustomerLookup {
    Found(grpc_api_types::payments::CustomerServiceCreateResponse),
    NotFound,
}

deepanshu-iiitu (dismissed):

crates/internal/composite-service/src/payments.rs
```diff
@@ -371,13 +371,7 @@ where
         let connector_customer_id = payload
             .state
             .as_ref()
-            .and_then(|state| state.connector_customer_id.clone())
-            .or_else(|| {
-                payload
-                    .customer
-                    .as_ref()
-                    .and_then(|c| c.connector_customer_id.clone())
-            });
```
Aishwariyaa-Anand, line 438: revert this

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -1665,6 +1666,10 @@ message EventServiceHandleResponse {
 
   // Integrity fields that downstream can use to verify the webhook payload.
   repeated IntegrityCheck supported_integrity_checks = 6;
+
+  // If true, caller must perform a PSync before trusting the webhook status.
+  // Set for connectors that send unsigned webhooks (e.g. Glomopay).
+  bool requires_payment_sync = 7;
```
Aishwariyaa-Anand: we can use source_verified field to decide on sync, this is not required

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -380,77 +576,67 @@ impl CustomerService for Customer {
...
+                                &connector_customer_request_data,
+                                &metadata_payload.connector,
+                                &service_name,
+                                &config,
+                                &metadata_payload.request_id,
+                                &metadata_payload.lineage_ids,
+                                &metadata_payload,
+                                test_context.clone(),
+                            )
+                            .await?
```
Aishwariyaa-Anand: CREATE_CONNECTOR_CUSTOMER_FLOW should not call GET_CONNECTOR_CUSTOMER_FLOW internally. This decision can be handled at composite layer if required but not in core flow

crates/types-traits/interfaces/src/connector_types.rs
```diff
@@ -210,6 +211,10 @@ pub trait ValidationTrait: ConnectorCommon {
         false
     }
 
+    fn should_lookup_connector_customer(&self) -> bool {
```
Aishwariyaa-Anand: let's stick with same name 
```suggestion
    fn should_get_connector_customer(&self) -> bool {
```

crates/types-traits/domain_types/src/types.rs
```diff
@@ -9148,8 +9160,8 @@ impl ForeignTryFrom<RefundWebhookDetailsResponse> for RefundResponse {
             error: Some(grpc_api_types::payments::ErrorInfo {
                 unified_details: None,
                 connector_details: Some(grpc_api_types::payments::ConnectorErrorDetails {
-                    message: value.error_message,
-                    code: value.error_code,
+                    message: value.error_message.clone(),
+                    code: value.error_code.clone(),
```
Aishwariyaa-Anand:
```suggestion
                    message: value.error_message,
                    code: value.error_code,
```

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4375,8 +4376,15 @@ message CustomerServiceGetRequest {
...
+  // Connector-assigned customer ID, when known. Optional because callers may
+  // instead identify the customer via `customer` fields (e.g. lookup by email).
+  optional string connector_customer_id = 2;
+
+  // Customer identity fields the connector can look up by (email, phone, name,
+  // connector_customer_id, etc.). The connector's GetConnectorCustomer flow
+  // picks which field it uses — Glomopay uses `email`, other connectors may
+  // key off `connector_customer_id` or other identifiers.
+  optional Customer customer = 3;
 }
```
Aishwariyaa-Anand:
```suggestion
  // Customer Information
  string connector_customer_id = 2;

  optional SecretString email = 3;   // Email address of the customer
  optional SecretString phone_number = 4;  // Phone number of the customer
}
```
let's follow similar pattern as CustomerServiceCreateRequest

crates/internal/composite-service/src/payments.rs
```diff
@@ -382,32 +383,68 @@ where
...
-                *create_customer_request.extensions_mut() = extensions.clone();
-
-                let create_customer_response = self
-                    .customer_service
-                    .create(create_customer_request)
-                    .await?
-                    .into_inner();
+        if !should_create_connector_customer {
+            return Ok(None);
+        }
```
Aishwariyaa-Anand: avoid early returns

crates/internal/composite-service/src/payments.rs
```diff
@@ -382,32 +383,68 @@ where
...
+        // For connectors that support lookup (e.g. Glomopay, which 4xxs on
+        // duplicate email), try GET first — if the customer already exists on
+        // the connector, reuse that ID instead of hitting CREATE and getting
+        // a duplicate error.
+        if connector_data.connector.should_lookup_connector_customer() {
+            let get_payload = CustomerServiceGetRequest {
+                merchant_customer_id: payload.customer.as_ref().and_then(|c| c.id.clone()),
+                connector_customer_id: None,
+                customer: payload.customer.clone(),
+            };
```
Aishwariyaa-Anand: use try froms

crates/internal/composite-service/src/payments.rs
```diff
@@ -382,32 +383,68 @@ where
...
+                    // Found on connector — synthesize a create-shaped response
+                    // so downstream code (which expects CustomerServiceCreateResponse)
+                    // is unchanged.
+                    return Ok(Some(CustomerServiceCreateResponse {
+                        merchant_customer_id: get_inner.merchant_customer_id,
+                        connector_customer_id: id,
+                        error: None,
+                        status_code: get_inner.status_code,
+                        response_headers: get_inner.response_headers,
+                    }));
```
Aishwariyaa-Anand: use try from

crates/internal/composite-service/src/payments.rs
```diff
@@ -382,32 +383,68 @@ where
...
+                }
+                // Get succeeded but returned no customer (or an error inside) —
+                // fall through to CREATE.
             }
-            false => None,
-        };
+            // Get failed (infra error, not-found, etc.) — fall through to CREATE.
+        }
```
Aishwariyaa-Anand: make this as a function

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,202 @@ pub struct Payments {
...
+    connector_config: &ConnectorSpecificConfig,
+    connector_customer_request_data: &ConnectorCustomerData,
+    connector: &ConnectorVariant,
+    service_name: &str,
+    config: &Arc<Config>,
+    request_id: &str,
+    lineage_ids: &lineage::LineageIds<'_>,
+    metadata_payload: &utils::MetadataPayload,
+    test_context: Option<external_services::service::TestContext>,
+) -> Result<grpc_api_types::payments::CustomerServiceGetResponse, tonic::Status> {
```
Aishwariyaa-Anand: follow the same pattern as other flow, and have this in trait

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,202 @@ pub struct Payments {
...
+// ============================================================================
+
+/// Build a `RouterDataV2` for a customer flow (Get or Create).
+/// The default response is `Err(ErrorResponse::default())` — connectors
+/// populate the real value during execution.
+fn build_customer_router_data<F>(
+    payment_flow_data: PaymentFlowData,
+    connector_config: ConnectorSpecificConfig,
+    request: ConnectorCustomerData,
+) -> RouterDataV2<F, PaymentFlowData, ConnectorCustomerData, ConnectorCustomerResponse> {
```
Aishwariyaa-Anand: follow similar pattern as other flows

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,202 @@ pub struct Payments {
...
+/// identical between them.
+fn build_customer_event_params<'a>(
+    flow_name: FlowName,
+    connector_name: &'a str,
+    service_name: &'a str,
+    config: &'a Arc<Config>,
+    request_id: &'a str,
+    lineage_ids: &'a lineage::LineageIds<'a>,
+    metadata_payload: &'a utils::MetadataPayload,
+) -> EventProcessingParams<'a> {
```
Aishwariyaa-Anand: follow similar pattern as other flows

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -297,6 +298,202 @@ pub struct Payments {
...
+    connector_config: &ConnectorSpecificConfig,
+    connector_customer_request_data: &ConnectorCustomerData,
+    connector: &ConnectorVariant,
+    service_name: &str,
+    config: &Arc<Config>,
+    request_id: &str,
+    lineage_ids: &lineage::LineageIds<'_>,
+    metadata_payload: &utils::MetadataPayload,
+    test_context: Option<external_services::service::TestContext>,
+) -> Result<grpc_api_types::payments::CustomerServiceCreateResponse, tonic::Status> {
```
Aishwariyaa-Anand: follow similar pattern as other flows

crates/internal/composite-service/src/transformers.rs
```diff
@@ -157,6 +188,30 @@ impl ForeignFrom<&CompositeAuthorizeRequest> for CustomerServiceCreateRequest {
...
+/// implementation picks which one to look up by (Glomopay uses email).
+impl ForeignFrom<&CompositeAuthorizeRequest> for CustomerServiceGetRequest {
+    fn foreign_from(item: &CompositeAuthorizeRequest) -> Self {
+        let customer = item.customer.as_ref();
+        Self {
+            merchant_customer_id: item
+                .merchant_customer_id
+                .clone()
+                .or_else(|| customer.and_then(|c| c.id.clone())),
+            connector_customer_id: None,
```
Aishwariyaa-Anand: map connector_customer_id from CompositeAuthorizeRequest, do not keep it None

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4376,15 +4399,15 @@ message CustomerServiceGetRequest {
...
 
-  // Customer identity fields the connector can look up by (email, phone, name,
-  // connector_customer_id, etc.). The connector's GetConnectorCustomer flow
-  // picks which field it uses — Glomopay uses `email`, other connectors may
-  // key off `connector_customer_id` or other identifiers.
-  optional Customer customer = 3;
+  // Identity fields the connector can look up by. The connector's
+  // GetConnectorCustomer flow picks which one it uses — Glomopay uses `email`.
+  optional SecretString email = 3;         // Email address of the customer
+  optional SecretString phone_number = 4;  // Phone number of the customer
```
Aishwariyaa-Anand: do not mention connector name in comment, also use existing comment for these

crates/internal/composite-service/src/payments.rs
```diff
@@ -383,61 +418,45 @@ where
...
-        if connector_data.connector.should_lookup_connector_customer() {
-            let get_payload = CustomerServiceGetRequest {
-                merchant_customer_id: payload.customer.as_ref().and_then(|c| c.id.clone()),
-                connector_customer_id: None,
-                customer: payload.customer.clone(),
+        let create_customer_response = if should_create_connector_customer {
+            // Try lookup first for connectors that support get-or-create
+            // semantics (e.g. Glomopay). If the customer already exists on
+            // the connector, reuse that ID instead of hitting CREATE.
+            let existing = if connector_data.connector.should_get_connector_customer() {
```
Aishwariyaa-Anand: use proper variable names

crates/internal/composite-service/src/payments.rs
```diff
@@ -383,61 +418,45 @@ where
...
+                match self
+                    .try_lookup_existing_connector_customer(payload, metadata, extensions)
```
Aishwariyaa-Anand: rename this to get_connector_customer

crates/internal/composite-service/src/transformers.rs
```diff
@@ -157,6 +188,30 @@ impl ForeignFrom<&CompositeAuthorizeRequest> for CustomerServiceCreateRequest {
...
+            email: item
+                .email
+                .clone()
+                .or_else(|| customer.and_then(|c| c.email.clone())),
+            phone_number: item
+                .phone_number
+                .clone()
+                .or_else(|| customer.and_then(|c| c.phone_number.clone())),
+        }
```
Aishwariyaa-Anand, line 235: use proper naming instead of |c|

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -355,131 +407,56 @@ impl CustomerService for Customer {
...
+    async fn get(
+        &self,
+        request: tonic::Request<grpc_api_types::payments::CustomerServiceGetRequest>,
+    ) -> Result<tonic::Response<grpc_api_types::payments::CustomerServiceGetResponse>, tonic::Status>
+    {
+        let service_name = request
+            .extensions()
+            .get::<String>()
+            .cloned()
+            .unwrap_or_else(|| "PaymentService".to_string());
```
Aishwariyaa-Anand:
```suggestion
            .unwrap_or_else(|| "CustomerService".to_string());
```

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -345,7 +398,6 @@ impl CustomerService for Customer {
         tonic::Response<grpc_api_types::payments::CustomerServiceCreateResponse>,
         tonic::Status,
     > {
-        info!("CREATE_CONNECTOR_CUSTOMER_FLOW: initiated");
```
Aishwariyaa-Anand, line 348: revert this, and add this log for get connector customer too

Aishwariyaa-Anand (dismissed):

Aishwariyaa-Anand (dismissed):

deepanshu-iiitu (dismissed):

crates/internal/composite-service/src/payments.rs
```diff
@@ -360,6 +369,32 @@ where
...
+        extensions: &tonic::Extensions,
+    ) -> ConnectorCustomerLookup {
+        let get_payload = CustomerServiceGetRequest::foreign_from(payload);
+        let mut get_request = tonic::Request::new(get_payload);
+        *get_request.metadata_mut() = metadata.clone();
+        *get_request.extensions_mut() = extensions.clone();
+
+        let get_response = match self.customer_service.get(get_request).await {
+            Ok(resp) => resp,
+            Err(_) => return ConnectorCustomerLookup::NotFound,
```
gopikrishna000: why do we return notfound on error case, doesnt it create duplicate customer., but ideally it could have happened due to code / network / connector error right, should we return error instead

crates/internal/composite-service/src/payments.rs
```diff
@@ -360,6 +369,32 @@ where
...
+        };
+
+        CustomerServiceCreateResponse::foreign_try_from(get_response.into_inner())
+            .map(|r| ConnectorCustomerLookup::Found(Box::new(r)))
+            .unwrap_or(ConnectorCustomerLookup::NotFound)
```
gopikrishna000: same

crates/internal/composite-service/src/payments.rs
gopikrishna000: why dont we have a concrete status to decide whether customer look up was found or not found in the proto And since i believe this is the first pr to uncomment GetCustomerCreate, should we have a enum to differentiate the status in its response in the `CustomerServiceGetResponse`

```
enum CustomerLookupStatus {
  CUSTOMER_LOOKUP_UNSPECIFIED = 0;
  CUSTOMER_FOUND     = 1;
  CUSTOMER_NOT_FOUND = 2;
}
```

crates/types-traits/grpc-api-types/proto/payment.proto
gopikrishna000: should this be optional

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+        // GlomoPay requires a non-blank HTTPS callback_url. If the caller-supplied
+        // return_url is HTTP (e.g. localhost during local testing), fall back to a
+        // placeholder HTTPS URL so GlomoPay's validator accepts the request.
+        let callback_url = Some(
+            router_data
+                .resource_common_data
+                .return_url
+                .as_deref()
+                .filter(|url| url.starts_with("https://"))
+                .unwrap_or("https://www.google.com")
```
JeevaRamu0104: **[S0] Silent fallback redirects the cardholder to an unrelated third party.**

When `return_url` is absent or not HTTPS, `callback_url` silently becomes `https://www.google.com`. The payment still completes at Glomopay, but the cardholder is dropped on a third-party domain and never returns to the merchant — with no error raised anywhere.

Per the comment at `transformers.rs:876-878`, Glomopay's redirect carries `paymentId` / `authToken` / `redirectUrl` as query params, so whatever it appends to this callback is disclosed to that third party.

`www.google.com` appears nowhere else in a production path in this repo — only inside `test.rs` fixtures (`adyen/test.rs:50`, `calida/test.rs:56`). The established convention for a required return URL is to fail fast: see `adyen/transformers.rs:6392`, `iatapay/transformers.rs:287`, `paysafe/transformers.rs:384`, which all use `.ok_or_else(missing_field_err)`.

Suggested fix:
```rust
let callback_url = router_data
    .resource_common_data
    .return_url
    .clone()
    .ok_or_else(crate::utils::missing_field_err("return_url"))?;
```
…and reject a non-HTTPS `return_url` explicitly rather than substituting one.

crates/integrations/connector-integration/src/connectors/glomopay.rs
```diff
@@ -0,0 +1,691 @@
...
+            &self,
+            req: &RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>,
+        ) -> CustomResult<Vec<(String, Maskable<String>)>, IntegrationError> {
+            self.build_headers(req)
+        }
+        fn get_url(
+            &self,
+            req: &RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>,
+        ) -> CustomResult<String, IntegrationError> {
+            let request_id = &req.resource_common_data.connector_request_reference_id;
```
JeevaRamu0104, line 344: **[S1] PSync can report an unrelated payment's status.**

Two gaps compound here:

1. `connector_request_reference_id` is interpolated into `payment?request_id={}` with no empty check. If it is ever empty the filter drops and Glomopay returns every payment on the account.
2. The response transformer (`glomopay/transformers.rs:938`) takes `response.data.into_iter().next()` — the first element — and never verifies it corresponds to the payment being synced.

This is exactly the hazard you documented and guarded against for RSync at `glomopay.rs:404-419`:

> *"Dropping the ?request_id filter would return every refund on the merchant's account and the response transformer would silently pick the first one — effectively reporting an unrelated refund's status as if it belonged to the current one."*

The same reasoning applies verbatim to PSync, but the guard was only applied to RSync. Please mirror the empty-`request_id` guard here, and in the transformer select the entry by matching against `PaymentsSyncData.connector_transaction_id` instead of taking the first element.

JeevaRamu0104 (commented):
## Code review — Glomopay connector

Reviewed the full diff (138 files) with a focus on payment correctness. The mechanical
`GetConnectorCustomer` additions across ~100 connector files are consistent with the new
`ConnectorServiceTrait` supertrait bound and look correct.

Findings below are ordered by severity: **1x S0, 5x S1, 3x S2, 1x S3**. The S0 and the PSync/RSync
findings are payment-correctness issues that I'd consider blocking. Existing review threads already
cover the style/architecture points, so I've tried not to duplicate them.

### Verified as *not* issues
Flagging these explicitly since they look suspicious but check out:

- **sandbox/dev sharing the production `base_url`** — normal; 20 of 98 connectors do this
  (Stripe, Razorpay, Mollie).
- **`verify_webhook_source` hardcoded to `Ok(false)`** — `source_verified=false` does not reject the
  webhook, it propagates to the caller (`ffi/src/services/payments.rs:799`). Matches the documented
  intent and `worldpayvantiv`.
- **`should_get_connector_customer()` defaults to `false`** — no extra GET for the other 97 connectors.
- **No `test.rs`** — only 6 of 98 connectors have one; the `specs.json` approach is current convention.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+    ) -> Result<Self, Self::Error> {
+        let response = item.response;
+        let router_data = item.router_data;
+        let connector_refund_id = router_data.request.connector_refund_id.clone();
+
+        let refund_entry = response
+            .data
+            .iter()
+            .find(|r| r.id == connector_refund_id)
+            .or_else(|| response.data.first());
```
JeevaRamu0104: **[S1] `.or_else(|| response.data.first())` reintroduces the hazard the RSync URL guard exists to prevent.**

When the id match misses, this adopts an arbitrary refund's status as if it belonged to the refund being synced — the precise outcome `glomopay.rs:404-419` calls out as the reason for failing loudly on an empty `request_id`.

The `None` branch immediately below already does the safe thing (`RefundStatus::Pending`, so the next sync retries). This fallback bypasses it.

Suggested fix — drop the fallback and let a miss fall through to `Pending`:
```rust
let refund_entry = response.data.iter().find(|r| r.id == connector_refund_id);
```

crates/internal/integration-tests/src/harness/connector_override/glomopay.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl ConnectorOverride for GlomopayConnectorOverride {
+    fn connector_name(&self) -> &str {
+        "glomopay"
+    }
+
+    fn transform_response(&self, suite: &str, _scenario: &str, response: &mut Value) {
+        match suite {
+            "PaymentService/Refund" => {
+                *response = serde_json::json!({
```
JeevaRamu0104: **[S1] Refund and RSync suites assert against a hardcoded stub, so they cannot fail.**

`transform_response` replaces the entire gRPC response with a fixed literal before assertions run, discarding whatever UCS actually returned. `PaymentService/Refund` and `RefundService/Get` therefore pass unconditionally — they would pass even if the refund transformers were completely broken. This is why the RSync and refund-webhook defects flagged elsewhere in this review are invisible to CI.

`connector_specs/glomopay/specs.json` nonetheless lists both as `supported_suites`, so coverage tooling reports them as exercised.

This is categorically broader than the three existing overrides, which normalize only what the sandbox genuinely cannot produce while still asserting on real connector behavior — `helcim.rs` promotes a non-standard transaction-id field, and `redsys.rs` generates a valid order-id format while explicitly noting:

> *"The production connector transformer correctly rejects anything that doesn't match — that's real validation we don't want to relax just to make tests pass."*

Either narrow this to the specific sandbox limitation, or remove `PaymentService/Refund` and `RefundService/Get` from `supported_suites` so the coverage signal stays honest.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+    for RouterDataV2<Authorize, PaymentFlowData, PaymentsAuthorizeData<T>, PaymentsResponseData>
+{
+    type Error = error_stack::Report<errors::ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<GlomopayAuthorizeResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let response = item.response;
+        let status = match response.status {
+            GlomopayPaymentStatus::ActionRequired => AttemptStatus::Failure,
```
JeevaRamu0104, line 869: **[S1] `action_required` maps to `Failure` here but `Pending` on every other path.**

Three code paths disagree on the same connector status:

| Path | `action_required` maps to |
|---|---|
| Authorize (this line) | `AttemptStatus::Failure` |
| PSync (`From` impl, `transformers.rs:125`, used at `:945`) | `AttemptStatus::Pending` |
| Webhook (`transformers.rs:222`) | `AttemptStatus::Pending` |

The Authorize response also contradicts itself: `redirection_data` is still populated from `next_steps` at `:869-880`, so it marks the attempt terminally failed while handing back a redirect for the customer to follow.

The failure scenario: Glomopay returns `action_required` on Authorize, UCS records `Failure`, the customer completes the redirect and is charged, and the subsequent webhook reports `Pending` for an attempt already recorded as terminally failed.

If Authorize-time `action_required` genuinely differs in meaning from the sync/webhook one, could you document why? Otherwise please align it with the `From` impl.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+    }
+
+    pub fn into_refund_webhook_details_response(
+        self,
+        http_code: u16,
+        raw_body: &[u8],
+    ) -> RefundWebhookDetailsResponse {
+        RefundWebhookDetailsResponse {
+            connector_refund_id: Some(self.data.id.clone()),
+            merchant_transaction_id: self.data.payment_id.clone(),
```
JeevaRamu0104: **[S1] Connector-side payment id assigned to `merchant_transaction_id`.**

`data.payment_id` is Glomopay's own identifier (`payt_...`), not a merchant-supplied one. `glomopay.rs:549` maps this exact field to `connector_transaction_id` in `get_webhook_event_reference` — so the same value is treated as a connector id in one place and a merchant id in the other. Per `RefundWebhookDetailsResponse`, this line is the incorrect one, and it will misidentify the refund's parent transaction downstream.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+            status: self.get_refund_status(),
+            connector_response_reference_id: Some(self.data.id),
+            error_code: None,
```
JeevaRamu0104: **[S2] Refund webhook never surfaces error details, even on `Failed`.**

`error_code` and `error_message` are hardcoded to `None`, so a failed refund webhook reaches consumers as an opaque failure. The payment webhook does this correctly at `transformers.rs:232-241`, promoting `error_code` / `error_message` / `error_description` on the `Failed` branch — worth mirroring here.

crates/internal/integration-tests/src/connector_specs/glomopay/specs.json
```diff
@@ -0,0 +1,12 @@
+{
+  "connector": "glomopay",
+  "supported_suites": [
+    "CustomerService/Create",
+    "PaymentService/CreateOrder",
+    "PaymentService/Authorize",
+    "PaymentService/Get",
+    "EventService/HandleEvent",
+    "PaymentService/Refund",
+    "RefundService/Get"
```
JeevaRamu0104, line 11: **[S2] `CustomerService/Get` missing from `supported_suites`.**

This PR implements the `GetConnectorCustomer` flow and `glomopay.rs:449` returns `should_get_connector_customer() -> true`, so the lookup runs in the composite authorize path — but the spec doesn't declare it, leaving the flow unexercised and under-reported by `check_connector_specs`.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+        let router_data = wrapper.router_data;
+
+        let amount = GlomopayAmountConvertor::convert(
+            router_data.request.minor_refund_amount,
+            router_data.request.currency,
+        )?;
+
+        Ok(Self {
+            payment_id: router_data.request.connector_transaction_id.clone(),
+            reason: "Requested By Customer".to_string(),
```
JeevaRamu0104: **[S2] Hardcoded refund reason discards the caller's value.**

`RefundsData` carries `reason: Option<String>`; every refund is sent to Glomopay as "Requested By Customer" regardless of what the merchant supplied. Suggest forwarding `router_data.request.reason` and falling back to the constant only when it is `None`.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+// ============================================================================
+// Authorize
+// ============================================================================
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum GlomopaySequence {
+    Initial,
+    Subsequent,
```
JeevaRamu0104, line 714: **[S3] `GlomopaySequence::Subsequent` is never constructed.**

`sequence` is hardcoded to `Initial` at `transformers.rs:843`. The phase-2 comment there explains the intent, so this is fine to leave — just flagging that the variant is dead until MIT/mandate support lands.

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
srishtijuspay: THe action_required case comes when the docuemnt upload is required from merchant on the glomopay dashboard , jf this is hte case we will get the status as action required in the create order itself and we will not initiate the payemnt , it is here just for safety guard . Also it is confimred frim the glomopay that the redirect url will not come if the status is action_required (so the payment will not by initiated to glomopay), if we gwt this as status in the psync then we cna keep it in pending untill they send us the success or fialure webhook

crates/internal/integration-tests/src/harness/connector_override/glomopay.rs
srishtijuspay: Removed it , I'd added for local testing with our mock server, not a real sandbox limitation.

crates/types-traits/grpc-api-types/proto/payment.proto
srishtijuspay: Kept it non-optional to match the convention used by other enums in this file (DocumentKind, PaymentStatus, etc.). The CUSTOMER_LOOKUP_UNSPECIFIED = 0 sentinel already serves as the "not set" signal â the consumer in payments.rs errors on that case, so the contract is still enforced.

crates/integrations/connector-integration/src/connectors/glomopay.rs
srishtijuspay: Sure I have removed the fisrt element matching check but not matching against the "connector_transaction_id" as we only rely on the euler side identifier for payment sync (send as request_id in the url)

JeevaRamu0104 (dismissed):

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
```diff
@@ -0,0 +1,1163 @@
...
+    type Error = error_stack::Report<errors::ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<GlomopayAuthorizeResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let response = item.response;
+        let status = match response.status {
+            GlomopayPaymentStatus::ActionRequired => AttemptStatus::Failure,
+            other => AttemptStatus::from(other),
+        };
```
JeevaRamu0104, line 871: **[S0] `action_required` is mapped to a terminal `Failure` in Authorize.**

This overrides the canonical `From` impl (which correctly maps `ActionRequired => Pending`, line 124) with `ActionRequired => AttemptStatus::Failure`.

`AttemptStatus::Failure` is terminal, so the attempt is abandoned with no PSync retry — yet the next few lines extract the 3DS redirect URL and attach it to this same response. The customer gets sent to authenticate, Glomopay charges the card, and UCS has recorded a failure. That's a charged-but-recorded-failed outcome.

Glomopay has a distinct `Failed` variant, so `action_required` is not a connector-declared failure — it's the state that *needs* customer authentication.

**Fix:** delete this override and map `ActionRequired => AttemptStatus::AuthenticationPending` in the canonical `From` impl, letting Authorize / PSync / webhook share one mapping. The same bug exists in CreateOrder (`transformers.rs:664-688`), where `action_required` is treated as terminal alongside `failed`/`expired`, producing false declines before authorization.

sdk/javascript/src/payments/_generated_grpc_client.ts
```diff
@@ -600,6 +602,11 @@ export class GrpcCustomerClient {
     return callGrpc(this.ffi, this.config, "customer/customer_create",
       req, types.CustomerServiceCreateRequest, types.CustomerServiceCreateResponse);
   }
+  /** CustomerService.Get — Retrieves customer details from the payment processor. Callers typically use this before Create to implement get-or-create semantics for connectors that reject duplicates (e.g. Glomopay). */
+  async customerGet(req: unknown): Promise<unknown> {
+    return callGrpc(this.ffi, this.config, "customer/customer_get",
+      req, types.CustomerServiceGetRequest, types.CustomerServiceGetResponse);
```
JeevaRamu0104, line 609: **[S1] The SDKs ship a `customerGet` method that cannot execute.**

Uncommenting `rpc Get` in `services.proto` regenerated all four gRPC clients (this one plus the Python / Java / Rust peers), each routing to `"customer/customer_get"`. But this PR touches **zero** files under `crates/ffi/`, and that route is registered nowhere — there are no `customer_get_req_transformer` / `customer_get_res_transformer` entries in `crates/ffi/ffi/src/services/payments.rs`, so the flow never made it into `_generated_ffi_flows.rs`.

Dispatch is a runtime string lookup, so this compiles cleanly and fails only when a caller actually invokes it.

Corroborating evidence that the flow is unwired end-to-end: `data/field_probe/glomopay.json` has no `customer_get` key, and `docs-generated/all_connector.md` consequently reports `Customer.Get = x` (Not Supported) for Glomopay — the very connector this flow was built for.

**Fix:** add the `customer_get` req/res transformers in `crates/ffi/ffi/src/services/payments.rs` mirroring the `customer_create` pair (with `flow_marker: GetConnectorCustomer` and `generate_response_fn: generate_get_connector_customer_response`), then re-run `make generate`. Alternatively, state explicitly that the flow is gRPC-only and keep it out of the generated SDK surface.

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -315,6 +334,40 @@ pub struct Payments {
...
+        flow_marker: CreateConnectorCustomer,
+        resource_common_data_type: PaymentFlowData,
+        request_data_type: ConnectorCustomerData,
+        response_data_type: ConnectorCustomerResponse,
+        request_data_constructor: ConnectorCustomerData::foreign_try_from,
+        common_flow_data_constructor: PaymentFlowData::foreign_try_from,
+        generate_response_fn: generate_create_connector_customer_response,
+        connector_data_type: ConnectorData<DefaultPCIHolder>,
+        all_keys_required: None
+    );
```
JeevaRamu0104, line 352: **[S1] This macro migration silently drops per-request connector config overrides.**

The inline body this replaces called `utils::connectors_with_connector_config_overrides(connector_config, &config)` before constructing `PaymentFlowData`. The macro instead passes `config.connectors.clone()` raw (`crates/grpc-server/grpc-server/src/utils.rs:950`), so the merchant's `connector_config` override never reaches `PaymentFlowData`. Override call-sites in this file went from 8 on `main` to 7 — this is the one that was lost.

That matters here because `connector_config_override_patch()` supplies the merchant's `base_url`, and Glomopay resolves its endpoint from `resource_common_data.connectors.glomopay.base_url`. Composite Authorize still applies the override (`payments.rs:646`), so a merchant with a `base_url` override gets **Authorize on the override host while CREATE/GET customer go to the TOML default** — split-brain inside a single composite payment, potentially sandbox vs production.

**Fix:** thread `metadata_payload.connector_config` through the macro's `common_flow_data_constructor` call so it matches the behavior of the code it replaced. The same applies to the `internal_get_connector_customer` invocation just below.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4399,7 +4400,18 @@ message CustomerServiceGetRequest {
...
+  optional string connector_customer_id = 2;
+  optional SecretString email = 3;         // Email address of the customer
+  optional SecretString phone_number = 4;  // Phone number of the customer
+}
+
+// Outcome of a connector-side customer lookup. Explicit signal so callers
+// don't have to infer presence-of-id vs absent as found-vs-not-found, which
+// would silently misroute malformed responses into a duplicate CREATE.
+enum CustomerLookupStatus {
+  CUSTOMER_LOOKUP_UNSPECIFIED = 0;
```
gopikrishna000: can we have uniform naming 
CUSTOMER_LOOKUP_STATUS_UNSPECIFIED
and all member to have one one prefix (say 'CUSTOMER_LOOKUP_STATUS_')

crates/internal/composite-service/src/payments.rs
```diff
@@ -418,6 +427,46 @@ where
...
+        let get_response = self.customer_service.get(get_request).await?.into_inner();
+
+        // Consume the explicit lookup_status enum rather than inferring
+        // found-vs-not-found from field presence. A malformed or partially
+        // populated response no longer silently routes to a duplicate CREATE.
+        match grpc_api_types::payments::CustomerLookupStatus::try_from(get_response.lookup_status) {
+            Ok(grpc_api_types::payments::CustomerLookupStatus::CustomerFound) => {
+                CustomerServiceCreateResponse::foreign_try_from(get_response)
+                    .map(|r| Ok(ConnectorCustomerLookup::Found(Box::new(r))))
+                    .unwrap_or(Ok(ConnectorCustomerLookup::NotFound))
```
gopikrishna000: why unwrap_or(

crates/internal/composite-service/src/payments.rs
```diff
@@ -418,6 +427,46 @@ where
...
+            }
+            Ok(grpc_api_types::payments::CustomerLookupStatus::CustomerNotFound) => {
+                Ok(ConnectorCustomerLookup::NotFound)
+            }
+            _ => Err(tonic::Status::internal(
```
gopikrishna000: should we explicitly match over unspecified variant

crates/integrations/connector-integration/src/connectors/glomopay/transformers.rs
srishtijuspay: Kept the Failure mapping intentionally. action_required at Glomopay means the merchant needs to upload compliance docs on the Glomopay dashboard before payment can proceed â it requires merchant intervention followed by a fresh transaction, not customer authentication or auto-resolution. Pending / AuthenticationPending would break this retry semantic since both imply the current attempt will resolve.

 The charged-but-failed scenario is guarded by Glomopay's contract that no redirect step is included in next_steps when status is action_required  confirmed directly with them. In practice this branch shouldn't fire at Authorize at all since action_required is caught at CreateOrder and we abort before initiating payment; the mapping here is a defensive fail-loud for the unexpected case.

crates/types-traits/grpc-api-types/proto/payment.proto
gopikrishna000: DocumentKind, PaymentStatus are enum field they are non optional cause it has a 0-sentinel for "unset".
But here Customer struct is genuinely absent on NOT_FOUND, a bare Customer customer misleads a proto reader into thinking it's always populated.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7804,7 +7808,15 @@ pub fn generate_payment_sync_response(
                         connector_transaction_id: e.connector_transaction_id.clone(),
                         status: None,
                     }),
-                    issuer_details: Some(grpc_payment_types::IssuerErrorDetails::from(&e)),
+                    issuer_details: Some(grpc_payment_types::IssuerErrorDetails {
```
gopikrishna000: why removed Some(grpc_payment_types::IssuerErrorDetails::from(&e)),
is it better to update the From impl

crates/integrations/connector-integration/src/connectors/glomopay.rs
```diff
@@ -343,7 +343,21 @@ macros::macro_connector_implementation!(
...
+            // Fail loudly on an empty reference id. Dropping the ?request_id
+            // filter would return every payment on the merchant's account and
+            // the response transformer would silently pick the first one —
+            // effectively reporting an unrelated payment's status as if it
+            // belonged to the current one.
+            if request_id.is_empty() {
+                Err(error_stack::report!(
+                    IntegrationError::MissingRequiredField {
+                        field_name: "connector_request_reference_id",
+                        context: Default::default(),
```
Aishwariyaa-Anand: specify error context

Aishwariyaa-Anand (commented):
include CustomerServiceGetRequest in CompositeAuthorizeRequest check at crates/internal/composite-service/tests/composite_request_schema_check.rs

Aishwariyaa-Anand (approved):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

PR #1948, opened 2026-07-15
feat(proto): add RawConnectorStatus for connector code/message/reason

gopikrishna000:
## Description

Adds a `RawConnectorStatus { code, message, reason }` message — the connector's raw status as returned — and:

- sets the field on the responses that carry a connector transaction status: `PaymentServiceAuthorizeResponse`, `PaymentServiceGetResponse` (sync), `PaymentServiceCaptureResponse`, `PaymentServiceVoidResponse`, `RefundResponse` (refund + refund-sync), `RecurringPaymentServiceChargeResponse`, `PaymentServiceSetupRecurringResponse`;
- adds the domain home on `PaymentFlowData` / `RefundFlowData` (`resource_common_data`), mirroring `settlement_status`, so any flow can populate it — connectors set it via struct-update, so none of them break;
- wires the domain→proto mapping to read `resource_common_data.raw_connector_status`, so a populated value reaches the wire.

Population per connector (setting `resource_common_data.raw_connector_status = Some(...)`) is a follow-up, done per connector as each starts reporting a code.

## Motivation and Context

On success/pending the connector still returns a status code (an approval code, a "still processing" advice) that the failure-only `error.connector_details` never carries. This adds a place for it, passed through as the connector returned it (not derived from the normalized status).

### Additional Changes

## How did you test it?

`cargo build --workspace --all-targets` — proto codegen and the full workspace compile cleanly. No connector files change (struct-update fills the new field). No runtime behaviour change yet (no transformer populates the field).


crates/types-traits/domain_types/src/types.rs
```diff
@@ -6039,6 +6046,7 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
                     .transpose()?;
 
                 PaymentServiceAuthorizeResponse {
+                    raw_connector_status: router_data_v2.resource_common_data.raw_connector_status.clone().map(|r| grpc_api_types::payments::RawConnectorStatus { code: r.code, message: r.message, reason: r.reason }),
```
Aishwariyaa-Anand: can we use proper naming instead of |r|

Aishwariyaa-Anand (dismissed):

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -110,6 +110,14 @@ message ConnectorErrorDetails {
   optional FlowStatus status = 5;
 }
 
+// The connector's reported status for this response: the code, message and
+// reason returned by the connector, when provided.
+message RawConnectorStatus {
```
JeevaRamu0104, line 115: Instead of `RawConnectorStatus` can we say `RawConnectorStatusDetails`

gopikrishna000: Thanks! I'd prefer to keep `RawConnectorStatus` (without `Details`), for a couple of reasons:

1. A proto `message` is already a group of fields — that's what a message is. Adding `Details` (which basically means "a bundle of fields") just repeats that, so it doesn't add any real meaning.

2. "Details" usually means "the detailed version of something" — it only makes sense when there's also a shorter/summary type beside it (like `X` and `XDetails`). We don't have that here, so `...Details` would hint that a plain `RawConnectorStatus` exists somewhere and this is its expanded form — which could send readers looking for a sibling that doesn't exist.

3. `RawConnectorStatus` already says exactly what it holds: the connector's raw status (code + message + reason).

Aishwariyaa-Anand (dismissed):

Aishwariyaa-Anand (approved):

JeevaRamu0104 (approved):

PR #1952, opened 2026-07-15
feat(paysafe): connector-native Card 3DS; prove Card no-3DS + Google Pay E2E

shuklatushar226:
## Summary

Paysafe on UCS: **connector-native Card 3DS**, plus **Card no-3DS** and **Google Pay** proven end-to-end HS → UCS → Paysafe. Companion HS change: hyperswitch#13352.

### Card 3DS (connector-native)
Paysafe runs 3DS itself from a `threeDs` object on the payment handle. The `ThreeDs` wire type already existed in `requests.rs` but was dead code — both construction sites hardcoded `None`. This populates it and routes a 3DS card to the handle-creation leg.

- `paysafe.rs` — Authorize `get_url` selects `v1/paymenthandles` vs `v1/payments` via `is_paysafe_handle_creation_leg`
- `transformers.rs` — new `is_paysafe_three_ds_card`; `is_paysafe_settle_leg` un-gated from APM-only so cards can reach the settle leg; the Card+3DS arm resolves the `card.three_ds` account and populates the existing `ThreeDs` type; dispatch routes card+3DS to the PaymentHandle body

UCS has no `CompleteAuthorize` flow, so HS's Authorize + CompleteAuthorize split is modelled as **two legs of one Authorize** (the pre-existing redirect-APM mechanism), extended to cards. `device_channel` is `BROWSER`: UCS carries no client-platform signal, so this matches the value HS's own `None` branch emits.

### Google Pay
Route the Google Pay payment-handle creation to **`v1/paymenthandles`** (was `v1/singleusepaymenthandles` → `403 / 5270 Unauthorized access`). Verified directly against Paysafe: the identical Google Pay decrypted-token body 403/5270s on `singleusepaymenthandles` but is created successfully (`status: PAYABLE`) on `v1/paymenthandles`. Same endpoint as Apple Pay and Skrill.

> Note: Google Pay on Paysafe is the **decrypted** path — HS decrypts the token and forwards `GPayPredecryptData`; UCS builds the `googlePay.googlePayPaymentToken...decryptedToken` handle body. End-to-end also needs the companion HS fix (hyperswitch#13352) that forwards the Tokenize handle token into the settle Authorize for wallets.

## How to test

Local HS → UCS stack, Paysafe in `ucs_only_connectors`. Route each payment straight to Paysafe with `routing.single`.

**1. Create the Paysafe merchant connector account**

```bash
curl --location 'localhost:8080/account/<merchant_id>/connectors' \
--header 'api-key: <api-key>' \
--data '{
  "connector_type": "payment_processor",
  "connector_name": "paysafe",
  "connector_account_details": { "auth_type": "BodyKey", "api_key": "<paysafe_api_key>", "key1": "<paysafe_key1>" },
  "test_mode": true,
  "disabled": false,
  "metadata": { "account_id": { "card": { "USD": { "three_ds": "<3ds_account_id>", "no_three_ds": "<no_3ds_account_id>" } } } },
  "connector_wallets_details": {
    "google_pay": {
      "provider_details": { "merchant_info": { "merchant_name": "juspay",
        "tokenization_specification": { "type": "DIRECT", "parameters": { "private_key": "<ec_p256_pkcs8_b64>", "recipient_id": "merchant:juspay" } } } },
      "cards": { "allowed_auth_methods": ["PAN_ONLY","CRYPTOGRAM_3DS"], "allowed_card_networks": ["VISA","MASTERCARD","AMEX","DISCOVER"] }
    }
  },
  "payment_methods_enabled": [
    { "payment_method": "card", "payment_method_types": [ { "payment_method_type": "credit" }, { "payment_method_type": "debit" } ] },
    { "payment_method": "wallet", "payment_method_types": [ { "payment_method_type": "google_pay", "payment_experience": "invoke_sdk_client" } ] }
  ],
  "business_country": "US",
  "business_label": "default"
}'
```

**2. Card — no 3DS**

```bash
curl --location 'localhost:8080/payments' \
--header 'api-key: <api-key>' \
--data '{
  "amount": 1000, "currency": "USD", "confirm": true, "capture_method": "automatic",
  "profile_id": "<profile_id>", "customer_id": "cust_card_test",
  "payment_method": "card", "payment_method_type": "credit", "authentication_type": "no_three_ds",
  "payment_method_data": { "card": {
    "card_number": "4000000000001091", "card_exp_month": "12", "card_exp_year": "2027", "card_cvc": "123", "card_holder_name": "Test User"
  } },
  "billing": { "address": { "line1": "100 Main St", "city": "New York", "state": "NY", "zip": "10001", "country": "US", "first_name": "Test", "last_name": "User" }, "email": "test@example.com" },
  "routing": { "type": "single", "data": { "connector": "paysafe", "merchant_connector_id": "<mca_id>" } }
}'
```

**3. Google Pay** — `<google_pay_token>` is the raw token from a Google Pay TEST sheet (DIRECT ECv2 to the recipient key above); mint fresh (single-use, short-lived) and JSON-escape it as the string value.

```bash
curl --location 'localhost:8080/payments' \
--header 'api-key: <api-key>' \
--data '{
  "amount": 1000, "currency": "USD", "confirm": true, "capture_method": "automatic",
  "profile_id": "<profile_id>", "customer_id": "cust_gpay_test",
  "payment_method": "wallet", "payment_method_type": "google_pay",
  "payment_method_data": { "wallet": { "google_pay": {
    "type": "CARD", "description": "Visa ****1234",
    "info": { "card_network": "VISA", "card_details": "1234" },
    "tokenization_data": { "type": "PAYMENT_GATEWAY", "token": "<google_pay_token>" }
  } } },
  "billing": { "address": { "line1": "100 Main St", "city": "New York", "state": "NY", "zip": "10001", "country": "US", "first_name": "Test", "last_name": "User" }, "email": "test@example.com" },
  "routing": { "type": "single", "data": { "connector": "paysafe", "merchant_connector_id": "<mca_id>" } }
}'
```

## Results (Paysafe sandbox, via HS → UCS)

| Flow | Path | Result |
|---|---|---|
| Card no-3DS | Tokenize `POST /v1/paymenthandles` (handle `PAYABLE`) → Authorize settle | **succeeded** |
| Card 3DS (challenge) | handle `INITIATED` + hosted ACS → shopper completes OTP → `PAYABLE` → settle | **COMPLETED** (authCode `553323`) |
| Google Pay | HS decrypts → Tokenize `POST /v1/paymenthandles` (`PAYABLE`) → Authorize forwards the handle token → settle | **succeeded** (`pay_fZ2haK2seRDaDVWMu64b`) |


deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/connectors/paysafe/transformers.rs
```diff
@@ -250,10 +260,76 @@ where
...
+                    Some(req_card.card_cvc.clone())
+                },
+                holder_name: req_card.card_holder_name.clone().or_else(|| {
+                    router_data
+                        .resource_common_data
+                        .get_optional_billing_full_name()
+                }),
+            };
+            // Paysafe rejects a `threeDs` body on a non-3DS account (error 5040); use 3DS account.
+            let account_id = account_id.get_account_id(PaysafeAccountKind::CardThreeDs, currency)?;
```
AmitsinghTanwar007: why account kink is hardcoded how you are deciding is it threeds or not

shuklatushar226: Good catch — the hardcoded `CardThreeDs` did read like a runtime 3DS decision, which it isn't. That `Card` arm is actually **unreachable**: this builder (`PaysafeSetupMandateRequest` from `Authorize`) is only entered via `is_paysafe_handle_creation_leg` (`transformers.rs:1672`), which is true only for redirect APMs — Skrill / Interac / paysafecard (`is_paysafe_redirect_apm`, `:153-164`; cards hit the `_ => false` arm). So a card never reaches line 310, and no card account-kind is ever chosen here.

The actual 3DS-vs-no-3DS decision lives in two places:

- `next_authentication_step` (`paysafe.rs:109`) routes `AuthenticationType::ThreeDs` + Card to the **PreAuthenticate** flow, which mints the `threeDs` handle in the PreAuthenticate builder (`transformers.rs:617`, `CardThreeDs`). That builder rejects non-cards (`:586`), so it's 3DS-card-only by construction.
- **No-3DS cards** mint via the `PaymentMethodToken` flow (`transformers.rs:939`, `CardNoThreeDs`) and settle through `PaysafePaymentsRequest`, which branches on `is_three_ds()` → `CardThreeDs` / `CardNoThreeDs` (`transformers.rs:1600-1610`).

So a no-3DS card never touches this arm and can't hit Paysafe's `5040`.

I've pushed `2c3da36` which replaces this dead `Card` arm with an explicit `NotImplemented` (mirroring the PreAuthenticate builder's non-card guard at `:586`), so the account-kind is never assumed in this builder and the question can't resurface.

AmitsinghTanwar007 (dismissed):

deepanshu-iiitu (dismissed):

AmitsinghTanwar007 (commented):
## L2 (Domain Types) Review — `crates/types-traits/domain_types/src/types.rs`

### 1. `payment_method_type` — silent error swallow (2 locations)

```rust
payment_method_type: payment_method_clone
    .and_then(|pm| <Option<PaymentMethodType>>::foreign_try_from(pm).ok())
    .flatten(),
```

This is in two shared `ForeignTryFrom` impls (PreAuthenticate + Authenticate data construction). A genuinely malformed proto value gets silently converted to `None` — no log, no trace. This affects every connector using these flows, not just Paysafe.

**Suggestion:** At minimum add a `tracing::warn!` before `.ok()` so malformed values surface in production. Better: only suppress when the field is actually absent (proto default/empty), and still error on a populated-but-invalid value.

### 2. `PaymentMethod` defaulted to `Card`

```rust
payment_method: value
    .payment_method
    .and_then(|pm| PaymentMethod::foreign_try_from(pm).ok())
    .unwrap_or(PaymentMethod::Card),
```

This is in the Authenticate `PaymentFlowData` construction — shared by ALL connectors. If any future connector uses Authenticate for a non-card flow, it silently becomes `Card`. The comment says "Authenticate transformers ignore it" but that's a Paysafe-local claim on a global path.

**Suggestion:** Either make the field `Option<PaymentMethod>` so it can be `None`, or keep the error and handle Paysafe's missing-PM case inside the Paysafe connector transformer instead of the shared domain layer.

### 3. Error message generalization — ✅ OK

Removing the Razorpay-specific doc_url and message is correct for a multi-connector platform. This change is fine.

shuklatushar226:
Thanks for the L2 review — both points were real. Addressed in `e51f9c8af`.

**1. `payment_method_type` silent error swallow (PreAuthenticate + Authenticate):** Went with your "better" option rather than the `tracing::warn!` minimum. The `.ok()` was there to tolerate the Paysafe post-redirect re-fetch legs, which legitimately carry no payment method — but it also absorbed populated-but-invalid values (e.g. `CardRedirect` with `Unspecified` type). Both sites now use:

```rust
payment_method_type: payment_method_clone
    .filter(|pm| pm.payment_method.is_some())
    .map(<Option<PaymentMethodType>>::foreign_try_from)
    .transpose()?
    .flatten(),
```

so absent **or proto-default (empty oneof)** → `None`, while any populated-but-invalid value propagates the error exactly as before this PR. Nothing is suppressed that isn't semantically "not provided", so no warn-and-continue path remains. This also brings the two sites in line with the PostAuthenticate conversion in this same PR, which already used the `.map(...).transpose()?` shape.

**2. `PaymentMethod` defaulted to `Card` (Authenticate `PaymentFlowData`):** Same narrowing applied — the `Card` fallback now only fires when the field is absent/empty; a populated-but-invalid value errors:

```rust
payment_method: value
    .payment_method
    .filter(|pm| pm.payment_method.is_some())
    .map(PaymentMethod::foreign_try_from)
    .transpose()?
    .unwrap_or(PaymentMethod::Card),
```

On your two suggested alternatives:
- **`Option<PaymentMethod>`** is the right long-term fix, but `PaymentFlowData.payment_method` (`connector_types.rs:710`) is non-optional and read by every connector, so Option-izing it is a cross-cutting change I'd rather not smuggle into a connector PR. I've left a comment at the site marking it as the follow-up.
- **Handling it in the Paysafe transformer** isn't reachable: this conversion runs in the shared domain layer *before* any connector transformer executes, so if the missing PM errored here, the Authenticate re-fetch would fail before Paysafe code ever ran. The tolerance has to live at the conversion site.

The residual risk is now narrow: a future connector using Authenticate only sees `Card` if its request omits `payment_method` entirely — in which case there is no value to mis-map — and any real value (valid or invalid) is handled loudly.

**3.** Agreed, no change needed.

`cargo check` + `clippy` clean on `domain_types`.

AmitsinghTanwar007 (approved):

AmitsinghTanwar007 (approved):

deepanshu-iiitu (approved):

PR #1955, opened 2026-07-16
feat(connector): [Absa] Add EFT payouts support for absa_sanlam

Saptak88:
### Additional Changes


XyneSpaces:
[blocking] The new payout enqueue path derives the Kafka topic from `base_url`, which can produce invalid topic names such as `https://api.absa.co.za_payouts_queue`. Kafka topic names cannot contain URL separators like `:` and `/`, so a configured HTTPS base URL can make payouts fail before they are enqueued.

Please use a dedicated configured topic name, or derive a short sanitized identifier rather than using the full base URL.

XyneSpaces:
[blocking] Please include the generated Rust fallout for the `payouts.proto` changes. This PR adds `EftBankTransferPayout`, `PayoutMethod.eft`, and `PayoutServiceTransferRequest.metadata`, and the domain conversion now references `grpc_api_types::payouts::EftBankTransferPayout`; without the matching generated `grpc-api-types/src` update in the same PR this will either fail to compile or ship stale wire types.

PR #1962, opened 2026-07-17
docs: add blog on the low-level design behind Prism's connector architecture

JeevaRamu0104:
Adds `docs/blogs/how-we-used-low-level-design-to-build-a-payments-library.md`.

## What this is

A blog post walking through the low-level design of the connector architecture, using real code from this repo. It opens with the problem — 97 processors that all do the same job and none agree how — then covers:

- **OOP Fundamentals** — encapsulation (`CardNumber` + the uninhabited `CardNumberStrategy` masking tag), enums as sum types (`ConnectorSpecificConfig`), interfaces/abstraction, inheritance via supertraits
- **Class Relationships** — association, aggregation, composition, dependency, realisation
- **UML** — class, sequence, and state-machine diagrams (ASCII, no image deps)
- **Design Patterns** — 15 patterns with code: Factory Method, Abstract Factory, Singleton, Builder, Adapter, Bridge, Facade, Decorator, Chain of Responsibility, Flyweight, Strategy, Template Method, Null Object, State, Observer

## Notes for review

- **Docs only.** No code changes. Not added to `SUMMARY.md`, matching the other posts in `docs/blogs/` — none of them appear in the gitbook nav.
- **Citations are permalinks pinned to `b406d5ee2`** (current `origin/main`), not `main`. Pinning keeps the `#L` anchors pointing at the code the prose describes; a `main` link would drift within weeks. All 31 links verified to return 200, and every cited line number was checked against that SHA specifically.
- **Ends with the `**Reference:**` line** matching `authorization-to-settlement.md`.

### DIFF
--- a/docs/blogs/how-we-used-low-level-design-to-build-a-payments-library.md
+++ b/docs/blogs/how-we-used-low-level-design-to-build-a-payments-library.md
@@ -1,0 +1,759 @@
+# How We Used Low-Level Design to Build a Payments Library for 100+ Processors
+
+Every payment processor does the same job. Take money from a card, hold it, capture it, refund it. None of them agree on how.
+
+Stripe calls it a PaymentIntent. Adyen calls it a payment. PayPal calls it an order. One returns `"Captured"`, another returns `"SETTLED"`, a third returns `"1"`. Prism supports **97 of them** behind a single API.
+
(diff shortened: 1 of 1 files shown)


XyneSpaces:
[should-fix] Fix the new blog’s source references before merging. The factory-method link points at `types.rs#L30` instead of the actual `get_connector_by_name` function around `#L44`, the decorator link points at the middleware struct instead of the `call` method around `#L53`, and the adapter prose says `TryFrom` while the shown/source example is `impl From<GlobalpayPaymentStatus> for AttemptStatus`.

PR #1963, opened 2026-07-17
feat(connectors): [interpayments] add surcharge strategy to get transaction fee request

AkshayaFoiger:
## Description

add surcharge strategy to get transaction fee request


XyneSpaces:
[should-fix] Omit `data` when no surcharge strategy needs to be sent. With this new `Option<Vec<_>>` field and no `skip_serializing_if`, the default `Apply`/absent-strategy path serializes as `"data": null` instead of preserving the previous request shape, which can be rejected by APIs expecting the optional array to be absent.

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub data: Option<Vec<InterpaymentsSurchargeStrategy>>,
```

PR #1964, opened 2026-07-19
fix(stripe): include on_behalf_of key in mandate_metadata for Authorize/Capture

### LINKED ISSUE #1965: [shadow-validation] stripe/capture: UCS↔HS request diff
Shadow-validation flagged a connector-request diff for **stripe / capture**.

`router.keyDiff:response.Ok.TransactionResponse.mandate_reference.mandate_metadata.on_behalf_of`

Fixed by UCS PR #1964.

10X-GRACE:
## What
Adds the missing `on_behalf_of` key to the `mandate_metadata` JSON built in Stripe's shared
`PaymentIntentResponse` → `RouterDataV2` mapping (used by both **Authorize** and **Capture**).

## Why
Shadow-validation reported a `keyDiff` on
`response.Ok.TransactionResponse.mandate_reference.mandate_metadata.on_behalf_of` for
`stripe`/`capture`: HS's `mandate_metadata` JSON always includes an `on_behalf_of` key
(populated or `null`), but UCS's Authorize/Capture path omitted the key entirely — it was
only being added on UCS's `SetupMandate` response path.

## Root cause
`crates/integrations/connector-integration/src/connectors/stripe/transformers.rs`, the
`json!` block that builds `mandate_metadata` for the `PaymentIntentResponse` impl (shared by
Authorize/Capture) was missing the `on_behalf_of` field that the corresponding HS transformer
(`crates/hyperswitch_connectors/.../stripe/transformers.rs`, both the PaymentIntentResponse
and SetupIntentResponse impls) and UCS's own `SetupMandateResponse` impl already include.

## Fix
One-line addition to the `json!` block — mirrors the existing `SetupMandate` block a few
hundred lines below in the same file:

```rust
Some(Secret::new(serde_json::json!({
    "transfer_account_id": stripe_split_data.transfer_account_id,
    "charge_type": stripe_split_data.charge_type,
    "application_fees": stripe_split_data.application_fees,
    "on_behalf_of": stripe_split_data.on_behalf_of,
})))
```

## Verification
Verified locally with a direct UCS gRPC `Capture` A/B test (Stripe captures aren't
idempotent, so shadow-diff replay isn't usable — same approach as prior capture
mandate_metadata fixes in this family):

- **Before**: capture response `mandate_reference.mandate_metadata` JSON had no `on_behalf_of`
  key at all when built with `splitPayments` set.
- **After**: capture response `mandate_reference.mandate_metadata` now includes
  `"on_behalf_of"` (matching HS's shape, `null` when not supplied in the split payment
  request, populated when supplied) — key now present on both sides, diff resolved.
- Capture without `splitPayments` unaffected on both legs (`mandate_metadata` stays `None`).

Local checks run clean: `cargo build`, `cargo clippy -p connector-integration --all-targets -- -D warnings`, `cargo fmt --all -- --check`, `cargo test -p connector-integration`.

## Category
L3 (UCS connector transformer only) — no proto/domain_types change needed, the field already
existed on the request side; this only fixes the response-side JSON shape.

Closes #1965


Shubhodip900:
changes are merged via https://github.com/juspay/hyperswitch-prism/pull/1972 PR

PR #1966, opened 2026-07-20
fix(nmi): drop customer-email fallback for email/shipping_email on repeat_payment (MIT)

10X-GRACE:
## What

NMI `repeat_payment` (MIT) request builder no longer falls back to the customer email for `email` / `shipping_email` — it now sources them from the billing / shipping address only, matching the hyperswitch Direct reference.

Fixes the shadow request diff in juspay/hyperswitch-cloud#20218 (parent #20217), `greenstone_rx` / `nmi` / `repeat_payment`, `235` occurrences.

## Root cause

Direct routes NMI MIT charges through the Authorize flow's `NmiPaymentsRequest` (the `MandatePayment` branch), which builds billing/shipping via `get_billing_details` / `get_shipping_details`:

- `email = get_optional_billing_email()` — billing address email only
- `shipping_email = get_optional_shipping_email()` — shipping address email only

Both structs are `#[skip_serializing_none]`, so when the address carries no email the field is **dropped from the wire body**. There is no fallback to the customer email.

Prism's `NmiRepeatPaymentRequest` builder (added in #1682) instead did:

```rust
email: common_data.get_optional_billing_email().or_else(|| router_data.request.email.clone()),
shipping_email: common_data.get_optional_shipping_email().or_else(|| router_data.request.email.clone()),
```

For a repeat charge whose billing/shipping addresses have **no** email but a customer-level email is set, prism populated `email`/`shipping_email` from the customer email while Direct sent nothing — producing:

```
body.keyDiff.email          = {"hyperswitch":false,"ucs":true}
body.keyDiff.shipping_email = {"hyperswitch":false,"ucs":true}
```

## Fix

Remove the `.or_else(|| router_data.request.email.clone())` fallback on both fields so the MIT request mirrors Direct byte-for-byte.

Reproduced end-to-end through the router → shadow-UCS → validation-service (`nmi_repro` merchant, CIT setup-mandate then MIT repeat_payment; billing address with **no** email + a top-level customer email, no shipping — the exact prod scenario).

**BEFORE (this branch reverted to `main`)** — `diff_result_repeat_payment:nmi:019f7cda-459d-76b1-a13c-fa7ceabc6b2d`:
```json
"bodyComparison": {
  "keyDiff": {
    "email":          {"hyperswitch": false, "ucs": true},
    "shipping_email": {"hyperswitch": false, "ucs": true}
  },
  "valueDiff": {}, "typeDiff": {}
}
```
(exactly the #20218 signature)

**AFTER (this branch)** — `diff_result_repeat_payment:nmi:019f7cdb-3a91-7162-bdbd-639ec3e3a633`:
```json
"bodyComparison": { "keyDiff": {}, "valueDiff": {}, "typeDiff": {} }
```
`bodyComparison.keyDiff == {}` ✅ (only the ignore-listed `via` / `x-merchant-id` headers remain in headerComparison, as expected). MIT charge succeeded (nmi `12323073431`).

Regression fixture: `fixtures/nmi/repeat_payment/20218.json`.

## Classification

UCS-only (prism transformer). No proto or hyperswitch-client change required.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/nmi/transformers.rs
@@ -1972,12 +1972,12 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
             zip: common_data.get_optional_billing_zip(),
             country: common_data.get_optional_billing_country(),
             phone: common_data.get_optional_billing_phone_number(),
-            // Prefer the billing-address email (mirrors the hyperswitch reference
-            // NMI MIT path), falling back to the top-level `RepeatPaymentData.email`
-            // so the customer email is still sent when no billing email is present.
-            email: common_data
-                .get_optional_billing_email()
-                .or_else(|| router_data.request.email.clone()),
+            // Source `email` from the billing address only, matching the hyperswitch
+            // reference NMI MIT path (`get_billing_details` -> `get_optional_billing_email`).
+            // No fallback to the top-level `RepeatPaymentData.email`: the reference omits
+            // the field entirely when the billing address carries no email, so falling
+            // back to the customer email would send `email` when Direct sends nothing.
+            email: common_data.get_optional_billing_email(),
             shipping_details: Some(NmiShippingDetails {
                 shipping_firstname: common_data.get_optional_shipping_first_name(),
                 shipping_lastname: common_data.get_optional_shipping_last_name(),
@@ -1987,11 +1987,10 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                 shipping_state: common_data.get_optional_shipping_state(),
                 shipping_zip: common_data.get_optional_shipping_zip(),
                 shipping_country: common_data.get_optional_shipping_country(),
-                // Same precedence for the shipping email: shipping-address email
-                // first, then the top-level `RepeatPaymentData.email` fallback.
-                shipping_email: common_data
-                    .get_optional_shipping_email()
-                    .or_else(|| router_data.request.email.clone()),
+                // Same as billing: shipping email comes from the shipping address only,
+                // matching the reference (`get_shipping_details` -> `get_optional_shipping_email`),
+                // with no `RepeatPaymentData.email` fallback.
+                shipping_email: common_data.get_optional_shipping_email(),
             }),
         })
     }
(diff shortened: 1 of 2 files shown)


10X-GRACE:
### Re-validated against issue hyperswitch-cloud#20218 (nmi / repeat_payment / greenstone_rx)

Rebased/confirmed on latest `main` (0 commits behind) and re-ran the full shadow reproduce → verify loop locally (router `:8086` → prism `:8000` shadow → validation service `:9100`). Both prism binaries (main vs this branch) were rebuilt and restarted so the before/after is non-vacuous.

**Repro scenario** (`fixtures/nmi/repeat_payment/20218.json`): CIT (`setup_future_usage=off_session` + `customer_acceptance`) → MIT/repeat_payment with a **billing address carrying no email** + a **top-level customer `email`** set, no shipping. This is exactly the condition under which the removed `.or_else(|| request.email.clone())` fallback fired.

#### BEFORE — `main` (fallback present) · x-request-id `019f7ce4-e506-7c32-a3e9-6753266e5dd9`
```
bodyComparison.keyDiff:
  email          : {"hyperswitch": false, "ucs": true}
  shipping_email : {"hyperswitch": false, "ucs": true}
  valueDiff: {}   typeDiff: {}
```
Reproduces the exact prod signature from #20218 — UCS/prism over-sends `email` + `shipping_email`; Direct omits them.

#### AFTER — this branch (fallback removed) · x-request-id `019f7ce6-30ef-7003-8744-9c16f1879ba6`
```
bodyComparison.keyDiff  : {}
bodyComparison.valueDiff: {}
bodyComparison.typeDiff : {}
```
Full request-body parity. (Header `keyDiff` retains only the ignore-listed `via` / `x-merchant-id` headers, unrelated to this fix.)

**RCA:** Direct routes NMI MIT through the Authorize-flow `NmiPaymentsRequest` MandatePayment branch, reading email from the billing/shipping **address only** (`get_optional_billing_email` / `get_optional_shipping_email`) with no top-level fallback; `#[skip_serializing_none]` drops the field when the address has no email. Prism's `NmiRepeatPaymentRequest` had added a `.or_else(|| request.email.clone())` fallback (#1682), so it sent the customer email while Direct sent nothing. This PR removes that fallback on both fields → parity. UCS-only; no proto / hs-client change. Grafana/Loki RCA was attempted but the prod log store is unavailable, so RCA is from the diff signature + code + this reproduction.

XyneSpaces:
[blocking] This fixture embeds raw card PAN/CVC values (`4111111111111111` / `999`) in a checked-in JSON file outside the connector test-spec tree. Replace them with the repo's test-token/fixture placeholders or move this into the sanctioned connector-spec fixture format so payment credentials are not committed as literal data.

PR #1961, opened 2026-07-17
fix(connector): Added stripe moto flag in Authorize, RepeatPayment &  SetupMandate request

Shubhodip900:
## Description

Added stripe moto flag in Authorize, RepeatPayment &  SetupMandate request
### Additional Changes
- [ ] This PR modifies the API contract
- [ ] This PR modifies application configuration/environment variables

## How did you test it?

Stripe CIT:-
Request:-
```json
{
    "amount": 100,
    "currency": "USD",
    "confirm": true,
    "capture_method": "automatic",
    "capture_on": "2022-09-10T10:11:12Z",
    "customer_id": "cus_CtLmH8RPwrLgww",
    "email": "guest@example.com",
    "name": "John Doe",
    "phone": "999999999",
    "setup_future_usage": "off_session",
    "payment_channel" : "telephone_order",
    "description": "Its my first payment request",
    "authentication_type": "no_three_ds",
    "return_url": "https://example.com",
    "payment_method": "card",
    "payment_method_type": "credit",
    "payment_method_data": {
        "card": {
            "card_number": "4242424242424242",
            "card_exp_month": "01",
            "card_exp_year": "28",
            "card_holder_name": "Joseph Doe",
            "card_cvc": "123"
        }
    },
    "billing": {
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US",
            "first_name": "john",
            "last_name": "Doe"
        }
    },
    "shipping": {
        "address": {
            "line1": "1467",
            "line2": "Harrison Street",
            "line3": "Harrison Street",
            "city": "San Fransico",
            "state": "California",
            "zip": "94122",
            "country": "US",
            "first_name": "john",
            "last_name": "Doe"
        }
    },
    "statement_descriptor_name": "joseph",
    "statement_descriptor_suffix": "JS",
    "metadata": {
        "udf1": "value1",
        "new_customer": "true",
        "login_date": "2019-09-10T10:11:12Z"
    },
    "browser_info": {
        "ip_address": "129.0.0.1",
        "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36",
        "accept_header": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8",
[...]

PR #1967, opened 2026-07-20
fix(stripe): send payment_method_options[card][moto] on setup_mandate

10X-GRACE:
## What

The Stripe **SetupMandate** (SetupIntent) request builder in prism never emitted the
`payment_method_options[card][moto]` field. Hyperswitch's Direct gateway sets it to `true` for
card payments initiated over a mail-order / telephone-order channel
(`payment_channel = MailOrder | TelephoneOrder`), so the shadow UCS leg diverged from Direct.

Fixes shadow-validation issue **juspay/hyperswitch-cloud#20285** (`stripe` / `setup_mandate`, prod,
merchant `abo_factory`).

### Before (prod diff signature)
```
body.keyDiff.payment_method_options[card][moto] = {"hyperswitch": true, "ucs": false}
```
`grep moto` over the prism Stripe connector returned **0 hits** — the key was structurally impossible
to emit on the UCS side, hence the keyDiff.

## Root cause

`hyperswitch/crates/hyperswitch_connectors/src/connectors/stripe/transformers.rs`
(`TryFrom<&SetupMandateRouterData> for SetupIntentRequest`) computes:
```rust
let is_moto = matches!(pmd, Card{..}) && payment_channel == Some(MailOrder)
              || payment_channel == Some(TelephoneOrder);
```
and emits `moto: is_moto` → `payment_method_options[card][moto]`. It also suppresses
`setup_future_usage` (`OnSession + moto → None`). The prism `SetupMandateRequest` had neither.
The source signal is already present: `SetupMandateRequestData.payment_channel` is wired through from
proto `PaymentServiceSetupRecurringRequest.payment_channel` (field 27). This is a UCS-only transformer
gap — no proto change needed.

## Fix

Add a top-level `moto` field to `SetupMandateRequest`, serialized as
`payment_method_options[card][moto]` with `skip_serializing_if = Option::is_none`, computed with the
exact same condition as hyperswitch, plus the matching `OnSession + moto → usage=None` coupling so the
`usage` field stays in parity too.

Fired the UCS leg directly at prism `:8000` `PaymentService/SetupRecurring` with
`payment_channel = MAIL_ORDER` + a card, and inspected the **outbound Stripe SetupIntent body**:

**After — MAIL_ORDER (MOTO):**
```
"expand[0]":"latest_attempt","payment_method_options[card][moto]":true
```
✅ UCS now emits `payment_method_options[card][moto]=true`, matching hyperswitch Direct → `keyDiff: {}`.

**Negative — ECOMMERCE (non-MOTO):**
```
moto key absent (skip_serializing_if) — no spurious payment_method_options[card][moto]=false
```
✅ No new diff introduced for non-MOTO traffic.

Regression check (`check_20285.py`): `moto:true occurrences: 6, moto:false occurrences: 0 → GREEN`.

## Notes
- No proto/two-repo change: `payment_channel` already flows into `SetupMandateRequestData`.
- No credentials or local config included in this PR.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/stripe/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/stripe/transformers.rs
@@ -277,6 +277,9 @@ pub struct SetupMandateRequest<
     pub expand: Option<ExpandableObjects>,
     #[serde(flatten)]
     pub browser_info: Option<StripeBrowserInformation>,
+    #[serde(rename = "payment_method_options[card][moto]")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub moto: Option<bool>,
 }
 
 #[derive(Debug, Eq, PartialEq, Serialize)]
@@ -4945,12 +4948,30 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
             .clone()
             .map(StripeBrowserInformation::from);
 
+        let is_moto = if matches!(
+            item.router_data.request.payment_method_data,
+            PaymentMethodData::Card(_)
+        ) && item.router_data.request.payment_channel
+            == Some(common_enums::PaymentChannel::MailOrder)
+            || item.router_data.request.payment_channel
+                == Some(common_enums::PaymentChannel::TelephoneOrder)
+        {
+            Some(true)
+        } else {
+            None
+        };
+
+        let usage = match (item.router_data.request.setup_future_usage, is_moto) {
+            (Some(common_enums::FutureUsage::OnSession), Some(true)) => None,
+            _ => item.router_data.request.setup_future_usage,
+        };
+
         Ok(Self {
             confirm: true,
             payment_data,
             return_url: item.router_data.request.router_return_url.clone(),
             off_session: item.router_data.request.off_session,
-            usage: item.router_data.request.setup_future_usage,
+            usage,
             payment_method_options: None,
             customer: item
                 .router_data
@@ -4962,6 +4983,7 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
             payment_method_types: Some(pm_type),
             expand: Some(ExpandableObjects::LatestAttempt),
             browser_info,
+            moto: is_moto,
         })
     }
 }


XyneSpaces:
[blocking] The `is_moto` condition currently treats any `TelephoneOrder` request as MOTO, even when `payment_method_data` is not `Card`, because `&&` binds tighter than `||`. A non-card telephone-order setup mandate will still serialize `payment_method_options[card][moto]=true`; group the channel checks under the card guard, e.g. `matches!(payment_method_data, PaymentMethodData::Card(_)) && matches!(payment_channel, Some(MailOrder | TelephoneOrder))`.

XyneSpaces:
[should-fix] In `crates/integrations/connector-integration/src/connectors/stripe/transformers.rs`, please parenthesize the MOTO channel check so `moto` is only sent for card setup mandates. With the current `Card && MailOrder || TelephoneOrder` precedence, any TelephoneOrder setup mandate sets `payment_method_options[card][moto]=true` even when `payment_method_data` is not `Card`, causing Stripe to receive a card-specific option for non-card flows.

Suggested shape:
```rust
let is_moto = if matches!(
    item.router_data.request.payment_method_data,
    PaymentMethodData::Card(_)
) && matches!(
    item.router_data.request.payment_channel,
    Some(common_enums::PaymentChannel::MailOrder)
        | Some(common_enums::PaymentChannel::TelephoneOrder)
) {
    Some(true)
} else {
    None
};
```

XyneSpaces:
[should-fix] Parenthesize the MOTO channel check so the card guard applies to both `MailOrder` and `TelephoneOrder`. As written, `&&` binds tighter than `||`, so any non-card setup mandate with `payment_channel == TelephoneOrder` will still serialize `payment_method_options[card][moto]=true`.

```suggestion
        let is_moto = if matches!(
            item.router_data.request.payment_method_data,
            PaymentMethodData::Card(_)
        ) && matches!(
            item.router_data.request.payment_channel,
            Some(
                common_enums::PaymentChannel::MailOrder
                    | common_enums::PaymentChannel::TelephoneOrder,
            )
        ) {
```

Shubhodip900:
This changes are merged via https://github.com/juspay/hyperswitch-prism/pull/1961 PR

PR #1968, opened 2026-07-20
fix(proto): add connector_reference_id and fix misuse as merchant_*_id

### LINKED ISSUE #1943: fix(payments): merchant_transaction_id incorrectly populated with connector response reference ID
The merchant_transaction_id field responses is being incorrectly overwritten with the connector_response_reference_id, causing data integrity issue. This field should preserve the merchant's original transaction reference ID but is instead being populated with the connector's response reference ID. This overwriting violates proto field semantics and results in failed integrity checks when validating responses against requests.

AmitsinghTanwar007:
## Summary

`connector_response_reference_id` (a connector-side secondary reference — e.g. Stripe charge ID `ch_xxx`, order codes, invoice numbers) was incorrectly mapped to `merchant_transaction_id`, `merchant_void_id`, `merchant_reverse_id`, `merchant_capture_id`, `merchant_refund_id`, and `merchant_dispute_id` in the gRPC response layer.

**91% of connectors** populate this field with **connector-side IDs**, not merchant IDs — the previous mapping was semantically wrong.

### The Three IDs

| ID | Source | Example | Proto field |
|----|--------|---------|-------------|
| `merchant_transaction_id` | Merchant sends in request | `order_abc123` | Existed |
| `connector_transaction_id` | Connector's primary ID | Stripe `pi_xxx`, Adyen PSP ref | Existed |
| `connector_reference_id` | Connector's secondary reference | Stripe `ch_xxx`, order codes, invoice numbers | **Added in this PR** |

### Changes

**Proto (`payment.proto`):**
- Added `optional string connector_reference_id` to 7 response messages:
  - `PaymentServiceAuthorizeResponse` (field 23)
  - `PaymentServiceGetResponse` (field 36)
  - `PaymentServiceVoidResponse` (field 15)
  - `PaymentServiceReverseResponse` (field 9)
  - `PaymentServiceCaptureResponse` (field 17)
  - `RefundResponse` (field 25)
  - `DisputeResponse` (field 19)

**Mapping fixes (`types.rs`):**
- Routed domain `connector_response_reference_id` to the new `connector_reference_id` proto field instead of `merchant_*_id` across all 13 response construction sites (authorize, void, reverse, psync, capture, refund, dispute — both OK and error paths)
- Set `merchant_*_id` to `None` where it was previously incorrectly populated with connector-side data

### Impact

- **Wire-compatible**: new optional field, no existing field removed
- **Hyperswitch consumer side**: currently reads `response.merchant_transaction_id` as a workaround (`hyperswitch_interfaces/src/unified_connector_service/transformers.rs:459,522`). After this PR merges and HS bumps the UCS client tag, HS needs a 2-line change:
  ```rust
  // Before:
  connector_response_reference_id: response.merchant_transaction_id,
  // After:
  connector_response_reference_id: response.connector_reference_id,
  ```

### Follow-up: Connector-side alignment

This PR fixes the **proto + domain mapping layer only**. A follow-up PR will align the individual connector transformers so each value lands in the correct proto field. Analysis of 75 connectors found:
- **33 connectors** match between UCS and HS
- **34 connectors** have mismatches (UCS populates where HS returns `None`, or uses a different field)
- **~12 connectors** are UCS-only

The follow-up PR will address:
1. **Bucket B connectors** (adyen psync, flywire, paysafe, phonepe, sanlam) — merchant-echoed IDs that should go to `merchant_transaction_id` instead of `connector_reference_id`
2. **Field mismatches** (payu, trustpayments, xendit, jpmorgan, silverflow, elavon, airwallex) — UCS and HS chose different response fields
3. **Coverage gaps** — connectors where UCS populates the field but HS doesn't (or vice versa)

## Test plan

- [x] Full workspace `cargo build` passes
- [ ] Verify Hyperswitch integration after HS bumps UCS client tag
- [ ] Validate with a Bucket A connector (e.g. Stripe) that `connector_reference_id` is populated correctly
- [ ] Validate that `merchant_transaction_id` is no longer polluted with connector-side IDs


XyneSpaces:
[blocking] This proto contract change adds new `connector_reference_id` fields, but the PR does not include the corresponding regenerated `crates/types-traits/grpc-api-types/src/` output. Consumers compiled from the checked-in generated types will not see fields like `PaymentServiceAuthorizeResponse.connector_reference_id`; regenerate and commit the proto fallout in the same PR.

XyneSpaces:
[blocking] `payment.proto` adds `connector_reference_id` fields, but this PR does not include the matching generated Rust updates under `crates/types-traits/grpc-api-types/src/`. The new `domain_types/src/types.rs` initializers set `connector_reference_id`, so a checkout that uses the committed generated types will fail with an unknown-field compile error.

Regenerate and commit the grpc-api-types output for these proto changes.

XyneSpaces:
[blocking] Please regenerate and commit the generated Rust/SDK files for the new `connector_reference_id` proto fields.

This PR changes `payment.proto`, but only `payment.proto` and domain mapping code are changed; without the generated fallout, consumers of the generated payment response types will not see these new fields consistently.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7204,6 +7209,7 @@ pub fn generate_payment_void_post_capture_response(
                 Ok(PaymentServiceReverseResponse {
                     connector_transaction_id: connector_reference_id.clone().unwrap_or_default(),
                     status: grpc_status.into(),
+                    connector_reference_id: None,
                     merchant_reverse_id: connector_reference_id,
```
JeevaRamu0104: **[S1] The bug this PR fixes is left in place on a line the PR edited.**

The local `connector_reference_id` here is destructured from `PaymentsResponseData::PostCaptureVoidResponse` — it is connector-side data (it is used for `connector_transaction_id` three lines up). This PR added `connector_reference_id: None` to this exact struct literal but left the connector ID sitting in `merchant_reverse_id`.

```rust
connector_reference_id: connector_reference_id.clone(),
merchant_reverse_id: None,
```

As written, consumers reading the new field get `None` on this branch while the old field still carries connector data — the opposite of the PR's intent.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -8417,6 +8427,7 @@ pub fn generate_refund_sync_response(
                 ),
                 connector_refund_id: response.connector_refund_id.clone(),
                 status: grpc_status as i32,
+                connector_reference_id: None,
                 merchant_refund_id: Some(response.connector_refund_id.clone()),
```
JeevaRamu0104: **[S1] Same bug class untouched in the direct refund paths.**

`response.connector_refund_id` is a connector-side ID being written into `merchant_refund_id`, in a struct literal this PR edited (`connector_reference_id: None` added directly above). Same issue on the error path at line 8471, where `merchant_refund_id: e.connector_transaction_id.clone()`.

Note the inconsistency inside a single flow: the fix *was* applied to the webhook variant (`ForeignTryFrom<RefundWebhookDetailsResponse>`, ~line 9161) but not to `generate_refund_sync_response` / `generate_refund_response`.

These should route the connector ID to `connector_reference_id` and set `merchant_refund_id` to `None` (or to the genuine merchant reference, see the `merchant_void_id` comment).

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7036,7 +7038,8 @@ pub fn generate_payment_void_response(
                         &grpc_resource_id,
                     ),
                     status: grpc_status.into(),
-                    merchant_void_id: connector_response_reference_id,
+                    connector_reference_id: connector_response_reference_id,
+                    merchant_void_id: None,
```
JeevaRamu0104: **[S1] `merchant_void_id` / `merchant_reverse_id` / `merchant_capture_id` become permanently unpopulated.**

After this PR no code path sets these on the Void / Reverse / Capture responses — every site is `None` (7042, 7090, 7182, 7248, 10420, 10469). That is an undeclared API surface removal: the field stays in the proto and silently returns null forever.

These also round-trip into requests elsewhere in this file — `PaymentServiceVoidRequest` builds `connector_request_reference_id` from `value.merchant_void_id` (~5305), and capture does the same from `merchant_capture_id` (~10321). A client echoing the response field back into the next request now sends empty.

Pick one and make it explicit:
1. Populate from `router_data_v2.resource_common_data.connector_request_reference_id` — the genuine merchant reference, which is exactly what the PSync error path does (~7661); or
2. Mark the fields `[deprecated = true]` in `payment.proto` with a migration note.

Leaving them silently null is the worst of the three.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -6101,7 +6102,8 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
                 connector_transaction_id: err.connector_transaction_id.clone(),
                 redirection_data: None,
                 network_transaction_id: None,
-                merchant_transaction_id: err.connector_transaction_id.clone(),
+                connector_reference_id: None,
+                merchant_transaction_id: None,
```
JeevaRamu0104, line 6216: **[S2] Authorize error path drops a merchant ID that is available.**

Set to `None` here, while the PSync error path populates the same field from `router_data_v2.resource_common_data.connector_request_reference_id` (~7661). That value is equally available in `generate_payment_authorize_response`.

Same concept handled two different ways across two flows — worth making consistent, whichever way you land on.

JeevaRamu0104 (commented):
Four findings on the mapping changes. Three are the same root cause: the fix was applied at most sites but skipped in struct literals this PR already edited, leaving connector-side IDs in `merchant_*_id` fields.

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: merchant_refund_id is now mapped to the id passed in request

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: all merchant specific ids are mapped with its corresponding id from request

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: In error scenario it is mapped to None

crates/types-traits/domain_types/src/types.rs
```diff
@@ -8571,7 +8611,8 @@ pub fn generate_refund_sync_response(
                 ),
                 connector_refund_id: response.connector_refund_id.clone(),
                 status: grpc_status as i32,
-                merchant_refund_id: Some(response.connector_refund_id.clone()),
+                connector_reference_id: None,
+                merchant_refund_id: None,
```
kanikac199: This sets `merchant_refund_id: None`, but `generate_void_post_refund_response` sets it to `Some(connector_request_reference_id`). same for `generate_refund_response`, which sets it to none

crates/types-traits/domain_types/src/types.rs
kanikac199: Authorize error path sets this to None, but the PSync error path sets merchant_transaction_id: Some(connector_request_reference_id)

kanikac199 (approved):

crates/types-traits/domain_types/src/types.rs
```diff
@@ -10633,7 +10688,8 @@ pub fn generate_payment_capture_response(
                 connector_transaction_id: extract_connector_request_reference_id(
                     &e.connector_transaction_id.clone(),
                 ),
-                merchant_capture_id: e.connector_transaction_id.clone(),
+                connector_reference_id: None,
+                merchant_capture_id: None,
```
Shubhodip900, line 10692: why None? some connectors sends transaction id during error cases, so we should populate the values

crates/types-traits/domain_types/src/types.rs
```diff
@@ -6206,7 +6212,8 @@ pub fn generate_payment_authorize_response<T: PaymentMethodDataTypes>(
                 connector_transaction_id: err.connector_transaction_id.clone(),
                 redirection_data: None,
                 network_transaction_id: None,
-                merchant_transaction_id: err.connector_transaction_id.clone(),
+                connector_reference_id: None,
```
Shubhodip900, line 6215:
```suggestion
                connector_reference_id: err.connector_transaction_id.clone(),
```

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7199,7 +7212,8 @@ pub fn generate_payment_void_response(
                 connector_transaction_id: extract_connector_request_reference_id(
                     &e.connector_transaction_id,
                 ),
-                merchant_void_id: e.connector_transaction_id.clone(),
+                connector_reference_id: None,
```
Shubhodip900, line 7215: +1, populate the connector_transaction_id during error case

crates/types-traits/domain_types/src/types.rs
```diff
@@ -7362,7 +7388,8 @@ pub fn generate_payment_void_post_capture_response(
                     &e.connector_transaction_id,
                 ),
                 status: status.into(),
-                merchant_reverse_id: e.connector_transaction_id.clone(),
+                connector_reference_id: None,
```
Shubhodip900, line 7391: +1

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: connector_transaction_id is mapped for error cases, connector_reference_id will be none

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: we already have connector_transaction_id for error scenarios, connector_reference_id is not populated in these cases

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: connector_transaction_id is already being populated in the respective field

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: we already have connector_transaction_id for error scenarios, connector_reference_id is not populated in these cases

Shubhodip900 (approved):

JeevaRamu0104 (approved):

PR #1970, opened 2026-07-20
feat(events): stamp deployment version metadata on every connector event

kanikac199:
What
----
Adds a version tag to every connector event so releases can be A/B-monitored (compare
old build vs new build). Each event now carries four top-level fields:

- version — the deployed build (A/B groups on this)
- application_name — the microservice (e.g. connector-service-http)
- deployment_id — the rollout / replicaset id
- pod_name — the pod name

How it is sourced (no deployment-config or infra changes required)
----
- version: injected once by the binary at startup from ucs_env::git_describe!()
  (compiled, per-build → correct for A/B grouping), via
  external_services::set_build_version(). Sourced from the binary rather than ucs_env
  directly to avoid a dependency cycle (ucs_env already depends on external-services),
  and rather than a config/env value because it must reflect the actual running build.
- application_name / deployment_id / pod_name: derived from the pod name in HOSTNAME,
  which k8s sets to <application_name>-<replicaset_hash>-<pod_suffix>.

This means the fields require nothing from the generate-env-configs pipeline (which only
emits CS__-prefixed vars from TOML) and no downward-API/valueFrom in the deployment
template — k8s already provides HOSTNAME and the version is compiled in.

Changes
----
- crates/common/common_utils/src/events.rs: add add_application_name / add_version /
  add_deployment_id / add_pod_name setters (masked, land in additional_fields).
- crates/common/external-services/src/service.rs: set_build_version() + version_metadata()
  (git_describe for version; HOSTNAME-derived for the rest); stamped in create_event.
- crates/grpc-server/grpc-server/src/main.rs: inject the compiled build version at startup.

Scope
----
- Applies to ConnectorCall-stage events (the connector events A/B compares).
- Works for both the grpc and http deployments (same grpc-server binary).

Testing
----
Live-verified against a real connector Authorize (2C2P PACO UAT, 200 CHARGED): the
ConnectorCall event carried all four fields, e.g.
application_name=connector-service-http, deployment_id=7d9f8b6c5,
pod_name=connector-service-http-7d9f8b6c5-x2k9p, version=2026.07.03.0-60-g433499aee.

Deployment note
----
No hyperswitch-infra changes needed. Requires deploying an image built from this branch;
HOSTNAME is set by k8s automatically.

### DIFF
--- a/crates/common/common_utils/src/events.rs
+++ b/crates/common/common_utils/src/events.rs
@@ -293,6 +293,34 @@ impl Event {
         });
     }
 
+    /// The application / microservice name (e.g. `connector-service-http`). Sourced from the
+    /// `APPLICATION_NAME` env, not the internal gRPC `service_name`.
+    pub fn add_application_name(&mut self, application_name: &str) {
+        self.push_version_field("application_name", application_name);
+    }
+
+    /// The deployed build version (e.g. `2026.07.08.0`). A/B groups on this field directly.
+    pub fn add_version(&mut self, version: &str) {
+        self.push_version_field("version", version);
+    }
+
+    /// The rollout / replicaset id (k8s downward API).
+    pub fn add_deployment_id(&mut self, deployment_id: &str) {
+        self.push_version_field("deployment_id", deployment_id);
+    }
+
+    /// The pod name (k8s downward API `metadata.name`).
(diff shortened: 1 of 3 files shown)


kanikac199:
Superseded by #1994 (same change, rebased onto the current main tip). Closing this stale-base duplicate.

PR #1951, opened 2026-07-15
feat(proto): Add authenticator connector support and bank account payment method details

Aishwariyaa-Anand:
## Summary
Adds proto definitions to support authenticator connectors (e.g., Plaid) for bank account linking and verification flows. Extends existing `CreateClientAuthenticationToken` and `PaymentMethodServiceGet` APIs with authenticator-specific context and bank account payment method details. Designed generically to support any account authentication connector (Plaid, Yodlee, MX, Finicity, etc.).

## Changes

- CreateClientAuthenticationToken

    1. Authenticator Client Context (CreateClientAuthenticationToken Request)
    2. Plaid Client Authentication Response (CreateClientAuthenticationToken Response)

- PaymentMethodServiceGet -  Bank Account Payment Method Details (Response)

### Additional Changes
- [ ] This PR modifies the API contract
- [ ] This PR modifies application configuration/environment variables

PR #1975, opened 2026-07-20
feat(connector): [PLAID] add Plaid as authenticator connector 

Aishwariyaa-Anand:
## Description
<!-- Describe your changes in detail -->

Integrates Plaid as authenticator connector in UCS, enabling bank account linking via three gRPC flows:                                                                                                                                 
                                                                       
  - Flow 1 — Link Token (MerchantAuthenticationService.CreateClientAuthenticationToken): Calls Plaid's POST /link/token/create and returns a link_token for initialising Plaid Link on the client side.
  - Flow 2 — Token Exchange (PaymentMethodService.Tokenize): Calls Plaid's POST /item/public_token/exchange to exchange a public_token (from Plaid Link) for a long-lived access_token, stored as payment_method_token.                             
  - Flow 3 — Auth Get (PaymentMethodService.Get): Calls Plaid's POST /auth/get using the access_token and returns BankAccountDetails with ACH/BACS/SEPA routing numbers.
                                                                                                                                                                                                                                                    
##  Changes                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                 
  Domain Types                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                    
  - Added AuthenticatorConnectorEnum (Plaid) and ConnectorVariant::Authenticator variant
  - Added ClientAuthenticationTokenData::Plaid and PlaidClientAuthenticationResponse
  - Added BankAccountRoutingDetails enum (ACH / BACS / SEPA)

  Additional Changes                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                    
  - This PR adds raw_connector_request and raw_connector_response in PaymentMethodService.Tokenize
 
### Additional Changes
<!--
Provide links to the files with corresponding changes.

Following are the paths where you can find config files:
1. `config`
-->

## How did you test it?
<!--
Did you write an integration/unit/API test to verify the code changes?
Or did you test this change manually (provide relevant screenshots)?
-->
**Create Link token**
```
curl --location 'http://localhost:8011/payments/client_authentication_token' \
--header 'x-connector-config: {"config":{"Plaid":{"client_id":"-","secret":"-"}}}' \
--header 'x-merchant-id: test_merchant' \
--header 'x-tenant-id: public' \
--header 'x-request-id: req-plaid-flow1-001' \
--data-raw '{
    "merchant_client_session_id": "session_plaid_001",
    "connector_feature_data": "{\"client_name\":\"My Test App\"}",
    "domain_context": {
      "Authenticator": {
        "country_codes": ["US"],
        "locale": "en",
        "customer": {
          "id": "customer_001",
          "name": "Jane Doe",
          "email": "jane@example.com"
        }
      }
    }
  }'
```
Response
```
{
    "session_data": {
        "sdk_type": {
            "AuthenticatorConnectorSpecific": {
                "connector": {
                    "Plaid": {
                        "link_token": "-",
                        "expires_in_seconds": 14399
                    }
                }
            }
        }
    },
    "status_code": 200,
    "raw_connector_response": "-",
    "raw_connector_request": "-"
}
```
**Exchange Public token**
```
curl --location 'http://localhost:8011/payments/create_payment_method_token' \
--header 'x-connector-config: {"config":{"Plaid":{"client_id":"-","secret":"-"}}}' \
--header 'x-merchant-id: test_merchant' \
--header 'x-tenant-id: public' \
--header 'x-request-id: req-plaid-flow2-001' \
--data '{
    "metadata": "{\"public_token\": \"-\"}",
    "payment_method": {
        "payment_method": {
            "ach_bank_transfer": {}
        }
    }
}'
```
Response
```
{
    "payment_method_token": "-",
    "status_code": 200,
    "response_headers": {
        "vary": "Accept-Encoding",
        "plaid-version": "2020-09-14",
        "x-content-type-options": "nosniff",
        "date": "Wed, 22 Jul 2026 10:01:58 GMT",
        "content-type": "application/json; charset=utf-8",
        "connection": "keep-alive",
        "x-xss-protection": "1; mode=block",
        "server": "nginx",
        "x-frame-options": "DENY",
        "x-envoy-upstream-service-time": "150",
        "strict-transport-security": "max-age=31536000; includeSubDomains; preload",
        "x-envoy-decorator-operation": "lazyxds.svc-apiv2:8080/*"
    },
    "merchant_payment_method_id": "-",
    "connector_payment_method_id": "aqr7eBwxAdTX7QQbQbnLI9qgvalv7NCRnby4j",
    "raw_connector_response": "-",
    "raw_connector_request": "-"
}
```
**Get bank Account Details**
```
curl --location 'http://localhost:8011/composite/payment_methods/get' \
--header 'x-connector-config: {"config":{"Plaid":{"client_id":"-","secret":"-"}}}' \
--header 'x-merchant-id: test_merchant' \
--header 'x-tenant-id: public' \
--header 'x-request-id: req-plaid-flow3-001' \
--header 'x-connector: plaid' \
--data '{
  "payment_method_token": "-",
  "payment_method_type": "ACH"
}'
```
Response
```
{
    "get_response": {
        "connector_payment_method_id": "aqr7eBwxAdTX7QQbQbnLI9qgvalv7NCRnby4j",
        "payment_method_details": {
            "details": {
                "BankAccount": {
                    "accounts": [
                        {
                            "account_name": "Plaid Checking",
                            "account_id": "-",
                            "bank_type": "BANK_TYPE_CHECKING",
                            "balance": {
                                "minor_amount": 11000,
                                "currency": "USD"
                            },
                            "available_balance": {
                                "minor_amount": 10000,
                                "currency": "USD"
                            },
                            "account_details": {
                                "Ach": {
                                    "account_number": "-",
                                    "routing_number": "-"
                                }
                            }
                        },
                        {
                            "account_name": "Plaid Saving",
                            "account_id": "-",
                            "bank_type": "BANK_TYPE_SAVINGS",
                            "balance": {
                                "minor_amount": 21000,
                                "currency": "USD"
                            },
                            "available_balance": {
                                "minor_amount": 20000,
                                "currency": "USD"
                            },
                            "account_details": {
                                "Ach": {
                                    "account_number": "-",
                                    "routing_number": "-"
                                }
                            }
                        },
                        {
                            "account_name": "Plaid Cash Management",
                            "account_id": "-",
                            "balance": {
                                "minor_amount": 1206000,
                                "currency": "USD"
                            },
                            "available_balance": {
                                "minor_amount": 1206000,
                                "currency": "USD"
                            },
                            "account_details": {
                                "Ach": {
                                    "account_number": "-",
                                    "routing_number": "-"
                                }
                            }
                        }
                    ]
                }
            }
        },
        "status_code": 200,
        "response_headers": {
            "vary": "Accept-Encoding",
            "server": "nginx",
            "content-type": "application/json; charset=utf-8",
            "connection": "keep-alive",
            "x-xss-protection": "1; mode=block",
            "x-frame-options": "DENY",
            "x-envoy-upstream-service-time": "349",
            "x-envoy-decorator-operation": "lazyxds.svc-apiv2:8080/*",
            "strict-transport-security": "max-age=31536000; includeSubDomains; preload",
            "date": "Wed, 22 Jul 2026 10:02:11 GMT",
            "x-content-type-options": "nosniff",
            "plaid-version": "2020-09-14"
        },
        "raw_connector_response": "-",
        "raw_connector_request": "-"
    }
}
```


XyneSpaces:
[blocking] Public enums in `domain_types` gain new variants (`ConnectorVariant::Authenticator`, `ClientAuthenticationTokenData::Plaid`, and `PaymentMethodDetails::BankAccount`), which can break downstream crates that exhaustively match these types. Please coordinate the consumer update in the same landing window, or make the external-match contract explicit with a transition strategy such as `#[non_exhaustive]` if downstream exhaustive matching is not supported.

XyneSpaces:
[blocking] Please don’t hardcode USD and `* 100` when converting Plaid balances in `crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs` lines 526–534. Plaid `/auth/get` can return non-USD accounts and currencies with exponents other than 2, so this can persist incorrect minor-unit balances.

Add `iso_currency_code` to `PlaidBalances`, parse it into `common_enums::Currency`, and convert using that currency’s exponent instead of forcing `Currency::USD`.

XyneSpaces:
[blocking] Please avoid converting Plaid balances through `f64` multiplication and hardcoded USD.

For a Plaid account with `current = 123.459`, `(amt * 100.0) as i64` can truncate due to binary floating-point representation, and non-USD accounts are reported as USD. Parse the connector amount as a decimal/minor-unit value and derive the currency from Plaid’s currency code before constructing `Money`.

`crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs`, lines 396 and 397.

XyneSpaces:
[blocking] Do not hard-code Plaid account balances as USD. `/auth/get` can return BACS and SEPA account details as well as ACH, so a UK/EU account with a balance here would be reported with `Currency::USD`; thread the actual account/request currency into both `balance` and `available_balance` instead.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    pub redirect_uri: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub webhook: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct PlaidUser {
+    pub client_user_id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub legal_name: Option<String>,
```
deepanshu-iiitu: Make PII fields secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub email_address: Option<String>,
```
deepanshu-iiitu: Use struct Email here

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub phone_number: Option<String>,
```
deepanshu-iiitu: Make phone_number as secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    pub products: Vec<String>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub country_codes: Vec<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub language: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub redirect_uri: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub webhook: Option<String>,
+}
```
deepanshu-iiitu, line 89: Make PII data secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PlaidItem {
+    pub item_id: String,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PlaidAccount {
+    pub account_id: String,
+    pub name: String,
```
deepanshu-iiitu: Should PlaidAccount.name be secret?

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    #[serde(default)]
+    pub international: Vec<PlaidSepaNumbers>,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PlaidAchNumbers {
+    pub account_id: String,
+    pub account: Secret<String>,
+    pub routing: String,
+}
```
deepanshu-iiitu, line 500: Make these fields secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+    pub account: Secret<String>,
+    pub sort_code: String,
+}
+
+#[derive(Debug, Deserialize, Serialize)]
+pub struct PlaidSepaNumbers {
+    pub account_id: String,
+    pub iban: Secret<String>,
+    pub bic: Option<Secret<String>>,
+}
```
deepanshu-iiitu, line 522: These fields should all be secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+        PaymentFlowData,
+        GetPaymentMethodData,
+        GetPaymentMethodResponseData,
+    >
+{
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(item: ResponseRouterData<PlaidAuthGetResponse, Self>) -> Result<Self, Self::Error> {
+        let res = item.response;
+        let request_currency = item
```
deepanshu-iiitu: use a util here

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+            .into_iter()
+            .map(|sepa| (sepa.account_id.clone(), sepa))
+            .collect();
+
+        // Only include accounts that have at least one routing scheme (ACH / BACS / SEPA).
+        let accounts: Vec<BankAccount> = res
+            .accounts
+            .into_iter()
+            .filter_map(|acct| {
+                // Prefer ACH → BACS → SEPA in that order; skip accounts with none.
```
deepanshu-iiitu: How was this preference decided? ACH → BACS → SEPA

Shouldn't we prioritize this based on the currency?

crates/types-traits/domain_types/src/payment_method_data.rs
```diff
@@ -1926,7 +1926,55 @@ pub struct CustomerInfoDetails {
...
+pub enum BankAccountRoutingDetails {
+    Ach(BankAccountAchDetails),
+    Bacs(BankAccountBacsDetails),
+    Sepa(BankAccountSepaDetails),
+}
+
+/// A single bank account with identity and routing details
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BankAccount {
+    pub account_name: String,
```
deepanshu-iiitu: Should this be secret?

deepanshu-iiitu (changes requested):

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+            .as_ref()
+            .and_then(|feature_data| {
+                feature_data
+                    .peek()
+                    .as_object()
+                    .and_then(|obj| obj.get("client_name"))
+                    .and_then(|val| val.as_str())
+                    .map(|client_name_str| client_name_str.to_owned())
+            })
+            .unwrap_or_else(|| "Hyperswitch".to_owned());
```
iemyashasvi: dont fall back to hardcoded , if this is required please throw an err

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+                        .resource_common_data
+                        .connector_request_reference_id
+                        .clone()
+                }),
+            legal_name: req
+                .customer_name
+                .as_ref()
+                .map(|name| name.peek().to_owned()),
+            email_address: req.email.as_ref().map(|email| email.peek().to_owned()),
+            phone_number: None,
```
iemyashasvi: why is this marked as None

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+                        account_number: bacs.account.clone(),
+                        sort_code: Secret::new(bacs.sort_code.clone()),
+                    })
+                } else {
+                    let sepa = sepa_map.get(&acct.account_id)?;
+                    BankAccountRoutingDetails::Sepa(BankAccountSepaDetails {
+                        iban: sepa.iban.clone(),
+                        bic: sepa.bic.clone(),
+                    })
+                };
```
iemyashasvi: nit : can we use match

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,614 @@
...
+                            super::PlaidAmountConvertor::convert_back(amt, currency)
+                                .ok()
+                                .map(|amount| common_utils::types::Money { amount, currency })
+                        });
+
+                Some(BankAccount {
+                    account_name: acct.name,
+                    account_id: Secret::new(acct.account_id),
+                    bank_type,
+                    bank_holder_type: None,
```
iemyashasvi: why None

JeevaRamu0104 (changes requested):
## Review — Plaid authenticator connector

**Decision: request_changes** · Risk: medium-high
Scenarios: `connector-new-integration` (primary) + `core-flow-framework`, `proto-api-contract`, `server-composite`, `sdk-ffi-codegen`.

The connector implementation itself is clean and well-structured (secrets use `Secret<String>`, proto changes are backward compatible, all 9 existing connectors updated for the new `PaymentMethodTokenResponse.connector_payment_method_id` field, cross-SDK maps consistent). One blocker below.

---

### Blocking

**[S1] Composite `process_get` unconditionally triggers a `tokenize` connector call for any request lacking `payment_method_token` — regression for existing (wallet) get flows**
`crates/internal/composite-service/src/payment_methods.rs` (`create_payment_method_token`)

```rust
if payload.payment_method_token.is_none() {
    let tokenize_inner = PaymentMethodServiceTokenizeRequest::foreign_from(payload);
    ... self.payment_method_service.tokenize(tokenize_request).await?  // error propagates
```

The gate is purely `payment_method_token.is_none()` with **no connector-variant check**. Previously `process_get` required `x-connector` and only did server-auth-token + get (no tokenize). Now any existing composite PM get (e.g. the documented "fetch wallet details" case, a payment connector) that does not pass a `payment_method_token` will unconditionally invoke `tokenize`. For a non-authenticator connector, `tokenize` → `handle_tokenize_internal` requires `payload.payment_method` (`server/payments.rs`: `.ok_or(MissingRequiredField { field_name: "payment_method" })?`). Existing get callers don't set the new `payment_method` field (proto field 16), so tokenize errors and `?` propagates — **the entire composite get fails**.

*Failure scenario:* consumer calls `CompositePaymentMethodService.Get` with `x-connector: <payment_connector>`, no `payment_method_token`, no `payment_method` → new code calls `tokenize` → `MissingRequiredField: payment_method` → composite get returns an error where it previously succeeded.

*Fix:* gate the tokenize on the authenticator variant — only call `create_payment_method_token` when the metadata resolves to `ConnectorVariant::Authenticator(_)` (`x-auth-connector`), rather than on `payment_method_token.is_none()` alone.

---

### Non-blocking

**[S2] `process_get` silently swallows a malformed `x-connector`.** The change from `connector_from_composite_authorize_metadata(&metadata).map_err(|err| *err)?` to `if let Ok(connector) = ...` means a typo'd/invalid `x-connector` now yields `access_token_response = None` and proceeds, instead of failing fast. Existing callers that relied on that validation error lose it. — `crates/internal/composite-service/src/payment_methods.rs`

**[S3] `PlaidErrorResponse::switch()` sets `http_status_code: None`.** Acceptable since `ErrorSwitch` has no access to the status code (and `build_error_response` populates it from `res.status_code`), but the diagnostic context is weaker on this path. — `plaid/transformers.rs`

**[S3] `get_currency_unit()` returns `CurrencyUnit::Minor` while the amount converter is `FloatMajorUnit`.** Inconsistent, though `get_currency_unit` is unused for these three flows (balances use `PlaidAmountConvertor::convert_back` directly). — `plaid.rs`

---

### Verified clean
- Proto backward compatible: new fields only (`PaymentMethodServiceTokenizeResponse` 8/9, `CompositePaymentMethodGetRequest` 12–16, `CompositePaymentMethodGetResponse` 3, `PlaidConfig` + `plaid = 139`). The `// GLOMOPAY = 137` / `// GIVEPAYMENTS = 138` edits are comment corrections only; no tag renumbering or reuse.
- `PaymentMethodTokenResponse.connector_payment_method_id` companion sync across all existing connectors — complete.
- Auth routing: flows 1 & 2 resolve `ConnectorVariant::Authenticator(Plaid)` via `x-connector-config`; dispatch methods branch correctly on `ConnectorVariant::Authenticator(_)`.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,638 @@
...
+
+// =============================================================================
+// FLOW 1: /link/token/create
+// =============================================================================
+
+#[derive(Debug, Serialize)]
+pub struct PlaidLinkTokenRequest {
+    pub client_id: Secret<String>,
+    pub secret: Secret<String>,
+    pub client_name: String,
```
iemyashasvi, line 79: shouldnt this be Secret<String>

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,638 @@
...
+    pub language: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub redirect_uri: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub webhook: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct PlaidUser {
+    pub client_user_id: String,
```
iemyashasvi, line 93: shouldnt this be Secret

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,638 @@
...
+        let user = PlaidUser {
+            client_user_id: req
+                .customer_id
+                .as_ref()
+                .map(|id| id.get_string_repr().to_owned())
+                .unwrap_or_else(|| {
+                    item.router_data
+                        .resource_common_data
+                        .connector_request_reference_id
+                        .clone()
```
iemyashasvi: is mapping connector_request_reference_id to user_id  as fallback correct ?

crates/integrations/connector-integration/src/authenticator_connectors/plaid.rs
```diff
@@ -0,0 +1,389 @@
...
+        Authenticate,
+        PostAuthenticate,
+        MandateRevoke,
+        CreateConnectorCustomer,
+        CreateOrder,
+        Accept,
+        DefendDispute,
+        SubmitEvidence,
+        PaymentMethodEligibility,
+    ]
```
iemyashasvi: can we classify them correctly ?

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: this is merchant's customer id

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: client_name is the application's display name shown to end users in the Plaid Link UI

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: removed

iemyashasvi (dismissed):

crates/integrations/connector-integration/src/authenticator_connectors/plaid.rs
```diff
@@ -0,0 +1,390 @@
...
+        MandateRevoke,
+        RepeatPayment,
+        CreateConnectorCustomer,
+        PreAuthenticate,
+        Authenticate,
+        PostAuthenticate,
+        PaymentMethodEligibility,
+    ],
+    not_supported: [
+        Capture,
```
gopikrishna000, line 379: if we are separating authentication connectors, how capture, RSync, etc will make sense for authentication only connector

crates/types-traits/grpc-api-types/proto/composite_payment.proto
```diff
@@ -571,12 +571,20 @@ message CompositePaymentMethodGetRequest {
   optional string merchant_access_token_id = 9;
   optional string merchant_request_id = 10; // Per-request idempotency / correlation token forwarded to the bootstrap call
   optional string payment_method_token = 11;  // PaymentMethod Token Information
+
+  // Tokenize-step fields
+  optional Money amount = 12;
```
gopikrishna000, line 576: why paymentmethodget has amount, payment_method itself?, etc

Aishwariyaa-Anand: this is compositePaymentMethodGet, amount field is added as we included granular paymentMethodTokenize in this composite flow

crates/integrations/connector-integration/src/authenticator_connectors/plaid.rs
Aishwariyaa-Anand: Plaid can be implemented as both authN connector and payment connector. These flows are supported by plaid as payment connector

JeevaRamu0104 (commented):
Review of the Plaid authenticator integration. Two blocking findings, both in the shared dispatch code rather than in the Plaid connector itself; the connector code is in good shape. Details inline.

**Blocking**
1. `composite-service/src/payment_methods.rs` — the new tokenize step in `process_get` fires for six existing payment connectors and hard-fails their composite get.
2. `domain_types/src/types.rs` — the new `None` arm for `domain_context` turns a previously-rejected request into a silent 0/USD SDK session.

**Missing companions**
- No unit tests for the three Plaid request builders or the three response parsers (repo convention: `connectors/{adyen,calida,ppro,razorpay,cashfree}/test.rs`).
- No regression test covering the new composite tokenize branch, which is where finding 1 lives.

**Scope**
Six scenarios in one PR (new connector + core-flow types + proto + composite server + SDK + config). The composite `payment_methods/get` tokenize step is the piece least related to "add Plaid", and is where both blocking findings' blast radius lands — splitting it out would let the connector merge on a much shorter review.

crates/internal/composite-service/src/payment_methods.rs
```diff
@@ -169,13 +178,71 @@ where
...
+        // Skip if the caller already has a token (e.g. a previously obtained access_token).
+        if payload.payment_method_token.is_none() {
+            let should_do_payment_method_token = {
+                let payment_method = payload
+                    .payment_method
+                    .as_ref()
+                    .map(|pm| common_enums::PaymentMethod::foreign_try_from(pm.clone()))
+                    .transpose()
+                    .into_grpc_status()?
+                    .unwrap_or_default();
```
JeevaRamu0104, line 198: **[S1 — blocking] This will hard-fail composite `payment_methods/get` for six existing payment connectors.**

`payload.payment_method` is `CompositePaymentMethodGetRequest` proto field 16, **added by this PR** — so every existing caller leaves it unset and `unwrap_or_default()` yields `PaymentMethod::Card` (`common_enums/src/enums.rs:1642-1644`, `Card` is `#[default]`).

`should_do_payment_method_token` then returns `true` for:

| Connector | Condition |
|---|---|
| `billwerk.rs:143`, `hipay.rs:106`, `paysafe.rs:89` | unconditional `true` |
| `braintree.rs:211`, `mollie.rs:204`, `getnet.rs:155` | `matches!(payment_method, Card)` |

The `true` branch below builds the tokenize request via `PaymentMethodServiceTokenizeRequest::foreign_from(payload)`, which copies `payment_method: item.payment_method.clone()` → `None`. `PaymentMethod::tokenize` then hits `grpc-server/src/server/payments.rs:2037`:

```rust
payload.payment_method.clone().ok_or(...MissingRequiredField { field_name: "payment_method" ... })?
```

and the `.await?` on the `tokenize` call propagates that status, failing the **entire composite get** — an endpoint that works today.

**Suggested fix:** return early when there is nothing to tokenize, and drop the defaulting:

```rust
let Some(payment_method_proto) = payload.payment_method.as_ref() else {
    return Ok(None);
};
let payment_method = common_enums::PaymentMethod::foreign_try_from(payment_method_proto.clone())
    .into_grpc_status()?;
```

Restricting the whole branch to `ConnectorVariant::Authenticator` would also work and matches the PR's stated scope.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -10119,102 +10121,112 @@ impl ForeignTryFrom<MerchantAuthenticationServiceCreateClientAuthenticationToken
...
+                        .shipping_cost
+                        .map(common_utils::types::MinorUnit::new),
+                    payment_method_type,
+                    permissions,
+                    webhook_url: None,
+                    country_codes: vec![],
+                    locale: None,
+                })
+            }
+            None => Ok(Self {
```
JeevaRamu0104: **[S1 — blocking] A missing `domain_context` now silently produces a zero-amount USD session instead of an error.**

Before this change the function rejected anything that was not a `Payment` context:

```rust
let payment_ctx = match value.domain_context.clone() {
    Some(DomainContext::Payment(ctx)) => ctx,
    _ => return Err(report!(IntegrationError::InvalidDataFormat { ... })),
};
```

That `_` arm covered `None`. This new `None` arm makes a **payment**-connector client-authentication-token request that omits `domain_context` succeed with `amount: 0` / `currency: USD`, rather than failing. The `Payment` arm just above still enforces `amount` via `MissingRequiredField` — so this arm is a silent bypass of that guard for exactly the callers most likely to trip it. Adyen / Braintree / Shift4 SDK sessions would be created against the wrong amount and currency.

The `Authenticator` arm legitimately has no amount; `None` does not.

**Suggested fix:**

```rust
None => Err(report!(IntegrationError::InvalidDataFormat {
    field_name: "domain_context",
    context: IntegrationErrorContext {
        additional_context: Some(
            "domain_context is required: use Payment for SDK sessions or Authenticator for bank linking".to_string(),
        ),
        ..Default::default()
    },
})),
```

crates/types-traits/domain_types/src/types.rs
```diff
@@ -14387,6 +14414,23 @@ pub fn generate_payment_sdk_session_token_response(
...
                         Some(convert_connector_specific_to_grpc(*data))
                     }
+                    ClientAuthenticationTokenData::Plaid(plaid_token) => {
+                        Some(grpc_api_types::payments::ClientAuthenticationTokenData {
+                            sdk_type: Some(
+                                grpc_api_types::payments::client_authentication_token_data::SdkType::AuthenticatorConnectorSpecific(
+                                    grpc_api_types::payments::AuthenticatorConnectorSpecificClientAuthenticationResponse {
+                                        connector: Some(grpc_api_types::payments::authenticator_connector_specific_client_authentication_response::Connector::Plaid(
+                                            grpc_api_types::payments::PlaidClientAuthenticationResponse {
+                                                link_token: plaid_token.link_token.expose(),
```
JeevaRamu0104: **[S2] `link_token` travels unmasked end-to-end, while the domain type treats it as a secret.**

This `.expose()` is forced because `proto/payment.proto:2182` declares `string link_token = 1` rather than `SecretString`. Consequences:

- `sdk/javascript/src/payments/_generated_grpc_client.ts` has **no** `PlaidClientAuthenticationResponse` entry in `_SECRET_STRING_FIELDS` — compare `StripeClientAuthenticationResponse: ["clientSecret"]` (line 149), whose proto counterpart is `SecretString`.
- `PlaidClientAuthenticationResponse.link_token` is `Secret<String>` in `connector_types.rs`, so the domain layer and the wire disagree about sensitivity.

Net effect: the token is logged and transported in the clear.

The proto line predates this PR (#1951), but this PR is its first producer, so it is the right place to fix it. Changing it to `SecretString link_token = 1` regenerates the JS mask automatically and lets this stay a `Secret<String>`.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4227,6 +4227,9 @@ message PaymentMethodServiceTokenizeResponse {
   optional ConnectorState state = 6;
 
   optional string connector_payment_method_id = 7; // Connector-issued payment method ID
+
+  optional SecretString raw_connector_response = 8;
```
JeevaRamu0104: **[S2] Worth an explicit decision: tokenize is the one flow that always carries raw instrument data, and this echoes its request back.**

These fields are populated at `domain_types/src/types.rs:12929-12937` from `extract_raw_connector_request` (`external-services/src/service.rs:1593-1637`), which serialises the body verbatim **and** de-masks headers (`Maskable::Masked(val) => val.clone().expose()`).

For this flow that means:
- Plaid puts `client_id` / `secret` in the request **body** (see `PlaidPublicTokenExchangeRequest`), so both land in `raw_connector_request`.
- Card tokenize bodies for Stripe / Braintree / Mollie carry PAN and CVV.

And it is on by default where it matters: `return_raw_connector_data = true` in both `config/production.toml:4` and `config/sandbox.toml:4`.

To be fair, this mirrors `PaymentMethodServiceCreateResponse`, which already exposes the same pair — so it is consistent with existing practice rather than novel, and the `SecretString` typing plus the JS `_SECRET_STRING_FIELDS` entry are both correct here. Flagging it because tokenize is the flow where the body is *always* sensitive, so it deserves to be a deliberate choice rather than an inherited one.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4354,7 +4357,7 @@ message BankAccountDetails {
 
 // Individual bank account credential with routing details
 message BankAccount {
-  string account_name = 1;                       // Account display name
+  SecretString account_name = 1;                  // Account display name
```
JeevaRamu0104, line 4450: **[S2] Wire-incompatible type change on field 1 of a released message.**

`string account_name = 1` → `SecretString account_name = 1`. `string` and `SecretString` are not interchangeable on the wire — `SecretString` is a message, so the payload gains an inner field tag and existing decoders will fail on this field.

`BankAccount` shipped in `d67d1b14f` (#1951, 2026-07-20) and is contained in tags from `2026.07.21.0` onward, so this is a change to a released contract, not to an unreleased draft.

Practical risk is low — this PR is the message's first producer, so there is unlikely to be a decoder in the wild yet. Raising it because the PR description carries no consumer-migration note for it, and a one-line confirmation that no consumer has adopted `BankAccount` yet would close this out.

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -2488,6 +2535,134 @@ impl PaymentMethod {
...
+                token: hyperswitch_masking::Secret::new(String::new()),
+            },
+        );
+        let tokenization_data = PaymentMethodTokenizationData::<DefaultPCIHolder> {
+            payment_method_data: dummy_pm_data,
+            metadata: request.metadata.clone(),
+            connector_feature_data: request
+                .connector_feature_data
+                .as_ref()
+                .and_then(|s| serde_json::from_str(s.peek()).ok()),
```
JeevaRamu0104: **[S2] Malformed `connector_feature_data` is silently discarded.**

`.ok()` swallows the parse error. The standard tokenize path uses the erroring conversion instead:

```rust
.connector_feature_data
.map(|m| ForeignTryFrom::foreign_try_from((m, "feature data")))
.transpose()?,
```

This matters concretely for Plaid: `client_name` is read out of `connector_feature_data` in `plaid/transformers.rs:126-141`. A merchant who sends malformed JSON gets `MissingRequiredField { field_name: "client_name" }` — pointing them at a field they *did* send — instead of a parse error naming the real problem.

Suggest using the same `ForeignTryFrom` conversion here for consistency with the non-authenticator path.

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -2488,6 +2535,134 @@ impl PaymentMethod {
...
+            amount: common_utils::types::MinorUnit::new(0),
```
JeevaRamu0104: **[S3] Inert placeholder values would read as real to the next authenticator connector.**

`amount: MinorUnit::new(0)`, `currency: Currency::USD`, and the `dummy_pm_data` `PaymentMethodToken` with an empty-string token are all unused by Plaid — `PlaidPublicTokenExchangeRequest` reads only `metadata.public_token` and the auth config.

That is fine today, but `PaymentMethodTokenizationData` is a shared struct and the next authenticator connector added here may well read `amount`/`currency` and silently get 0/USD. A short comment naming these as required-but-unused by the authenticator tokenize path would prevent that.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,646 @@
...
+                        })
+                    })?,
+                    Some(common_enums::Currency::EUR) => sepa_map.get(id).map(|sepa| {
+                        BankAccountRoutingDetails::Sepa(BankAccountSepaDetails {
+                            iban: sepa.iban.clone(),
+                            bic: sepa.bic.clone(),
+                        })
+                    })?,
+                    // Unsupported currency — skip this account
+                    _ => return None,
```
JeevaRamu0104: **[S2] Accounts are silently dropped from the `/auth/get` response.**

Two separate paths discard accounts with no error and no log:

1. This `_ => return None` drops any account whose currency is not USD/GBP/EUR. Plaid does return CAD accounts (with `eft` numbers), and this is also the fallthrough when `iso_currency_code` is null **and** `get_currency()` returns `None` — which is the case for a composite get that carries no `amount`.
2. The `?` on each `ach_map.get(id)` / `bacs_map.get(id)` / `sepa_map.get(id)` above drops accounts whose currency matched but which have no corresponding entry in `numbers` — e.g. a USD account not enabled for auth.

The caller receives a shorter `accounts` list and cannot tell the difference between "this item has two accounts" and "this item has five accounts, three of which we couldn't map."

The filtering itself is reasonable; the silence is the issue. Suggest counting the skipped accounts and emitting a `tracing::warn!` with the count and the unmapped `account_id`s so this is diagnosable in production.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,646 @@
...
+// =============================================================================
+
+#[derive(Debug, Serialize)]
+pub struct PlaidLinkTokenRequest {
+    pub client_id: Secret<String>,
+    pub secret: Secret<String>,
+    pub client_name: String,
+    pub user: PlaidUser,
+    pub products: Vec<String>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
JeevaRamu0104: **[S2] `country_codes` can serialise away even though Plaid requires it.**

`skip_serializing_if = "Vec::is_empty"` means an empty `country_codes` is omitted from the body entirely, but `/link/token/create` mandates the field — so the request fails at Plaid with `INVALID_FIELD` rather than at the UCS boundary.

This is inconsistent with how the other two required inputs in this same builder are handled: `client_name` (line 126) and `customer` (line 146) both raise `MissingRequiredField` with `suggested_action` and `doc_url` before the call is made.

Suggest validating it alongside those two so the error contract is uniform:

```rust
if country_codes.is_empty() {
    return Err(report!(IntegrationError::MissingRequiredField {
        field_name: "country_codes",
        context: IntegrationErrorContext {
            additional_context: Some("Plaid /link/token/create requires at least one country code".to_owned()),
            suggested_action: Some("Set country_codes in the Authenticator domain context, e.g. [\"US\"]".to_owned()),
            doc_url: Some("https://plaid.com/docs/api/tokens/#linktokencreate-country-codes".to_owned()),
        },
    }));
}
```

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,646 @@
...
+        item: ResponseRouterData<PlaidLinkTokenResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let res = item.response;
+
+        let expires_in_seconds: Option<i64> = res.expiration.as_deref().and_then(|exp| {
+            time::OffsetDateTime::parse(exp, &time::format_description::well_known::Rfc3339)
+                .ok()
+                .map(|expiry_time| {
+                    let now = time::OffsetDateTime::now_utc();
+                    (expiry_time - now).whole_seconds()
```
JeevaRamu0104: **[S2] The proto comment for this value contradicts what is computed here.**

This computes a **relative duration** — seconds from now until expiry. But `proto/payment.proto:2183` documents the field as:

```proto
optional int64 expires_in_seconds = 2;  // Expiration timestamp (seconds since epoch)
```

The code and the domain-type doc-comment (`connector_types.rs`: *"Seconds until the link_token expires"*) agree with each other; only the proto comment is wrong, and it is the one consumers read. A client following it would compute an expiry ~56 years in the past.

Fix the proto comment to "Seconds until the link_token expires".

Minor, same line: a `expiration` string that fails RFC3339 parsing silently becomes `None` via `.ok()`. Not worth failing the flow over, but a `tracing::debug!` would help if Plaid ever changes the format.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,646 @@
...
+            .iter()
+            .map(|country_code| country_code.to_string())
+            .collect::<Vec<_>>();
+
+        Ok(Self {
+            client_id: auth.client_id,
+            secret: auth.secret,
+            client_name,
+            user,
+            products: vec!["auth".to_owned()],
```
JeevaRamu0104, line 202: **[S3]** `products` is hardcoded to `["auth"]`, which is correct for this PR's scope (the `/auth/get` flow is the only consumer).

Noting it for when `identity` / `transactions` / `balance` are needed — that will require plumbing it through `connector_feature_data` alongside `client_name`, since the product set has to be fixed at link-token creation time and cannot be changed afterwards.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,646 @@
+use common_enums::{BankHolderType, BankType};
```
JeevaRamu0104, line 1: **[Missing companion] No tests for this file.**

Three request builders and three response parsers land with zero coverage, and the repo has the convention for it — `connectors/{adyen,calida,ppro,razorpay,cashfree}/test.rs`.

Highest-value cases, in rough priority order:

1. The currency → routing-scheme selection in the `/auth/get` parser (USD→ACH, GBP→BACS, EUR→SEPA) and the account-dropping behaviour flagged at line 561 — this is the only non-trivial logic in the file.
2. `PlaidAuthType::try_from` rejecting a non-`Plaid` `ConnectorSpecificConfig`.
3. The three `MissingRequiredField` paths: `client_name`, `customer`, `metadata.public_token`.
4. `expires_in_seconds` computation, including the unparseable-`expiration` fallback.

A regression test for the new composite tokenize branch in `composite-service` would be worth adding too — that is where the blocking finding on `payment_methods.rs:198` lives, and `composite_request_schema_check.rs` only covers the schema, not the branch.

crates/internal/composite-service/src/utils.rs
```diff
@@ -98,6 +127,10 @@ pub fn connector_variant_from_composite_metadata(
         return Ok(ConnectorVariant::Surcharge(connector));
     }
 
+    if let Some(connector) = authenticator_connector_from_composite_authorize_metadata(metadata)? {
```
JeevaRamu0104, line 137: **[S3] Header-priority order differs between the two resolvers.**

This one resolves `frm → surcharge → auth → payment`. The other, `ucs_interface_common/src/metadata.rs:209`, resolves `connector → surcharge → payout → frm → auth` — so `auth` is last there but third here, and `frm` is fourth there but first here.

No live divergence today, since composite never handled payout and requests carrying two specialised headers are not a real case. Flagging it because two independent orderings of the same header set will drift; a shared helper (or at least a comment in each pointing at the other) would keep them honest.

Aishwariyaa-Anand: @JeevaRamu0104 which order should we follow?

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: since Plaid has been integrated as AuthN connector, this was hardcoded

crates/types-traits/grpc-api-types/proto/payment.proto
Aishwariyaa-Anand: account_name should be secret and this field is not consumed currently

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: this pattern is changed to - build an account metadata map first, then iterate numbers and look up from it. this will now return all the  `accounts` with corresponding entry in `numbers`

crates/internal/composite-service/src/payment_methods.rs
Aishwariyaa-Anand: not all connector implementing tokenize will require PaymentMethod, this has to be optional and give out an error where it is required, will raise a separate PR for this concern

crates/internal/composite-service/src/utils.rs
JeevaRamu0104: you can follow the second order @Aishwariyaa-Anand

crates/types-traits/domain_types/src/types.rs
```diff
@@ -12892,9 +12902,11 @@ impl
...
             attempt_id: "IRRELEVANT_ATTEMPT_ID".to_string(),
             status: common_enums::AttemptStatus::Pending,
-            payment_method: PaymentMethod::foreign_try_from(
-                value.payment_method.unwrap_or_default(),
-            )?,
+            payment_method: value
+                .payment_method
+                .map(PaymentMethod::foreign_try_from)
+                .transpose()?
+                .unwrap_or_default(),
```
JeevaRamu0104, line 13010: This changes validation for every connector's tokenize flow, not just the authenticator one.

Before, this was `foreign_try_from(value.payment_method.unwrap_or_default())` — an absent `payment_method` became the unspecified proto enum, and `foreign_try_from` rejected it. Now the `None` case short-circuits before the conversion and we fall through to `PaymentMethod::default()`, which is `Card`.

So a tokenize request that omits `payment_method` entirely now silently succeeds as a card tokenization instead of erroring. That's the behaviour the new authenticator path wants, but it applies to all the existing payment connectors too — and it's what makes the new composite `create_payment_method_token` gating resolve to `Card` and fire a tokenize call for billwerk/hipay/paysafe/braintree/getnet/mollie/stax/finix.

Could we keep the strict behaviour here and scope the relaxation to the authenticator dispatch instead?

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,706 @@
...
+                        error = %e,
+                        raw_expiration = %exp,
+                        "Plaid link_token expiration is not RFC3339; expires_in_seconds will be omitted"
+                    );
+                })
+                .ok()
+                .map(|expiry_time| {
+                    let now = time::OffsetDateTime::now_utc();
+                    (expiry_time - now).whole_seconds()
+                })
```
JeevaRamu0104: If the `link_token` has already expired by the time we parse the response, `expiry_time - now` is negative and `expires_in_seconds` goes back to the client as a negative number.

Unlikely in practice (Plaid link tokens are ~4h and we're reading this immediately), but clock skew or a retried/replayed response would get there. Clamping to zero — or returning `None` once it's non-positive, which matches how an unparseable expiration is already handled — would keep the field's contract honest for SDK consumers.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
```diff
@@ -0,0 +1,706 @@
...
+}
+
+impl common_utils::errors::ErrorSwitch<ConnectorError> for PlaidErrorResponse {
+    fn switch(&self) -> ConnectorError {
+        ConnectorError::ResponseHandlingFailed {
+            context: domain_types::errors::ResponseTransformationErrorContext {
+                http_status_code: None,
+                additional_context: self.error_message.clone(),
+            },
+        }
```
JeevaRamu0104, line 704: `http_status_code` is hardcoded to `None` here, so any error surfaced through `ErrorSwitch` loses the status code. `build_error_response` in `plaid.rs` gets this right (it has `res.status_code` to hand), so the two paths produce inconsistently detailed diagnostics for the same underlying Plaid error.

Not really fixable inside this impl — `switch(&self)` has no access to the response — so it'd need a signature change to thread the status through. Fine to leave as-is for this PR, just flagging it so it doesn't get mistaken for an intentional omission.

JeevaRamu0104 (commented):
A few smaller things I noticed while reading through the authenticator wiring — none of these block on their own, but the first one is worth a second look since it quietly changes behaviour for connectors that have nothing to do with Plaid. Left them inline.

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: `switch(&self)` has no access to the HTTP response, so the status code cannot be threaded through here. `build_error_response` in plaid.rs uses `res.status_code` and produces richer diagnostics for the same error.

crates/types-traits/domain_types/src/types.rs
Aishwariyaa-Anand: fireing a tokenize call for billwerk/hipay/paysafe/braintree/getnet/mollie/stax/finix will happen only if should_do_payment_method_token is true. should_do_payment_method_token is only called in composite service, where the payment method is taken from request and not paymentflowdata

JeevaRamu0104 (dismissed):

crates/integrations/connector-integration/src/authenticator_connectors/plaid/transformers.rs
Aishwariyaa-Anand: The preference order no longer applies in the updated implementation. Instead of picking one account type per account, we now collect all available routing details ACH, BACS, and SEPA and return them together in the accounts list. So the currency-based prioritization is no longer used; the caller receives the full set of accounts

deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/authenticator_connectors/plaid.rs
gopikrishna000: in that case, plaid code will be present in both payment connector and authentication connector folder?., similar to payment and payout supported connectors in both folder

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -2078,6 +2079,18 @@ impl PaymentMethodService for PaymentMethod {
...
+                            &config,
+                            payload,
+                            &metadata_payload,
+                            &request_data.masked_metadata,
+                            &service_name,
+                            request_id,
+                        )).await?;
+                        return Ok(tonic::Response::new(result));
+                    }
+
```
AmitsinghTanwar007: Why can’t we use a `match` here and explicitly return “not supported” for the other variants, such as `Payout`?

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -2842,6 +3019,41 @@ impl MerchantAuthenticationOperational for MerchantAuthentication {
...
+        error_stack::Report<ucs_env::error::GrpcError>,
+    > {
+        if matches!(
+            request.extracted_metadata.connector,
+            ConnectorVariant::Authenticator(_)
+        ) {
+            self.internal_sdk_session_token_authenticator(request).await
+        } else {
+            self.internal_sdk_session_token_payment(request).await
+        }
```
AmitsinghTanwar007: +1

gopikrishna000 (dismissed):

AmitsinghTanwar007 (dismissed):

gopikrishna000 (dismissed):

AmitsinghTanwar007 (dismissed):

AmitsinghTanwar007 (approved):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

PR #1978, opened 2026-07-20
fix(docs): mark Linux x86_64 as unsupported in platform matrices

XyneSpaces:
## Summary
The platform-support tables across the repo advertised Linux x86_64 (64-bit Intel) as a supported target, which is not accurate. This PR removes/updates that claim so the docs no longer advertise an unsupported platform.

## Changes
- `README.md` — remove `| Linux | x86_64 |` row from Platform Support table
- `docs/README.md` — same removal
- `sdk/README.md` — `Linux | x86_64, aarch64` → `Linux | aarch64`
- `sdk/python/README.md`, `sdk/java/README.md` — `✅ Linux (x64, arm64)` → `✅ Linux (arm64)`
- `sdk/javascript/README.md` — `Linux (x64, arm64)` → `Linux (arm64)`

Interpreted "Linux 64 not supported" as Linux x86_64 (64-bit Intel); Linux aarch64 (64-bit ARM) left intact. If aarch64 should also be dropped, say so and I'll amend.

## Type
Docs only — no code changes, no build impact.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -246,7 +246,6 @@ The `hyperswitch-prism` SDK contains platform-specific native libraries compiled
 | Platform | Architecture |
 |----------|--------------|
 | macOS (Apple Silicon) | arm64 |
-| Linux | x86_64 |
 
 ### Reporting Vulnerabilities
 Please report security issues to [security@juspay.in](mailto:security@juspay.in).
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -246,7 +246,6 @@ The `hyperswitch-prism` SDK contains platform-specific native libraries compiled
 | Platform | Architecture |
 |----------|--------------|
 | macOS (Apple Silicon) | arm64 |
-| Linux | x86_64 |
```
JeevaRamu0104, line 249: we support this version. we need to support arm64 as well.

PR #1980, opened 2026-07-21
ci: harden release pipeline with CI gates, authorization, pinned actions and provenance

shuklatushar226:
## Summary

> **Rebased onto #1973** (npm OIDC trusted publishing): the `release-javascript.yml` deletion and npm provenance from this PR were superseded by that work and dropped; the SHA pins were re-applied on top of the new trusted-publishing code, including its new `setup-node` steps.

An audit of the release pipeline (nightly CalVer → Docker/GHCR, stable SemVer → Maven/PyPI/npm, hotfix flow) found that every publish path ships on trust: nothing verifies CI passed on the released commit, the release authorization check is commented out, and all third-party actions float on mutable refs while holding publish secrets. This PR closes those gaps without changing the release cadence or day-to-day flow.

### Changes

| # | Fix | Files |
|---|---|---|
| 1 | **Validate release tags before publish** — a `validate-tag` job rejects any pushed tag that isn't strict CalVer (`YYYY.MM.DD.N[-hotfixN]`) or SemVer (`vX.Y.Z`) before any GHCR-push job runs, and the tag trigger globs use `+` quantifiers so they admit only digits and dots. (An earlier `verify-ci` CI-result gate was dropped after review: nightly/stable tags land on `chore(version)` CHANGELOG-only commits whose `CI Result` passes by construction — path filters skip the real jobs and `ci-gate` counts skipped as pass — so it verified nothing. See discussion.) | `docker-to-ghcr-publish.yml` |
| 2 | **Re-enable release authorization** — restores the team-membership check (same pattern as upstream `juspay/hyperswitch`) via a GitHub App token, on the stable release and both hotfix workflows. Fails closed. The check itself lives in a single reusable `workflow_call` workflow that each entry point gates on with `needs: [authorize]`, so the policy has one source of truth (per review); secrets are passed explicitly rather than with `secrets: inherit`, and the allowed team is defined in the reusable workflow with no inputs, so callers cannot diverge. | `authorize-release.yml`, `release-stable-version.yml`, `create-hotfix-*.yml` |
| 3 | ~~Pin mutable refs~~ — reverted after review: action refs stay on version tags (`@v4`), matching the `juspay/hyperswitch` convention (verified: upstream has zero SHA-pinned actions). SHA-pinning is listed under follow-ups as an org-level decision. | — |
| 4 | **Provenance** — `actions/attest-build-provenance` on the multi-arch Docker manifest. (npm provenance now arrives automatically via the OIDC trusted publishing introduced in #1973, so no explicit flag is needed.) | `docker-to-ghcr-publish.yml` |
| 5 | **Robust version computation** — next version read from cog stdout instead of regex-scraping mixed log output; `cog get-version` failure now fails hard instead of silently falling back to `0.0.0` (which would have produced a full-history changelog); hotfix workflows share a `concurrency` group so concurrent dispatches can't race on tags; CHANGELOG marker is verified before the sed insertion that previously no-op'd silently. | `release-stable-version.yml`, `create-hotfix-*.yml` |
| 6 | ~~Remove duplicate npm publish path~~ — superseded: #1973 deleted `release-javascript.yml` on main first (this branch had the identical deletion; it was dropped on rebase). | — |
| 7 | **Stable Docker images** — the tag trigger now also matches `v*` SemVer tags, so stable releases produce a container, not just nightlies. | `docker-to-ghcr-publish.yml` |
| 8 | **Small bugs** — export the `stripe` path-filter output that `sdk-test` already consumes (the trigger was dead since it was never exported); align `sdk/python/setup.py` placeholder version with `pyproject.toml`. | `ci.yml`, `sdk/python/setup.py` |

Shell-injection hygiene: workflow inputs and `github` context values used in `run:` steps are routed through `env:` vars — including `docker-to-ghcr-publish.yml`'s `github.ref_name` uses, which previously were interpolated inline after GHCR login in a `packages: write` job (flagged in review). The `validate-tag` job also rejects tags that are not strict CalVer (`YYYY.MM.DD.N[-hotfixN]`) or SemVer (`vX.Y.Z`) before any build job starts, and the tag trigger globs use `+` quantifiers so they admit only digits and dots.

### Authorization credentials (fix 2)

The check uses the existing `HYPERSWITCH_BOT_APP_ID` / `HYPERSWITCH_BOT_APP_PRIVATE_KEY` secrets — already configured on this repo (used by `ci.yml` since #1141) and the same app upstream `juspay/hyperswitch` uses for its identical release-authorization check. **No maintainer setup is needed.**

The app token is authentication only: it holds org-wide *Organization members: read* and is used solely to query team membership. Authorization is decided by the workflow's `allowed_teams` list — membership in `connector-service-maintainers` is required; hyperswitch maintainers gain no release rights here.

### Post-merge verification

1. Dispatch **Release a stable version** with `dry_run: true` on the latest CalVer tag — exercises the auth gate and new version computation without publishing anything.
2. Dispatch **Docker Build and Push to GHCR** from a branch — produces a harmless `{branch}-{sha}` image and exercises `validate-tag` + attestation.
3. `gh attestation verify oci://ghcr.io/juspay/hyperswitch-prism:<tag> --owner juspay` once the first gated image ships.

### Follow-ups (not in this PR)

PyPI trusted publishing (needs pypi.org-side config) · SHA-pinning actions (org-level convention decision — upstream hyperswitch uses version tags everywhere; if adopted, the nightly reusable workflow ref pointing at another repo's mutable `@main` is the highest-value pin) · medusa release workflows pushing version bumps directly to the dispatched ref · CI path-filter escape hatches · running `ci.yml` on `hotfix-*` branch pushes so hotfix commits get any CI coverage at all (they currently get none) · requiring the `CI Result` check via branch protection (needs repo admin) · gating the remaining dispatchable release workflows (`release-sdks.yml`, `release-medusa-custom-payments*.yml`) on `authorize-release.yml` — they have no authorization check today, and adding one is now a two-line `needs:`, but it is a policy change rather than part of this refactor.

### DIFF
--- a/.github/workflows/create-hotfix-branch.yml
+++ b/.github/workflows/create-hotfix-branch.yml
@@ -8,89 +8,17 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  authorize:
+    uses: ./.github/workflows/authorize-release.yml
+    secrets:
+      HYPERSWITCH_BOT_APP_ID: ${{ secrets.HYPERSWITCH_BOT_APP_ID }}
+      HYPERSWITCH_BOT_APP_PRIVATE_KEY: ${{ secrets.HYPERSWITCH_BOT_APP_PRIVATE_KEY }}
+
   create_branch:
     runs-on: ubuntu-latest
+    needs: [authorize]
 
     steps:
-      - name: Generate GitHub app token
-        id: generate_app_token
-        uses: actions/create-github-app-token@v2
-        with:
-          app-id: ${{ secrets.HYPERSWITCH_BOT_APP_ID }}
-          private-key: ${{ secrets.HYPERSWITCH_BOT_APP_PRIVATE_KEY }}
-
-      - name: Check if user is authorized to trigger workflow
-        shell: bash
-        env:
-          GH_TOKEN: ${{ steps.generate_app_token.outputs.token }}
-          ORG_NAME: ${{ github.repository_owner }}
-          TRIGGERING_ACTOR: ${{ github.triggering_actor }}
-        run: |
-          echo "::add-mask::${GH_TOKEN}"
-
-          # 200 can also be returned for a pending invitation, so require
-          # state == "active"; 404 means not a member. Anything else is a
-          # credential/API/network problem — fail the workflow as an
-          # infrastructure error instead of reporting it as "not authorized".
-          function is_user_active_team_member() {
-            username="${1}"
-            team_slug="${2}"
-            response_file="$(mktemp)"
-
-            status_code="$(
-              curl \
-                --location \
-                --silent \
-                --output "${response_file}" \
-                --write-out '%{http_code}' \
-                --header 'Accept: application/vnd.github+json' \
-                --header 'X-GitHub-Api-Version: 2022-11-28' \
-                --header "Authorization: Bearer ${GH_TOKEN}" \
-                "https://api.github.com/orgs/${ORG_NAME}/teams/${team_slug}/memberships/${username}" \
-                || true
-            )"
-
-            case "${status_code}" in
-              200)
-                state="$(jq --raw-output '.state // "unknown"' "${response_file}")"
-                rm -f "${response_file}"
-                if [[ "${state}" == "active" ]]; then
-                  return 0
-                fi
-                echo "${username} has membership state '${state}' in team ${team_slug}; only active members are authorized."
-                return 1
-                ;;
-              404)
-                rm -f "${response_file}"
-                return 1
-                ;;
-              *)
-                echo "::error::Membership check for team '${team_slug}' failed with HTTP status '${status_code}' — credential, API, or network problem, not an authorization decision. Response: $(head --bytes 300 "${response_file}")"
-                rm -f "${response_file}"
-                exit 1
-                ;;
-            esac
-          }
-
-          allowed_teams=('connector-service-maintainers')
-          is_user_authorized=false
-          username="${TRIGGERING_ACTOR}"
-
-          for team in "${allowed_teams[@]}"; do
-            if is_user_active_team_member "${username}" "${team}"; then
-              is_user_authorized=true
-              break
-            fi
-          done
-
-          if ${is_user_authorized}; then
-            echo "${username} is authorized to trigger workflow"
-          else
-            printf -v allowed_teams_comma_separated '%s, ' "${allowed_teams[@]}"
-            echo "::error::${username} is not authorized to trigger workflow; must be a member of one of these teams: ${allowed_teams_comma_separated%, }"
-            exit 1
-          fi
-
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/create-hotfix-branch.yml
+++ b/.github/workflows/create-hotfix-branch.yml
@@ -16,8 +16,8 @@ jobs:
         id: generate_app_token
         uses: actions/create-github-app-token@v2
         with:
-          app-id: ${{ secrets.CONNECTOR_SERVICE_BOT_APP_ID }}
-          private-key: ${{ secrets.CONNECTOR_SERVICE_BOT_APP_PRIVATE_KEY }}
+          app-id: ${{ secrets.HYPERSWITCH_BOT_APP_ID }}
+          private-key: ${{ secrets.HYPERSWITCH_BOT_APP_PRIVATE_KEY }}
 
       - name: Check if user is authorized to trigger workflow
         shell: bash
--- a/.github/workflows/create-hotfix-branch.yml
+++ b/.github/workflows/create-hotfix-branch.yml
@@ -10,6 +10,9 @@ concurrency:
 jobs:
   authorize:
     uses: ./.github/workflows/authorize-release.yml
+    # The membership check authenticates with a GitHub app token minted
+    # from the secrets below and never uses GITHUB_TOKEN.
+    permissions: {}
     secrets:
       HYPERSWITCH_BOT_APP_ID: ${{ secrets.HYPERSWITCH_BOT_APP_ID }}
       HYPERSWITCH_BOT_APP_PRIVATE_KEY: ${{ secrets.HYPERSWITCH_BOT_APP_PRIVATE_KEY }}
--- a/.github/workflows/create-hotfix-branch.yml
+++ b/.github/workflows/create-hotfix-branch.yml
@@ -28,25 +28,48 @@ jobs:
         run: |
           echo "::add-mask::${GH_TOKEN}"
 
-          function is_user_team_member() {
+          # 200 can also be returned for a pending invitation, so require
+          # state == "active"; 404 means not a member. Anything else is a
+          # credential/API/network problem — fail the workflow as an
+          # infrastructure error instead of reporting it as "not authorized".
+          function is_user_active_team_member() {
             username="${1}"
             team_slug="${2}"
+            response_file="$(mktemp)"
 
-            # The API returns 200 if the user is a member of the team and 404
-            # otherwise, so compare on the HTTP status code.
             status_code="$(
               curl \
                 --location \
                 --silent \
-                --output /dev/null \
+                --output "${response_file}" \
                 --write-out '%{http_code}' \
                 --header 'Accept: application/vnd.github+json' \
                 --header 'X-GitHub-Api-Version: 2022-11-28' \
                 --header "Authorization: Bearer ${GH_TOKEN}" \
-                "https://api.github.com/orgs/${ORG_NAME}/teams/${team_slug}/memberships/${username}"
+                "https://api.github.com/orgs/${ORG_NAME}/teams/${team_slug}/memberships/${username}" \
+                || true
             )"
 
-            [[ status_code -eq 200 ]]
+            case "${status_code}" in
+              200)
+                state="$(jq --raw-output '.state // "unknown"' "${response_file}")"
+                rm -f "${response_file}"
+                if [[ "${state}" == "active" ]]; then
+                  return 0
+                fi
+                echo "${username} has membership state '${state}' in team ${team_slug}; only active members are authorized."
+                return 1
+                ;;
+              404)
+                rm -f "${response_file}"
+                return 1
+                ;;
+              *)
+                echo "::error::Membership check for team '${team_slug}' failed with HTTP status '${status_code}' — credential, API, or network problem, not an authorization decision. Response: $(head --bytes 300 "${response_file}")"
+                rm -f "${response_file}"
+                exit 1
+                ;;
+            esac
           }
 
           allowed_teams=('connector-service-maintainers')
@@ -54,7 +77,7 @@ jobs:
           username="${TRIGGERING_ACTOR}"
 
           for team in "${allowed_teams[@]}"; do
-            if is_user_team_member "${username}" "${team}"; then
+            if is_user_active_team_member "${username}" "${team}"; then
               is_user_authorized=true
               break
             fi
--- a/.github/workflows/create-hotfix-branch.yml
+++ b/.github/workflows/create-hotfix-branch.yml
@@ -3,11 +3,71 @@ name: Create hotfix branch
 on:
   workflow_dispatch:
 
+concurrency:
+  group: hotfix-release
+  cancel-in-progress: false
+
 jobs:
   create_branch:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Generate GitHub app token
+        id: generate_app_token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.CONNECTOR_SERVICE_BOT_APP_ID }}
+          private-key: ${{ secrets.CONNECTOR_SERVICE_BOT_APP_PRIVATE_KEY }}
+
+      - name: Check if user is authorized to trigger workflow
+        shell: bash
+        env:
+          GH_TOKEN: ${{ steps.generate_app_token.outputs.token }}
+          ORG_NAME: ${{ github.repository_owner }}
+          TRIGGERING_ACTOR: ${{ github.triggering_actor }}
+        run: |
+          echo "::add-mask::${GH_TOKEN}"
+
+          function is_user_team_member() {
+            username="${1}"
+            team_slug="${2}"
+
+            # The API returns 200 if the user is a member of the team and 404
+            # otherwise, so compare on the HTTP status code.
+            status_code="$(
+              curl \
+                --location \
+                --silent \
+                --output /dev/null \
+                --write-out '%{http_code}' \
+                --header 'Accept: application/vnd.github+json' \
+                --header 'X-GitHub-Api-Version: 2022-11-28' \
+                --header "Authorization: Bearer ${GH_TOKEN}" \
+                "https://api.github.com/orgs/${ORG_NAME}/teams/${team_slug}/memberships/${username}"
+            )"
+
+            [[ status_code -eq 200 ]]
+          }
+
+          allowed_teams=('connector-service-maintainers')
+          is_user_authorized=false
+          username="${TRIGGERING_ACTOR}"
+
+          for team in "${allowed_teams[@]}"; do
+            if is_user_team_member "${username}" "${team}"; then
+              is_user_authorized=true
+              break
+            fi
+          done
+
+          if ${is_user_authorized}; then
+            echo "${username} is authorized to trigger workflow"
+          else
+            printf -v allowed_teams_comma_separated '%s, ' "${allowed_teams[@]}"
+            echo "::error::${username} is not authorized to trigger workflow; must be a member of one of these teams: ${allowed_teams_comma_separated%, }"
+            exit 1
+          fi
+
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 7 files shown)


shuklatushar226:
**Re: why the `actions/checkout@v4` → `actions/checkout@<sha> # v4` change, and its reversal**

**Why it was originally done:** a version tag like `@v4` is a movable pointer — the action's author (or anyone who compromises their account) can repoint it to different code, and every workflow using it runs that new code on the next trigger with no visible change on our side. There is real precedent (the `tj-actions/changed-files` compromise, March 2025, repointed tags to secret-exfiltrating code), and these workflows carry npm/PyPI/Maven and GPG credentials. A full commit SHA is immutable, which is why GitHub's security-hardening guide recommends SHA-pinning for third-party actions; the `# v4` comment preserved readability.

**Why it's reverted:** checked all 17 workflows in `juspay/hyperswitch` — upstream uses version tags everywhere (zero SHA pins), so this repo now follows the same convention. The pinning question is better taken as an org-level decision than a single-repo divergence; it's listed in this PR's follow-ups. If the org ever adopts it, the highest-value pin is `release-nightly-version.yml`'s reusable-workflow ref, which executes another repository's `@main` with tag-push rights here.

The PR now touches 6 files: the CI gates, release authorization, Docker attestation, version-computation robustness, hotfix concurrency guards, and two small bug fixes — no ref-style changes.

JeevaRamu0104 (dismissed):

.github/workflows/create-hotfix-branch.yml
```diff
@@ -3,11 +3,71 @@ name: Create hotfix branch
...
 jobs:
   create_branch:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Generate GitHub app token
+        id: generate_app_token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.CONNECTOR_SERVICE_BOT_APP_ID }}
```
gopikrishna000: are these secret there?

.github/workflows/docker-to-ghcr-publish.yml
```diff
@@ -3,14 +3,62 @@ name: Docker Build and Push to GHCR
...
   workflow_dispatch:
   
-  # Automatically run when nightly tags are created.
-  # Accepts formats like: 2025.12.03.0 or 2025.01.09.12
+  # Automatically run when nightly (CalVer) or stable (SemVer) tags are created.
+  # Accepts formats like: 2025.12.03.0, 2025.01.09.12, or v1.2.3
   push:
     tags:
       - "20[0-9][0-9].[0-1][0-9].[0-3][0-9].*"
+      - "v[0-9]*.[0-9]*.[0-9]*"
```
gopikrishna000: Blocking: this is a GitHub glob, not a SemVer regex, so it accepts valid Git refs beyond vX.Y.Z, including a ref containing quote/semicolon shell syntax. The Set version step later substitutes github.ref_name directly into the generated shell script after GHCR login, in a job with packages: write. That allows the ref to break out of the quoted assignment and execute commands. The sink was already reachable through the permissive CalVer pattern; this new trigger expands its reach to v-prefixed tags. Please pass REF_NAME, REF_TYPE, and COMMIT_SHA through step env, use only the shell variables, and validate REF_NAME against strict CalVer/SemVer patterns before login/build.

.github/workflows/create-hotfix-branch.yml
shuklatushar226: Good catch — no, they weren't. `CONNECTOR_SERVICE_BOT_APP_ID` / `CONNECTOR_SERVICE_BOT_APP_PRIVATE_KEY` were never provisioned (the name has no history in this repo; the original commented-out check used `CONNECTOR_SERVICE_CI_PAT`, which lacks `read:org`).

Fixed in e3769a8cb: switched to the existing `HYPERSWITCH_BOT_APP_ID` / `HYPERSWITCH_BOT_APP_PRIVATE_KEY` secrets, which are already configured on this repo (used by `ci.yml` since #1141) and are the same app upstream `juspay/hyperswitch` uses for its identical release-authorization check — so the app is installed here and holds the org-wide **Organization members: read** permission the check needs. No new secrets or maintainer setup required.

Note this doesn't change *who* is authorized: the app token is only used to read team membership; the gate itself still requires membership in `connector-service-maintainers` (unchanged in all three workflows), so hyperswitch maintainers don't gain release rights here.

.github/workflows/create-hotfix-branch.yml
```diff
@@ -3,11 +3,71 @@ name: Create hotfix branch
...
+                --silent \
+                --output /dev/null \
+                --write-out '%{http_code}' \
+                --header 'Accept: application/vnd.github+json' \
+                --header 'X-GitHub-Api-Version: 2022-11-28' \
+                --header "Authorization: Bearer ${GH_TOKEN}" \
+                "https://api.github.com/orgs/${ORG_NAME}/teams/${team_slug}/memberships/${username}"
+            )"
+
+            [[ status_code -eq 200 ]]
```
gopikrishna000: The gate fails closed, but this status-only check has two correctness/diagnostic gaps (also repeated in the other two workflows). First, the membership response includes state, and a pending team invitation can have an HTTP 200 response; authorization should require state == active, not only status 200. Second, every non-200 result is later reported as user not authorized, so invalid/under-scoped credentials, API errors, rate limits, and network failures are mislabeled as a membership denial. Please retain the response body, distinguish 404 (not a member) from transport/auth/API failures, require an active membership, and emit an infrastructure/configuration error for unexpected statuses.

.github/workflows/docker-to-ghcr-publish.yml
gopikrishna000: - "v[0-9]*.[0-9]*.[0-9]*"

  should be:

  - "v[0-9]+.[0-9]+.[0-9]+"

  The distinction is:

  [0-9]*  = one digit, followed by any number of arbitrary characters
  [0-9]+  = one or more digits

.github/workflows/docker-to-ghcr-publish.yml
```diff
@@ -3,14 +3,62 @@ name: Docker Build and Push to GHCR
...
 
 jobs:
+  # Tag pushes fire this workflow directly, so nothing upstream guarantees the
+  # tagged commit passed CI. Block publishing until the "CI Result" check
+  # (ci.yml's ci-gate job) has completed successfully on this exact SHA.
+  # The CI run and the tag push start near-simultaneously, so poll rather
+  # than checking once.
+  verify-ci:
```
gopikrishna000: do we need this, nightly tag is always created from main, which is always ci passed.
its just changelog commit, which always will pass

.github/workflows/docker-to-ghcr-publish.yml
shuklatushar226: Agreed — fixed in fa58b8fbc:

- **Env-routed context**: every `github` context value used by a `run:` step in this workflow (`ref_name`, `ref_type`, `sha`, `repository`) now arrives via step/job `env:` and only shell variables appear in scripts. This also covers the pre-existing CalVer-path sink in `Set version` and the `run:` uses of `env.VERSION` downstream.
- **Strict validation before login/build**: a new first step in `verify-ci` rejects any tag not matching `^(v[0-9]+\.[0-9]+\.[0-9]+|20[0-9]{2}\.[0-9]{2}\.[0-9]{2}\.[0-9]+(-hotfix[0-9]+)?)$`. Both build jobs `need: verify-ci`, so nothing with `packages: write` starts (and no GHCR login happens) for a malformed tag.
- **Tightened trigger globs** (per your other comment) so the workflow doesn't even fire for refs outside strict CalVer/SemVer.

Branch-dispatch builds keep working: branch names aren't format-validated (they're arbitrary), but they're no longer interpolated anywhere, and the Docker-tag sanitization now restricts to `[A-Za-z0-9._-]`.

.github/workflows/create-hotfix-branch.yml
shuklatushar226: Both points taken — fixed in 8e467ac82, in all three workflows (this one, `create-hotfix-tag.yml`, `release-stable-version.yml`):

- The response body is kept and `200` now requires `state == "active"` (`jq`); a pending invitation logs the state and is treated as not authorized.
- `404` alone means "not a member of this team" and continues the team loop.
- Any other status — bad/under-scoped credentials, rate limits, API errors, network failures (curl `000`) — fails the workflow immediately with a `::error` explicitly labeled as a credential/API/network problem (including the response body), so it can't be mistaken for a membership denial. Still fail-closed either way.

Verified the three paths against the live API: active member → authorized; bad token → the labeled 401 infrastructure error; non-member → 404 denial.

.github/workflows/docker-to-ghcr-publish.yml
shuklatushar226: Applied in fa58b8fbc — `v[0-9]+.[0-9]+.[0-9]+`, and the CalVer glob got the same treatment (`20[0-9][0-9].[0-1][0-9].[0-3][0-9].[0-9]+`) since its trailing `.*` had the identical "dot then anything" problem. Hotfix tags (`YYYY.MM.DD.N-hotfixN`) previously matched via that `.*`, so they keep triggering builds through an explicit third pattern rather than silently losing Docker images.

The globs are now the first line of defense and the `verify-ci` regex validation (other thread) is the backstop, since filter patterns can't express everything a strict format check can.

.github/workflows/docker-to-ghcr-publish.yml
shuklatushar226: For the nightly-bot happy path you're right — that tag is on a commit whose CI just ran, and the gate resolves in one API call within seconds. The gate exists for every *other* way this workflow fires, because the trigger is a raw tag push, not the nightly workflow:

- anyone with write access can push a matching tag on **any** commit — an old SHA, a revert, or a commit from an unmerged branch that never passed CI;
- hotfix tags are cut on `hotfix-*` branches, not main, so "main is green" says nothing about them;
- stable `v*` tags (added in this PR) likewise aren't guaranteed to be on a CI-passed commit;
- branch-protection required checks don't apply to tag pushes, so there's no other enforcement point tying "image published" to "CI passed on this exact SHA".

So the cost is ~one API poll on the common path, and it's the only thing standing between a stray/forced tag and a published image. I'd keep it — but if the team prefers trusting tag discipline, it's an isolated job that's trivial to drop later.

gopikrishna000: I understand the value of verifying the exact tagged SHA, but the current implementation does not support the hotfix case mentioned. `create-hotfix-tag.yml` creates a new CHANGELOG commit on a `hotfix-*` branch, while `ci.yml` only runs on pushes to `main`. Therefore, the hotfix SHA never receives a `CI Result`, and `verify-ci` will wait for 90 minutes and block the Docker image. Also, matching CalVer/SemVer tag creation is already restricted by the repository ruleset, so users with write access cannot freely push these tags. This gate should be removed or adjusted so it does not block hotfix publishing.

shuklatushar226: You're right on both counts — thanks for pressure-testing this.

**Hotfix deadlock: confirmed and fixed in 1da61fe34.** `ci.yml` triggers only on `push: main`, `pull_request`, and `merge_group`, so hotfix-branch commits can never carry a `CI Result` check and the wait could only time out. Hotfix tags (`*-hotfix*`) now skip the CI wait with an explicit `::notice`; the PR body's Known-limitation section documents it, and "run `ci.yml` on `hotfix-*` pushes, then lift the exemption" is added to the follow-ups (that would also give the cherry-picked fix commits CI coverage they currently never get).

**Ruleset: confirmed** — the "CalVer and SemVer Tags" ruleset restricts creation/update/deletion of `v*` and CalVer tags, so my earlier "anyone with write access can push a matching tag" was wrong for these patterns. Retracted.

On removal vs. adjustment, I'd still keep the (now hotfix-exempt) gate, for one narrow reason: the nightly reusable workflow (`juspay/hyperswitch` `release-nightly-version-reusable.yml`) tags main HEAD without checking CI — its only guard is "HEAD isn't already tagged". Required checks run at merge time, but the post-merge `push` CI run on main can still go red (flaky/env/merge-skew), and the nightly would tag and ship that tip. The ruleset governs *who* can create tags; the gate is the only check on *whether the tagged commit was green*. For stable `v*` tags it's redundant (the release workflow verifies CI before tagging) but free, since that commit is on main with a `CI Result`.

That said, it's an isolated job — if you still prefer dropping it, no objection.

shuklatushar226: You're completely right, and thanks for not letting this one go — I traced it end to end and the gate was verifying nothing:

- `release-nightly-version-reusable.yml` creates a `chore(version): <tag>` **CHANGELOG-only** commit, tags *that* commit, and pushes it to `main`.
- `ci.yml` runs on it (`push: main`), but the CHANGELOG-only diff makes the path filters skip `check`/`test`/`sdk-test`, and `ci-gate` treats skipped as pass (`ci.yml:766` — "passed (or were skipped due to path filtering)"; it only fails on `failure`/`cancelled`).
- So `CI Result: success` on a nightly tag is green **by construction**, exactly as you said — "only the changelog commit is added, it can't fail CI on already-passed main."

And it wasn't just the Docker job: `release-stable-version.yml`'s own "Verify CI passed on tagged commit" step had the identical hole, since the CalVer tag it checks *is* that changelog commit.

Removed both in a957d7ec5. What stays on the Docker side is `validate-tag`, which keeps only the strict CalVer/SemVer **format** check — that's the guard for the injection sink (the tag name becomes the image version), not a CI gate, so it's real work with no dependency on `CI Result`. The rest of the PR (auth re-enable, provenance/attestation, the ref-injection fix, tighter tag globs, robust version computation, stable Docker images, small bugfixes) is untouched.

PR description and the "known limitation" section updated to match. If it'd be useful separately, running `ci.yml` on `hotfix-*` branches (so hotfix commits get any CI coverage at all — today they get none) is noted as a follow-up.

.github/workflows/create-hotfix-branch.yml
```diff
@@ -3,11 +3,94 @@ name: Create hotfix branch
...
+concurrency:
+  group: hotfix-release
+  cancel-in-progress: false
+
 jobs:
   create_branch:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Generate GitHub app token
```
gopikrishna000: Non-blocking: this fail-closed authorization implementation is duplicated in `create-hotfix-branch.yml`, `create-hotfix-tag.yml`, and `release-stable-version.yml`. The active-membership and HTTP-status fixes in this PR already had to be applied to all three copies, so future security fixes could drift.

Could we centralize it as a reusable `workflow_call` workflow (for example, `.github/workflows/authorize-release.yml`) and make each independently dispatchable release job depend on it with `needs: authorize`? Each entry point would remain protected, but the policy and error handling would have one source of truth. Please pass only the App ID/private-key secrets explicitly rather than using `secrets: inherit`, and keep `connector-service-maintainers` defined in the reusable workflow so callers cannot accidentally diverge.

shuklatushar226: Good catch — and the duplication was worse than "three similar copies": when I extracted the block, all three were **byte-for-byte identical**, which is exactly the drift risk you're describing (they only stayed identical because I hand-applied the same two fixes three times in 8e467ac82).

Done in 81fade4af, following your spec exactly:

- **`.github/workflows/authorize-release.yml`** — `on: workflow_call`, one `authorize` job holding the app-token generation and the membership check. The bash moved verbatim (diffed byte-for-byte against the deleted block), so there's no behavioural change to re-verify.
- **Explicit secrets, no `inherit`** — the workflow declares `HYPERSWITCH_BOT_APP_ID` and `HYPERSWITCH_BOT_APP_PRIVATE_KEY` as `required: true`, and each caller maps just those two. That also keeps `CONNECTOR_SERVICE_CI_PAT` out of the called workflow entirely, which `inherit` would not have.
- **Team not overridable** — `allowed_teams=('connector-service-maintainers')` stays in the reusable workflow and there is **no `inputs:` block at all**, so a caller has no mechanism to pass a different team.
- **All three entry points gated** via `needs: [authorize]` (`create_branch`, `create_tag`, `tag-release`). Fail-closed is preserved by `needs:` semantics: if `authorize` fails, dependents are skipped, never run. `publish-sdks` inherits the gate transitively through `needs: [tag-release]`.
- **`permissions: {}`** on the authorize job — it authenticates with the app token and never touches `GITHUB_TOKEN`, and this matters because `release-stable-version.yml` sets workflow-level `contents: write`/`id-token: write`, which would otherwise form the ceiling handed to the called workflow.

Net: −237/+124 lines, and the calling convention matches the existing `release-nightly-version.yml` (explicit `secrets:`, never `inherit`). `actionlint` on the full workflow directory reports the same 39 pre-existing shellcheck hints as before the change, none in the new file.

One thing your comment surfaced indirectly: `release-sdks.yml` and the two `release-medusa-custom-payments*.yml` workflows are also `workflow_dispatch`-triggered and have **no** authorization check at all. Gating them is now a two-line `needs:`, but it would block people who can dispatch them today — that's a policy call rather than part of this refactor, so I've listed it under follow-ups. Happy to fold it in here if you'd prefer.

gopikrishna000 (approved):

JeevaRamu0104 (approved):

PR #1986, opened 2026-07-21
feat(paysafe): Apple Pay / Google Pay recurring + encrypted GPay passthrough

shuklatushar226:
## What

Adds connector-side support for **Paysafe wallet recurring (CIT→MIT)** for Apple Pay and Google Pay, plus an **encrypted Google Pay token passthrough**. Card recurring already worked; this closes the wallet gaps.

## Changes

**Wallet vault conversion (leg 2).** Paysafe's customer-vault endpoint rejects raw `applePay`/`googlePay` payloads (`5068 "CARD object must be present"`), so a reusable wallet handle is produced by converting the single-use handle via `POST v1/customers/{id}/paymenthandles {paymentHandleTokenFrom}`. The Tokenize request is now an untagged enum `PaysafePaymentMethodTokenRequest::{Handle, VaultFromHandle}`; the conversion leg fires when a wallet + connector customer + echoed handle (`connector_feature_data`) are present, and the URL selector routes it to the customer endpoint. The Tokenize proto carries feature data on the request (not `PaymentFlowData`), so `paysafe_parse_feature_data_handle_token` reads it from there.

**Mandate account resolution.** A converted (`paymentType: CARD`) wallet handle has no account binding, so both the CIT settle and the MIT must send an explicit `accountId`, and the MIT must replay the **same** account as the CIT (`3061 "Invalid initial transaction reference"` otherwise). A `payment_method` discriminator — serialized as `pm` to fit the hyperswitch `VARCHAR(128)` mandate column, with a `payment_method` serde `alias` for older mandates — is stamped into the mandate reference at CIT time and drives resolution: Apple Pay → apple_pay slot (decrypt→encrypt→card fallback), Card / Google Pay → card `no_three_ds`.

**Encrypted Google Pay.** `PaysafeGooglePayTokenizationData` becomes an untagged `Decrypted`/`Encrypted` enum; encrypted tokens are passed through for Paysafe to decrypt gateway-side instead of being rejected (mirrors Adyen). The decrypted-token builder is extracted to a helper.

## Testing

Verified end-to-end on Paysafe sandbox (account `1002671400`) through hyperswitch → UCS → Paysafe:
- **Google Pay** real DIRECT-decrypt token: CIT succeeded (pm saved) → MIT ×2 succeeded.
- **Apple Pay** real Safari PKPaymentToken: CIT succeeded (pm saved) → MIT ×2 succeeded.
- **Card** CIT→MIT regression: green.

Paired with the hyperswitch-side changes (connector-customer gate for wallets, leg-2 conversion orchestration, one-shot tokenization, mandate config).

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/paysafe/requests.rs
+++ b/crates/integrations/connector-integration/src/connectors/paysafe/requests.rs
@@ -322,13 +322,27 @@ pub struct PaysafeGooglePayCardInfo {
 }
 
 #[derive(Debug, Serialize, Clone, PartialEq)]
-#[serde(rename_all = "camelCase")]
-pub struct PaysafeGooglePayTokenizationData {
-    /// Always "PAYMENT_GATEWAY"
-    #[serde(rename = "type")]
-    pub token_type: String,
-    /// The decrypted Google Pay token data
-    pub decrypted_token: PaysafeGooglePayDecryptedToken,
+#[serde(untagged)]
+pub enum PaysafeGooglePayTokenizationData {
+    /// Pre-decrypted token forwarded as `decryptedToken` (upstream decrypted).
+    Decrypted {
+        /// Always "PAYMENT_GATEWAY"
+        #[serde(rename = "type")]
+        token_type: String,
+        /// The decrypted Google Pay token data
+        #[serde(rename = "decryptedToken")]
+        decrypted_token: PaysafeGooglePayDecryptedToken,
+    },
+    /// Raw encrypted Google Pay SDK token passed through for Paysafe to
+    /// decrypt gateway-side (requires the merchant's Google Pay
+    /// gatewayMerchantId to be provisioned with Paysafe).
+    Encrypted {
+        /// Always "PAYMENT_GATEWAY"
+        #[serde(rename = "type")]
+        token_type: String,
+        /// The raw `tokenizationData.token` string from the Google Pay SDK
+        token: Secret<String>,
+    },
 }
 
 #[derive(Debug, Serialize, Clone, PartialEq)]
@@ -344,8 +358,8 @@ pub struct PaysafeGooglePayDecryptedToken {
 pub struct PaysafeGooglePayPaymentMethodDetails {
     pub auth_method: PaysafeGooglePayAuthMethod,
     pub pan: Secret<String>,
-    pub expiration_month: u8,
-    pub expiration_year: u16,
+    pub expiration_month: Secret<String>,
+    pub expiration_year: Secret<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub cryptogram: Option<Secret<String>>,
 }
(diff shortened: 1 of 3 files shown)


AmitsinghTanwar007 (dismissed):

crates/integrations/connector-integration/src/connectors/paysafe/requests.rs
```diff
@@ -322,13 +322,27 @@ pub struct PaysafeGooglePayCardInfo {
...
+    },
+    /// Raw encrypted Google Pay SDK token passed through for Paysafe to
+    /// decrypt gateway-side (requires the merchant's Google Pay
+    /// gatewayMerchantId to be provisioned with Paysafe).
+    Encrypted {
+        /// Always "PAYMENT_GATEWAY"
+        #[serde(rename = "type")]
+        token_type: String,
+        /// The raw `tokenizationData.token` string from the Google Pay SDK
+        token: String,
```
deepanshu-iiitu: Lets make this a secret

crates/integrations/connector-integration/src/connectors/paysafe/transformers.rs
```diff
@@ -107,8 +107,177 @@ pub struct PaysafeMandateMetadata {
...
+    /// Serialized as `pm` for the same VARCHAR(128) budget reason; `alias`
+    /// keeps mandates minted with the long key decodable.
+    #[serde(
+        default,
+        rename = "pm",
+        alias = "payment_method",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub payment_method: Option<PaysafeMandatePaymentMethod>,
+}
```
deepanshu-iiitu, line 135: Why do we need this?

crates/integrations/connector-integration/src/connectors/paysafe/transformers.rs
```diff
@@ -107,8 +107,177 @@ pub struct PaysafeMandateMetadata {
...
+        .peek()
+        .parse::<u16>()
+        .map_err(|_| IntegrationError::InvalidDataFormat {
+            field_name: "google_pay_decrypted_data.card_exp_year",
+            context: IntegrationErrorContext {
+                additional_context: Some(
+                    "Google Pay PAN expiration year must be a numeric YYYY value.".to_string(),
+                ),
+                ..Default::default()
+            },
```
deepanshu-iiitu: Use utls properly here

deepanshu-iiitu (approved):

deepanshu-iiitu (changes requested):
.

crates/integrations/connector-integration/src/connectors/paysafe/transformers.rs
shuklatushar226: Good question — but `connector_mandate_id` isn't the field being deprecated here. What's deprecated is only the `MandateReference` proto *oneof wrapper*, replaced by the flat `MandateReferenceDetails` — and that new message still carries `connector_mandate_id` as field 1. It's the canonical CIT→MIT channel: ~30 connectors read `get_connector_mandate_id()` on the MIT/RepeatPayment path (cybersource, stripe, adyen, nmi, braintree, authorize.net, fiuu, worldpay, …). `payment_method_id` is effectively vestigial on this path — only Stripe writes it (mirroring `connector_mandate_id` "for backward compatibility"), and only Airwallex reads it.

The `pm` discriminator is needed because a converted (customer-vaulted, `paymentType: CARD`) Paysafe handle has **no** account binding, so the MIT must send an explicit `accountId` that matches the account the CIT settled under (Paysafe returns 3061 "Invalid initial transaction reference" otherwise). Apple Pay settles under a specific encrypt/decrypt processing account, so we record which slot to replay at MIT; Card / Google Pay resolve to card `no_three_ds`, so `pm` is omitted for them (hence it's `Option`, `skip_serializing_if`).

The natural home for this would be `mandate_metadata`, but that's dropped on the gRPC recurring path (`Charge → RepeatPaymentData` hardcodes it to `None`), which is why we pack `token + initial_transaction_id + discriminator` into `connector_mandate_id` — the same composite-in-`connector_mandate_id` approach authorize.net uses (`"{customer}-{payment}"`). Keys are shortened (`t`/`i`/`pm`) with `alias` for backward-compat specifically to stay under the `VARCHAR(128)` `connector_mandate_id` column.

Happy to revisit if the platform direction is genuinely to move mandate references onto `payment_method_id` — but as the code stands today, `connector_mandate_id` is the correct/only reliable channel.

crates/integrations/connector-integration/src/connectors/paysafe/requests.rs
shuklatushar226: Done — `token` is now `Secret<String>`, and the passthrough construction wraps it (`Secret::new(encrypted.token.clone())`). Pushed in `21d68c0`.

crates/integrations/connector-integration/src/connectors/paysafe/transformers.rs
shuklatushar226: Done — the builder now takes the masked expiry straight from the shared utils (`get_expiry_month()` / `get_four_digit_expiry_year()`), and `expiration_month`/`expiration_year` are `Secret<String>` instead of plaintext `u8`/`u16`. Dropped the hand-rolled `.peek().parse()` + custom error blocks. This matches Paysafe's own card path (`PaysafeCardExpiry { month, year }`) and how cybersource/bankofamerica build predecrypted wallet tokens. Pushed in `21d68c0`.

AmitsinghTanwar007 (approved):

deepanshu-iiitu (approved):

PR #1991, opened 2026-07-22
fix(tsys_transit): infer card family from PAN when network is absent (Visa MIT NTI)

awasthi21:
## Summary

A MIT card fetched from the locker (`CardDetailsForNetworkTransactionId`) can
arrive **without `card_network`**. The `tsys_transit` connector derives its
`CardFamily` purely from the supplied network, so such a card was classified as
`CardFamily::Unknown` — and the Visa-only `cardOnFileTransactionIdentifier` (the
stored network transaction id) was **skipped** on unscheduled / merchant-initiated
Visa transactions, losing the link to the original stored credential.

This adds a BIN fallback: when the network is not supplied, infer the family from
the card number via `get_card_issuer`.

- `CardFamily::from_card_number(&str)` — map the PAN's issuer → family.
- `CardFamily::from_network_or_number(network, pan)` — prefer the explicit
  network, fall back to BIN detection.
- `derive_for_authorize` uses the fallback for both `Card` and
  `CardDetailsForNetworkTransactionId`.

## Verification (live, TSYS staging)

An unscheduled Visa MIT (`recurring_details: payment_method_id`,
`mit_category: unscheduled`) — before: no `cardOnFileTransactionIdentifier`;
after: `<cardOnFileTransactionIdentifier>000000003878032</cardOnFileTransactionIdentifier>`,
TSYS `A0000`.

## Test

`card_family_falls_back_to_bin_when_network_absent` — asserts Visa is recovered
from the PAN, and that an explicit network still wins over BIN detection.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/tsys_transit/profile/card_family.rs
+++ b/crates/integrations/connector-integration/src/connectors/tsys_transit/profile/card_family.rs
@@ -33,6 +33,35 @@ impl CardFamily {
         }
     }
 
+    /// Fallback when the card network is not supplied — infer the family from
+    /// the card number's BIN. A MIT card fetched from the locker
+    /// (`CardDetailsForNetworkTransactionId`) can arrive without a network, and
+    /// cert-critical fields (cardOnFile / cardOnFileTransactionIdentifier) are
+    /// gated on the family, so the network must still be recognised from the PAN.
+    pub fn from_card_number(card_number: &str) -> Self {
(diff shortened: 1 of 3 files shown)


awasthi21:
Note: this card-family-from-PAN change is now also included in #1988 (extended with a 16-digit Diners BIN fallback), since the recurring/installment MIT cert cases there depend on it. This PR can be merged first and #1988 rebased, or closed in favour of #1988 — whichever the team prefers.

PR #1994, opened 2026-07-22
feat(events): stamp deployment version metadata on every connector event

kanikac199:
## What

Stamp typed deployment / runtime metadata on every connector event so A/B monitoring can group by build and deployment:

- **`version`** — the compiled build (`ucs_env::git_describe!()`), set in `main`; always present.
- **`application_name` / `deployment_id` / `pod_name`** — optional, supplied by the deployment platform via **`CS__RUNTIME_METADATA__*`** config (k8s Downward API: `metadata.name`, deployment labels). Omitted when a platform can't provide them.

## Design

- A typed **`RuntimeMetadata`** struct (`common_utils::events`) `#[serde(flatten)]`'d into `Event` → top-level keys, compiler-checked (no loose `additional_fields` string keys).
- The same struct is the `[runtime_metadata]` config section; `version` is set in `main` from the binary, never read from config.
- Threaded through `EventProcessingParams` from **application state** — no process-global env or `OnceLock`.
- **No `HOSTNAME` inference.** Deployment-owned identity is provided explicitly, never reverse-engineered from a pod name. Missing optional values never fail startup or event creation.

## Testing

- **Behavior-level serialization tests** (`common_utils`): all supplied metadata reaches top-level event fields; absent optionals are omitted; `version` is always present; minimal metadata neither fails event creation nor serialization. **`test result: ok. 3 passed`.**
- `cargo clippy -p grpc-server -p external-services --all-targets -- -D warnings` (default features) — **clean**. Verified locally.

## Deployment (separate infra change)

The `CS__RUNTIME_METADATA__*` env injection (`POD_NAME` ← `metadata.name`, static `APPLICATION_NAME`, and `DEPLOYMENT_ID` where a valid source exists) lives in `hyperswitch-infra` (`infra-configurations/hyperswitch-ucs/*-values.yaml`) — **not in this code PR**. Until wired, only `version` is populated, which is the intended "omit rather than infer" behavior.


crates/common/common_utils/src/events.rs
```diff
@@ -293,6 +293,34 @@ impl Event {
...
+        self.push_version_field("deployment_id", deployment_id);
+    }
+
+    /// The pod name (k8s downward API `metadata.name`).
+    pub fn add_pod_name(&mut self, pod_name: &str) {
+        self.push_version_field("pod_name", pod_name);
+    }
+
+    /// Insert a masked top-level version-metadata field into `additional_fields`.
+    fn push_version_field(&mut self, key: &str, value: &str) {
```
gopikrishna000: should this name be generic like add_additional_fields, as the impl is not tied to version

kanikac199:
@gopikrishna000 addressed in `29295106a`:

- **Rename:** `push_version_field` → `push_additional_field` (the helper only masks + inserts one field into `additional_fields`; nothing version-specific), doc comment made generic.
- **CI fix:** Compilation Check + Clippy were failing on `-D warnings` in the **no-`injector-client` (FFI) build** — `version_metadata` and `derive_pod_identity` were dead code there (their only consumer `create_event` is gated on `injector-client`). Gated those items (and the unit-test module) behind the same feature. Verified locally: `cargo check`/`clippy -p external-services --no-default-features -- -D warnings` both pass; the 5 unit tests still run under default features.

crates/common/external-services/src/service.rs
```diff
@@ -1085,9 +1085,158 @@ fn create_event(
...
+fn version_metadata() -> &'static VersionMetadata {
+    static VERSION_METADATA: std::sync::OnceLock<VersionMetadata> = std::sync::OnceLock::new();
+    VERSION_METADATA.get_or_init(|| {
+        let version = BUILD_VERSION
+            .get()
+            .cloned()
+            .or_else(|| std::env::var("VERSION").ok());
+
+        // k8s sets HOSTNAME to the pod name: `<application_name>-<replicaset_hash>-<pod_suffix>`.
+        let (application_name, deployment_id, pod_name) = match std::env::var("HOSTNAME") {
```
gopikrishna000: Blocking: HOSTNAME cannot reliably define application or deployment semantics. With the supplied Infraswitch pod connector-service-http-d02a2bac0e-c9d9c4945-h6r4g, the deployment template already tells us app=connector-service-http and version=d02a2bac0e; this parser instead emits application_name=connector-service-http-d02a2bac0e and deployment_id=c9d9c4945. The application identity therefore changes per deployment and the rollout identifier becomes a ReplicaSet hash. Kubernetes also permits spec.hostname to differ from metadata.name. Please consume explicit optional values such as CS__RUNTIME_METADATA__APPLICATION_NAME, CS__RUNTIME_METADATA__DEPLOYMENT_ID, and CS__RUNTIME_METADATA__POD_NAME. In Kubernetes these can come from labels and metadata.name through fieldRef; when a deployment platform cannot provide one, omit it rather than emit inferred data.

crates/common/external-services/src/service.rs
```diff
@@ -1085,9 +1085,158 @@ fn create_event(
...
+            Some("connector-service-http-79f5d6b8c4-x2n7q")
+        );
+    }
+
+    #[test]
+    fn deployment_name_with_build_token_absorbed_into_app() {
+        // Live sbx pod: the deployment name itself embeds a build token (`d02a2bac0e`), so it is
+        // absorbed into `application_name` and the replicaset hash remains the deployment_id.
+        let (app, dep, pod) =
+            derive_pod_identity("connector-service-http-d02a2bac0e-c9d9c4945-h6r4g");
```
gopikrishna000: This test currently codifies the incorrect interpretation of the live pod name. For this exact Infraswitch example, the known semantic values are application connector-service-http, deployment version/rollout d02a2bac0e, ReplicaSet hash c9d9c4945, and the complete pod name. If deployment_id is intentionally meant to be the ReplicaSet hash, please name it replicaset_hash and define that contract explicitly; application_name still must not absorb the deployment token. A parser test can prove string splitting, but it cannot prove that the segments have these meanings.

crates/common/external-services/src/service.rs
```diff
@@ -1085,9 +1085,158 @@ fn create_event(
...
+}
+
+#[cfg(feature = "injector-client")]
+fn version_metadata() -> &'static VersionMetadata {
+    static VERSION_METADATA: std::sync::OnceLock<VersionMetadata> = std::sync::OnceLock::new();
+    VERSION_METADATA.get_or_init(|| {
+        let version = BUILD_VERSION
+            .get()
+            .cloned()
+            .or_else(|| std::env::var("VERSION").ok());
```
gopikrishna000: Please remove the VERSION fallback and the global initialization path. The binary already has the authoritative compiled value from ucs_env::git_describe!(). An environment fallback can contradict the running artifact and hides missed initialization. The two OnceLocks also create an ordering bug: if version_metadata() initializes before set_build_version(), it permanently caches the fallback or None, while the setter silently accepts being ineffective. Construct metadata in main with a non-optional compiled version and pass it through application/service state.

crates/common/common_utils/src/events.rs
```diff
@@ -293,6 +293,34 @@ impl Event {
         });
     }
 
+    /// The application / microservice name (e.g. `connector-service-http`). Sourced from the
```
gopikrishna000: These four names are now part of the serialized event contract, but they are implemented as unrelated string keys in additional_fields; the compiler cannot detect a typo, collision, or incomplete propagation. Please model them as typed optional event metadata, for example a RuntimeMetadata struct flattened into Event if the fields must remain top-level. This comment has already drifted as well: it says APPLICATION_NAME even though the implementation reads and parses HOSTNAME.

crates/common/external-services/src/service.rs
```diff
@@ -1085,9 +1085,158 @@ fn create_event(
     event.add_service_name(event_params.service_name);
     event.add_tenant_id(event_params.tenant_id);
 
+    // Version metadata: stamp the four deployment fields on every event so A/B can group builds.
+    // `version` is the compiled build (injected by the binary); `application_name`/`deployment_id`/
+    // `pod_name` are derived from the pod name in `HOSTNAME`. Computed once and cached.
+    let version_metadata = version_metadata();
```
gopikrishna000: The current tests cover only the parsing helper. Please add a behavior-level serialization test proving that: all supplied metadata reaches the expected top-level event fields; absent application/deployment/pod values are omitted; missing optional metadata does not fail startup or event creation; and the compiled version remains present. Passing metadata as an input/state value instead of process-global environment state will also make these cases deterministic and independently testable.

gopikrishna000 (commented):
The CI and helper-naming issues from the earlier head are resolved. The remaining concerns are about the correctness and maintainability of the metadata contract. In particular, deployment-owned semantics should not be reverse-engineered from a hostname. Suggested design: construct a typed RuntimeMetadata in main, set version directly from ucs_env::git_describe!(), read application/deployment/pod identity from optional CS__RUNTIME_METADATA__... configuration, and pass it through application state. Kubernetes deployers can populate those optional values from labels/metadata through the Downward API; non-Kubernetes OSS users may configure or omit them. Missing optional metadata must never prevent UCS startup or event emission.

kanikac199:
Reworked per the review — thanks, this is a better contract. Pushed in `89d4e6f2a`:

1. **HOSTNAME inference removed.** No more parsing pod names into app/deployment semantics. `application_name`/`deployment_id`/`pod_name` now come from explicit optional **`CS__RUNTIME_METADATA__*`** config (k8s Downward API: `metadata.name`, deployment labels). Absent ⇒ omitted, never inferred, never fails startup.
2. **Wrong-interpretation test deleted** along with the parser it tested.
3. **`version` = compiled `git_describe!()`, set in `main`** — removed the `VERSION` env fallback and the `BUILD_VERSION`/`version_metadata` OnceLock ordering hazard.
4. **Typed, not string keys.** New `RuntimeMetadata` struct `#[serde(flatten)]`'d into `Event` (compiler-checked); drifted `HOSTNAME`/`APPLICATION_NAME` comments gone.
5. **Threaded through application state** via `EventProcessingParams`, not process globals.
6. **Behavior-level serialization tests** added: supplied metadata → top-level event fields; absent optionals omitted; `version` always present; minimal metadata neither fails creation nor serialization.

The `CS__RUNTIME_METADATA__*` env wiring (Downward API) is a separate change not in this code PR.

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

AmitsinghTanwar007:
Do we need the test code as i don't think, we have any workflow where we test this code during pr is raised or merged.

JeevaRamu0104 (approved):

PR #2003, opened 2026-07-23
feat(connector): add PaymentMethodService.Refresh (account updater) with Juspay card sync

### LINKED ISSUE #2014: feat(connector): add PaymentMethodService.Refresh (account updater) with Juspay card sync
Add PaymentMethodService.Refresh (account updater) with Juspay card sync

kanikabansal08:
## What

Adds **`PaymentMethodService.Refresh`** — a new RPC for refreshing a stored payment method against the provider — and the first implementation of it: **Juspay Card Account Updater**.

The account updater answers one question about a stored card: *is it still current, and if not, what changed?* Networks reissue cards constantly (expiry bumps, replacement PANs after a breach), and a vaulted card silently rots until a payment fails. This flow lets Hyperswitch ask before charging rather than after declining.

Prism does not decide what to do with the answer. It returns **what the network said** (`outcome`) and **the card that resulted** (`card`); applying it is Hyperswitch's call.

## Why `Refresh` and not `Get`

**`Get` dereferences a reference. `Refresh` submits the instrument.**

Use **`Get`** when the connector owns the record and you hold an identifier for it — `connector_payment_method_id`, `merchant_payment_method_id`, a token. You are asking the connector to resolve a handle into data you do not have; its copy is the truth.

Use **`Refresh`** when the inverse holds, in all three parts:

1. **you already hold the complete payment method details**, not a handle;
2. **you depend on no provider-side record** — there is no identifier in existence that could have been sent instead;
3. **the connector exposes an endpoint that accepts the complete details** and answers a question about them.

Card account updating meets all three. Hyperswitch's locker holds the card. Juspay's account updater stores no payment method, issues no identifier, and requires the full PAN and expiry on **every** call — it is a query service, not a store. So this is not a stylistic preference between two workable options: `PaymentMethodServiceGetRequest` is identifiers, customer, and `payment_method_type`, and there is no reference to put in it and nowhere to put a card. Extending it would also place a full PAN permanently in scope for a message whose purpose is identification, visible to every connector implementing `Get`.

Two consequences follow from the same rule, and explain shapes further down that would otherwise look arbitrary:

- **The response can return a whole card only because the request was one.** Overlaying the network's changes onto the submitted card is possible precisely because the complete card was sent. A reference-based flow could only return a delta and push the merge onto every caller.
- **"Send the complete payment method" is the contract, not advice.** Anything omitted from the request is absent from the response, and UCS cannot tell a deliberately sparse card from a complete one.

## Flow

```mermaid
sequenceDiagram
    participant HS as Hyperswitch
    participant UCS as Prism (UCS)
    participant JP as Juspay

    HS->>UCS: Refresh(payment_method.card_with_no_cvc)
    Note over UCS: seal {accountNumber, expiryMonth, expiryYear}<br/>JWE RSA-OAEP-256 + A128GCM<br/>with Juspay's public key
    UCS->>JP: POST /cardAccountUpdater<br/>{network, cardData, keyId}
    Note over JP: decrypts with their private key,<br/>asks the card network
    JP-->>UCS: {status, responseCode, responseMessage, payload}
    Note over UCS: decrypt payload with our private key<br/>(selected by keyId)
    UCS-->>HS: {outcome, card, status_code}
```

Nothing between the two parties can read the card: `cardData` and `payload` are opaque JWE strings at every hop — proxies, load balancers, request logs, APM traces. TLS alone would leave a PAN in plaintext at every terminating hop.

### 1. New RPC — `PaymentMethodService.Refresh`

`services.proto`:

```protobuf
rpc Refresh(PaymentMethodServiceRefreshRequest) returns (PaymentMethodServiceRefreshResponse);
```

Request carries only the instrument; response carries the verdict plus the resulting card:

```protobuf
message PaymentMethodServiceRefreshRequest  { PaymentMethod payment_method = 1; }

message PaymentMethodServiceRefreshResponse {
  RefreshResult result = 1;
  uint32 status_code = 2;
  optional ErrorInfo error = 3;
  map<string, string> response_headers = 4;
  optional SecretString raw_connector_response = 5;
  optional SecretString raw_connector_request = 6;
}

message RefreshResult     { oneof result { CardRefreshResult card = 1; } }
message CardRefreshResult { CardRefreshOutcome outcome = 1; CardDetailsWithNoCvc card = 2; }
```

`RefreshResult` is a `oneof` so non-card instruments can be added later without reshaping the response. The raw request/response fields carry the wire bodies for debugging, where the card is a JWE in both directions — neither can hold a plaintext PAN — and capture stays opt-in via `return_raw_connector_data`.

The flow is bound in the **FFI crate** as well as over gRPC, so the embedded SDK path can call it too.

### 2. Outcome contract

`CardRefreshOutcome` publishes the **card network's** verdict — six values that transfer to any future account-updater provider — rather than Juspay's operational vocabulary.

| `responseCode` | `CardRefreshOutcome` | `card` returned | Meaning for HS |
|---|---|---|---|
| `ACCOUNT_UPDATED` | `CARD_REFRESH_ACCOUNT_UPDATED` | new PAN (+ expiry if sent) | Card reissued — replace it |
| `EXPIRY_UPDATED` | `CARD_REFRESH_EXPIRY_UPDATED` | same PAN, new expiry | Update the expiry |
| `NO_CHANGE` | `CARD_REFRESH_NO_CHANGE` | submitted card | Stored card is current — keep using it |
| `CARD_CLOSED` | `CARD_REFRESH_CLOSED` | submitted card | Account closed — stop using it, ask for a new method |
| `CARD_NOT_FOUND` | `CARD_REFRESH_NOT_FOUND` | submitted card | No update available — not found, **or the BIN isn't participating**. Don't update; this is *not* a signal to retire the card |
| `CONTACT_ISSUER` | `CARD_REFRESH_CONTACT_ISSUER` | submitted card | Issuer advises contacting the cardholder |
| *unmapped code* | `CARD_REFRESH_OUTCOME_UNSPECIFIED` | submitted card | Still a **successful inquiry**. Payload is **not** decrypted — its meaning under an unknown code is undefined |
| `status: FAILURE` (4 codes) | `CARD_REFRESH_OUTCOME_UNSPECIFIED` + **`error` set** | submitted card | Could not ask. Branch on `error`, not on the outcome |

### 3. Juspay card sync

`POST {base_url}cardAccountUpdater`, `Authorization: Basic base64("<api_key>:")` (password half empty), `Content-Type: application/json`.

```json
{ "network": "VISA", "cardData": "<compact JWE>", "keyId": "<registered key id>" }
```

Validation is limited to what the response layer actually consumes — payload presence when an update code implies a delta, and structural validity of any card/expiry present. Violations are **invalid gateway responses**, never guessed at or silently downgraded to no-change:

- `ACCOUNT_UPDATED` without a replacement PAN → rejected
- `EXPIRY_UPDATED` without a replacement expiry → rejected
- a **terminal** outcome carrying a card number or expiry → rejected (Juspay must not send card data under a non-update outcome)
- returned PAN is Luhn-validated; `updatedExpiryDate` is parsed as `MMYY`

### 4. Crypto (`juspay/crypto.rs`)

JWE-only — **RSA-OAEP-256 + A128GCM**, one key per direction, no signing. This shares neither the algorithms nor the two-layer shape of the existing `jose.rs` sign-then-encrypt helper, so it is a separate module rather than a dedup candidate.

- `alg`/`enc` are asserted on the protected header **before** decryption — the guard against algorithm confusion, since josekit otherwise takes `enc` from the header.
- RSA keys below **2048 bits** are rejected on both directions rather than silently sealing a PAN under a weak key.
- **The plaintext inside `cardData` is double-JSON-encoded.** Juspay's reference client (`encryptVAUData.py`) serialises the payload and then serialises that string again, so the recipient parses twice. Their written spec shows the bare object; their implementation does not, and the implementation is what the sandbox honours. Confirmed by measurement — the `cardData` ciphertext on a working sandbox call is 94 bytes, and A128GCM preserves plaintext length exactly, so single-encoding (80 bytes) could not have produced it.
- **The response direction is single-encoded.** The asymmetry is the contract, not an oversight, and is pinned by `seals_the_plaintext_as_a_json_string`.

### 5. Config

Three new **optional** per-merchant fields on `JuspayConfig`, carried in the `x-connector-config` header. No `config/*.toml` changes — only the merchant credential shape:

| Field | Holder | Used for |
|---|---|---|
| `juspay_encryption_public_key` | Juspay's public key, given to us | sealing `cardData` outbound |
| `response_decryption_private_key` | **our** private key | opening the response `payload` |
| `card_sync_key_id` | id of our public key registered with Juspay | telling Juspay which key to seal the response with |

They are optional because Juspay's existing payment flows don't need them; `JuspayCardSyncAuthType` requires all three and fails with a named `MissingRequiredField` if any is absent. The two keys are **not** a pair and must not be — they serve opposite directions.

## Additional Changes

API contract: new `Refresh` RPC in `services.proto`, plus `PaymentMethodServiceRefresh{Request,Response}`, `RefreshResult`, `CardRefreshResult`, `CardRefreshOutcome` in `payment.proto`. All additive.

Configuration: three optional `JuspayConfig` fields above. Merchants must provision them before a Refresh call can succeed; every other Juspay flow is unaffected.

## How did you test it?

**Unit — 61 tests** (`cargo test -p connector-integration juspay`), covering every documented success code, the documented `FAILURE` codes, unknown `responseCode`s, each consistency violation, malformed and tampered JWE, wrong and undersized keys, and the double-encoded plaintext shape. Error-path cases were mutation-checked — reverting the echo to `result: None` fails them and correctly leaves the success case green.

**Live sandbox — full path.** Local gRPC server → Juspay sandbox → JWE decrypt → mapped outcome, driven through Postman:

https://github.com/user-attachments/assets/0a783b70-3355-4705-917d-e9d867f875ba


crates/types-traits/domain_types/src/types.rs
```diff
@@ -13268,6 +13268,167 @@ impl
...
+            status_code: u32::from(value.status_code),
+            error: None,
+            response_headers: std::collections::HashMap::new(),
+        })
+    }
+}
+
+fn card_with_no_cvc_to_proto(
+    card: payment_method_data::CardWithNoCvc,
+) -> grpc_api_types::payments::CardDetailsWithNoCvc {
```
ThisIsMani: I have a feeling that this would already be there...

ThisIsMani (commented):
Contract and impl looks good to me.

ThisIsMani (dismissed):

JeevaRamu0104 (commented):
add this new flow in the FFI crate

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4314,6 +4314,41 @@ message PaymentMethodServiceGetResponse {
   optional SecretString raw_connector_request = 10;
 }
 
+// Refresh a stored card against the network updater.
+message PaymentMethodServiceRefreshRequest {
+  PaymentMethod payment_method = 1;
```
JeevaRamu0104, line 4354: will we not passing any id or reference_id in the request

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4314,6 +4314,41 @@ message PaymentMethodServiceGetResponse {
...
+}
+
+message PaymentMethodServiceRefreshResponse {
+  RefreshResult result = 1;
+  uint32 status_code = 2;                   // HTTP status code from the connector
+  optional ErrorInfo error = 3;             // Error details if the call failed
+  map<string, string> response_headers = 4; // Response headers from the connector
```
JeevaRamu0104, line 4361: should we return the raw request and raw response also in the response

crates/types-traits/domain_types/src/types.rs
```diff
@@ -13268,6 +13268,148 @@ impl
...
+        Ok(Self {
+            result,
+            status_code: u32::from(value.status_code),
+            error: None,
+            response_headers: std::collections::HashMap::new(),
+        })
+    }
+}
+
+fn card_with_no_cvc_to_proto(
```
JeevaRamu0104: impl is already present

crates/types-traits/grpc-api-types/proto/payment.proto
kanikabansal08: yes, we will not because refresh request carries the details itself, that's the distinction between this and Get.

crates/integrations/connector-integration/src/connectors/juspay.rs
```diff
@@ -1,3 +1,5 @@
+pub mod crypto;
+pub mod test;
```
shuklatushar226: remove test file

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
```diff
@@ -1338,3 +1344,505 @@ impl TryFrom<ResponseRouterData<JuspayVoidResponse, Self>>
...
+impl TryFrom<&ConnectorSpecificConfig> for JuspayCardSyncAuthType {
+    type Error = error_stack::Report<errors::IntegrationError>;
+
+    fn try_from(config: &ConnectorSpecificConfig) -> Result<Self, Self::Error> {
+        let missing = |field_name: &'static str| {
+            error_stack::report!(errors::IntegrationError::MissingRequiredField {
+                field_name,
+                context: Default::default(),
+            })
+        };
```
shuklatushar226, line 1365: add addition context

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
```diff
@@ -1338,3 +1344,505 @@ impl TryFrom<ResponseRouterData<JuspayVoidResponse, Self>>
...
+                    .to_owned()
+                    .ok_or_else(|| missing("response_decryption_private_key"))?,
+                card_sync_key_id: card_sync_key_id
+                    .to_owned()
+                    .ok_or_else(|| missing("card_sync_key_id"))?,
+            }),
+            _ => Err(error_stack::report!(
+                errors::IntegrationError::FailedToObtainAuthType {
+                    context: Default::default()
+                }
```
shuklatushar226, line 1402: +1

crates/integrations/connector-integration/src/connectors/juspay.rs
kanikabansal08: I checked it against the repo, and Juspay is already following the established convention exactly. we add tests. if you recommend I can remove it too.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -17213,3 +17442,264 @@ impl From<connector_types::WebhookResourceReference> for grpc_api_types::payment
...
+            card_network: Some(common_enums::CardNetwork::Visa),
+            card_holder_name: Some(Secret::new("A Cardholder".to_string())),
+            ..Default::default()
+        }
+    }
+
+    fn router_data_with(
+        payment_method_data: payment_method_data::PaymentMethodData<
+            payment_method_data::DefaultPCIHolder,
+        >,
```
AmitsinghTanwar007: i think you can remove test code.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -17073,6 +17222,86 @@ pub fn generate_create_payment_method_response(
...
+                result: submitted_card.map(|card| grpc_api_types::payments::RefreshResult {
+                    result: Some(grpc_api_types::payments::refresh_result::Result::Card(
+                        grpc_api_types::payments::CardRefreshResult {
+                            outcome: grpc_api_types::payments::CardRefreshOutcome::Unspecified
+                                .into(),
+                            card: Some(
+                                grpc_api_types::payments::CardDetailsWithNoCvc::foreign_from(card),
+                            ),
+                        },
+                    )),
```
AmitsinghTanwar007: can we have foreigntryfrom here

crates/types-traits/domain_types/src/types.rs
kanikabansal08: Removed.

crates/types-traits/domain_types/src/types.rs
```diff
@@ -13346,6 +13346,155 @@ impl
...
+                        ProtoOutcome::CardRefreshNoChange
+                    }
+                    connector_types::CardRefreshOutcome::Closed => ProtoOutcome::CardRefreshClosed,
+                    connector_types::CardRefreshOutcome::NotFound => {
+                        ProtoOutcome::CardRefreshNotFound
+                    }
+                    connector_types::CardRefreshOutcome::ContactIssuer => {
+                        ProtoOutcome::CardRefreshContactIssuer
+                    }
+                };
```
AmitsinghTanwar007: this can be a separate tryfrom

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
kanikabansal08: Here I think, it's sufficient information to report to hyperswitch. Added addition context in `FailedToObtainAuthType`

shuklatushar226 (dismissed):

AmitsinghTanwar007 (dismissed):

shuklatushar226 (dismissed):

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
```diff
@@ -1338,3 +1344,518 @@ impl TryFrom<ResponseRouterData<JuspayVoidResponse, Self>>
...
+) -> Result<(Secret<String>, Secret<String>), error_stack::Report<errors::IntegrationError>> {
+    let raw = expiry.peek().trim();
+    if raw.len() != 4 {
+        return Err(error_stack::report!(invalid_gateway_response(
+            "updated_expiry_date"
+        )))
+        .attach_printable("updatedExpiryDate is not in MMYY form");
+    }
+
+    let (month, year) = raw.split_at(2);
```
JeevaRamu0104, line 1618: `raw.len() != 4` is a byte-length check, but `split_at(2)` needs a char boundary — not the same thing. A 4-byte value like `"aé1"` (`a`=1, `é`=2, `1`=1) gets past the guard and then panics:

```
byte index 2 is not a char boundary; it is inside 'é' (bytes 1..3) of `aé1`
```

Confirmed by running it. So an `EXPIRY_UPDATED` / `ACCOUNT_UPDATED` response carrying a non-ASCII expiry takes down the request inside the tonic handler instead of returning `InvalidDataFormat`. Everything else on this path is parsed fallibly — which is the whole point of the block, since Juspay's response isn't trusted — so this is the one line that can't fail gracefully.

Can we gate on ASCII digits before splitting?

```rust
if raw.len() != 4 || !raw.bytes().all(|b| b.is_ascii_digit()) {
    return Err(error_stack::report!(invalid_gateway_response(
        "updated_expiry_date"
    )))
    .attach_printable("updatedExpiryDate is not four ASCII digits in MMYY form");
}
```

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
```diff
@@ -1338,3 +1344,518 @@ impl TryFrom<ResponseRouterData<JuspayVoidResponse, Self>>
...
+
+    let month = normalize_expiry_month(&Secret::new(month.to_string()))
+        .change_context(invalid_gateway_response("updated_expiry_date"))?;
+    let year =
```
JeevaRamu0104, line 1622: The year half isn't validated at all. `expand_expiry_year_to_four_digits` just prefixes the century for any 2-char string with no digit check, so `"12ab"` comes back as `CARD_REFRESH_EXPIRY_UPDATED` with `card_exp_year = "20ab"` and we hand that to Hyperswitch as the refreshed card. The month goes through `CardExpirationMonth::try_from`; the year goes through nothing.

The description commits to "structural validity of any card/expiry present" and "parsed as `MMYY`" — this is the gap. Same root cause as the boundary issue above, and the ASCII-digit guard fixes both.

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
kanikabansal08: Good catch, confirmed and fixed in eded7a33c. Guard is now `raw.len() != 4 || !raw.bytes().all(|b| b.is_ascii_digit())`, so the non-char-boundary input is rejected before `split_at(2)` and returns `InvalidDataFormat` as intended.

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
kanikabansal08: Right, the year half had no digit check at all. The same ASCII-digit guard in eded7a33c covers it, so "12ab" is now rejected instead of coming back as `card_exp_year = "20ab"`.

crates/integrations/connector-integration/src/connectors/juspay.rs
```diff
@@ -196,12 +238,19 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize> Conn
         let code = response
             .error_code
             .clone()
+            .or_else(|| {
+                response
+                    .error_info
+                    .as_ref()
+                    .and_then(|info| info.code.clone())
+            })
             .or_else(|| response.status.clone())
```
JeevaRamu0104, line 247: Worth calling out that this isn't scoped to Refresh — `get_error_response_v2` is shared by Authorize, PSync, Capture, Refund, RSync and Void, so slotting `error_info.code` ahead of `response.status` changes what all six report. An error that used to surface as `code = "FAILURE"` now surfaces the specific `error_info.code`. Same shape on line 253, where `user_message` jumps ahead of `error_info.user_message`, and the extra fallback on 271.

It's probably the more useful mapping, but it's a live behaviour change on six existing flows that the description doesn't mention and no test pins — and `code` is what downstream unified error mapping keys on. Was this deliberate as part of this PR, or would it sit better as its own change where it can be reviewed against the Juspay error-code table on its own terms?

crates/types-traits/domain_types/src/types.rs
```diff
@@ -13346,6 +13346,198 @@ impl
...
+    ) -> Result<Self, error_stack::Report<Self::Error>> {
+        // The request carries no identifier of its own, so the caller's request
+        // id is the only correlation handle. Matches the webhook-verify precedent.
+        let request_id = metadata
+            .get_raw(common_utils::consts::X_REQUEST_ID)
+            .unwrap_or_default();
+
+        Ok(Self {
+            connectors,
+            connector_request_reference_id: format!("refresh_{request_id}"),
```
JeevaRamu0104: When `x-request-id` is absent, `unwrap_or_default()` makes this the literal `"refresh_"` for every such call, so the reference id collapses to a constant. Given the request carries no id of its own (which is the whole point of Refresh vs Get), this is the only handle tying a log line back to a specific call — a constant makes it useless exactly when you'd want it.

Could we fall back to a generated uuid rather than an empty string? Or treat a missing `x-request-id` as an error on this flow, since there's nothing else to correlate on.

crates/integrations/connector-integration/src/connectors/juspay/transformers.rs
```diff
@@ -1338,3 +1344,518 @@ impl TryFrom<ResponseRouterData<JuspayVoidResponse, Self>>
...
+    let unchanged = |outcome| RefreshPaymentMethodResponseData {
+        result: Some(RefreshPaymentMethodResult::Card(CardRefreshResult {
+            outcome,
+            card: submitted_card.clone(),
+        })),
+        status_code,
+    };
+
+    // Undefined what a payload means under an unmapped code, so don't decrypt it.
+    if outcome == CardRefreshOutcome::Unrecognized {
```
JeevaRamu0104, line 1671: Returning early here drops the provider code on the floor — the caller gets `CARD_REFRESH_OUTCOME_UNSPECIFIED`, no `error` set, and nothing telling them which code actually arrived. `as_provider_code()` already renders it, but it's only used on the failure path (line 1761).

So the day Juspay adds a response code, it's invisible: no error to branch on, and no way to find out what came back short of turning on `return_raw_connector_data`. Not decrypting the payload under an unknown code makes sense, but could we at least surface the code itself — an `attach_printable` so it lands in the logs would be enough to notice it.

crates/integrations/connector-integration/src/connectors/juspay.rs
kanikabansal08: Deliberate, and it can't cleanly be split out: Refresh registers the same `get_error_response_v2` default (juspay.rs:601), so this handler is Refresh's error path too. `user_message` wasn't even on `JuspayErrorResponse` before this PR — it was added because card sync returns it. Moving the change out would mean either shipping Refresh with a mapping we know is wrong for card-sync errors, or duplicating the handler.

On `code` specifically: `error_info.code` only fires when `error_code` is absent, which is exactly where the old chain emitted `response.status` — a lifecycle status (`FAILURE`, `NEW`), not an error code. Since unified mapping keys on `code`, `"FAILURE"` gives it nothing to key on while `error_info.code` is Juspay's actual error identifier. Where `error_code` is present, nothing changes.

The part I'd concede is the ordering on 253 and 271: I put top-level `user_message` ahead of `error_info.user_message`, and that's two fields of the same kind with no real argument for either order. Happy to move `user_message` last in both chains — that makes those two purely additive, firing only where the old code fell through to `status` or the HTTP fallback, and narrows the behaviour change on the six existing flows to the `code` case above. Say the word and I'll push it.

Fair point that nothing pins this in tests, and I can't tell from the code how often the payment flows return `error_info.code` without `error_code` — that's an API-docs question.

kanikabansal08: To put a bound on how much actually changes for the six flows — these are fallback chains, so a newly inserted candidate only fires when the field ahead of it is absent:

- `code`: changes only when Juspay sends no `error_code` but does send `error_info.code`. That is exactly the case where the old chain fell through to `response.status` and reported `"FAILURE"`. Whenever `error_code` is present — the normal case — the value is identical to main.
- `message` / `reason`: change only when `error_message` (resp. `error_info`'s messages) is absent and top-level `user_message` is present. `user_message` wasn't deserialized at all before this PR, so that content was simply being dropped.

And nothing in this repo branches on any of the three. `code`, `message` and `reason` are copied straight into `UnifiedErrorDetails` at domain_types/src/errors.rs:618, and there's no Juspay-keyed error-code table anywhere in the tree — so no behaviour here keys off the change, only the strings that reach the caller.

Not claiming zero impact: if something downstream in Hyperswitch keys on `"FAILURE"` for Juspay, that's where it would be felt, and I can't verify that from this repo. But the blast radius is error bodies that are missing their primary field, not errors generally.

JeevaRamu0104 (dismissed):

crates/types-traits/domain_types/src/types.rs
```diff
@@ -13346,6 +13346,199 @@ impl
...
+        // The request carries no identifier of its own, so the caller's request
+        // id is the only correlation handle. Matches the webhook-verify precedent.
+        let request_id = metadata
+            .get_raw(common_utils::consts::X_REQUEST_ID)
+            .filter(|request_id| !request_id.trim().is_empty())
+            .unwrap_or_else(common_utils::fp_utils::generate_uuid_v7);
+
+        Ok(Self {
+            connectors,
+            connector_request_reference_id: format!("refresh_{request_id}"),
```
AmitsinghTanwar007: Hey, I missed this.
I have a question: why are we adding the `refresh_` prefix here?
Also, I thought `connector_request_reference_id` is the merchant's reference (correct me if I'm wrong).
Since UCS is supposed to be transparent, changing or creating a new reference in between doesn't seem right to me.

kanikabansal08: For context on where the prefix came from webhook verification does exactly the same thing today (grpc-server/src/server/events.rs:971,  format!("webhook_verify_{request_id}")) for VerifyWebhookSourceFlowData, which has the identical shape. So it followed an existing pattern rather than being invented here. But I agree it's on the wrong side of your line, and nothing on the Refresh path reads the field anyway, so the value wasn't buying anything.

AmitsinghTanwar007: instead can't we make the field connector req ref id  as optional so that instead of string::new we can directly pass it as none to avoid empty string

AmitsinghTanwar007 (approved):

crates/types-traits/domain_types/src/types.rs
kanikabansal08: the ConnectorRequestReference trait signature returns &str, so the empty string comes back at the trait boundary regardless.we will be making the struct honest, not eliminating "" from the system. Fully eliminating it means changing the trait to return Option<&str> and updating all ~10 impls — including FrmFlowData, which returns a hardcoded "" today for exactly this reason

JeevaRamu0104 (approved):

shuklatushar226 (approved):

PR #2004, opened 2026-07-23
feat(kount): enrich pre_risk_check (Evaluate Order) response — full masked risk body in logs

kanikac199:
## What & why

Enrich the **pre_risk_check** (Evaluate Order, `POST /commerce/v2/orders?riskInquiry=true`) response so the full documented Kount risk body is captured — masked — in the connector event log and `raw_connector_response`. Previously `KountOrderResponse` kept only `decision`/`omniscore`/`reason` and discarded the rest (persona, device, policyManagement, transactions, fulfillment…).

**The notify flows stay on PATCH.** An earlier revision of this branch moved the FRM notify flows (`FrmPaymentOutcome`/`FrmRefundProcessed`) to POST Evaluate Order; that was a mistake and has been reverted — notify is back to `PATCH /commerce/v2/orders/{orderId}` (Update Order), exactly as on `main`. Net change vs `main` is confined to `transformers.rs`.

## Changes (transformers.rs only)
- `KountOrderResponse` / `KountOrder` / `KountRiskInquiry` model the full documented Evaluate Order schema (+ ~20 nested types), using `#[serde(rename_all = "camelCase")]`.
- PII/card fields wrapped in `Secret` so they mask in logs: `deviceSessionId`, device id, IPs, card `bin`/`last4`/`paymentToken`/`cardBrand`, and granular device location (coordinates, postalCode, areaCode, city, region, regionCode). Country-level fields stay plaintext.
- Number-or-string tolerant parsing (`de_stringy`) for count/score fields; `warnings` typed as `Vec<serde_json::Value>` to tolerate structured shapes.
- `raw_connector_response` built via `masked_serialize` so the stored/egressed blob never carries cleartext card/PII.
- `PreRiskCheck` response reads (`decision`/`omniscore`/`reason`/`order_id`) are unchanged — backward compatible.
- **No proto or shared-type changes.**

## Not changed
- Notify (`FrmPaymentOutcome`/`FrmRefundProcessed`) → `PATCH /orders/{orderId}`, original body/response (`kount.rs` identical to `main`).
- prism gRPC contract unchanged. Full risk body is surfaced in logs + `raw_connector_response` only (no proto change to return it in `PreRiskCheckResponse`).

## Verification
- `cargo +nightly fmt --all --check` → clean; `RUSTFLAGS="-D warnings" cargo clippy --all-features --all-targets` → clean.
- Live sandbox: pre_risk_check returns the full `riskInquiry`; parsed into the enriched model with `deviceSessionId`/card fields masked (`*** … ***`) and `decision`/`omniscore` visible.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/kount/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/kount/transformers.rs
@@ -219,7 +219,7 @@ pub struct KountOrderResponse {
     pub version: Option<String>,
     pub order: Option<KountOrder>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub warnings: Option<Vec<String>>,
+    pub warnings: Option<Vec<serde_json::Value>>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -1476,7 +1476,13 @@ impl TryFrom<ResponseRouterData<KountOrderResponse, Self>>
     fn try_from(item: ResponseRouterData<KountOrderResponse, Self>) -> Result<Self, Self::Error> {
         // Always surface the raw Kount response (independent of the global
         // `return_raw_connector_data` flag), mirroring twoc_twop_paco / ppro.
-        let raw_connector_response = serde_json::to_string(&item.response).ok().map(Secret::new);
+        // Mask through the PII serializer so the stored/egressed blob never carries
+        // cleartext card/PII — a plain `serde_json::to_string` emits `Secret` fields
+        // exposed. `None` on serialization failure (degrades the audit trail rather
+        // than failing the flow).
+        let raw_connector_response = hyperswitch_masking::masked_serialize(&item.response)
+            .ok()
+            .map(|value| Secret::new(value.to_string()));
         let order = item.response.order.as_ref();
         let risk = order.and_then(|order| order.risk_inquiry.as_ref());
         Ok(Self {
@@ -1538,21 +1544,21 @@ fn build_notify_evaluate_order(
 
 /// Require an order reference from a notify request; it is the Kount
 /// `merchantOrderId` and the basis for the correlating `deviceSessionId`.
-fn notify_order_id(primary: Option<&String>, fallback: Option<&String>) -> Result<String, Error> {
+fn notify_order_id(
+    primary: Option<&String>,
+    fallback: Option<&String>,
+    primary_field: &'static str,
+) -> Result<String, Error> {
     primary.or(fallback).cloned().ok_or_else(|| {
         error_stack::report!(errors::IntegrationError::MissingRequiredField {
-            field_name: "merchant_transaction_id",
+            field_name: primary_field,
             context: errors::IntegrationErrorContext {
-                additional_context: Some(
-                    "Kount notify (Evaluate Order POST) needs a merchant/connector transaction \
-                     id; it is the order reference and the basis for the deviceSessionId that \
-                     correlates to the pre-auth order"
-                        .to_owned(),
-                ),
-                suggested_action: Some(
-                    "Set merchant_transaction_id (or connector_transaction_id) on the notify request"
-                        .to_owned(),
-                ),
+                additional_context: Some(format!(
+                    "Kount notify (Evaluate Order POST) needs {primary_field}; it is the order \
+                     reference and the basis for the deviceSessionId that correlates to the \
+                     pre-auth order"
+                )),
+                suggested_action: Some(format!("Set {primary_field} on the notify request")),
                 doc_url: Some(KOUNT_DOC_URL.to_owned()),
             },
         })
@@ -1599,6 +1605,7 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         let order_id = notify_order_id(
             req.merchant_transaction_id.as_ref(),
             req.connector_transaction_id.as_ref(),
+            "merchant_transaction_id",
         )?;
         let order_total =
             super::KountAmountConvertor::convert(req.amount.amount, req.amount.currency)?;
@@ -1618,7 +1625,7 @@ pub struct KountUpdateOrderResponse {
     pub version: Option<String>,
     pub order: Option<KountOrder>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub warnings: Option<Vec<String>>,
+    pub warnings: Option<Vec<serde_json::Value>>,
 }
 
 impl TryFrom<ResponseRouterData<KountUpdateOrderResponse, Self>>
@@ -1635,7 +1642,13 @@ impl TryFrom<ResponseRouterData<KountUpdateOrderResponse, Self>>
         item: ResponseRouterData<KountUpdateOrderResponse, Self>,
     ) -> Result<Self, Self::Error> {
         // Surface the raw Kount response for audit parity with PreRiskCheck.
-        let raw_connector_response = serde_json::to_string(&item.response).ok().map(Secret::new);
+        // Mask through the PII serializer so the stored/egressed blob never carries
+        // cleartext card/PII — a plain `serde_json::to_string` emits `Secret` fields
+        // exposed. `None` on serialization failure (degrades the audit trail rather
+        // than failing the flow).
+        let raw_connector_response = hyperswitch_masking::masked_serialize(&item.response)
+            .ok()
+            .map(|value| Secret::new(value.to_string()));
         Ok(Self {
             response: Ok(FrmPaymentOutcomeResponse {
                 status_code: item.http_code,
@@ -1683,13 +1696,17 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         >,
     ) -> Result<Self, Self::Error> {
         let req = &item.router_data.request;
-        // Refund requests carry no merchant_transaction_id; use the connector txn
-        // id (payment reference), falling back to the refund ids.
+        // Refund requests carry no merchant_transaction_id, so the order id (and
+        // thus deviceSessionId) is derived from the connector txn id, falling back
+        // to the refund ids. Correlation to the pre-auth order is therefore
+        // best-effort: unless connector_transaction_id == the pre-auth
+        // merchant_transaction_id, this logs a fresh, uncorrelated evaluation.
         let order_id = notify_order_id(
             req.connector_transaction_id.as_ref(),
             req.merchant_refund_id
                 .as_ref()
                 .or(req.connector_refund_id.as_ref()),
+            "connector_transaction_id",
         )?;
         let order_total =
             super::KountAmountConvertor::convert(req.amount.amount, req.amount.currency)?;
@@ -1709,7 +1726,7 @@ pub struct KountRefundUpdateResponse {
     pub version: Option<String>,
     pub order: Option<KountOrder>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub warnings: Option<Vec<String>>,
+    pub warnings: Option<Vec<serde_json::Value>>,
 }
 
 impl TryFrom<ResponseRouterData<KountRefundUpdateResponse, Self>>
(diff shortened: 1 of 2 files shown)


crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -224,27 +229,415 @@ pub struct KountOrder {
...
+    pub channel: Option<String>,
+    /// Device/session identifier — PII, masked.
+    #[serde(rename = "deviceSessionId", skip_serializing_if = "Option::is_none")]
+    pub device_session_id: Option<Secret<String>>,
+    #[serde(rename = "creationDateTime", skip_serializing_if = "Option::is_none")]
+    pub creation_date_time: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub transactions: Option<Vec<KountRespTransaction>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fulfillment: Option<Vec<KountRespFulfillment>>,
```
iemyashasvi, line 295: nit : `serde (rename_all="camelCase")` can be used at top of struct

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -224,27 +229,415 @@ pub struct KountOrder {
...
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct KountRiskInquiry {
```
iemyashasvi, line 300: nit: same rename_all=camelcase

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -224,27 +229,415 @@ pub struct KountOrder {
...
+        rename = "maxVelocity",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub max_velocity: Option<String>,
+    #[serde(rename = "riskiestRegion", skip_serializing_if = "Option::is_none")]
+    pub riskiest_region: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct KountRespDevice {
```
iemyashasvi, line 428: use rename_all=camelCase

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -224,27 +229,415 @@ pub struct KountOrder {
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct KountDeviceLocation {
+    #[serde(rename = "areaCode", skip_serializing_if = "Option::is_none")]
+    pub area_code: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub city: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub country: Option<String>,
+    #[serde(rename = "countryCode", skip_serializing_if = "Option::is_none")]
+    pub country_code: Option<String>,
```
iemyashasvi, line 500: should these be secret?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -224,27 +229,415 @@ pub struct KountOrder {
...
+    )]
+    pub longitude: Option<Secret<String>>,
+    #[serde(rename = "postalCode", skip_serializing_if = "Option::is_none")]
+    pub postal_code: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub region: Option<String>,
+    #[serde(rename = "regionCode", skip_serializing_if = "Option::is_none")]
+    pub region_code: Option<String>,
+    #[serde(rename = "localeCountryCode", skip_serializing_if = "Option::is_none")]
+    pub locale_country_code: Option<String>,
```
iemyashasvi, line 520: +1

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1343,26 +1696,39 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+        // thus deviceSessionId) is derived from the connector txn id, falling back
+        // to the refund ids. Correlation to the pre-auth order is therefore
+        // best-effort: unless connector_transaction_id == the pre-auth
+        // merchant_transaction_id, this logs a fresh, uncorrelated evaluation.
+        let order_id = notify_order_id(
+            req.connector_transaction_id.as_ref(),
+            req.merchant_refund_id
+                .as_ref()
+                .or(req.connector_refund_id.as_ref()),
+            "connector_transaction_id",
```
iemyashasvi: why is this hardcoded?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1107,97 +1506,72 @@ impl TryFrom<ResponseRouterData<KountOrderResponse, Self>>
...
+                    "Kount notify (Evaluate Order POST) needs {primary_field}; it is the order \
+                     reference and the basis for the deviceSessionId that correlates to the \
+                     pre-auth order"
+                )),
+                suggested_action: Some(format!("Set {primary_field} on the notify request")),
+                doc_url: Some(KOUNT_DOC_URL.to_owned()),
+            },
+        })
+    })
 }
```
iemyashasvi, line 1165: is it fine we are removing the status mapping , was it dead code in the first place

iemyashasvi (changes requested):

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Done in a9af718bf — applied `#[serde(rename_all = "camelCase")]` at the struct level and dropped the redundant per-field `rename`s.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Done (a9af718bf) — same struct, now `rename_all = "camelCase"`.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Done (a9af718bf) — `KountPersona` now uses `rename_all = "camelCase"` (kept the per-field `deserialize_with = de_stringy`).

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Good catch — masked the granular device-location fields (`areaCode`, `city`, `region`, `regionCode`) as `Secret` in a9af718bf, alongside the already-masked coordinates/postalCode. Country-level fields (`country`/`countryCode`/`localeCountryCode`) stay plaintext since they aren't user-identifying.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Done — `region`/`regionCode` are now `Secret` too (a9af718bf).

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: It's the `field_name` surfaced in the `MissingRequiredField` error, not request data. It's a per-flow constant because each notify flow has a fixed order-ref source: payment notify reads `merchant_transaction_id`, refund notify reads `connector_transaction_id` (refunds carry no merchant_transaction_id). This was the earlier-review fix so the error names the field the flow actually reads.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: It was effectively dead. The old PATCH sent top-level `paymentStatus`/`frmDisposition`, but those aren't in Kount's `OrderPatchBody` schema — verified against sandbox that Kount silently ignores unknown fields (a bogus field name also returns 200), so the mapping never affected anything. The new POST is an Evaluate Order, which has no payment-status field, so there's nothing to map. Net: not a regression — an inert call replaced by a valid, risk-data-bearing one. (Genuinely forwarding the outcome to Kount would need `transactions[].transactionStatus`/`authorizationStatus` on an Update Order, i.e. a proto change — out of scope here.)

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -247,6 +282,376 @@ pub struct KountUpdateOrderResponse {
...
+        deserialize_with = "de_stringy",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub unique_devices: Option<String>,
+    #[serde(
+        default,
+        deserialize_with = "de_stringy",
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub unique_emails: Option<String>,
```
iemyashasvi, line 401: should we use Email struct here ?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -247,6 +282,376 @@ pub struct KountUpdateOrderResponse {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub fulfillment_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub merchant_fulfillment_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub shipping: Option<KountRespShipping>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub status: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub access_url: Option<String>,
```
iemyashasvi: should this be secret

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Good catch to check — but `uniqueEmails` here is a *count* (a persona velocity metric = number of distinct emails Kount has seen linked to this instrument, e.g. `"3"`), not an email address. So the `Email` type doesn't apply. It's typed `Option<String>` via `de_stringy` because Kount returns these counts as a *string* in the docs but a *number* on the live API — `de_stringy` normalises both. Same for the other `unique*`/`totalBank*`/`maxVelocity` counts. Happy to switch them to a number-tolerant `i64` if you'd prefer stronger typing, though they can also come back empty (`""`), so the lenient String is safest for this log/audit data.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
kanikac199: Done in 85f0d92f0 — `access_url` is now `Option<Secret<String>>` (a digital-delivery URL can carry a download token), so it masks in the log/masked-serialize paths like the other PII fields.

iemyashasvi (dismissed):

iemyashasvi (dismissed):

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -247,6 +341,377 @@ pub struct KountUpdateOrderResponse {
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct KountSegment {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub priority: Option<i64>,
+}
```
Aishwariyaa-Anand, line 543: should they be secret?

Aishwariyaa-Anand (approved):

iemyashasvi (approved):

PR #2016, opened 2026-07-28
chore: OIDC release

JeevaRamu0104:
### Additional Changes


XyneSpaces:
[should-fix] The workflow switches the JavaScript publish step to npm trusted publishing/OIDC, but the workflow/job does not grant `id-token: write`. Add job-level permissions such as `id-token: write` and `contents: read` to `publish-javascript`; otherwise repositories with restricted default `GITHUB_TOKEN` permissions will fail to mint the OIDC token during release.

PR #2020, opened 2026-07-29
fix: preserve connector HTTP client error semantics

gopikrishna000:
## Description

- Map common connector HTTP error statuses to canonical gRPC codes, including `409 -> ABORTED`.
- Treat otherwise-unmapped connector 4xx responses as `INVALID_ARGUMENT` instead of `UNKNOWN`.
- Keep unclassified connector 5xx responses as `UNKNOWN` so they are not mislabeled as internal UCS failures.
- Preserve the exact upstream status in `ConnectorError.http_status_code`.
- Add one table-driven unit test for the mapping and fallback boundaries.

## Motivation and Context

PPRO can return HTTP 409 when an idempotency key is reused with a different request body. Previously, 409 fell through to gRPC `UNKNOWN`; UCS HTTP mode then translated that to HTTP 500 even though the structured connector details correctly contained `http_status_code: 409`.

This change maps generic conflicts to gRPC `ABORTED`, which the HTTP bridge returns as HTTP 409, and prevents other connector 4xx responses from being reported as UCS 500 errors.

### Additional Changes

## How did you test it?

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p ucs_env`
- `cargo clippy -p ucs_env --all-targets -- -D warnings`

### DIFF
--- a/crates/common/ucs_env/src/error.rs
+++ b/crates/common/ucs_env/src/error.rs
@@ -381,9 +381,7 @@ impl IntoGrpcStatus for Report<ConnectorFlowError> {
 
 #[cfg(test)]
 mod tests {
-    use super::{connector_http_status_to_grpc_code, ToGrpcStatus};
-    use domain_types::{errors::ConnectorError, router_data::ErrorResponse};
-    use prost::Message;
+    use super::connector_http_status_to_grpc_code;
     use tonic::Code;
 
     #[test]
@@ -412,36 +410,6 @@ mod tests {
                 expected_grpc_code,
                 "unexpected gRPC mapping for connector HTTP status {http_status}"
             );
-        }
-    }
-
-    #[test]
-    fn leaves_non_error_and_non_http_statuses_unknown() {
-        for status in [0, 200, 399, 600, u16::MAX] {
-            assert_eq!(
-                connector_http_status_to_grpc_code(status),
-                Code::Unknown,
-                "unexpected gRPC mapping for non-error status {status}"
-            );
-        }
-    }
-
-    #[test]
-    fn preserves_connector_409_in_grpc_error_details() {
-        let connector_error = ConnectorError::ConnectorErrorResponse(ErrorResponse {
-            code: "409".to_string(),
-            message: "idempotency key reused with a different request body".to_string(),
-            status_code: 409,
-            ..ErrorResponse::default()
-        });
-
-        let status = connector_error.to_grpc_status_unlogged();
-        let details = grpc_api_types::payments::ConnectorError::decode(status.details());
-
-        assert_eq!(status.code(), Code::Aborted);
-        assert!(details.is_ok(), "connector error details should decode");
-        if let Ok(details) = details {
-            assert_eq!(details.http_status_code, Some(409));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/common/ucs_env/src/error.rs
```diff
@@ -243,6 +236,28 @@ impl ToGrpcStatus for ConnectorError {
...
+        401 => tonic::Code::Unauthenticated,
+        403 => tonic::Code::PermissionDenied,
+        404 => tonic::Code::NotFound,
+        408 | 504 => tonic::Code::DeadlineExceeded,
+        409 => tonic::Code::Aborted,
+        412 => tonic::Code::FailedPrecondition,
+        429 => tonic::Code::ResourceExhausted,
+        501 => tonic::Code::Unimplemented,
+        502 | 503 => tonic::Code::Unavailable,
+        400..=499 => tonic::Code::InvalidArgument,
```
Aishwariyaa-Anand: Any connector 4xx not explicitly matched in connector_http_status_to_grpc_code (e.g. 400, 405, 410, 415, 422, 423, etc.) gets classified as InvalidArgument and surfaces to the UCS caller as HTTP 400 — regardless of what the connector's actual status was.

gopikrishna000: this is http status to grpc code conversion, there are no wider enums to support all possible http status code
But for response for ucs-http endpoint, if we need to extract connector's real status code, then we need to rely on existing ConnectorError.CONNECTOR_ERROR_RESPONSE.status_code

crates/common/ucs_env/src/error.rs
```diff
@@ -243,6 +236,28 @@ impl ToGrpcStatus for ConnectorError {
...
+///
+/// The exact connector status is retained in `ConnectorError.http_status_code`.
+/// Unknown connector 4xx responses remain caller-visible errors rather than being
+/// converted to `Unknown`, which the HTTP bridge exposes as a UCS 500 response.
+fn connector_http_status_to_grpc_code(status_code: u16) -> tonic::Code {
+    match status_code {
+        401 => tonic::Code::Unauthenticated,
+        403 => tonic::Code::PermissionDenied,
+        404 => tonic::Code::NotFound,
+        408 | 504 => tonic::Code::DeadlineExceeded,
```
Aishwariyaa-Anand: DeadlineExceeded will be 504, should we include 408 also here?

Aishwariyaa-Anand (dismissed):

crates/grpc-server/grpc-server/src/http/error.rs
```diff
@@ -187,17 +187,23 @@ fn extract_error_details_from_status(status: &tonic::Status) -> Option<ErrorDeta
     None
 }
 
-/// Extract the connector's exact HTTP error status when the gRPC details identify a
+/// Extract the connector's exact HTTP error status when the decoded details identify a
 /// connector-originated 4xx/5xx response.
-fn connector_http_status_from_grpc_status(status: &tonic::Status) -> Option<StatusCode> {
-    let connector_error =
-        grpc_api_types::payments::ConnectorError::decode(status.details()).ok()?;
+fn connector_http_status_from_error_details(details: Option<&ErrorDetails>) -> Option<StatusCode> {
```
Aishwariyaa-Anand, line 192: do not use early returns

Aishwariyaa-Anand (approved):

AmitsinghTanwar007 (approved):

PR #2028, opened 2026-07-31
fix(airwallex): carry connector_order_id and access token in dedicated proto fields

shuklatushar226:
## Why

Raised in review on [hyperswitch#13278](https://github.com/juspay/hyperswitch/pull/13278#discussion_r3689706509) — connector-specific logic had leaked into the connector-agnostic UCS transformer on the Hyperswitch side.

Two proto gaps were being worked around by overloading `connector_feature_data`:

| Message | Missing | Workaround |
|---|---|---|
| `RecurringPaymentServiceChargeRequest` | `connector_order_id` | caller wrote `{"connector_order_id": "..."}`, connector read the key back |
| `CustomerServiceCreateRequest` | `state` | caller wrote `{"access_token": "..."}`, connector read the key back |

Airwallex needs both: its MIT leg confirms a pre-created PaymentIntent (so it needs the id from `CreateOrder`), and its customer API authenticates with a Bearer token separate from the merchant credentials.

Everywhere else `connector_feature_data` is an opaque, connector-owned blob round-tripped between legs. Overloading it with router-synthesized magic keys gave one field two contracts, with nothing but a string literal holding the two sides together — rename it on either side and it fails at runtime, not at build time.

## What

- `payment.proto`: add `RecurringPaymentServiceChargeRequest.connector_order_id = 39` and `CustomerServiceCreateRequest.state = 10`
- `types.rs`: map both into `PaymentFlowData` (the `state` → `access_token` extraction is the same one the recurring-charge impl already uses)
- `composite-service`: thread `state` into the customer-create request, mirroring the neighbouring `CustomerServiceGetRequest` impl
- `airwallex.rs`: drop both fallback reads; `get_access_token()` now goes through the normal path like the other nine call sites in the file

Net **+32/−31** — this removes more than it adds.

## Notes

- Both fields are new `optional` tags (`39` and `10` are free at `2026.07.07.2`, `2026.07.28.0`, and `main`), so the change is wire-compatible in both directions — old servers ignore them, new servers see `None`.
- The composite authorize path previously had **no** way to pass a token to customer-create; it hardcoded nothing and the `connector_feature_data` workaround never covered it. Threading `state` closes that gap.
- Verified: `cargo clippy --workspace --all-targets` clean.
- The Hyperswitch side of this is held until a tag is cut from this — it pins `2026.07.07.2`, which lacks both fields.

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/airwallex.rs
+++ b/crates/integrations/connector-integration/src/connectors/airwallex.rs
@@ -27,7 +27,7 @@ use domain_types::{
     types::Connectors,
 };
 use error_stack::ResultExt;
-use hyperswitch_masking::{ExposeInterface, Maskable, PeekInterface};
+use hyperswitch_masking::{ExposeInterface, Maskable};
 use interfaces::{
     api::ConnectorCommon, connector_integration_v2::ConnectorIntegrationV2, connector_types,
     decode::BodyDecoding, verification::SourceVerification,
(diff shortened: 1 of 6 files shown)


XyneSpaces:
[blocking] `payment.proto` adds new fields to `RecurringPaymentServiceChargeRequest` and `CustomerServiceCreateRequest`, but this PR does not include the corresponding generated Rust under `crates/types-traits/grpc-api-types/src/`. The domain conversions now read `connector_order_id` / `state`, so consumers built from the checked-in generated API will not see the new fields.

Regenerate and commit the gRPC Rust artifacts for these proto changes.

XyneSpaces:
[blocking] Please regenerate and commit the Rust gRPC artifacts for the new `payment.proto` fields. This PR adds `CustomerServiceCreateRequest.state` and `RecurringPaymentServiceChargeRequest.connector_order_id`, and the Rust code now reads those fields; without the matching `crates/types-traits/grpc-api-types/src/` updates, the workspace will fail to compile.

Issue #1940, opened 2026-07-14
Publish an arm64 (aarch64) Linux binary for libconnector_service_ffi


alondayan2:
## Summary

`hyperswitch-prism` cannot run on **arm64 Linux (aarch64)** because the package ships no aarch64-linux native binary and the loader does not branch on architecture. On any Linux host it tries to load the bundled **x86-64** `.so`, which fails on arm64. This blocks adoption on the increasingly common arm64 server platforms (AWS Graviton, GCP Tau/Axion, Ampere, arm64 CI runners).

## Environment

- Package: `hyperswitch-prism@0.2.0` (latest)
- Target runtime: Node.js on **Linux arm64 / aarch64** (e.g. AWS Graviton)
- Works: macOS arm64 (loads the `.dylib`), Linux x86-64 (loads the `.so`)
- Fails: Linux arm64

## What the package ships

Only two native binaries are included:

```
dist/src/payments/generated/libconnector_service_ffi.so     # ELF 64-bit, x86-64
dist/src/payments/generated/libconnector_service_ffi.dylib  # Mach-O 64-bit, arm64 (macOS)
```

There is no `aarch64` Linux binary.

## Root cause

The FFI loader selects the binary purely by `process.platform`, with **no `process.arch` branch**:

```js
// dist/src/payments/uniffi_client.js
const ext = process.platform === "darwin" ? "dylib" : "so";
libPath = path.join(_dirname, "generated", `libconnector_service_ffi.${ext}`);
```

The same platform-only selection is used in `dist/src/payments/_generated_grpc_client.js`, so `GrpcClient` is affected too — there is no pure-JS fallback. On arm64 Linux this resolves to the x86-64 `.so` and the process fails to load the native library.

## Steps to reproduce

On an arm64 Linux host (e.g. an AWS Graviton instance or `--platform=linux/arm64` container):

```bash
mkdir prism-arm64-repro && cd prism-arm64-repro
npm init -y
npm install hyperswitch-prism@0.2.0
node -e "const { PaymentClient } = require('hyperswitch-prism'); new PaymentClient({ connectorConfig: {} });"
# => fails loading libconnector_service_ffi.so (wrong ELF architecture: x86-64 loaded on aarch64)
```

The same install and require succeed on macOS arm64 and on Linux x86-64.

## Request

1. Publish an **`aarch64-unknown-linux-gnu`** build of `libconnector_service_ffi`.
[...]

PR #2029, opened 2026-08-01
fix(sdk): select gRPC native library by platform and arch

Utkal059:
Follow-up to #1940, which reported two arch-blind native loaders in the JavaScript SDK. #1996 fixed the UniFFI one — it resolves `generated/<platform>-<arch>/` now, and the release workflow builds `aarch64-unknown-linux-gnu`. The gRPC loader was left as it was:

```ts
// sdk/javascript/src/payments/_generated_grpc_client.ts
const ext = process.platform === "darwin" ? "dylib" : "so";
libPath = path.join(_dirname, "generated", `libhyperswitch_grpc_ffi.${ext}`);
```

No `process.arch`, and a flat path with nowhere to put a second architecture. This brings it in line.

### Changes

- `sdk/javascript/src/payments/native_lib.ts` (new): one `resolveNativeLib()` shared by both loaders, so the platform/arch mapping can't drift between them.
- `_generated_grpc_client.ts` resolves per-arch. The edit is in `templates/javascript/grpc_client.ts.j2`, regenerated with `generate.py --lang grpc`; the Python, Kotlin and Rust outputs regenerate byte-identical.
- `make generate-grpc-bindings` stages the native under `generated/$(NODE_TARGET)/`, the way `generate-bindings` already does, so staging and lookup agree.
- An unsupported runtime now fails with something actionable. The reporter in #1940 had to run `file` on the `.so` to find the mismatch; now it is:

```
hyperswitch-prism: libconnector_service_ffi is not bundled for linux-s390x
(expected .../generated/linux-s390x/libconnector_service_ffi.so).
This package bundles: darwin-arm64, linux-arm64, linux-x64.
```

### On scope

The gRPC part is consistency rather than a crash anyone hits today: staging and loading were both flat, so they agreed for any single-arch build. It starts to matter once a package carries more than one architecture, which is what `make dist` now produces for the UniFFI native. The diagnostic is the part users feel immediately.

One thing I ran into and left alone: `make -C sdk/javascript dist` stages only `libconnector_service_ffi`, and `release-sdks.yml` never calls `generate-grpc-bindings`, so no gRPC native ships in the npm tarball on any platform. Shipping a second native reads as a packaging decision rather than a loader fix, so I have not made it here. Happy to follow up if it is wanted.

### Testing

`make -C sdk test-grpc` in ci.yml covers this path end to end — it stages the native, packs a tarball, installs it, and loads through the resolver.

Locally:

- `tsc --noEmit` matches `main`: the same two pre-existing `generated/proto` errors, nothing new.
- Resolver checked against a simulated multi-arch tarball with `process.platform` and `process.arch` overridden. Each of linux-arm64, linux-x64 and darwin-arm64 resolves to its own directory and extension; aarch64 never falls through to the x86-64 binary; a stray flat `generated/*.so` is ignored; both failure paths produce the message above.
- `make -C sdk/javascript generate-grpc-bindings` stages into `generated/linux-x64/`.

I do not have aarch64 hardware, so this has not run against a real Graviton box — @alondayan2 offered to test an arm64 build in #1940.

Refs #1940


Utkal059:
CI note on the failing `SDK Tests` check: it stops at `Credentials file not found: creds.json`. The job log reports `Secret source: None` — a PR from a fork doesn't receive `CONNECTOR_SPECIFIC_AUTH`, so the "Create connector credentials" step writes nothing and the Rust and Kotlin gRPC smoke tests can't start. The JavaScript and Python harnesses fall back to placeholder credentials, and both passed:

```
══ gRPC Results Summary ══
  ✗ Rust (exit 2)
  ✓ JavaScript
  ✓ Python
  ✗ Kotlin (exit 2)
```

JavaScript is the only SDK this PR touches, and that run exercises the changed path end to end — `generate-grpc-bindings` stages the native, `pack-archive` builds the tarball, the test installs it, and the client loads through the new resolver.

The diff touches no Rust, Kotlin, or workflow files, so I've left CI alone rather than working around the missing credentials. Happy to rebase or re-run whenever it suits.

XyneSpaces:
[blocking] The JavaScript package manifest still only includes `dist/src/payments/generated/*`, but this PR moves native libraries into `generated/<platform>-<arch>/...`. npm's `files` glob will not include those nested directories, so a published package can install without the native library and `resolveNativeLib` will fail at runtime; please update `sdk/javascript/package.json` to include the recursive generated tree (for example `dist/src/payments/generated/**/*`).

XyneSpaces:
Automated review note: I found line-specific issues in this PR, but this environment only exposes PR-level comment creation and not inline diff-comment posting. Per review policy, I’m not bundling line-specific findings into a PR-level comment.

XyneSpaces:
[blocking] `dist` still stages only `libconnector_service_ffi`, but `_generated_grpc_client.ts` now resolves `generated/<platform>-<arch>/libhyperswitch_grpc_ffi.*`. Please update the distribution packaging loop to stage `libhyperswitch_grpc_ffi` for every supported target too; otherwise the published JS package will fail to load the gRPC client on every platform.

Utkal059:
Pushed `fc83e72` — thanks for both. I think the second one holds and the first doesn't; evidence for each below.

**`dist` staging the gRPC native** — correct, and fixed.

One wrinkle the suggestion doesn't cover: `release-sdks.yml` builds and uploads only `libconnector_service_ffi`, so staging `libhyperswitch_grpc_ffi` in `dist` alone would have hit the existing fail-closed check and aborted every release run. The commit does both — `JS_NATIVE_LIBS` in the staging loop, plus a `build-grpc-ffi-lib` step and a widened upload glob in the workflow. A published package now carries a gRPC native for the first time.

**`package.json` `files` glob** — I don't think this one is real, so I've left it unchanged.

`npm-packlist` includes a matched *directory* recursively, so `dist/src/payments/generated/*` already covers `generated/<platform>-<arch>/`. Three `npm pack` runs over identical trees:

| `files` entry | natives packed |
| --- | --- |
| `generated/*` (current) | all 6 |
| generated line removed | none — confirms that entry is what ships them |
| `generated/**/*` (suggested) | identical to current |

And from a real `make -C sdk/javascript dist` tarball, with `package.json` untouched:

```
package/dist/src/payments/generated/darwin-arm64/libconnector_service_ffi.dylib
package/dist/src/payments/generated/darwin-arm64/libhyperswitch_grpc_ffi.dylib
package/dist/src/payments/generated/linux-arm64/libconnector_service_ffi.so
package/dist/src/payments/generated/linux-arm64/libhyperswitch_grpc_ffi.so
package/dist/src/payments/generated/linux-x64/libconnector_service_ffi.so
package/dist/src/payments/generated/linux-x64/libhyperswitch_grpc_ffi.so
```

Worth noting the per-arch layout isn't new in this PR — #1996 shipped it for the UniFFI native, so a glob that couldn't reach those subdirectories would already be visible in released packages today.

Happy to switch to `**/*` regardless if you'd prefer it as explicit documentation of intent, but it's a no-op.

alondayan2:
Hi @Utkal059 - any news on this?
thanks?

Utkal059:
@alondayan2 sorry for the wait — rebased onto current `main` just now, diff unchanged at 6 files / +86 −22.

On the two red checks: they're one failure, and it isn't this diff.

`CI Result` is only an aggregator — its log is `sdk-test: failure` and every other input is `success` or `skipped`. So the single failure is `SDK Tests`, and it stops at:

```
Secret source: None
...
Configuration error: Credentials file 'creds.json' not found. Provide it with --creds-file
Exception in thread "main" java.lang.IllegalArgumentException: Credentials file not found: creds.json
```

A pull request from a fork receives no `secrets.CONNECTOR_SPECIFIC_AUTH`, so the "Create connector credentials" step writes nothing and logs *"running in dry-run mode"*. Four harnesses then diverge on what dry-run means:

```
══ gRPC Results Summary ══
  ✗ Rust (exit 2)      <- hard-fails on the missing file
  ✓ JavaScript          <- placeholder creds, connector errors recorded as skipped
  ✓ Python              <- same
  ✗ Kotlin (exit 2)     <- hard-fails on the missing file
```

Three things say this is environmental rather than the change:

1. #2080 is an unrelated PR from a fork and fails identically — same two SDKs, same `creds.json` message, same `Secret source: None`.
2. This diff touches no Rust and no Kotlin.
3. JavaScript is the only SDK it touches, and that run covers the changed path end to end — `generate-grpc-bindings` stages the native, `pack-archive` builds the tarball, the test installs it and loads through the new resolver.

Worth noting the gap is only in *where* the file is read, not in severity handling. Rust already classifies a connector error the same way JavaScript does (`is_transport_error` → failed, otherwise `skipped (connector_error)`, connector still passes), so with placeholder credentials it would land on the JS/Python result. The blockers are `build_config` erroring on the absent file (`sdk/rust/smoke-test/src/grpc_smoke_test.rs`) and `loadCredentials` throwing (`sdk/java/smoke-test/src/main/kotlin/SmokeTest.kt`) — and the FFI harnesses look like they'd need the same treatment (`run_tests` panics in `sdk/rust/smoke-test/src/main.rs`, and Kotlin's FFI path shares the same `loadCredentials`).

I've deliberately not put that in this PR — it's four code paths across two languages, unrelated to a JavaScript native-library loader, and I can't build the Rust workspace or run the Kotlin harness locally to verify it. Happy to open it as a separate PR if that's useful; it would unblock CI for any fork PR touching `sdk/`. Alternatively, since maintainer edits are enabled, pushing this branch to a branch on `juspay/` would give it a run with secrets.

### Re-verified after the rebase

- `tsc --noEmit` — identical to `main`: the same two pre-existing `generated/proto` errors, nothing new.
- Resolver driven against a simulated `make dist` tree with `process.platform`/`process.arch` overridden — 12 assertions, all passing: each of `darwin-arm64`, `linux-arm64`, `linux-x64` resolves to its own directory and extension for both natives; `linux-arm64` never falls through to `linux-x64`; a stray flat `generated/*.so` is ignored; both failure paths produce the diagnostic; `win32` maps to `.dll`.
- `npm pack --dry-run` on npm 11.8.0 over a multi-arch `dist/` — all 6 natives packed with the existing `files` entry, re-confirming that `dist/src/payments/generated/*` already includes the per-arch subdirectories recursively.

```
dist/src/payments/generated/darwin-arm64/libconnector_service_ffi.dylib
dist/src/payments/generated/darwin-arm64/libhyperswitch_grpc_ffi.dylib
dist/src/payments/generated/linux-arm64/libconnector_service_ffi.so
dist/src/payments/generated/linux-arm64/libhyperswitch_grpc_ffi.so
dist/src/payments/generated/linux-x64/libconnector_service_ffi.so
dist/src/payments/generated/linux-x64/libhyperswitch_grpc_ffi.so
```

alondayan2:
@Utkal059 got it - so how do we proceed here?

Utkal059:
Two parts, and then it's a maintainer call.

**The red check.** It isn't this diff, and it isn't only this PR. `SDK Tests` dies in its first test step because a fork PR gets no `CONNECTOR_SPECIFIC_AUTH`, so no `creds.json` is written and the four harnesses disagree about the missing file — Rust and Kotlin abort, JavaScript and Python fall back to placeholders. The two steps behind it need that file just as much and never run either (`run_smoke_tests_parallel.py:499` copies it unconditionally; all four FFI harnesses abort on it). I've opened https://github.com/juspay/hyperswitch-prism/pull/2129 for that: one line in `ci.yml`, copying the `creds_dummy.json` the mock tests already use when the secret is absent. Being a fork PR itself, its own run exercises the branch it adds.

**This PR.** Unchanged and ready — 6 files, +86 −22, rebased onto `main` three days ago. JavaScript is the only SDK it touches, and `SDK Tests` covers that path end to end: `generate-grpc-bindings` stages the native, `pack-archive` builds the tarball, the test installs it and loads through the new resolver. What it's waiting on is two approving reviews from `juspay/connector-service-maintainers`.

So either order works:

- merge the CI fix first, then re-run this one and read a green check, or
- review this one now on the evidence above and let CI catch up after.

Happy to rebase onto current `main` whenever you'd like a fresh run.

alondayan2:
@Utkal059 im not a hyperswitch maintainer - I thought you are
who we can talk to?

gopikrishna000:
> @Utkal059 im not a hyperswitch maintainer - I thought you are who we can talk to?

hi, we are looking into the pr., sorry for the late reply.
@Utkal059 Thanks for the contribution, and for opening #2129 alongside it.

gopikrishna000 (approved):

gopikrishna000:
LGTM.
@Utkal059  Good catch on the release workflow. Staging the grpc native without building it would have broken every release, and the earlier comment missed that.
You were right about the files glob too — 0.4.0 already ships the per-arch dirs, so nothing to change there.
The red check is the fork/secrets thing, not your change. Rust and Kotlin abort when creds.json isn't there.

I'll do a preview release run to cover the aarch64 build, then get a second approval and merge. Thanks for sticking with it.

JeevaRamu0104 (approved):

alondayan2:
Hi @gopikrishna000 @JeevaRamu0104, i see the CI is still failing for this - do you know when it should be merged roughly?

alondayan2:
Hi @gopikrishna000 @JeevaRamu0104 , any news?

PR #2030, opened 2026-08-02
feat: add payment method headers in shadow mode

AmitsinghTanwar007:
## Summary
- add `x-payment-method` and `x-payment-method-type` header constants
- carry optional `payment_method_type` on `PaymentFlowData` from payment service requests
- add PM/PMT headers to UCS outbound connector requests only when shadow mode is enabled

Related to juspay/hyperswitch#13500


XyneSpaces:
[blocking] Please update every explicit `PaymentFlowData { ... }` construction (or introduce a constructor/default path) when adding `payment_method_type`. A repo-wide search shows many existing `PaymentFlowData` struct literals outside the three updated conversions, and Rust will fail to compile any explicit literal that does not set the new non-default field.

XyneSpaces:
[blocking] Adding `payment_method_type` to `PaymentFlowData` requires updating every struct literal that builds this type, not only the three conversions touched here. Code search still shows other `PaymentFlowData { ... }` initializers in connector tests/examples; please update them or add a defaulted construction path so the workspace does not fail with a missing-field error.

XyneSpaces:
[should-fix] Possible issue — worth verifying: the new shadow headers are emitted from `PaymentFlowData::payment_method`, but several `PaymentFlowData` constructors still use placeholder `PaymentMethod::Card` values for non-authorize flows. For a sync/void/capture shadow call on a non-card payment, this can send `x-payment-method: card`; please only add this header when the method is derived from real request data, or skip it for flows where it is not known.

XyneSpaces:
[blocking] Adding `PaymentFlowData.payment_method_type` makes every existing `PaymentFlowData { ... }` literal supply this field, but this PR only updates the authorize/setup-recurring/recharge paths. The other constructors in `crates/types-traits/domain_types/src/types.rs` still omit it, so this branch should fail to compile; please add `payment_method_type: None` or the appropriate converted value to every `PaymentFlowData` construction.

XyneSpaces:
[should-fix] The new `x-payment-method` header is populated from `PaymentFlowData.payment_method`, but the updated request conversions still hard-code `payment_method: PaymentMethod::Card`. Non-card requests will therefore be tagged as card in shadow-mode traffic; please derive `payment_method` from the incoming request alongside `payment_method_type`, or omit the header when the value is not known.

XyneSpaces:
[should-fix] `convert_optional_payment_method_type` silently discards unrecognized `payment_method_type` enum values via `.ok()` and turns them into `None`. That masks bad input and makes the new shadow-mode header unreliable; propagate the conversion error instead of swallowing it.

`crates/types-traits/domain_types/src/types.rs`, added line 79.

XyneSpaces:
[blocking] Gate the new `x-payment-method` and `x-payment-method-type` headers behind the existing shadow-mode condition, or update the implementation if these are intended for all traffic. The PR says these headers are for shadow mode, but the added request-building code injects them unconditionally, which will send the new headers to live connector traffic.

gopikrishna000 (dismissed):

gopikrishna000 (dismissed):

iemyashasvi (dismissed):

gopikrishna000 (approved):

iemyashasvi (approved):

shuklatushar226 (approved):

PR #2034, opened 2026-08-03
feat(connector): Finix Apple Pay Tokenize + SetupRecurring, Apple Pay & Google Pay webhooks

shuklatushar226:
## Summary

Implements four Finix wallet flows in UCS, with a full logic-parity pass against the reference implementation in `hyperswitch/crates/hyperswitch_connectors/src/connectors/finix`:

- **Apple Pay Tokenize** (`PaymentMethodToken`)
- **Apple Pay SetupRecurring** (`SetupMandate`)
- **Apple Pay Webhooks**
- **Google Pay Webhooks**

Flow-name mapping for reviewers: `Tokenize` / `SetupRecurring` are the UCS gRPC rpc names; the legacy Hyperswitch flow types are `PaymentMethodToken` / `SetupMandate`. Parity was assessed against the latter.

The work was implemented flow-by-flow, then reviewed by an **independent adversarial audit** that validated claims against the real Finix OpenAPI (`https://docs.finix.com/_spec/api/index.yaml`) rather than only against the reference implementation. The audit found two regressions introduced by the first four commits; both are fixed in `5fdd260dd`, which also adds a status-matrix test to pin the behaviour.

## Commits

| Commit | Scope |
|---|---|
| `44b89f158` | Apple Pay Tokenize — token structs, `build_apple_pay_instrument_request()`, Tokenize + SetupMandate arms |
| `4f59b3895` | SetupRecurring parity — 4-digit card expiry, Google Pay name/address, real disabled-instrument errors |
| `f809ad890` | Incoming-webhook parity — unify status mapping, fix dispute/tags decoding, stop mis-classifying ledger transfers |
| `6c8aae17c` | Dispute webhook decodes a null amount; Google Pay webhook verification |
| `5fdd260dd` | Audit remediation — restore `CANCELED` failure semantics, reconcile webhook event type with payload status |
| `158f877c2` | Record verified Finix documentation links |

## Defects fixed along the way

These were found while implementing the four flows and are fixed here:

1. **Ledger transfers were resolved as payments.** A `SETTLEMENT` transfer returned `PAYMENT_INTENT_SUCCESS` with a payment reference — a merchant payout would have been applied to a payment attempt as a charge. Now `WebhookEventTypeNotFound`.
2. **Every dispute webhook was undecodable.** `currency` was required but Finix omits it; because `FinixEmbedded` is `#[serde(untagged)]`, the failure took down `get_event_type` and `get_webhook_event_reference` too.
3. **`FinixDisputes.amount` was required** but Finix declares it `nullable: true` — same untagged-decode collapse.
4. **`FinixPaymentStatus` had no `#[serde(other)]`**, so a `RETURNED` state broke PSync/Refund deserialization outright.
5. **Card `expiration_year` was sent verbatim** — a merchant sending `"30"` stored a card-on-file mandate with `expiration_year: 30`. Now expanded via the existing shared util. Verified live: sent `"30"`, read back `2030`.
6. **Google Pay dropped billing `name`/`address`** while the Apple Pay arm sent them.
7. **`disabled_instrument_error` collapsed two distinct failures** into `NO_ERROR_CODE`/`NO_ERROR_MESSAGE`.
8. **Webhook HMAC used `String::from_utf8_lossy`** — U+FFFD substitution would turn a valid signature into a mismatch. Now byte-exact, which Finix's docs require.

## Structural cleanup

Deleted the webhook-only `FinixState` enum so `FinixPaymentStatus` is the single deserialization source for Finix `state`, and routed all 8 flows (Authorize, PSync, Capture, RSync, Void, Refund, RepeatPayment, webhooks) through one canonical `get_finix_attempt_status()`, deleting 8 hand-rolled inline matches. The webhook **event-type** mapper is now defined *in terms of* the status mapper, so the two cannot drift apart by hand.

## Testing

- **Webhooks: 32/32 live assertions** via gRPC `EventService/ParseEvent` + `/HandleEvent` with real HMAC-SHA256 `Finix-Signature` headers. The same suite scores 18-passed / 14-failed against the pre-remediation binary; the 14 are exactly the regression rows.
- **Card end-to-end against Finix sandbox:** Create → Tokenize → Authorize(MANUAL) → PSync → Capture → PSync, plus Authorize → Void → PSync, plus a decline path (`4000000000009987` → 402 → `AUTHORIZATION_FAILED`). Byte-identical before and after remediation.
- **Recurring, live via Card:** `SetupRecurring` → 201 `CHARGED` with `mandateReference`, then an MIT `RecurringPaymentService/Charge` against that mandate → 201.
- **New `finix/test.rs`** pins a 54-triple `(flow, state, is_void) → AttemptStatus` matrix, plus `canceled_without_void_is_a_payment_failure`, `unknown_is_never_terminal`, `pending_void_stays_pending`.
- `cargo build`, `cargo clippy --all-targets`, `cargo +nightly fmt --check`, and 7 unit tests all clean.

### Known testing limitation

**No wallet-token-bearing call can return 2xx in this environment.** Finix cryptographically verifies the Apple-issued PassKit signature server-side and publishes no sandbox wallet token; a structurally-correct request returns `422 INVALID_FIELD "Invalid Apple Pay token"` while the identical Card request succeeds. Apple Pay / Google Pay **request construction is therefore verified structurally, not live** — request JSON shape, error surfacing with real code/message/reason, and fast-failing negative paths. Webhook flows are unaffected and fully live-tested, since they are payload-driven.

Both wallet token shapes were verified against the vendor OpenAPI examples: Apple Pay is the wrapped `{"token":{...}}` object serialized to a string, Google Pay is the raw ECv2 token string with no wrapper.

## Intentional deviations from the reference

Every deviation below is deliberate and was reviewed by the independent audit.

| # | Deviation | Rationale |
|---|---|---|
| 1 | `UNKNOWN` → `Pending` (reference: terminal failure) | An indeterminate state must never be terminal. The reference is already self-inconsistent here — it does exactly this for DEBIT transfers. |
| 2 | `PENDING` void stays `Pending` (reference: `Voided`) | The reference reports funds released before Finix says so. |
| 3 | `CANCELED` → `Voided` **only** when `is_void` is set | Restored in `5fdd260dd`. On non-void paths there is no void in flight, and collapsing both contexts discards the decline reason. |
| 4 | `RETURNED` modelled explicitly (reference has it commented out) | Finix documents it in `transfer.state`; leaving it unmodelled broke PSync deserialization. |
| 5 | HMAC over raw body bytes, not `from_utf8_lossy` | Finix docs require byte-exact. The reference has a latent bug here. |
| 6 | Apple Pay header fields are `Secret<String>` (reference: plain `String`) | Cryptogram material. Wire bytes identical. |
| 7 | DEBIT/REVERSAL guards in the webhook content builders | No reference equivalent, but load-bearing in UCS: the shared dispatcher falls back to the payment builder for unrecognised event types, so without the guard a FEE or SETTLEMENT would post to a payment attempt. |
| 8 | Void hardcodes `is_void = Some(true)` | The reference would report `SUCCEEDED → Charged` on a void. |
| 9 | SetupMandate sends `tags{merchant_reference}` (reference: `tags: None`) | `POST /payment_instruments` has no `idempotency_id`; the tag is the only correlation handle. |
| 10 | SetupMandate rejects a disabled instrument / missing id (reference: unconditional `Charged`) | Persisting a disabled instrument as a mandate strands every later MIT. |
| 11 | RepeatPayment rejects `NetworkMandateId` / `NetworkTokenWithNTI` | Confirmed against the OpenAPI: Finix exposes no `network_transaction_id` property at all; it chains by its own transfer id. |
| 12 | `EventNotSupported` → `IncomingWebhookEventUnspecified` | Architecture-forced; the Hyperswitch UCS client maps it back. |
| 13 | `connector_dispute_id` stays `None` on the dispute reference | Lets the normaliser resolve `dispute.transfer`, matching the reference's `PaymentId(ConnectorTransactionId(..))`. |
| 14 | No Google-Pay-specific webhook code | Verified, not assumed — Finix's event catalogue has no wallet entity, its own Google Pay guide publishes a transfer field-identical to a card transfer, and the OpenAPI `transfer` schema has 47 properties with no wallet discriminator. |
| 15 | No webhook timestamp-freshness check | Absent in both implementations; adding it unilaterally breaks parity and risks rejecting Finix's own 15-minute retries. Recommended as a framework-level follow-up. |

## Known issues NOT fixed here

Deliberately out of scope. Each is pre-existing and/or needs a change outside `connectors/finix*`.

- **`credential_on_file` is never sent on RepeatPayment (CRITICAL).** The code asserts *"No MIT-specific fields are required"*; the Finix OpenAPI contradicts this — `POST /transfers` and `/authorizations` accept `credential_on_file{type, initial_transfer_id}`, with `initial_transfer_id` **required** for both `RECURRING` and `UNSCHEDULED`. Month-2 subscription charges therefore post as ordinary CNP transactions, risking soft declines, loss of recurring interchange tier, and Visa Stored Credential Framework / Mastercard MIT assessments. **`hyperswitch` Direct has the identical gap, so fixing it here would deliberately break parity** — it needs a coordinated change in both, and should be raised against Direct too. The plumbing already exists unused: `RepeatPaymentData.mit_category` maps 1:1 onto Finix's enum and is already consumed by shift4/checkout/dlocal/tsys_transit.
- **Dispute webhooks cannot report an amount (HIGH).** `process_dispute_webhook` needs a currency that does not exist in Finix's 19-property dispute schema. It is not fixable in connector scope: domain `EventContext` carries only `capture_method`, the proto `DisputeEventContext` is an **empty message**, and `dispute.transfer` would need an outbound `GET`. Related finding: `ForeignTryFrom<DisputeWebhookDetailsResponse> for DisputeResponse` hardcodes `dispute_amount: None`, so **every** connector currently computes a dispute amount/currency that no caller consumes. Rather than fabricate a value, the error now names the real gap (`WebhookMissingRequiredContext`) instead of falsely claiming Finix omitted a field it does not define.
- **Endpoint-verification probes fail in `HandleEvent`.** `webhook_utils.rs:72-81` routes any non-payment/refund/dispute event to the payment builder, which cannot parse `{}`. `ParseEvent` handles the probe correctly. Affects every connector.
- **No webhook replay window / non-constant-time signature compare.** Finix recommends a 300s window. Framework-level, reference-parity.
- **Webhook signing-key rotation unsupported.** Finix's `updateWebhook` exposes `generate_new_secret` / `previous_secret_active_hours`, so during rotation either key may be valid; both implementations use only `secret` and ignore `additional_secret`.

## Router-data shadow validation

Not run. `juspay/ucs-shadow-validation-service` requires a live legacy Hyperswitch router **and** a live UCS grpc-server both hitting real Finix sandbox — it has no fixture or replay mode — and the Docker path is unavailable in the build environment. Separately, **webhook parity is not measurable with that service as it stands**: `WebhookShadowSnapshot` carries no `payment_id`, and `routerDataHandler` rejects any payload missing it (`VS_42`) before comparing. A complete runbook for the cheapest viable path, including the exact `[comparison_service]` config and wallet rollout keys, is available and can be attached on request.


XyneSpaces:
[should-fix] Dispute webhook handling now requires `dispute.currency` and `dispute.amount`, but the added Finix dispute model treats both as optional and the code comments note the resource may not carry currency. A real Finix dispute webhook missing either field will fail before producing `DisputeWebhookDetailsResponse`; please resolve these from event context or return an explicit unsupported event instead of wiring a path that always errors for live payloads.

PR #2036, opened 2026-08-03
feat(observability): add typed connector request/response fields

### LINKED ISSUE #2038: api to respond the typed connector response instead or along with raw connector response
proto changes are required

AmitsinghTanwar007:
## Summary

Adds `typed_connector_request` and `typed_connector_response` fields across the entire connector service stack to provide structured, domain-typed observability data **separate from** raw wire-format payloads.

### Problem
Currently, only `raw_connector_request` / `raw_connector_response` exist — these carry the literal HTTP bytes sent/received. For debugging and analytics, we also need the **typed representation** (the Rust domain struct, masked and serialized) so consumers can reason about connector data in a connector-agnostic way without parsing raw JSON/XML from different providers.

### What changed

**Proto definitions** (`payment.proto`, `frm.proto`):
- Added `typed_connector_request` and `typed_connector_response` optional fields to all response messages (Authorize, Get, Void, Capture, Refund, CreateOrder, SetupRecurring, MandateRevoke, FRM pre/post risk check, etc.)

**Domain types** (`connector_types.rs`, `frm_types.rs`, `surcharge_types.rs`, `payouts_types.rs`, `merchant_authentication_flow_data.rs`):
- Added `typed_connector_request` / `typed_connector_response` fields to all flow data structs (`PaymentFlowData`, `RefundFlowData`, `DisputeFlowData`, `FrmFlowData`, `SurchargeFlowData`, `VerifyWebhookSourceFlowData`, `RefreshPaymentMethodFlowData`, `PayoutFlowData`)
- Extended the `RawConnectorRequestResponse` trait with `get/set_typed_connector_request` and `get/set_typed_connector_response` methods
- Implemented the new trait methods for all flow data structs

**Connector integration trait** (`connector_integration_v2.rs`):
- Added `get_typed_connector_request()` default method returning `None`
- Wired it into `build_request_v2()` via `set_typed_connector_request()` on the request builder

**Request infrastructure** (`request.rs`):
- Added `typed_connector_request` field to `Request` and `RequestBuilder`
- Added `set_typed_connector_request()` builder method

**Connector macros** (`macros.rs`):
- Added `serialize_typed_connector_payload()` utility — masked serialization of typed structs
- Added `expand_fn_get_typed_connector_request!` macro — generates `get_typed_connector_request()` for each flow
- Modified `expand_fn_handle_response!` to capture and set `typed_connector_response` on the flow data
- Wired both macros into all 4 `macro_connector_implementation!` variants

**Connector-specific** (`bamboraapac/transformers.rs`):
- Added `Serialize` derive to request structs that were missing it (required for typed serialization)

**Service layer** (`service.rs`):
- Set `typed_connector_request` on flow data from the request object after connector call

**gRPC response generators** (`types.rs`, `frm/types.rs`, `payments.rs`):
- All `generate_*_response()` functions now extract and pass typed fields to proto responses

**Server layer** (`events.rs`):
- Added `typed_connector_request/response: None` to `VerifyWebhookSourceFlowData` construction

### Key design principle: raw ≠ typed
- **raw**: the literal bytes off the wire (HTTP body as-is)
- **typed**: the domain Rust struct, masked-serialized to JSON — these MUST be different values

## Test plan
- [ ] `cargo check` passes across all crates
- [ ] Existing connector tests (razorpay, calida, bamboraapac) compile with new fields
- [ ] Verify typed fields are `None` by default for connectors not using the macro system
- [ ] Verify typed fields are populated for macro-based connectors (e.g., bamboraapac)
- [ ] Confirm raw and typed values differ when both are populated


crates/grpc-server/grpc-server/src/server/payments.rs
gopikrishna000, line 3867: why was these functions added in this file instead of type.rs
not related to this pr though, but how did we missed

config/development.toml
```diff
@@ -18,7 +18,7 @@ buffer_limit = 100000
 
 [server]
 host = "0.0.0.0"
-port = 8000
+port = 8003
```
JeevaRamu0104: revert these changes.

crates/grpc-server/grpc-server/src/server/payments.rs
AmitsinghTanwar007: Don't know but i think in future we need to write clippy rules so that the structure is maintained

crates/integrations/connector-integration/src/connectors/macros.rs
```diff
@@ -369,6 +385,75 @@ macro_rules! expand_fn_get_request_body {
...
+                    router_data: req.clone()
+                };
+                bridge
+                    .request_body(input_data)
+                    .ok()
+                    .and_then(|request| crate::connectors::macros::serialize_typed_connector_payload(
+                        &request,
+                        "typed_connector_request",
+                    ))
+            }
```
JeevaRamu0104: **[S1] The typed request is reconstructed, not captured — it can differ from what was actually sent.**

This re-invokes `bridge.request_body(req.clone())` a second time, independently of `get_request_body`. For connectors with per-call non-determinism the two diverge:

- `airwallex/transformers.rs:1264` mints a fresh `uuid::Uuid::new_v4()` per call, with a comment stating reuse is rejected as `duplicate_request`.
- `nuvei/transformers.rs:82-86` derives a checksum from `date_time::now()`.

The observability record will show a `request_id`/timestamp/signature that was never on the wire — misleading for exactly the debugging use case this PR exists for.

It also doubles request-transformation cost, including a full `RouterDataV2` clone, on the payment hot path.

Capture the typed value once inside `get_request_body` and thread it through instead.

crates/common/external-services/src/service.rs
```diff
@@ -1081,15 +1115,14 @@ where
...
-                .masked_serialize()
-                .unwrap_or(json!({ "error": "failed to mask serialize connector request"})),
-            RequestContent::FormData(_) => json!({"request_type": "FORM_DATA"}),
-            RequestContent::RawBytes(_) => json!({"request_type": "RAW_BYTES"}),
-        },
+        Some(request) => request
+            .masked_serialize_inner()
+            .map(|(v, _)| v)
+            .unwrap_or_else(|| match request {
+                RequestContent::FormData(_) => json!({"request_type": "FORM_DATA"}),
```
gopikrishna000, line 1167: for FormData and RawByte, can we still pass the another parameter of typed struct similar to other types

crates/common/external-services/src/service.rs
```diff
@@ -441,7 +443,7 @@ where
                             .set_connector_response_headers(body.headers.clone());
                     }
 
-                    let error_response = match body.status_code {
+                    let (error_response, typed_response) = match body.status_code {
```
gopikrishna000: for 4xx or 5xx, since we added the new field inside ConnectorError proto, so should adding the typed_response field fiedl inside ErrorResponse struct be sufficient? 
Here we setting inside resource_common_data, and taking back to set ErrorResponse

crates/integrations/connector-integration/src/connectors/macros.rs
```diff
@@ -393,13 +437,23 @@ macro_rules! expand_fn_handle_response {
                 ))?;
 
             let response_body = bridge.response(response_bytes, res.status_code)?;
-            event_builder.map(|i| i.set_connector_response(&response_body));
+            // Serialize once: masked Value for event logging, String for typed_connector_response
```
gopikrishna000, line 423: bro, actually for event also we need string type only., we are doing stringy at downstream and that dont handle all cases like array, float etc
So we can actually send string only to events as well, so no need of this divergence 
(Both request and response)

crates/integrations/connector-integration/src/connectors/macros.rs
```diff
@@ -76,6 +76,49 @@ pub(crate) fn validate_xml_structure(xml_str: &str) -> Result<(), String> {
...
+            );
+        })
+        .ok()
+}
+
+/// Masked-serialize once, returning both the `serde_json::Value` (for event
+/// logging via `set_connector_response`) and the stringified form (for
+/// `typed_connector_response`).  This avoids calling `masked_serialize` twice.
+pub(crate) fn masked_serialize_connector_response<T: serde::Serialize>(
+    payload: &T,
```
gopikrishna000: this fucntion is similar to from_masked_optional that we have as impl for MaskedSerdeValue type?

crates/integrations/connector-integration/src/connectors/macros.rs
```diff
@@ -393,13 +437,23 @@ macro_rules! expand_fn_handle_response {
                 ))?;
 
             let response_body = bridge.response(response_bytes, res.status_code)?;
-            event_builder.map(|i| i.set_connector_response(&response_body));
+            // Serialize once: masked Value for event logging, String for typed_connector_response
+            let serialized =
+                crate::connectors::macros::masked_serialize_connector_response(&response_body);
+            if let Some(evt) = event_builder {
+                if let Some((ref value, _)) = serialized {
+                    evt.set_connector_response(value);
```
gopikrishna000: also this set_connector_response, already does MaskedSerdeValue::from_masked_optional

crates/integrations/connector-integration/src/utils.rs
```diff
@@ -75,6 +75,38 @@ macro_rules! with_response_body {
     };
 }
 
+/// Serialize the connector response once, build `RouterDataV2` via `TryFrom<ResponseRouterData>`,
+/// set both event response data and `typed_connector_response`, and return `Ok(result)`.
+///
+/// Replaces the 12-line boilerplate block in every manual `handle_response_v2`.
+#[macro_export]
+macro_rules! set_typed_response {
```
gopikrishna000: name is set_typed_response
But we do more than that
we can name better

crates/integrations/connector-integration/src/connectors/macros.rs
```diff
@@ -76,6 +76,49 @@ pub(crate) fn validate_xml_structure(xml_str: &str) -> Result<(), String> {
     Ok(())
 }
 
+pub(crate) fn serialize_typed_connector_payload<T: serde::Serialize>(
```
gopikrishna000, line 79: it is similar to from_masked_optional that we have.
It comes out of box if we use typed MaskedSerdeValue instead of serde::json, so mistaken unmasked value also not possible

crates/common/ucs_env/src/error.rs
```diff
@@ -221,7 +221,8 @@ impl ToGrpcStatus for ConnectorError {
         let _ = connector_error.encode(&mut buf);
 
         match self {
-            Self::ConnectorErrorResponse(error_response) => match error_response.status_code {
+            Self::ConnectorErrorResponse { error_response, .. } => match error_response.status_code
```
gopikrishna000: having raw_connector_response field inside error_response struct would be cleaner?

crates/common/external-services/src/service.rs
AmitsinghTanwar007: I will handle this change in later PR as this may require change in L3 layer which i need to go through,
already the pr has many changes.

crates/integrations/connector-integration/src/utils.rs
```diff
@@ -75,6 +75,40 @@ macro_rules! with_response_body {
...
+            }
+        }
+        let mut result = RouterDataV2::try_from(ResponseRouterData {
+            response: $response,
+            router_data: $data.clone(),
+            http_code: $status_code,
+        })
+        .map_err(|e| {
+            e.change_context($crate::ConnectorError::ResponseHandlingFailed {
+                context: Default::default(),
```
gopikrishna000: can we update default context

gopikrishna000: also why map_Err? already RouterDataV2::try_from has Report ConnectorError?

JeevaRamu0104 (dismissed):

gopikrishna000 (dismissed):

JeevaRamu0104 (approved):

gopikrishna000 (approved):

shuklatushar226 (approved):

PR #2057, opened 2026-08-04
feat(core): add connector_request_reference_id to WebhookDetailsResponse

Aishwariyaa-Anand:
## Description

PR juspay/hyperswitch-prism#1968 introduced `connector_reference_id` as a field. It was subsequently reverted in juspay/hyperswitch-prism#2048 at domain types due to backward compatibility

`WebhookDetailsResponse` does not carry this field in `connector_request_reference_id` for merchant_txn_id.
This PR adds the missing field so that once PR #1968 is re-applied, webhooks will continue to function correctly.

**All connector implementations**
- For every connector that constructs `WebhookDetailsResponse`, mapped `connector_request_reference_id` to the same value as `connector_response_reference_id` (with appropriate clone/move handling per Rust ownership rules)

### Additional Changes


crates/integrations/connector-integration/src/connectors/phonepe.rs
```diff
@@ -223,7 +223,8 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
                 payload.transaction_id.clone(),
             )),
             status,
-            connector_response_reference_id: Some(payload.merchant_transaction_id),
+            connector_response_reference_id: Some(payload.merchant_transaction_id.clone()),
+            connector_request_reference_id: Some(payload.merchant_transaction_id),
```
gopikrishna000, line 227: i see both field connector_response_reference_id and connector_request_reference_id populating same data
is it not dont as per connector docs
Ideally both are different meaning right

connector_request_reference_id is for merchant txn id
and connector_response_reference_id for connector txn id of that payment

Aishwariyaa-Anand: yes, in domain types the `merchant txn id` mapping is incorrect where in merchant txn id is mapped to connector_response_reference_id instead of connector_request_reference_id. 

Even if we fix this mapping there is no field equivalent to connector_request_reference_id in Webhook Response resulting to mapping merchant txn id as None, which can break for existing merchant.
Which is why i have added new field connector_request_reference_id and mapped it with the same value as connector_response_reference_id.

gopikrishna000 (approved):

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #2058, opened 2026-08-04
POC: log api_message as structured JSON

AmitsinghTanwar007:
## Summary
- updates the log_utils dependency to the fork commit containing SerdeValue
- records external service api_message with log_utils::SerdeValue so JSON log mode keeps it as structured JSON instead of a string

## Framework dependency
- Depends on https://github.com/juspay/framework-libs-rs/pull/20
- This PoC currently points log_utils to https://github.com/AmitsinghTanwar007/framework-libs-rs at rev f4cebf380e9fd3a0080f390736817e17be369b1b
- After the framework PR merges, this should be switched back to the upstream framework-libs-rs repo/rev

## Verification
- cargo fmt --check
- cargo check -p grpc-server

Screenshot 2026-08-04 at 8 42 09 PM


XyneSpaces:
[blocking] The `log_utils` dependency is being switched to `https://github.com/AmitsinghTanwar007/framework-libs-rs` in production crates. Pulling runtime logging code from a personal fork bypasses the org-owned dependency chain and can block security updates from `juspay/framework-libs-rs`.

Please point this back to the org-owned repository and land the required `log_utils` changes there first.

PR #2070, opened 2026-08-06
fix(netcetera): fix Authenticate response parity and card-proxy expiry handling

ShankarSinghC:
## Summary
- Always populate `authentication_data` (with `trans_status`) on the Authenticate response for both challenge and frictionless outcomes, mirroring hyperswitch's own Netcetera connector — it was previously nulled out for the challenge case specifically, deferring everything to the RReq webhook and leaving callers with no `trans_status` to read.
- Drop the redundant `acsSignedContent`/`acsReferenceNumber`/`acsTransID` duplication into `RedirectForm.form_fields` — that data already travels via `connector_feature_data`, and `acs_transaction_id` is now reliably present on `authentication_data`. `RedirectForm` stays scoped to what a browser actually submits (`endpoint` + `creq`).
- `card_exp_month`/`card_exp_year` on the `ProxyCardDetails -> Card<VaultTokenHolder>` conversion now pass through the real value directly instead of a `"{{$card_exp_month}}"` vault-alias placeholder — Netcetera's own connector derives a combined YYMM `cardholderAccount.cardExpiryDate` by truncating each field, and that derivation is a no-op on opaque placeholder text, producing a wrong-length date for any card whose stored expiry year isn't already 2 digits.
- Route to `AuthenticationStep::Authorize` instead of looping back into `PreAuthenticate` when the initial request completes `PreAuthenticate` with no browser redirect needed (e.g. only 3DS1 supported, or the vault-token/card_proxy path skips DDC).

## Test plan
- [ ] Netcetera 3DS challenge flow: confirm `trans_status`/`acs_url`/`creq` resolve correctly end-to-end through the vault-proxy (card_proxy) path
- [ ] Netcetera 3DS frictionless flow: confirm `trans_status`/`cavv` still resolve correctly
- [ ] Saved/vaulted card with a 4-digit stored expiry year: confirm `cardholderAccount.cardExpiryDate` is the correct 4-character YYMM
- [ ] PreAuthenticate-only completion (no DDC redirect) proceeds straight to Authorize instead of looping


prasunna09 (approved):
domain type changes looks good to me

AmitsinghTanwar007:
Hey @ShankarSinghC , I’m okay with this approach.

As a follow-up, I think UCS should provide reusable scripts/config templates for these proxy integrations. For example, VGS appears to support this kind of transformation through Larky, so clients could directly use a UCS-provided script. That would give us a cleaner contract between UCS and external proxy providers like VGS.

context: about the card exp change

AmitsinghTanwar007 (approved):

gopikrishna000 (approved):

PR #2076, opened 2026-08-07
feat(logs): euler-parity golden-line sources + mapping config (on #2075)

kanikac199:
Stacked on top of #2075 (config-driven log-field engine). Adds the euler **source fields** the app must emit so the config engine can rename/nest them, plus the actual euler field-mapping config.

## What this adds

**Code**
- `http/error.rs`: expose `http_status_for_status()` — the golden line reports the connector's exact 4xx/5xx status, not the coarse gRPC code.
- Incoming line (`utils.rs`):
  - `res_code` — numeric euler status (`200` on success, connector-aware HTTP status on error), recorded additively via the Storage API. `status_code` is left untouched (see compatibility note below).
  - `latency_ms` — additive numeric latency (euler `latency` is a number), recorded via a shared `record_latency_ms()` helper used by both the streaming and non-streaming wrappers.
  - Euler id sources — `merchant_order_id` / `customer_id` (`customer.id`) / `merchant_transaction_id` extracted from the masked payload and recorded as **flat** span fields (the engine sources by flat key; it can't reach into `request_body`). Written via the storage API, so no per-handler `#[instrument]` edits.
  - `action` — the real HTTP verb on the request span (gRPC is always POST; the HTTP gateway carries the true verb) → feeds `api_details.method`.
  - request/response **bodies + headers recorded as structured JSON** (merged from #2094): `record_json_fields_on_span` + `maskable_headers_to_json`.

**Config (`development.toml`)**
- `[log.fields]` with a runtime `enabled` toggle (default opt-out — only `development.toml` sets `enabled = true`; `production.toml`/`sandbox.toml`/`docker_compose.toml` have no `[log.fields]` block).
- Per-direction `[log.fields.incoming]` / `[log.fields.outgoing]` tables. Each entry is one of:
  - `{ source = "<span field>" }` — copy/rename a flat span field. Dotted target keys (e.g. `"api_details.api_tag"`) build the nested `api_details` object.
  - `{ value = "<literal>" }` — inject a constant (e.g. `"api_details.req_type" = { value = "INTERNAL" }` incoming / `"EXTERNAL"` outgoing).
- Renders both golden lines in euler shape: flat `latency` / `entity` / `udf_*` keys plus the nested `api_details` object, split per direction.

> Note: an earlier draft of this description referenced `[log.transformations.incoming|outgoing]` + `[log.static_values]`; the implemented schema is `[log.fields.incoming|outgoing]` with `{ source = … }` / `{ value = … }` (updated here so operators aren't misled about the TOML keys).

## `status_code` compatibility — additive, not breaking
`status_code` is left **exactly as `main`** emits it: the gRPC code-name string (e.g. `"InvalidArgument"`) on error, and unset (`Empty`) on success. The numeric euler value lives in a **new additive `res_code` field** (recorded via the Storage API, no `#[instrument]` edits), and the config sources `resp_code` / `api_details.res_code` from `res_code`. No existing consumer of `status_code` breaks.

Verified live: success → `res_code=200`, `status_code` absent; error → `status_code="…"` unchanged, `res_code=<http status>`.

## Verified live
- **authorizedotnet** (Authorize): incoming `action="POST"` → `api_details.method="POST"`, `category="INCOMING_API"`, `udf_*` populated, `entity="Authorize"`, full nested `api_details`; outgoing `category="OUTGOING_API"`, `api_details.url` = connector URL.
- **twoc_twop_paco** (2C2P PACO, euler→UCS end-to-end): authorize (3DS) → charged → refund (pre-settlement VOID). Golden line carries `api_details.{api_tag, req_type, method, latency, url}`, `latency_ms`, `entity`, and structured req/res bodies. Real 2c2p sandbox responses (invoice `CEBU…`, `PC-B050000 Success`).

## Notes
- The `api_details → message` rename stays downstream (Vector/LP): euler reads `message.*`, but `message` is reserved in `log_utils`, so the app stages the object under `api_details`.
- `hostname`/`version`/`service` need no mapping — euler derives `_version`/`_service` from the pod-name `hostname` (which `log_utils` auto-emits); euler has no standalone `version` field.
- Build prereqs: rustc ≥ 1.96 and `CARGO_NET_GIT_FETCH_WITH_CLI=true` (the pinned `log_utils` rev is a bare sha).


config/development.toml
```diff
@@ -16,6 +16,63 @@ brokers = ["localhost:9092"]
...
+# euler overrides the real `tenant_id` (uuid) per-request via x-config-override.
+"tenant_name" = { source = "tenant_id" }
+# euler emits `latency` at the top level (in addition to the nested one); mirror it here.
+"latency" = { source = "latency_ms" }
+"api_details.url" = { source = "uri" }
+"api_details.method" = { source = "action" }
+"api_details.res_code" = { source = "res_code" }
+"api_details.req_body" = { source = "request_body" }
+"api_details.res_body" = { source = "response_body" }
+"api_details.error" = { source = "error_message" }
```
AmitsinghTanwar007: this cant be in open source file you have to have this in your own deployment toml file

crates/common/external-services/src/service.rs
```diff
@@ -11,6 +11,8 @@ use common_utils::{
     request::TransportType,
 };
 use common_utils::{
+    // `CompiledLogFields` is used by the ungated `EventProcessingParams` struct, so it must be
+    // imported unconditionally (the ffi build compiles this crate without `injector-client`).
```
AmitsinghTanwar007: no need of this comment

crates/common/external-services/src/service.rs
```diff
@@ -579,6 +581,7 @@ pub struct EventProcessingParams<'a> {
         response.status_code = Empty,
         message_ = "Golden Log Line (outgoing)",
         latency = Empty,
+        latency_ms = Empty,
```
AmitsinghTanwar007, line 584: what is the difference between both the above fields

crates/common/external-services/src/service.rs
```diff
@@ -1087,6 +1090,8 @@ where
 
     let elapsed = start.elapsed().as_millis();
     tracing::Span::current().record("latency", elapsed);
+    // Additive numeric latency (euler `latency` is a number). `latency` (string) left as-is.
```
AmitsinghTanwar007: dont mention euler in comments

crates/common/external-services/src/service.rs
```diff
@@ -1087,6 +1090,8 @@ where
...
+    tracing::Span::current().record("latency_ms", u64::try_from(elapsed).unwrap_or(u64::MAX));
```
AmitsinghTanwar007: is there any need of this because we have time out and i dont think so it will ever cross `18,446,744,073,709,551,615`

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -41,6 +41,9 @@ pub fn record_fields_from_header<B: hyper::body::Body>(request: &Request<B>) ->
         "request",
         uri = %url_path,
         version = ?request.version(),
+        // euler's `action` = the real HTTP verb (GET/POST/…); euler maps it to the `_method`
+        // column. gRPC-over-HTTP2 is always POST, the HTTP gateway carries the true verb.
```
AmitsinghTanwar007: dont mention euler avoid this comments

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -276,21 +279,54 @@ where
...
     };
     current_span.record("gateway", connector_name);
     current_span.record("merchant_id", merchant_id);
     current_span.record("tenant_id", tenant_id);
     current_span.record("request_id", request_id);
+
+    // Euler secondary-key id sources (udf_order_id / udf_customer_id / udf_txn_uuid).
+    // Recorded as flat span fields via the storage API because the log transform sources by flat
+    // key (it cannot reach into `request_body`) and this avoids declaring the fields on every
+    // handler's `#[instrument]`. (`category` is set per-direction via static values.)
```
AmitsinghTanwar007: same here

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -276,21 +279,54 @@ where
...
+                "customer_id",
+                body.get("customer")
+                    .and_then(|customer| customer.get("id"))
+                    .and_then(Value::as_str),
+            ),
+            (
+                "merchant_transaction_id",
+                body.get("merchant_transaction_id").and_then(Value::as_str),
+            ),
+        ]
```
AmitsinghTanwar007: the two merchant id are euler specific id's if they know the values they can send it in runtime headers?

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -276,21 +279,54 @@ where
...
+    }) {
+        log_utils::Storage::with_current_span_mut(|storage| {
+            for (key, value) in ids {
+                if let Some(value) = value {
+                    storage.record_value(key, Value::String(value.to_owned()));
+                }
+            }
+        });
+    }
```
AmitsinghTanwar007: dont call this function instead use the existing function record_json_fields_on_span

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -306,6 +342,12 @@ pub fn log_after_initialization<T>(
 
     match &result {
         Ok(response) => {
+            // Additive numeric euler `res_code` (success). `status_code` is left untouched
+            // (main parity) so existing consumers of the gRPC-code field don't break.
+            log_utils::Storage::with_current_span_mut(|storage| {
+                storage.record_value("res_code", Value::from(200_i64));
+            });
```
AmitsinghTanwar007: use the existing function

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -340,7 +382,14 @@ pub fn log_after_initialization<T>(
...
         Err(status) => {
             current_span.record("error_message", status.message());
+            // Backward-compatible: keep main's gRPC code-name string on `status_code`.
             current_span.record("status_code", status.code().to_string());
+            // Additive numeric euler `res_code`: connector-aware HTTP status (e.g. 422) —
+            // matches the HTTP response the caller receives, not the coarse gRPC code.
+            let http_status = crate::http::error::http_status_for_status(status).as_u16();
+            log_utils::Storage::with_current_span_mut(|storage| {
+                storage.record_value("res_code", Value::from(i64::from(http_status)));
+            });
```
AmitsinghTanwar007: use the existing function

crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -356,6 +405,17 @@ pub fn log_after_initialization<T>(
...
+/// Record the additive numeric `latency_ms` on the current span (euler `latency` is a number).
+/// Shared by the streaming and non-streaming logging wrappers to avoid drift.
+fn record_latency_ms(duration: u128) {
+    log_utils::Storage::with_current_span_mut(|storage| {
+        storage.record_value(
+            "latency_ms",
+            Value::from(u64::try_from(duration).unwrap_or(u64::MAX)),
+        );
+    });
+}
```
AmitsinghTanwar007, line 380: is this needed

AmitsinghTanwar007 (changes requested):

crates/common/external-services/src/service.rs
kanikac199: Both point to the same elapsed time. `latency` is the human-readable string we already had; `latency_ms` is the exact same duration but as a plain number of milliseconds, so downstream consumers that want a numeric latency can read it directly instead of parsing the string. One is the string form, the other the numeric form.

crates/common/external-services/src/service.rs
kanikac199: Agreed it will never realistically overflow. `Duration::as_millis()` returns a `u128` and the workspace lints forbid `as` casts and `.unwrap()`, so I still need a checked conversion to `u64`. I dropped the misleading `u64::MAX` fallback and switched to `unwrap_or_default()` (i.e. 0), so on the impossible overflow it is just 0 rather than a fake max value.

crates/grpc-server/grpc-server/src/utils.rs
kanikac199: These are not extra euler-only values. `merchant_order_id`, `customer.id` and `merchant_transaction_id` are already part of the standard request body UCS receives for every payment. I am just reading them out of the body we already have and putting them on the span as flat fields so the log mapping can pick them up (it cannot reach into the nested `request_body`). So the caller does not need to resend them as headers, we already have them. I reworded the comment to make that clear.

crates/grpc-server/grpc-server/src/utils.rs
kanikac199: It is a small helper but it is called from two places (the streaming and non-streaming logging wrappers), so it exists to keep those two identical and avoid drift. I have simplified it to just call `record_json_fields_on_span`, so it is now a 3-line wrapper. Happy to inline it at both call sites if you prefer.

Aishwariyaa-Anand (dismissed):

crates/grpc-server/grpc-server/src/utils.rs
AmitsinghTanwar007: its ok they are part of request body 
but if you need them in logging can't you use the header of log transformation?

kanikac199: Right now the log-transform can only read a field by its exact flat name, it can't look inside a nested object. These three ids sit inside the request_body object (like customer.id), so writing source = "request_body.customer.id" in the config/header wouldn't find them today. That's why I copy them out to flat fields in code first, so the normal source = "merchant_order_id" mapping (which you can still override via the header) can read them.

AmitsinghTanwar007 (approved):

Aishwariyaa-Anand (approved):

PR #2086, opened 2026-08-10
ci: make connector certification a merge-blocking gate with regression attribution

### LINKED ISSUE #2140: REGRESSION
Full change inventory for **#2086** — everything the branch does that `main` does not, and why.

The PR description carries the motivation and the design; this is the exhaustive list, for review and for anyone picking the work up later.

---

## A. The certification gate

`main` has **no live-sandbox gate at all**. A connector change merges on compile + unit tests; nothing ever calls the connector.

**1. `.github/scripts/certify-connectors.sh` (new)** — runs live scenarios and decides, per failure, whether this PR is responsible.

- **Which connectors:** git facts only. Connector files touched → those connectors. `rust-core`/`proto` touched → every connector with `live_in_production: true`. It never names suites or scenarios; the framework derives those from each connector's `specs.json`.
- **Pass 1** runs each connector whole (`--connector X`, no suite/scenario flags) and reads `report.json` for `is_dependency != true && assertion_result == "FAIL"`.
- **Pass 2** re-runs each failure once. A single flake would otherwise trigger an arbitration rebuild and, worse, could read as a regression if the base happens to pass.
- **Pass 3** checks out merge-base sources (`crates config Cargo.toml Cargo.lock` only — never the scripts, which would rewrite the running file mid-read), rebuilds in the same target dir, restarts the server, re-runs the failure there.
- **Verdict, deliberately collapsed to two:** *passes at base + fails at HEAD* → `REGRESSION`, exit 1. **Everything else** → `NOT_ATTRIBUTABLE`, warned with both error strings printed. An observed pass at the base is the only thing that *proves* causation; a classifier comparing error text would guess, and a wrong guess blocks an innocent PR.
- Non-zero exit with no FAIL row (timeout, crash) emits a connector-level row so it cannot read as a pass.
- Missing credentials are a hard `::error::` **once certification is actually selected**. Falling back to placeholders lets a scenario "pass" without reaching the connector.

**2. `.github/scripts/verify-new-connectors.sh` (new)** — a connector added in this PR has no merge base, so every failure is its own by definition and arbitration cannot judge it. Runs before arbitration can swap binaries. Hard-fails on: no `specs.json`; empty `supported_suites`; an `alpha_connectors.json` entry with no `reason`; no entry there *and* no entry in the credentials file; or verification failing.

**3. `ci.yml` wiring** — a `new-connectors` output (spec dir absent at `base.sha` = new, so it is correct across multi-commit PRs); binaries built once; pinned `grpcurl` with SHA-256 verification; the two script steps; and a `github-script` step that posts/updates a single marker comment listing the connectors merging unproven. All of it lives in job `test`, and `ci-gate` — the single required check — has `needs: [… test …]`, so certification blocks merge by inheritance with no new required check to configure.

> **Why a missing credentials entry is fatal:** the sweep path *skips* a connector whose credentials fail to load. A connector silently vanishing from the report reads as "no regression".

---

## B. Three new declaration fields — the connector describes itself

`main`'s `specs.json` carries only `connector` + `supported_suites`.

**4. `alpha_connectors.json`** — the connectors CI has no credentials for and therefore never certifies: 101 of 103 today. One file rather than a flag per connector, so the size of the unproven set is visible in one place, and the file carries a TODO to remove a connector once its credentials reach CI. Replaces the `live_creds` flag this PR originally added per connector.

**5. `live_in_production: bool`** — carries real traffic; the set a shared-code change certifies. The full 100+ sweep belongs to the nightly. Currently `stripe` and `fiserv`.

A `certification-inputs` paths filter widens "shared change" to the six paths certification reads at run time — `connector-creds`, the test binaries, `scripts/`, `config/`, `.github/scripts/`, `webhook_signatures.rs`. None matched a connector path or `rust-core`, so a PR editing only the gate itself selected no connectors and went green. Kept as its own filter because seven other jobs gate on `rust-core`.

**6. `supported_payment_methods: Vec<String>`** — the `payment_method` oneof variant names (`card`, `klarna`, `upi_intent`). Filter applied in `run_suite_test_with_options`: a scenario naming a method runs only if declared; one naming none (Capture, Void, Get, Refund — they inherit their dependency's method) always runs; **empty or absent runs everything**, so it is opt-in and changes no existing connector's behaviour.

Without it, a card-only connector fails every UPI scenario and the only way through is an `assert` override — which permanently hides real regressions on that scenario.

**Why fields and not files:** `live-connectors.json` and `select-certification-targets.sh` both existed mid-PR and are both gone. Facts about a connector belong with that connector; a shared list rots the moment someone adds a connector and forgets it.

---

## C. Runner plumbing

**7. `scripts/grpc-server.sh` (new)** — sourced rather than executed, so a caller owns the lifecycle across many runs instead of paying start+shutdown per scenario. **Bounded shutdown:** the previous unbounded `wait` held a CI job open for two hours after its scenario had already passed. Also exits early if the process dies before the port opens. Only ever signals its own recorded `$GRPC_SERVER_PID` — never `pkill` by name.

**8. `scripts/run-tests --no-build`** — uses prebuilt `target/debug/{grpc-server,test_ucs}` and errors loudly if they are absent. The gRPC call takes seconds; the cost is the build, and arbitration must not rebuild between scenarios.

**9. `test_ucs`: `run_outcome` / `requires_execution`** — a *named* scenario that was skipped (unmet dependency, filter mismatch) left `passed=0, failed=0` and exited **0**: success reported for work that never happened, inherited by everything keyed off the exit code. Now a named selection with nothing executed is an error that names the scenario, while a sweep may legitimately skip everything. Four unit tests.

**10. `#[serde(deny_unknown_fields)]` on `ScenarioOverridePatch` and `PreRequestHttpHook`** — `{"unsupported": true, "reason": "…"}` parsed cleanly with every field `None`. The diff read as a declaration; the run ignored it. It now errors and names the field.

---

## D. One credentials file, one reader

`main` had **two** loaders reading **two** shapes: the harness read proto-native entries, while the `grpc-server` tests read a legacy `connector_account_details` wrapper and rebuilt `ConnectorAuthType` + `x-api-key`/`x-key1`/`x-auth` headers.

**11. New crate `crates/internal/connector-creds`** — one public fn, `connector_config_header(path, connector) -> Result<String>`, producing `{"config":{"Stripe":{"api_key":"sk_…"}}}`. It does exactly three things: reject the legacy shape by name, strip `metadata`, unwrap `{"value": …}`, then wrap under the PascalCase oneof variant.

- The `{"value": …}` unwrap is a **pre-pass over `serde_json::Value`, not a `Deserialize` impl**, and has to be: `build.rs` maps `.types.SecretString` → `hyperswitch_masking::Secret<String>`, whose `Deserialize` delegates to bare `String`, and the orphan rule forbids implementing `Deserialize` for it here. The alternatives are `deserialize_with` on all **713** `SecretString` fields, re-enumerated whenever one is added, or an `extern_path` newtype that changes all 713 public types.
- **Dropped-field check:** the generated config messages have no `deny_unknown_fields`, so a *misspelled* field is silently discarded and the connector receives an empty config — arriving as an auth failure that looks like a connector defect. Field names differ per connector (`authorizedotnet`: `name`/`transaction_key`; `paysafe`: `username`/`password`/`account_id`), so that is the likelier mistake. The parsed config is round-tripped and the offending path named. It **warns** rather than fails: the sweep skips a connector whose credentials do not load, so rejecting would make it disappear from the report, and an absent connector reads as nothing to see.
- Carries a **TODO to delete the crate** once the credentials artifact is stored in header shape; the transformation is the only part worth sharing, and with a passthrough file each caller reads it in three lines.

**12. `harness/credentials.rs`** — 181 lines of duplicated logic deleted; delegates to the crate.

**13. Eight `grpc-server` test files migrated** (`authorizedotnet`, `cashtocode`, `helcim`, `novalnet`, `payload`, `paysafe`, `stripe`, `xendit`) from `x-api-key`/`x-auth` to `x-connector-config` — the path a real caller uses, and the one **97** connector transformers consume via `TryFrom<&ConnectorSpecificConfig>`.

**14. `credential_utils.rs`: 480 → 59 lines**, and it **panics** rather than returning `Err` when the credentials file exists. `Err` invites `.ok()` + an early `return`, which Rust records as a passing test. `Result` is kept so a local checkout with no file still skips.

**15. Four Paysafe tests** reported PASS in **~15ms** for months, never issuing a request: the old loader could not produce a `BodyKey` auth type, so each took its "credentials unavailable" early return and counted as coverage. Reading credentials correctly made them run for the first time — and they fail, because the flow needs a payment handle the tests never obtain. `#[ignore]`d with that reason, so the gap is visible rather than counted as coverage.

**16. Both consumers now read the same proto-native credentials artifact** — the one the report pipeline already used. Two artifacts meant a rotated secret had to be applied twice, and applying it once surfaced as a live-sandbox auth failure that read like a connector defect.

---

## E. `check_connector_specs` hardening

**17. Reads both declaration forms** — `create_all_prerequisites!` *and* hand-written `impl ConnectorIntegrationV2<Flow, …>`. Many connectors use both at once (the macro for payment flows, hand-written impls for access-token and authentication flows). Reading only the macro let a connector implement a flow with no `specs.json` entry and still pass.

**18. `OUT_OF_SCOPE_FLOWS`, and unknown flows now fail.** The previous `_ => None` silently skipped any unmapped flow — which is how **`VoidPC` came to be implemented by 14 connectors and certified by none**. A flow now opts out only by decision; anything in neither list fails and prints connector + flow.

**19. `no_macro` split** into `no_flows_declared` (SKIP) and `unverifiable` (**FAIL** — an unreadable connector source is not a pass).

**20. Phase 2b: webhook coverage** — a connector declaring `EventService/HandleEvent` must ship a `webhook_payload.json`. No heuristic: its own declaration is the trigger.

**21. `flow_to_suites` fixes** — `GetConnectorCustomer` was mapped to `CustomerService/Create` (wrong) → `CustomerService/Get`; `PaymentMethodEligibility` was missing entirely.

---

## F. What the hardening then exposed — all real gaps

**22. Twelve connectors' `supported_suites` were incomplete** — `braintree` (Authorize), `cybersource`, `hipay`, `itaubank` (declared *nothing*), `kount`, `paypal`, `razorpay`, `razorpayv2`, `redsys`, `tamara`, `worldpayvantiv`, and `qwikcilver` (Eligibility). Each implements the flow; none declared the suite; so none were ever certified for it.

**23. `phone_number: "auto_generate"` → `{"value": "auto_generate"}`** across ~25 scenarios — it is a `SecretString`, and the bare string never matched the proto schema.

**24. `connector_customer_id: ""` removed** everywhere it appeared (customer, state, and `RecurringPaymentService/Charge`'s root). An empty string is not "unset", and connectors reading it got an empty ID.

**25. `PaymentMethodService/Eligibility` scenarios were fabricated** — they used `CARD`/`WALLET` (not `PaymentMethodType` members) plus nonexistent `country_code`/`currency`. Now `CREDIT`/`GOOGLE_PAY`, `country`, and the required `order_details`.

**26. `PaymentService/Authorize` suite_spec: the `PaymentMethodService/Tokenize` dependency dropped.** No Authorize scenario in the suite uses a token; `TokenAuthorize` has the correct wiring (`connector_token.value`). It was a wasted live call for **10 connectors**, and its context injection landed nowhere.

**27. phonepe webhook support** — a `SignatureContext` struct replaces the bare `timestamp: Option<i64>` argument. phonepe's `X-VERIFY` is `sha256(inner_base64_body + api_path + salt) + "###" + key_index` — it hashes the *inner* base64 `response` field, not the outer body. Per-connector request-shape knowledge is confined to `signature_payload_bytes` / `signature_context`, so `apply_webhook_payload_overrides` never branches on connector name. Plus new `phonepe/webhook_payload.json`, `boost/webhook_payload.json` and `boost/override.json` (boost genuinely has no source verification, so its fixture asserts `source_verified: false` rather than overriding the global expectation).

---

## G. Defects found in the gate's own first real run, and fixed

The first live run of the gate reported *"No certification regression attributable to this PR"* — on zero evidence. Three separate defects, all now fixed:

**a. `run_connector` captured the test runner's own stdout.** `failures=$(run_connector …)` captured everything `./scripts/run-tests` printed, so every line of the run's output became a bogus failure target — ~120 of them, each producing a `jq` parse error and a re-check invoked with an empty connector name. Worse, the real scenario detail never reached the log. The run's output now goes to stderr.

**b. Arbitration never ran.** The workflow checks out at `fetch-depth: 1`, so the merge base was not in the checkout and every confirmed failure fell through to `NOT_ATTRIBUTABLE`. `prepare_base` now does a targeted `git fetch --depth=1 origin "$BASE_SHA"` first — cheaper than deepening history on every run.

**c. The report file was read with the wrong shape.** `report.json` is `{"runs": [...]}`, an object. The filter read it as a bare array, so `.[]?` yielded the `runs` array itself, `select(.is_dependency …)` errored on an array, and `|| true` swallowed the error. Every failure silently degraded to whole-connector granularity — *"the run failed with status 1 and named no scenario"* — while 27 stripe and 39 fiserv scenario failures sat in the report unread. A whole connector almost never passes at the merge base, so **no regression could ever have been proven again**: the gate would have stayed permanently unable to block. Fixed to `(.runs // .)[]?` in both filters.

**d. The self-tests could not have caught any of these**, because the stub runner was more convenient than the real thing: it wrote nothing to stdout, and it wrote a bare array rather than `{"runs": …}`. Both corrected. Correcting the stub alone was still not enough for (c) — the whole-connector fallback produces the *same exit codes*, and the suite only asserted exit codes. A ninth case now asserts **granularity**: the failure line must name the scenario (`stripe (S / sc)`), not the connector (`(- / -)`). With the fixes 9/9; without them 8/9 and 4/8 respectively.

The gate failed **open** throughout, never wrongly blocking anyone — but it also proved nothing, which is the failure mode we said must not be silent.

---

## Net-zero — in the commit history, not in the final diff

`--skip-dependencies`, `verified_scenarios`, `has_live_creds`, `live-connectors.json`, `select-certification-targets.sh`, the failure-class classifier, and the oneof injection guard were all built and then removed. Grep confirms zero references remain.

---

## Open items

1. **Stripe's failures at HEAD are real and unexplained.** The first end-to-end run gave 133 passed / 27 failed for stripe and 19 / 39 for fiserv, against 48 / 130 and 0 / 58 at the merge base — so the PR roughly triples stripe's pass rate, and every remaining failure predates it on the evidence so far. The clusters worth reading: `EventService/HandleEvent` 0/4, `IncrementalAuthorization` 1/5, `RecurringPaymentService/Charge` 2/6, `Authorize` 35/43. Per-scenario arbitration can only now answer which, if any, this PR caused.
2. **The blocking direction has never occurred on a real runner** — a scenario that fails at HEAD and passes at the base. Stubs cover it; reality has not produced one.
3. **`CustomerService/Create` is bare in all 14 declaring suites** — 57 wasted live calls per stripe sweep. Needs a domain answer on whether connectors need a customer to exist, not a code change.
5. **Nightly baseline gate** (post-merge): must treat a connector *vanishing* from the report as a regression.
6. **Reshape the credentials artifact** to header shape, then delete `connector-creds`.

gopikrishna000:
**Full change inventory:** #2140

## Why

A `specs.json` entry says a connector *claims* to support a flow. Nothing today says the flow still works — a connector change merges on compile and unit tests, and no PR ever calls the connector.

This makes certification a merge-blocking behaviour gate: selected scenarios run against live connector sandboxes on every relevant PR, and a failure blocks the merge **only when the PR is shown to have caused it**.

Three constraints shaped it:

- **A gate that blocks must be trustworthy.** One connector's sandbox outage must not block an unrelated author.
- **A gate that runs on every PR must be cheap.** It cannot add meaningful wall-clock to the common path.
- **Proof must not be self-granted.** A new connector should not declare itself verified with nothing behind it.

## Attributing a failure

The core of the change. A failing scenario is not assumed to be the PR's fault:

1. Run at `HEAD`.
2. On failure, re-run at `HEAD`. Passes → a flake; reported, not blocking.
3. Still failing → check out the merge base, rebuild, run the same scenario there.
   - Fails there too → **not attributable**. Reported with both errors; does not block.
   - Passes there → **the PR breaks it**. Blocks.

A controlled comparison, not a heuristic: same scenario, same credentials, same sandbox, minutes apart, only the code differing. A sandbox outage fails on both sides and blocks nobody, which also removes the need for a quarantine list.

Nothing else is compared. An earlier revision classified each failure as an outage or a content failure and compared the two sides — a build could then fail on a judgement the data does not support, since two failures that look alike may be unrelated. The only question now is whether the scenario passed at the base, which is an observed fact. Every other outcome is reported with both error messages for a reviewer to judge.

The trade is deliberate: a regression introduced while a sandbox is down will merge. The nightly gate covers that.

## What gets certified

CI decides which **connectors** and nothing more. What each one then runs — suites, scenarios, payment methods — the test framework derives from that connector's own `specs.json`, exactly as a local or nightly run does.

- Connector-only changes certify the connectors they touch, unless the connector is listed in `alpha_connectors.json` — the connectors CI has no credentials for.
- Shared changes certify the connectors marked `live_in_production` — the widest blast radius; the full sweep belongs to the nightly.
- Changes to certification's own inputs (`connector-creds`, the test binaries, `scripts/`, `config/`, `.github/scripts/`) certify the same production set. Without this a PR editing only the gate itself would certify nothing and go green.

`alpha_connectors.json` is one file rather than a flag per connector, so how much is unproven is visible in one place: 101 of 103 today. A connector leaves the list when its credentials reach CI, and the file carries a TODO saying so.

Certification previously rebuilt the workspace to make a call lasting under two seconds. Binaries are now built once and reused (`run-tests --no-build`).

## New connectors

Arbitration cannot judge them — a new connector does not exist at the merge base, so every failure is attributable to the PR that introduced it. They are gated separately: a non-empty `supported_suites`, and either credentials in CI — which runs everything declared against the sandbox — or an entry in `alpha_connectors.json` with a reason, surfaced in the job summary and a sticky PR comment rather than only in the logs. Enforced only for connectors absent at the base commit, so the rule needs no exception list.

## Credentials

Two encrypted artifacts were maintained in parallel, because the `grpc-server` tests parsed a legacy shape while the report pipeline read the proto-native one. A rotated secret had to be applied twice, and applying it once surfaced as a live-sandbox auth failure that read like a connector defect.

Both consumers now read the proto-native artifact through one crate, `crates/internal/connector-creds`, and the tests send the entry as `x-connector-config` — the header a real caller uses, and the one 97 connector transformers consume via `TryFrom<&ConnectorSpecificConfig>`. `credential_utils` drops from 480 lines to 59. The read validates against `ConnectorSpecificConfig` before the call, so a bad entry names itself and its connector locally, and the legacy shape is rejected by name so pointing at the wrong artifact is diagnosed as such.

One check is deliberately not fatal. Deserializing catches a field of the wrong *type* but not one of the wrong *name*: unknown keys are dropped, the connector receives an empty config, and it arrives later as an authentication failure. Round-tripping names the offending path — as a warning, because the sweep skips any connector whose credentials fail to load, and an absent connector reads as nothing to see.

## What this exposed

Hardening the checks turned up gaps that already existed:

- **Twelve connectors implement flows their `specs.json` never declared**, so no scenario was ever required for them — `braintree` (Authorize), `worldpayvantiv` (Capture, Refund, Void, RSync), `razorpayv2`, `razorpay`, `redsys`, and single flows on `cybersource`, `hipay`, `kount`, `paypal`, `itaubank`, `qwikcilver`. Corrected here. The specs checker now reads hand-written `ConnectorIntegrationV2` impls as well as the `create_all_prerequisites!` macro, and a flow in neither the suite mapping nor the out-of-scope list now fails instead of being skipped — which is how `VoidPC` came to be implemented by 14 connectors and certified by none.
- **Four `paysafe` tests reported a pass in ~15ms** in every previous run — not enough time to reach a sandbox. The old loader could not produce their auth type, so each took its "credentials unavailable" early return and counted as coverage. Reading credentials correctly made them run for the first time, and they fail: paysafe needs a handle from `PaymentMethodService.Tokenize` and the tests send a raw card. `#[ignore]`d with that reason — a visible gap can be closed; one counted as a pass cannot.
- **`test_ucs` exited zero for a named scenario that never ran.** A scenario the harness declined to run counted as skipped, so a gate keyed off the exit code could report proof for a call that never reached the connector. The exit status now depends on what was asked for: a sweep may skip everything, a named scenario must have run.
- **`PaymentService/Authorize` depended on `PaymentMethodService/Tokenize`** for 10 connectors, though no Authorize scenario uses a token — a wasted live call whose context injection landed nowhere. Dropped.
- Connectors declaring webhook support must now ship a `webhook_payload.json`; `boost` was missing one.

Also: `grpcurl` pinned and checksum-verified; the gRPC server shutdown bounded, after an unbounded `wait` held a job open for two hours past its passing scenario; missing credentials now fail only the steps that need them, rather than reddening CI for PRs needing none.

## Verification

Local: self-tests for selection, arbitration and the new-connector gate (9/9, 8/8, 7/7) over stubbed outcomes covering pass, flake, failure on both sides, PR-caused regression, timeout, missing merge base, and per-scenario granularity; the full specs check; clippy and the workspace test suite.

On a real runner, all four phases — certify, re-check, prepare merge base, arbitrate — have now executed end to end. Stripe runs 133 passed / 27 failed at HEAD against 48 / 130 at the merge base; fiserv 19 / 39 against 0 / 58. The scenario corrections in this PR account for the difference.

**Not yet proven in the blocking direction.** Getting here took four defects in the gate itself, fixed across `a4cb0000c` and `2cc04ffbb` and detailed in #2140 — the worst being that failures were degrading to whole-connector granularity, which would have left the gate permanently unable to prove a regression. It failed *open* throughout, so nothing was wrongly blocked, but a scenario that passes at the base and fails at HEAD has still never been observed on a runner. **That is the blocker for merge.**

## Follow-ups

1. Drive the attribution paths end to end on a real runner, including the new-connector path.
2. Maintainer-approval gate for connectors merged without live credentials.
3. Validate the credentials artifact itself — presence and freshness per certified connector. Stale secrets currently surface only as a connector-side 401 mid-run.
4. Nightly baseline gate, which must treat a connector *vanishing* from the report as a regression.
5. Reshape the credentials artifact to the header shape, then delete `connector-creds`.


XyneSpaces:
Review findings (aggregated; inline placement not available via this tooling):

[should-fix] `.github/workflows/ci.yml:508`
The credential-setup step now `exit 1`s if the decrypted creds file is missing, turning the previous graceful fallback into a hard failure for every `RUN_TESTS=true` run. S3/GPG misconfiguration or transient decryption failures will break the whole `test` job. Move the fail-fast behavior into the new certification step so the generic test job can still degrade safely.

[should-fix] `.github/workflows/ci.yml:571-577`
grpcurl is downloaded from a hard-coded GitHub release (`1.9.3`, `linux_x86_64`) with no checksum verification. This breaks on non-x86_64 runners and supplies no integrity guarantee. Pin via the runner's package manager or verify the archive SHA256 and gate on `runner.arch`.

[should-fix] `.github/workflows/ci.yml:585-592`
The Stripe certification run passes `--skip-dependencies` for a `PaymentService/Authorize` scenario; if the fixture relies on prerequisite context (e.g., a created customer), the certification may issue an invalid request or report a misleading pass/fail. Either remove `--skip-dependencies` or confirm `no3ds_auto_capture_credit_card` has no required prerequisites.

XyneSpaces:
[should-fix] CI downloads grpcurl release tarball without checksum verification

The new "Install grpcurl" step in `.github/workflows/ci.yml` curls `grpcurl_${GRPCURL_VERSION}_linux_x86_64.tar.gz` directly from GitHub Releases and extracts it without verifying a SHA-256 or signature. Add a pinned checksum comparison (or use a verified GitHub Action) before extracting.

XyneSpaces:
[should-fix] Nested credential loader selects first match with undefined ordering

`extract_connector_block` iterates `obj.values()` and returns the first nested object containing `connector_account_details`. If the credentials file has multiple environments, `serde_json::Map`'s iteration order determines which account is used, which can run certification against the wrong environment. Explicitly resolve the intended environment key and fail if more than one candidate is present.

AmitsinghTanwar007:
## Review: Test trigger scope needs rethinking

The current approach intentionally skips certification when core/proto/shared files are touched:

> "Firing this per-PR gate on every core/proto change would tax every unrelated PR"

I think this is backwards — **core/proto changes are the most dangerous** because they can silently break multiple connectors at once. A connector-specific change has limited blast radius, but a change to `router_data.rs`, `domain_types`, `external-services`, or proto definitions can break every connector.

### Suggested approach (3 tiers):

**Tier 1 — Core/Proto changes touched**
→ Run tests for **all certified/hot/prod connectors**. These are high-value connectors you can't afford to break silently. Yes it adds CI time, but that's the point — catching regressions before merge.

**Tier 2 — Connector-specific files touched**
→ Run tests for **that connector only**. If someone changes `stripe/transformers.rs`, test Stripe.

**Tier 3 — New connector added**
→ Run its declared `verified_scenarios` as a one-time proof gate (this part is already in the PR).

The "tax" of running a few sandbox calls with retries on core changes is minimal compared to the cost of shipping a broken core change that silently breaks production connectors. The Jenkins nightly job catches it eventually, but by then the damage is already in `main`.

Shubhodip900:
What This PR Does

This PR introduces a live-sandbox certification gate that blocks a PR merge if it can be proven to cause a connector regression. The core design is a 3-pass system in certify-connectors.sh:

1. Pass 1 — Run every selected connector's scenarios at HEAD
2. Pass 2 — Re-run failures to filter flakes
3. Pass 3 — Check out the merge base, rebuild, re-run the confirmed failures. Only if a scenario passed at the base and fails at HEAD does it block the merge.

The PR also introduces a new connector-creds crate that consolidates credential loading into a single place (replacing 480 lines of duplicated logic), fixes 12 connectors whose specs.json was missing flows they actually implemented, and exposes 4 Paysafe tests that had been silently "passing" in 15ms by never issuing a request.

---
Issues

1. The blocking direction has never been exercised on a runner — this is the stated merge blocker.

The PR description is honest about it:

▎ "A scenario that passes at the base and fails at HEAD has still never been observed on a runner. That is the blocker for merge."

This is the most important thing to resolve before approving. The entire value of the gate depends on Pass 3 working correctly end-to-end. The author acknowledges four bugs in the gate were found and fixed during development (a4cb0000c, 2cc04ffbb), which raises the question: could there be another one that only manifests on the blocking path?

---
2. build_binaries truncates to tail -5, which may be too little to debug a failure.

build_binaries() {
  cargo build --workspace --bins 2>&1 | tail -5
  return "${PIPESTATUS[0]}"
}

A Rust build failure can have the actual error buried many lines above the last 5. If the merge-base doesn't compile, arbitration degrades to "inconclusive" — which is the right fallback — but diagnosing why it failed is very hard with only 5 lines. Consider tail -30 or redirecting stderr to a temp file and uploading it as an artifact.

---
3. verify-new-connectors.sh starts a fresh gRPC server per connector, while certify-connectors.sh shares one.

certify-connectors.sh calls grpc_server_start once before its loop and passes --no-server to each run_connector. verify-new-connectors.sh omits --no-server, meaning run-tests starts and stops a server for every new connector:

# verify-new-connectors.sh
local args=(--skip-setup --no-build --connector "${name}" --interface grpc --report)
# no --no-server here

If multiple new connectors are added in a single PR this is slow. More importantly, if the gRPC port from a previous step hasn't fully released, the next server start fails with a bind error that reads as a test failure. Consider starting the server once in verify-new-connectors.sh as well.

---
4. pascal_connector_name only capitalizes the first character, which will silently produce the wrong variant for any future mul

fn pascal_connector_name(connector: &str) -> String {
    let mut chars = connector.chars();
    match chars.next() {
        Some(first) => first.to_uppercase().collect::<String>() + chars.as_str(),
        ...
    }
}

worldpayvantiv → Worldpayvantiv. Today this is fine because all connector names map 1:1 to their proto variant (all lowercase + first-cap). But the moment a connector name differs from its proto variant (e.g., underscores, compound words), this silently produces a wrong variant name, which then fails deserialization and becomes a credential-load error — not caught until CI runs. Worth at least a comment or an assertion.

---
5. Dropped fields during credential loading are a warning, not an error — the connector still gets called with an empty config.

if !dropped.is_empty() {
    eprintln!(
        "[credentials] '{connector}': {} ignored by the config message ({}). ...",
        ...
    );
}
Ok(wrapped.to_string()) // still returns Ok

A field with the wrong name (e.g., api_key when the proto expects apiKey) is silently dropped, the connector receives an empty config, and the test fails at the connector level with an auth error that looks like a connector defect. The warning in the log is only visible if you know to look for it. Consider making this a hard error in the grpc-server tests' connector_config_header wrapper (since those tests have a credentials file present), while keeping it a warning in the harness sweep path (where an absent connector reads as "nothing to see").

---
6. The NO_CREDS loop sets blocking=1 with an unnecessary iteration.

for x in ${NO_CREDS[@]+"${NO_CREDS[@]}"}; do
  blocking=1
done

This is the only blocking category where the loop body doesn't use $x. A simpler and more readable form: [[ ${#NO_CREDS[@]} -gt 0 ]] && blocking=1. The current form works, but it's inconsistent with how any other "set a flag if array is non-empty" check would be written.

---
7. The sticky PR comment always overwrites, even if the content is unchanged.

if (existing) {
  await github.rest.issues.updateComment({ owner, repo, comment_id: existing.id, body });
} else {
  await github.rest.issues.createComment({ owner, repo, issue_number, body });
}

Every CI re-run that finds unproven connectors will update the existing comment — producing a "last edited" notification even if nothing changed. Add a content check before updating: if (existing.body !== body) { ... update ... }.

---
Minor Notes

- The connector_config_header wrapper in credential_utils.rs uses assert! inside a Result-returning function to panic when the credentials file is present but the connector isn't loadable. This is intentional (to prevent the Paysafe-style silent-skip pattern) and documented, but it's unusual enough to warrant a second set of eyes to confirm the #[allow(clippy::panic_in_result_fn)] suppression is appropriate here.
- alpha_connectors.json carries a TODO saying 101 of 103 connectors are unproven. The cert gate is therefore only live for 2 connectors right now. The operational value grows as credentials are added — worth calling out in the PR that the gate's coverage is narrow at merge time.
- The grpcurl binary is pinned and checksum-verified — good practice.
- Security note: connector file names from fork PRs reach the shell through CONNECTOR_FILES env var, not direct interpolation — the injection fix is correctly applied.

---
Summary

The design is sound and the PR description is unusually thorough. The main blocker is the one the author already named: the regression-attribution path (Pass 3 failing the build) must be exercised on a real runner before merge. Items 2, 3, and 5 are the most likely to cause subtle false-negatives or hard-to-debug failures in production.

AmitsinghTanwar007 (dismissed):

Shubhodip900 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (approved):

Shubhodip900 (approved):

PR #2094, opened 2026-08-11
feat(logs): Euler-compatible log structure with structured JSON fields

### LINKED ISSUE #2044: dynamic keys in the logs
key mapping based on env

AmitsinghTanwar007:
## Summary

- **Unified log field transformations**: Configurable incoming/outgoing log field mappings via `[log.fields]` config, with support for static values and span source references (e.g., `"api_details.url" = { source = "request.url" }`)
- **Runtime enabled flag**: `log.fields.enabled` (default `false`) allows toggling log field application at runtime without rebuilding. Overridable per-request via `x-config-override` header
- **Structured JSON in log lines**: `request_body`, `response_body`, `request.headers`, `request.body`, `response.body`, and `response.headers` are now recorded as `serde_json::Value` directly into span storage via `record_json_fields_on_span()`, producing proper nested JSON instead of escaped strings
- **Docker + CI**: `log-transformations` feature added to Dockerfile build and CI test commands

## Key changes

- `record_json_fields_on_span()` helper in `common_utils::events` — safe single-lock pattern for writing structured JSON to span storage
- `CompiledLogFieldsConfig` with `enabled` flag propagated through `EventProcessingParams`
- `Maskable<String>` headers converted to JSON objects for structured logging
- Feature-gated with `log-transformations` — no impact when feature is off

## Test plan

- [x] `cargo check` / `cargo clippy` pass with and without `log-transformations` feature
- [x] `cargo fmt --check` passes
- [x] Tested `enabled = true` — all configured log fields appear in golden log lines as proper JSON
- [x] Tested `enabled = false` — no custom log fields appear
- [x] Per-request override via `x-config-override` header works
- [x] `request_body` and `response_body` appear as nested JSON objects (not escaped strings) in log output


gopikrishna000 (approved):

XyneSpaces:
:warning: On `masked_serialize` error, `response_body` falls back to `tracing::field::debug(res_ref)`, which can write unmasked secrets to logs. Remove the unmasked fallback or mask the response before recording.

:bulb: `serde_json::to_value(&evt.headers)` failure is silently ignored. If the headers are not masked-safe, they may be lost or exposed. Handle the error explicitly or use a masking serializer consistently.

JeevaRamu0104 (approved):

PR #2096, opened 2026-08-12
fix(connector): Add missing payment method type mapping and Stripe MIT parity fix

### LINKED ISSUE #2097: Add missing payment method type mapping  in PSync flow and Stripe MIT parity fix
Add missing payment method type mapping  in PSync flow and Stripe MIT parity fix

Shubhodip900:
## Description

Add missing payment method type mapping and Stripe MIT parity fix
### Additional Changes

## How did you test it?

Used `cargo build` to verify the build.


crates/integrations/connector-integration/src/connectors/stripe/transformers.rs
```diff
@@ -5354,35 +5354,31 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize + Ser
                 }
             });
 
-        // If the Split Payment Request in MIT mismatches with the metadata from CIT, throw an error
+        // If the Split Payment Request in MIT mismatches with the metadata from CIT, throw an error.
+        // `application_fees` is intentionally NOT compared: it can legitimately vary per
+        // transaction (Stripe allows the platform to set a different application fee on each
+        // charge), while charge_type / transfer_account_id / on_behalf_of must remain consistent
+        // with the ones used at mandate creation.
```
pixincreate, line 5361: can this be confirmed with the docs, once? asking because, the same exists in the oss. if it is removed here, it might have to removed there too since that would cause a diff.

pixincreate: also, nit, this comment can be removed?

pixincreate (commented):
other than that, lgtm.

crates/integrations/connector-integration/src/connectors/stripe/transformers.rs
Shubhodip900: https://docs.stripe.com/connect/marketplace/tasks/app-fees
verified from the doc, it says application_fees can be different for different request.

Also this is not present in OSS there also it's removed

pixincreate (approved):

XyneSpaces:
:bulb: MIT split-payments parity change removes `application_fees` from the mismatch check, altering money-movement validation semantics. Include a test or spec that exercises this path so future changes don't regress the fee check.

:mag: New `PaymentMethodType` mappings are exhaustive at compile time, but no unit tests cover the new mappings. Add a round-trip test or field-probe scenario for the new method types.

gopikrishna000 (approved):

Vani-1107 (approved):

PR #2106, opened 2026-08-13
fix(connector): read calida shop name from connector specific config

### LINKED ISSUE #2119: [BUG] CALIDA Authorize call flowing due to missing required field
shop name field is missing.

pixincreate:
## Description
Calida authorize requests require a `shop_name` field, which UCS previously read via `connector_feature_data`. When that field was absent, the request transformation failed with a missing-required-field error, causing shadow-mode diffs for bluecode wallet payments.

This PR:
- Adds `shop_name` to the `CalidaConfig` gRPC message (payment.proto) and to `ConnectorSpecificConfig::Calida` in domain types, so the shop name travels with the connector specific config (the same mechanism used for Mifinity brand fields).
- Updates the calida request transformer to read `shop_name` exclusively from `ConnectorSpecificConfig::Calida`, returning a clear `MissingRequiredField` error when it is absent. There is intentionally no fallback path: the old `connector_feature_data` read was the broken production behavior being fixed (hyperswitch does not populate it for regular authorizes).

## Motivation and Context
Bluecode payments through the calida connector fail in shadow mode because the shop name is never populated over gRPC for regular authorize flows; hyperswitch only populated `connector_feature_data` for specific flows. The connector specific config is the canonical transport for per-merchant connector metadata (see the Mifinity pattern), so the shop name belongs there. The hyperswitch companion change (populating `ConnectorSpecificConfig::Calida.shop_name` from MCA metadata) is required for this fix to take effect in production.

### Additional Changes

This PR adds an optional field to the `CalidaConfig` gRPC message and the corresponding domain type. The change is backward compatible at the wire level: the new field is optional.

## How did you test it?
- `cargo check -p domain_types -p connector-integration` passes.
- `cargo test -p connector-integration calida` passes: 3 tests (valid request with `shop_name` in connector specific config, missing-fields error case asserting `MissingRequiredField{shop_name}`).
- End-to-end through UCS: calida authorize with `bluecode_redirect` passed against the calida sandbox using the test credentials (shop_name resolved from connector specific config).


crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -4798,6 +4798,7 @@ message BamboraapacConfig {
 message CalidaConfig {
   SecretString api_key = 1;
   optional string base_url = 50;
+  optional SecretString shop_name = 51;
```
Shubhodip900:
```suggestion
SecretString api_key = 1;
optional SecretString shop_name = 2;
optional string base_url = 50;
```

JeevaRamu0104 (approved):

deepanshu-iiitu (approved):

Shubhodip900 (approved):

PR #2126, opened 2026-08-13
fix(core): allow Authorize without payment_method on the 3DS return leg

shuklatushar226:
## Problem

`PaymentService::authorize` required `payment_method` unconditionally:

```rust
proto_payload.payment_method.clone().ok_or(MissingRequiredField { field_name: "payment_method" })?
```

Connectors that finalise 3-D Secure **through the Authorize flow** — Ilixium, NMI, Airwallex,
Nexixpay and others that key off `request.redirect_response.is_some()` — are re-invoked after
the ACS posts back to `TermUrl`. At that point the caller no longer holds the raw instrument, so
Hyperswitch sends `payment_method_data: None` and omits the proto field (it is optional in both
the proto and Hyperswitch's `CompleteAuthorizeData`).

UCS therefore rejected the completion call with `InvalidArgument: Missing required field:
payment_method` **before any connector code ran**, making 3DS completion impossible for every
connector that uses this pattern.

## Fix

`payment_method` is now required only when the call is *not* a redirect-return leg. The leg is
identified by `redirection_response`, the same signal the connectors themselves already use.
A plain authorize with no `payment_method` is still rejected exactly as before.

`PaymentsAuthorizeData::payment_method_data` is not optional, so the absent instrument is
modelled explicitly as a new `PaymentMethodData::NoInstrumentAfterRedirect` variant rather than
substituted with a default (e.g. `MandatePayment`) that a connector could mistake for a payable
instrument.

Adding the variant is compiler-enforced: all 51 existing matches on `PaymentMethodData` now
handle it through their **existing unsupported-payment-method rejection arm**, so no connector's
behaviour changes and a connector that later wants to accept this leg has to opt in deliberately.
That accounts for the file count — the connector diffs are one added pattern alternative each.

`PaymentsAuthorizeData::foreign_try_from` derived `payment_method_type` from the same field and
had the same unconditional gate; it now returns `None` for the instrument-less leg only (matched
on `NoInstrumentAfterRedirect`), and errors as before otherwise.

## Verification

Driven end to end against a mock Ilixium server (Hyperswitch → UCS → mock):

- **3DS**: `POST /payments` → `requires_customer_action` → ACS challenge → complete-authorize →
  mock received `POST /platform/ili/direct/threedcomplete` (merchant digest verified) → HS
  payment status `succeeded`. This call was rejected with `InvalidArgument` before the fix.
- **Direct gRPC**: `PaymentService/Authorize` with `payment_method` omitted and
  `redirection_response` present now reaches the connector and returns `CHARGED`.
- **No-3DS regression**: unchanged, payment `succeeded`.
- **Validation preserved**: `Authorize` with no `payment_method` and no `redirection_response`
  still fails with `InvalidArgument: Missing required field: payment_method`.

`cargo check --workspace --all-targets`, `cargo fmt --all` and `cargo clippy` on the touched
crates are clean.

## Context

Found while integrating the Ilixium connector (#2125). Raised separately because it is a
core-flow change that unblocks 3DS completion for every UCS connector using this pattern, not
just Ilixium.

### DIFF
--- a/crates/grpc-server/grpc-server/src/server/payments.rs
+++ b/crates/grpc-server/grpc-server/src/server/payments.rs
@@ -889,14 +889,25 @@ impl PaymentService for Payments {
                 // Convert proto request to intermediate type
                 let payload: AuthorizationRequest = proto_payload.clone().into();
 
-                let payment_method_data_action = PaymentMethodDataAction::get_payment_method_data_action(proto_payload.payment_method.clone().ok_or(ucs_env::error::GrpcError::from(IntegrationError::MissingRequiredField { field_name: "payment_method", context: domain_types::errors::IntegrationErrorContext::default() }))?)
-                    .map_err(|err| {
-                        tracing::error!("PAYMENT_AUTHORIZE_FLOW: failed to get payment method data action - error: {:?}", err);
-                        ucs_env::error::GrpcError::from(IntegrationError::InvalidDataFormat { field_name: "payment_method", context: domain_types::errors::IntegrationErrorContext::default() })
-                    })?;
+                // `payment_method` is required for an ordinary authorization, but not for the
+                // second leg of a redirect-based one: connectors that finalise 3DS through
+                // Authorize are re-invoked after the ACS posts back to `TermUrl`, and the
+                // caller no longer holds the raw instrument at that point. That leg is
+                // identified by `redirection_response`, exactly as the connectors themselves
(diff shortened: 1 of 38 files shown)


shuklatushar226:
Superseded — folding this into the single Ilixium PR #2125 rather than raising a separate PR. The identical change is now on `feat/grace-Illexium` as commit 664e8abf3, kept as its own commit so it stays reviewable in isolation.

PR #2129, opened 2026-08-13
ci(sdk): stage placeholder credentials when secrets are unavailable

Utkal059:
Fork pull requests cannot pass `SDK Tests`, and the reason is one line in this workflow rather than anything in `sdk/`.

A PR from a fork receives no `secrets.CONNECTOR_SPECIFIC_AUTH`, so "Create connector credentials" writes nothing and logs *"running in dry-run mode"*. Nothing downstream implements that mode. Every SDK harness resolves `creds.json`, and the four disagree on what a missing file means:

```
══ gRPC Results Summary ══
  ✗ Rust (exit 2)      Configuration error: Credentials file 'creds.json' not found
  ✓ JavaScript         loadCreds() returns {} → placeholder api_key
  ✓ Python             same
  ✗ Kotlin (exit 2)    IllegalArgumentException: Credentials file not found: creds.json
```

`Run SDK gRPC Tests` is the first test step, so the job dies there — and the two steps behind it need the file just as much:

- `Run SDK FFI Tests`: all four harnesses abort. `run_tests` panics with `Credentials file not found` (`sdk/rust/smoke-test/src/main.rs`), `loadCredentials` throws (`sdk/java/smoke-test/src/main/kotlin/SmokeTest.kt`), and the JavaScript and Python FFI harnesses raise as well — only their *gRPC* harnesses degrade.
- `Run SDK Mock Tests`: `scripts/run_smoke_tests_parallel.py:499` copies `repo_root/creds.json` unconditionally.

So three of the job's steps are unreachable without secrets, and the message a contributor sees points at their own diff.

## Change

Copy `creds_dummy.json` — the file the mock tests already use — when the secret is absent, so all four harnesses take one path.

```yaml
cp creds_dummy.json creds.json
```

**The values stay `<REPLACE_WITH_YOUR_VALUE>` rather than being scrubbed to `"placeholder"`**, and that detail is the whole fix. `has_valid_credentials()` reads a recognised placeholder as *nothing configured*, so every harness would skip every connector, and all four then agree on this:

```rust
if passed == 0 && skipped > 0 { return 1; }   // "All tests skipped (no valid credentials found)"
```

A scrubbed file turns four crashes into four green-looking skips that still exit 1. The unscrubbed dummy makes the flows actually run: the connector rejects the key, each harness classifies that as `skipped (connector error)` with the connector itself passing, exactly as the JavaScript and Python gRPC harnesses do on fork PRs today. Packaging, native-library loading, transport and request building stay covered.

Runs that hold the secret are untouched — the live branch is unchanged, and the placeholder branch is reachable only when `CONNECTOR_SPECIFIC_AUTH` is empty.

## Verified

Loading the real predicates out of `sdk/python/smoke-test/test_smoke.py` (AST-extracted, so it is the shipped code, not a re-implementation):

```
load_credentials(missing)                 : FileNotFoundError — today's fork failure
load_credentials(after cp)                : OK, 72 connectors, stripe present
has_valid_credentials(stripe dummy entry) : True    → flows run, no all-skipped exit 1
is_placeholder('<REPLACE_WITH_YOUR_VALUE>'): False
is_placeholder('placeholder')             : True    → why scrubbing would break it
```

The Rust, Kotlin and JavaScript harnesses carry the same two predicates and the same `passed == 0 && skipped > 0` rule; the Rust and Kotlin gRPC harnesses classify a non-transport error as `skipped (connector error)` and leave the connector `passed`, so they land where JavaScript and Python already do.

`bash -n` on the changed step, and the workflow parses.

This PR is its own test: it changes `.github/**`, which sets `changes.ci`, which is one of the triggers for `sdk-test` — and being a fork PR, it runs the placeholder branch it adds.

## Not included

The harness divergence is real on its own — `make -C sdk test-grpc` without a `creds.json` gives a contributor ✓✓✗✗ locally too. Converging that is four code paths across two languages I cannot build here, and it is a separate concern from unblocking CI; happy to follow up if it is wanted.

One cosmetic leftover: `Run SDK FFI Tests` still branches on `[[ -f "creds.json" ]]` to choose its log line, which is now always true, so a fork run prints "with connector credentials". Say the word and I will make it read the mode instead.

Refs #2029

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -698,7 +698,15 @@ jobs:
             echo "${CONNECTOR_SPECIFIC_AUTH}" > creds.json
             echo "Connector credentials created"
           else
-            echo "No CONNECTOR_SPECIFIC_AUTH secret found, running in dry-run mode"
+            # A pull request from a fork receives no secrets, and the four smoke-test
+            # harnesses disagree about a missing creds.json — Rust and Kotlin abort,
+            # JavaScript and Python fall back to placeholders. Stage the dummy file
+            # the mock tests already use so all four take one path: the connector
+            # rejects the placeholder key and every harness records that as
+            # "skipped (connector error)", leaving packaging, native-library
+            # loading, transport and request building still covered.
+            cp creds_dummy.json creds.json
+            echo "No CONNECTOR_SPECIFIC_AUTH secret found, using placeholder credentials from creds_dummy.json"
           fi
 
       - name: Run SDK gRPC Tests


Utkal059:
Worth recording why the two obvious alternatives don't hold, since staging a credentials file isn't the first thing that comes to mind for this failure.

**`make ... --dry-run || true`.** `--dry-run` there is GNU make's `-n`: it prints the recipes and runs none of them, so the step would pass without executing a single test. `|| true` then makes the step incapable of failing at all. `SDK Tests` is a required check — this would leave it green permanently.

**Skipping the step when credentials are absent.** That drops SDK gRPC coverage for every fork PR, including the run #2029 leans on to show its resolver change works end to end (stage native → pack → install → load). Fork PRs are where most SDK contributions come from; they'd be the ones losing the check.

**"The harnesses already support dry-run and mock."** Only the FFI ones. `SmokeTest.kt:503-504` is the `Args` data class of the *FFI* harness; the step that fails is the gRPC one, and neither gRPC harness accepts those flags:

```
GrpcSmokeTest.kt:150-153    --creds-file  --connectors  --all        --help
grpc_smoke_test.rs:198-210  --creds-file  --endpoint    --connector  --help
```

Even in the FFI harness the modes don't help: `runTests` installs the mock intercept and then calls `loadCredentials`, so `--mock` throws on the missing file just the same. Which is really the point — mock mode has always needed a credentials file, and `make test-package-mock` supplies one by copying `creds_dummy.json` (`sdk/java/Makefile:176`). This PR does that one level up so all four harnesses get it, instead of each deciding for itself what a missing file means.

The distinction that matters: this keeps the tests running and failing honestly. The alternatives get the check green by making it stop looking.

Utkal059:
Update after re-reading the latest run: the placeholder credentials do what they're meant to — all four gRPC harnesses now take one path and the gRPC step passes, with flows running and landing on `skipped (connector error)` rather than aborting on a missing file.

The job still fails, one step later. `Run SDK FFI Tests` gets through `sdk/rust` and then dies in `sdk/python`:

    FAILED: Invalid API Key provided: <REPLACE*************LUE>
    make[1]: *** [Makefile:165: test-package] Error 1

So the Python FFI harness treats an invalid key as a hard failure while its own gRPC harness treats it as a skip. That's the harness divergence I listed under "Not included" — it turns out to block the job rather than just being untidy, so this PR unblocks the first of the three steps rather than all of them.

Happy to extend it to converge that classification, or to keep this scoped to the gRPC step and file the FFI side separately — whichever you prefer. Flagging it rather than leaving the PR claiming more than it delivers.

PR #2122, opened 2026-08-13
feat(connector): [WORLDPAYWPG] wallets, stored credentials and recurring flows

pixincreate:
## Description

Worldpay XML was **card-only, one-off-only**: `PaymentMethodData::Wallet` was rejected outright, nothing marked an authorisation as part of a stored-credential agreement, and `SetupMandate` / `RepeatPayment` were both in the `not_implemented` list. This adds wallets and stored-credential recurring.

**Wallets**
- Add `PAYWITHGOOGLE-SSL`, `APPLEPAY-SSL` and `EMVCO_TOKEN-SSL` payment methods.
- Wallet data arriving **already decrypted** goes over `EMVCO_TOKEN-SSL` as a network token — the only shape Worldpay registers against a stored credential.
- Wallet data arriving **still encrypted** (Worldpay decrypts at its end) is forwarded on its wallet-specific element.
- Google Pay decrypted without a cryptogram (PAN_ONLY) falls back to `CARD-SSL`, so `cvc` is now `Option` — always `Some` for real cards, so card requests are unchanged.
- `customer_name` is carried on the Google Pay elements and omitted on the decrypted Apple Pay token, matching the connector's contract.

**Recurring**
- CIT orders carry `<createToken tokenScope="shopper">` alongside `storedCredentials usage="FIRST"`.
- Responses parse `<token><tokenDetails><paymentTokenID>` into a `MandateReference`, chained via `schemeResponse/transactionIdentifier`.
- MIT submits the token over `TOKEN-SSL` with `usage="USED"`, `merchantInitiatedReason` and `schemeTransactionIdentifier`.
- `SetupMandate` / `RepeatPayment` registered; `authenticatedShopperID` sent, which Worldpay requires for shopper-scoped tokens.

**Shared field-probe fixture fix (incidental)**

The Apple Pay probe stub decoded to a token with no `header`, which is mandatory in Apple's format, so any connector that parses the envelope rejected it. Correcting it flipped **datatrans** and **noon** from `error` to `supported` alongside worldpayxml. Google Pay's `PAYMENT_GATEWAY` token is gateway-specific, so `[wallet_token_overrides.<connector>]` was added for connectors that forward the raw envelope rather than a gateway-shaped one.

## Motivation and Context

Wallet payments and merchant-initiated recurring were unsupported on this connector, blocking any merchant that needs Apple Pay / Google Pay or subscription billing through Worldpay WPG.

[...]

PR #2130, opened 2026-08-14
fix(connector): [WORLDPAYWPG] fix PSync/RSync response parsing

pixincreate:
> **Stacked on #2122.** Base is `feat/worldpayxml-wallet-decrypt-recurring`; retarget to `main` once that merges.

## The bug

`PaymentService/Get` (PSync) and `RefundService/Get` (RSync) returned `RESPONSE_HANDLING_FAILED` for **every** worldpayxml order — including plain card orders with no token, so this is independent of the recurring work in #2122 and predates it.

The sync body was deserialized directly **into** `#[serde(untagged)] WorldpayxmlTransactionResponse`. Untagged forces serde's `deserialize_any`, whose buffered value loses XML sequence and text semantics, so the `Payment` variant could never match any payload:

- `balance: Option<Vec<_>>` — a single `<balance>` buffers as a map, not a 1-element sequence.
- `<lastEvent>CAPTURED</lastEvent>` buffers as `{"$text": "CAPTURED"}`, which a unit-variant enum cannot be built from.

`WorldpayxmlWebhookResponse`, whose fields were all `Option`, then matched unconditionally — it matches literally any input, including unrelated XML. Every real parse error was silently swallowed and resurfaced as a generic failure at `transformers.rs:1699`, naming neither the failing type nor the field.

Proven in a standalone repro on the same quick-xml version: even the Authorize response that parses fine today resolves to the `Webhook` variant when routed through that enum. Authorize/Capture/Void/Refund work only because they are typed directly.

## The fix

Stop routing XML through an untagged enum. `WorldpayxmlTransactionResponse` no longer derives `Deserialize`; a hand-written impl parses a concrete, element-driven body, wraps a `<paymentService>` reply into `Payment` manually, and only then falls back to the order-notification body. Concrete type first, enum as a manually-constructed carrier, no `deserialize_any` anywhere.

`WorldpayxmlWebhookResponse` gains a required `OrderCode` and a typed `PaymentStatus` — the all-`Option` shape is what made it match everything.

### Response hardening, folded in

- `LastEvent`: added `Settled`, `SentForAuthorisation`, `SentForFastRefund`, `RefundRequested`, `RefundedByMerchant`, `QueryRequired`, `CancelReceived`, `RefundReceived`, plus `#[serde(other)] Unknown`. Without this PSync would have broken again the moment an order settled.
- Status mappers take `previous_status` and have an explicit `Unknown` arm that logs a warning and retains the prior status instead of erroring or assuming a terminal state.
- Previously-mandatory response fields wrapped in `Option` (`AuthorisationId.@id`, `ResultCode.@description`, `Balance.@accountType`/`amount`, `PaymentMethodDetail.card`, `CardResponse.@type`).

`Expired` was **kept** even though the reference implementation lacks it — dropping it would silently reroute a real `EXPIRED` event into `Unknown`/retain-previous, a behaviour regression.

`payout_connectors/worldpayxml/transformers.rs` is touched only because its status mapper matches the same shared enum and would not otherwise compile.

`parse_last_event` is removed entirely — with `WorldpayxmlLastEvent` now deserialized as a typed field, the helper is dead code. (#2122 fixes an unsafe JSON interpolation in it for the interim.)

## Testing

Verified against the Worldpay sandbox. Clippy warning-free, no `unwrap`/`expect`, no suppressions, no test files.

| Check | Before | After |
|---|---|---|
| Authorize → PSync | `RESPONSE_HANDLING_FAILED` | `CHARGED`, amount returned |
| CIT order with `<token>` → PSync | `RESPONSE_HANDLING_FAILED` | `CHARGED` |
| Refund → RSync | `RESPONSE_HANDLING_FAILED` | `REFUND_PENDING` (correct — Worldpay had not yet reflected it) |
| Parent-branch Authorize / SetupRecurring / MIT Charge | pass | pass — no regression |

The payload that used to fail now parses, including the `balance` array and `<lastEvent>` that broke `deserialize_any`:

```xml
<orderStatus orderCode="..."><payment>
  <lastEvent>CAPTURED</lastEvent><AuthorisationId id="..."/>
  <paymentMethodDetail><card type="creditcard"/></paymentMethodDetail>
  <balance accountType="IN_PROCESS_CAPTURED"><amount value="1000" currencyCode="GBP" exponent="2"/></balance>
</payment></orderStatus>
```

An early PSync on a not-yet-available order returned Worldpay's `<error code="5">` shape and correctly yielded `PENDING` rather than a hard failure.

---

## Review follow-ups (status mapping hardening)

Beyond the parsing fix, this PR now narrows the `lastEvent` mappings to the set the connector actually reports, and makes them fallible instead of silently coercing:

- **Payout status** narrowed to the six FastAccess disbursement events (`PUSH_REQUESTED`, `PUSH_PENDING`, `PUSH_APPROVED`, `PUSH_REFUSED`, `ERROR`, `SETTLED_BY_MERCHANT`, `CANCEL_RECEIVED`); the refund family is not part of a disbursement lifecycle and mapping it was inventing semantics. Anything outside the set now errors.
- **Authorize / SetupMandate / Refund status mappers** are `Result`-returning and reject events outside each flow's lifecycle, rather than reporting them as `Failure`.
- `EXPIRED` is kept as an explicit terminal `Failure` in all three mappers — it is a routine outcome for an authorisation left uncaptured, so it must not fall into the error arm.
- `Unknown` continues to retain the previous status, so an unmodelled event never invents an outcome.

Also fixed while here:

- **PSync no longer downgrades a terminal status.** An `<orderStatus>` carrying an inquiry-level `<error>` with no `<payment>` used to force `Pending`, which could walk a `Charged` attempt backwards. It now preserves the existing status.
- **Auto-capture `AUTHORISED` reports `Charged`** rather than `Pending` (pre-existing on `main`; the attempt no longer needs a later sync to reach a terminal state).

### Test evidence

Sandbox merchant with tokenisation enabled. Card `4111111111111111`, GBP unless stated.

| Flow | Result |
|---|---|
| Authorize AUTOMATIC → PSync | `CHARGED` → `CHARGED`; wire `captureDelay="0"`, `action="SALE"` |
| Authorize MANUAL → Get → Capture → Get | `AUTHORIZED` → `AUTHORIZED` → `CAPTURE_INITIATED` → `AUTHORIZED` (Worldpay batch had not advanced `lastEvent`) |
| Void authorised order → Get | `VOID_INITIATED` → `AUTHORIZED` (same batch latency) |
| SetupRecurring (CIT), zero and non-zero amount | `CHARGED` with mandate reference + scheme transaction id |
| RecurringPaymentService/Charge (MIT) | `CHARGED`; wire carries `@merchantInitiatedReason` `RECURRING` and `INSTALMENT` for the respective `mit_category` |
| Declined SetupRecurring | `FAILURE` with `code "5"`, `message "REFUSED"` — the new failure branch surfaces the ISO 8583 detail |
| Refund → RSync | `REFUND_PENDING`, parsed cleanly through the now-fallible refund mapper |
| Empty billing address | Worldpay DTD rejection (`code 2`, incomplete `address` element) — the request is well-formed otherwise |
| JPY Authorize | wire `@exponent="0"` — exponent handling correct for a zero-decimal currency |

### Known gaps

- **`EXPIRED` and the mappers' reject arms are compile-verified only.** Worldpay cannot be driven into those states on demand.
- **`SETTLED` / `SETTLED_BY_MERCHANT`** are batch-driven overnight; the manual-capture and void tests above show `AUTHORIZED` because the sandbox batch had not run.
- **Payout (`FF_DISBURSE-SSL`) mappings are untested** — this merchant is a payments profile and FastAccess needs separate enablement. The payout mapper was rewritten here and has not been near a sandbox.
- **Apple Pay / Google Pay** are being verified separately across both the wallet-decryption and connector-decryption paths.

### Follow-ups, not addressed here

- **Authorize does not surface ISO 8583 decline detail.** A declined Authorize returns a bare `Failure` with no error object, while `SetupMandate` and `RepeatPayment` both surface `iso8583_return_code`. Same gap, most common decline path.
- **`build_error_response` cannot parse Worldpay errors.** `WorldpayxmlErrorResponse` is a single-variant untagged enum fed through a JSON parser, while Worldpay returns XML on non-2xx, so HTTP-level errors surface as a generic parse failure.

---

PSync

image

RSync

image


Shubhodip900 (dismissed):

Nithin1506200 (dismissed):

crates/integrations/connector-integration/src/payout_connectors/worldpayxml/transformers.rs
```diff
@@ -56,20 +56,34 @@ fn map_worldpayxml_payout_status(
...
+        | WorldpayxmlLastEvent::RefundReceived
+        | WorldpayxmlLastEvent::RefundRequested => common_enums::PayoutStatus::Pending,
         WorldpayxmlLastEvent::Cancelled => common_enums::PayoutStatus::Cancelled,
-        WorldpayxmlLastEvent::SentForRefund | WorldpayxmlLastEvent::Refunded => {
-            common_enums::PayoutStatus::Reversed
-        }
+        WorldpayxmlLastEvent::SentForRefund
+        | WorldpayxmlLastEvent::SentForFastRefund
+        | WorldpayxmlLastEvent::Refunded
+        | WorldpayxmlLastEvent::RefundedByMerchant => common_enums::PayoutStatus::Reversed,
```
Vani-1107: Are these statuses expected in case of payouts ? Can we please confirm with WPG - fast access payout doc ?

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1046,44 +1046,75 @@ fn map_worldpayxml_authorize_status(
...
+        | WorldpayxmlLastEvent::RefundReceived
+        | WorldpayxmlLastEvent::RefundFailed
+        | WorldpayxmlLastEvent::CancelReceived
+        | WorldpayxmlLastEvent::QueryRequired
+        | WorldpayxmlLastEvent::Expired
+        | WorldpayxmlLastEvent::Error
+        | WorldpayxmlLastEvent::PushRequested
+        | WorldpayxmlLastEvent::PushPending
+        | WorldpayxmlLastEvent::PushApproved
+        | WorldpayxmlLastEvent::PushRefused => AttemptStatus::Failure,
```
Vani-1107: These statuses are not expected in case of setup mandate flow right? why are we not throwing an error in these cases ?

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1029,15 +1029,15 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
 fn map_worldpayxml_authorize_status(
     last_event: &WorldpayxmlLastEvent,
     is_auto_capture: bool,
-    capture_sync_status: Option<&AttemptStatus>,
+    previous_status: Option<&AttemptStatus>,
 ) -> AttemptStatus {
     match last_event {
         WorldpayxmlLastEvent::Authorised => {
             if is_auto_capture {
                 AttemptStatus::Pending
```
Vani-1107: This should be captured

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1046,44 +1046,75 @@ fn map_worldpayxml_authorize_status(
...
         WorldpayxmlLastEvent::RefundFailed => AttemptStatus::Failure,
         WorldpayxmlLastEvent::Expired => AttemptStatus::Failure,
         WorldpayxmlLastEvent::Error => AttemptStatus::Failure,
 
         WorldpayxmlLastEvent::PushRequested
         | WorldpayxmlLastEvent::PushPending
         | WorldpayxmlLastEvent::PushApproved
-        | WorldpayxmlLastEvent::PushRefused
-        | WorldpayxmlLastEvent::SettledByMerchant => AttemptStatus::Failure,
+        | WorldpayxmlLastEvent::PushRefused => AttemptStatus::Failure,
```
Vani-1107: why are we not throwing an error here ?

crates/integrations/connector-integration/src/connectors/worldpayxml/responses.rs
```diff
@@ -310,20 +325,102 @@ pub struct WorldpayxmlCancelOrRefundReceived {
...
 pub enum WorldpayxmlTransactionResponse {
     Payment(Box<WorldpayxmlAuthorizeResponse>),
     Webhook(WorldpayxmlWebhookResponse),
 }
 
+/// Concrete (never `deserialize_any`) shape covering both bodies a sync call can return, so the
+/// right [`WorldpayxmlTransactionResponse`] variant can be picked without buffering the payload.
+#[derive(Debug, Deserialize)]
+#[serde(rename = "paymentService")]
+struct WorldpayxmlSyncResponseBody {
```
Vani-1107, line 343: can we use `rename_all = "PascalCase"` instead of using rename each individual field ?

crates/integrations/connector-integration/src/connectors/worldpayxml/responses.rs
```diff
@@ -310,20 +325,102 @@ pub struct WorldpayxmlCancelOrRefundReceived {
...
+    }
+}
+
+/// Order-notification body Worldpay can deliver for an order.
+///
+/// `order_code` and `payment_status` are mandatory: making them optional is what let this shape
+/// match every payload and mask genuine parse failures.
 #[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(rename_all = "camelCase")]
 pub struct WorldpayxmlWebhookResponse {
```
Vani-1107, line 398: lets use `#[serde(rename_all = "PascalCase")]` instead of using rename on individual field

Vani-1107 (changes requested):

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1124,45 +1156,46 @@ fn get_worldpayxml_mandate_reference(
...
+        WorldpayxmlLastEvent::Refunded | WorldpayxmlLastEvent::RefundedByMerchant => {
+            Ok(RefundStatus::Success)
         }
+        WorldpayxmlLastEvent::SentForRefund
+        | WorldpayxmlLastEvent::RefundRequested
+        | WorldpayxmlLastEvent::SentForFastRefund => Ok(RefundStatus::Pending),
+        WorldpayxmlLastEvent::RefundFailed => Ok(RefundStatus::Failure),
+        // An expired order can no longer be refunded, so the refund will never complete.
+        WorldpayxmlLastEvent::Expired => Ok(RefundStatus::Failure),
+        WorldpayxmlLastEvent::Captured | WorldpayxmlLastEvent::Settled => Ok(RefundStatus::Pending),
```
Vani-1107, line 1241: in case of refunds, why are we mapping these to pending ?

crates/integrations/connector-integration/src/payout_connectors/worldpayxml/transformers.rs
```diff
@@ -70,17 +70,34 @@ fn map_worldpayxml_payout_status(
         | WorldpayxmlLastEvent::PushRefused
         | WorldpayxmlLastEvent::Expired
         | WorldpayxmlLastEvent::Error => common_enums::PayoutStatus::Failure,
+        // Exhaustiveness only: the shared lastEvent enum gained variants for the payment flows.
+        // Every mapping above is unchanged from before those variants existed.
+        _ => common_enums::PayoutStatus::Pending,
```
Vani-1107: why do we want to default to pending ?

Vani-1107 (dismissed):

JeevaRamu0104 (dismissed):

gopikrishna000 (dismissed):

Nithin1506200 (dismissed):

crates/integrations/connector-integration/src/payout_connectors/worldpayxml/transformers.rs
```diff
@@ -49,38 +52,62 @@ impl TryFrom<&ConnectorSpecificConfig> for WorldpayxmlAuthType {
...
-    } else {
-        "2".to_string()
-    }
+fn worldpayxml_amount_exponent(
+    currency: common_enums::Currency,
+) -> Result<String, Report<IntegrationError>> {
+    currency
+        .number_of_digits_after_decimal_point()
+        .map(|digits| digits.to_string())
+        .map_err(|_| {
```
Shubhodip900: can we log the actual error as well

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -294,14 +292,28 @@ fn get_worldpayxml_wallet_payment_method(
...
-    } else {
-        "2".to_string()
-    }
+fn get_worldpayxml_exponent(
+    currency: common_enums::Currency,
+) -> Result<String, Report<IntegrationError>> {
+    currency
+        .number_of_digits_after_decimal_point()
+        .map(|digits| digits.to_string())
+        .map_err(|_| {
```
Shubhodip900: +1

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -435,7 +446,13 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
                     .customer_name
                     .clone()
-                    .map(|name| crate::utils::normalize_cardholder_name(Secret::new(name)));
+                    .map(Secret::new)
+                    .or_else(|| {
+                        router_data
+                            .resource_common_data
+                            .get_optional_billing_full_name()
+                    })
+                    .map(crate::utils::normalize_cardholder_name);
```
Shubhodip900:
```suggestion
                    .map(utils::normalize_cardholder_name);
```

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -618,7 +630,13 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
```
Shubhodip900:
```suggestion
                    .map(utils::normalize_cardholder_name);
```

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1084,65 +1092,89 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
+        // left uncaptured, say). The money can never be taken, so this is terminal — not a
+        // protocol violation to error on.
+        WorldpayxmlLastEvent::Expired => Ok(AttemptStatus::Failure),
+        WorldpayxmlLastEvent::Cancelled => Ok(AttemptStatus::Voided),
+        WorldpayxmlLastEvent::Captured
+        | WorldpayxmlLastEvent::Settled
+        | WorldpayxmlLastEvent::SettledByMerchant => Ok(AttemptStatus::Charged),
+        WorldpayxmlLastEvent::SentForAuthorisation => Ok(AttemptStatus::Authorizing),
+        WorldpayxmlLastEvent::Unknown => Ok(retain_previous_attempt_status(previous_status)),
+        _ => Err(crate::utils::unexpected_response_fail(
```
Shubhodip900:
```suggestion
        _ => Err(utils::unexpected_response_fail(
```

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1084,65 +1092,89 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
+        // See [`map_worldpayxml_authorize_status`]: an expired authorisation is terminal.
+        WorldpayxmlLastEvent::Expired => Ok(AttemptStatus::Failure),
+        WorldpayxmlLastEvent::Cancelled => Ok(AttemptStatus::Voided),
+        WorldpayxmlLastEvent::Authorised
+        | WorldpayxmlLastEvent::Captured
+        | WorldpayxmlLastEvent::Settled
+        | WorldpayxmlLastEvent::SettledByMerchant => Ok(AttemptStatus::Charged),
+        WorldpayxmlLastEvent::SentForAuthorisation => Ok(AttemptStatus::Authorizing),
+        WorldpayxmlLastEvent::Unknown => Ok(retain_previous_attempt_status(previous_status)),
+        _ => Err(crate::utils::unexpected_response_fail(
```
Shubhodip900:
```suggestion
        _ => Err(utils::unexpected_response_fail(
```

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1412,9 +1478,13 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
             crate::utils::response_deserialization_fail(item.http_code, "worldpayxml: response body did not match the expected format; confirm API version and connector documentation."),
         )?;
 
-        let status = AttemptStatus::from(&payment.last_event);
+        let status = map_worldpayxml_setup_mandate_status(
+            &payment.last_event,
+            Some(&router_data.resource_common_data.status),
```
Shubhodip900, line 1483: the previous status is by default Pending in the `resource_common_data`. Is this expected to fall back to pending state?

crates/integrations/connector-integration/src/connectors/worldpayxml/transformers.rs
```diff
@@ -1183,45 +1215,46 @@ fn get_worldpayxml_mandate_reference(
...
+        WorldpayxmlLastEvent::Unknown => {
+            // An unrecognised event says nothing about the refund, so hold the status we already
+            // had rather than reporting a terminal one.
+            tracing::warn!(
+                retained_status = ?previous_status,
+                "worldpayxml: unknown lastEvent received; retaining previous refund status"
+            );
+            Ok(previous_status)
+        }
+        _ => Err(crate::utils::unexpected_response_fail(
```
Shubhodip900:
```suggestion
        _ => Err(utils::unexpected_response_fail(
```

Shubhodip900 (approved):

Nithin1506200 (approved):

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

PR #2133, opened 2026-08-14
log response headers

Saptak88:
### Changes
- Introduces MaskableHeadersToJson implementations for HeaderMap, Vec, and HashSet.
- Adds get_unmasked_response_headers() to ConnectorIntegrationV2.
- Allows connectors to explicitly declare response headers that are safe to log.
- Masks all headers when no allowlist is provided.
- Keeps non-UTF-8 header values masked even when allowlisted.
- Removes the old response-header serialization path and uses the new trait for success and error responses.
- Adds tests covering allowlisted, sensitive, non-UTF-8, and absent-allowlist behavior.

### Additional Changes

### DIFF
--- a/crates/common/common_utils/src/events.rs
+++ b/crates/common/common_utils/src/events.rs
@@ -1,8 +1,8 @@
 #[cfg(feature = "log-transformations")]
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
-use hyperswitch_masking::ErasedMaskSerialize;
+use hyperswitch_masking::{ErasedMaskSerialize, Maskable, Secret};
 use serde::{Deserialize, Serialize};
 
 use crate::errors::EventPublisherError;
@@ -19,6 +19,7 @@ use crate::{
     lineage,
     types::TimeRange,
 };
+use reqwest::header::HeaderMap;
 
 /// Wrapper type that enforces masked serialization for Serde values
 #[derive(Debug, Clone, Serialize)]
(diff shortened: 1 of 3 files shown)


AmitsinghTanwar007 (commented):
## Review: Response Header Logging

### Overall
Good direction — masking response headers by default with a per-connector allowlist is the right pattern. A few concerns:

---

### 1. `reqwest::header::HeaderMap` import in `events.rs` — wrong dependency direction

```rust
use reqwest::header::HeaderMap;
```

`common_utils` (a low-level utility crate) is importing from `reqwest` (an HTTP client). The `Response` struct in `domain_types/router_response_types.rs` uses `http::HeaderMap`, not `reqwest::header::HeaderMap`. These are the same underlying type (reqwest re-exports from `http`), but the import should be `http::HeaderMap` to keep `common_utils` independent of `reqwest`.

---

### 2. `Vec` and `HashSet` impls ignore `unmasked_keys`

```rust
impl MaskableHeadersToJson for Vec<(String, Maskable<String>)> {
    fn maskable_headers_to_json(&self, _unmasked_keys: Option<&[&str]>) -> serde_json::Value {
        maskable_header_values_to_json(self.iter())
    }
}
```

The `_unmasked_keys` parameter is silently ignored for `Vec` and `HashSet` impls. This is confusing — a caller passing `Some(&["x-request-id"])` would expect unmasking but get none. Either:
- Use the parameter consistently across all impls, or
- Make the trait method not take `unmasked_keys` for these types (separate methods)

---

### 3. Response headers now logged outside the event guard

Before:
```rust
if let Some(evt) = event.as_deref_mut() {
    // response body + headers logged here
}
```

After:
```rust
if let Some(evt) = event.as_deref_mut() {
    // only response body
}
// response headers logged outside, directly from body.headers
```

This changes behavior — previously, headers were only logged when an event context existed. Now they are always logged. Is this intentional? Probably fine, but worth calling out.

---

### 4. Error path response headers — good addition

The PR adds response header logging on the error path (Err(body) branch) which was missing before. This is a good catch.

---

### 5. Test for non-UTF-8 asserts assert_ne — hard to reason about

```rust
fn non_utf8_headers_remain_masked_even_when_allowlisted() {
    ...
    assert_ne!(headers_json["x-request-id"], "<non-UTF-8 header value>");
}
```

This asserts the value is NOT equal to the literal string but the actual value is a Secret-formatted version of that string. The test name says "remain masked" but assert_ne against the unmasked placeholder does not clearly prove masking. Consider asserting the value contains the masking format explicitly.

---

### 6. get_unmasked_response_headers returns static slice

This is fine for now, but means the allowlist is hardcoded per connector type at compile time. If this ever needs to be config-driven, the signature will need to change. Not a blocker, just noting.

---

### Summary
The approach is sound. Main fix needed: use `http::HeaderMap` instead of `reqwest::header::HeaderMap` in `events.rs`. The ignored `_unmasked_keys` on Vec/HashSet impls is a footgun worth addressing.

AmitsinghTanwar007 (commented):
## Follow-up: Not reusing existing masking pattern

The PR's `HeaderMap` impl manually creates `Secret::new(value)` and formats with `format!("{:?}", Secret::new(...))` inline:

```rust
Ok(value) => format!("{:?}", Secret::<String>::new(value.to_string())),
Err(_) => format!("{:?}", Secret::<String>::new("<non-UTF-8 header value>".to_string())),
```

This bypasses the existing `Maskable` pattern that the codebase already uses everywhere for header masking. The established approach is:

1. Convert headers to `Vec<(String, Maskable<String>)>` — mark sensitive headers as `Maskable::Masked(Secret::new(val))` and allowlisted ones as `Maskable::Normal(val)`
2. Pass through the existing `maskable_headers_to_json()` function

Instead of a new trait + 3 impls, you could add a simple conversion function:

```rust
fn header_map_to_maskable(
    headers: &http::HeaderMap,
    unmasked_keys: Option<&[&str]>,
) -> Vec<(String, Maskable<String>)> {
    headers.iter().map(|(key, value)| {
        let is_unmasked = unmasked_keys.is_some_and(|keys| {
            keys.iter().any(|k| key.as_str().eq_ignore_ascii_case(k))
        });
        let val = value.to_str().unwrap_or("<non-UTF-8>").to_string();
        if is_unmasked {
            (key.to_string(), Maskable::Normal(val))
        } else {
            (key.to_string(), Maskable::Masked(Secret::new(val)))
        }
    }).collect()
}
```

Then call the existing `maskable_headers_to_json()` on the result. This:
- Reuses the existing masking infrastructure instead of reimplementing it
- Keeps the masking format consistent (if `Secret`'s debug format ever changes, it changes everywhere)
- Eliminates the need for the `MaskableHeadersToJson` trait and its 3 separate impls
- Avoids the `_unmasked_keys` parameter being silently ignored on `Vec`/`HashSet` impls

AmitsinghTanwar007 (commented):
## Review Comments

1. **Wrong import** — Use `http::HeaderMap` instead of `reqwest::header::HeaderMap` in `events.rs`. `common_utils` shouldn't depend on `reqwest`.

2. **Don't reimplement masking** — The `HeaderMap` impl manually does `Secret::new()` + `format!("{:?}")`. Instead, convert `HeaderMap` to `Vec<(String, Maskable<String>)>` and reuse the existing `maskable_headers_to_json()`. This eliminates the need for the new `MaskableHeadersToJson` trait entirely.

3. **`_unmasked_keys` ignored** — `Vec` and `HashSet` impls silently ignore the `unmasked_keys` parameter. A caller passing allowlisted keys would expect unmasking but get none.

4. **Weak test** — The non-UTF-8 test uses `assert_ne` which doesn't actually prove masking. Assert the value contains the masking format instead.

PR #2134, opened 2026-08-14
fix(maya): capture missing response fields + surface raw connector response

kanikac199:
## Description

Two related connector-response improvements for the **Maya** connector (one commit each):

**1. Capture missing PSync & refund response fields via the money framework**
Maya's payment-inquiry (PSync) and refund responses returned fields the connector was neither deserializing nor handing back, so they were dropped from the emitted response.
- **PSync** (`MayaWebhookBody`): add `isPaid`, `amount`, `currency`, `canCapture`, `createdAt`, `updatedAt`. `amount` is modeled as `StringMajorUnit` (Maya serializes it as a JSON **string** `"100"`, despite the API reference documenting a number), run through `AmountConvertor::convert_back` and surfaced via `PaymentSynIntegrityObject`. The non-money flags/timestamps go through `connector_metadata`.
- **Refund** (`MayaRefundResponse`): add `amount` (also a string) + `currency` → `RefundIntegrityObject`.
- RSync's integrity object is identity-based, so its amount is modeled/consumed but not wired to an amount check.

**2. Surface raw connector response on all flows**
Populate `raw_connector_response` in connector code for Authorize / PSync / Void / Refund / RSync (mirrors flywire/kount and Maya's existing webhook path), plus the error path (`get_error_response`). This captures the raw response independent of the `return_raw_connector_data` config flag (off in development).

3. Removed user_id from request 
No need to pass user_id , since it is optional and not needed for current implementation.

## Motivation and Context

The connector was silently dropping response data Maya returns. Because the emitted response body is re-serialized from the typed structs, any unmodeled field never reached the caller — losing amount/currency (needed for money-framework integrity verification) and operational flags (`isPaid`, `canVoid`, `canRefund`, `canCapture`) as well as the raw connector payload. This models those fields and routes amount/currency through the money framework so PSync/refund amounts are verified against intent, and makes the raw connector response available on every flow.

### Additional Changes

(No proto/contract change — the response fields already exist; they are now populated. No config changes committed.)

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/maya/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/maya/transformers.rs
@@ -571,17 +571,6 @@ impl TryFrom<ResponseRouterData<MayaWebhookBody, Self>>
             _ => None,
         };
 
-        // Operational fields Maya returns that have no typed slot on the payments
-        // response are surfaced verbatim through `connector_metadata`.
-        let connector_metadata = serde_json::json!({
-            "isPaid": payment.is_paid,
-            "canVoid": payment.can_void,
-            "canRefund": payment.can_refund,
-            "canCapture": payment.can_capture,
-            "createdAt": payment.created_at.clone(),
-            "updatedAt": payment.updated_at.clone(),
-        });
-
         let settlement_status = maya_settlement_status(payment.can_void, payment.can_refund);
 
         let status = common_enums::AttemptStatus::from(payment.status.clone());
@@ -607,7 +596,7 @@ impl TryFrom<ResponseRouterData<MayaWebhookBody, Self>>
                 resource_id,
                 redirection_data: None,
                 mandate_reference: None,
-                connector_metadata: Some(connector_metadata),
+                connector_metadata: None,
                 network_txn_id: None,
                 connector_response_reference_id: Some(connector_response_reference_id),
                 incremental_authorization_allowed: None,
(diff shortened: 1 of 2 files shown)


crates/integrations/connector-integration/src/connectors/maya/transformers.rs
```diff
@@ -534,6 +555,33 @@ impl TryFrom<ResponseRouterData<MayaWebhookBody, Self>>
...
+        // Operational fields Maya returns that have no typed slot on the payments
+        // response are surfaced verbatim through `connector_metadata`.
+        let connector_metadata = serde_json::json!({
+            "isPaid": payment.is_paid,
+            "canVoid": payment.can_void,
+            "canRefund": payment.can_refund,
+            "canCapture": payment.can_capture,
+            "createdAt": payment.created_at.clone(),
+            "updatedAt": payment.updated_at.clone(),
+        });
```
Aishwariyaa-Anand: do we need this

shuklatushar226 (dismissed):

crates/integrations/connector-integration/src/connectors/maya/transformers.rs
kanikac199: Good call — removed in f3742d5. It was needed when first added (those fields had no other home), but the later commit that sets `raw_connector_response` on all flows means isPaid/canVoid/canRefund/canCapture/createdAt/updatedAt are already surfaced in both `raw_connector_response` and `typed_connector_response` (the full payment object). Nothing consumed the metadata back, so it was pure duplication.

Aishwariyaa-Anand (approved):

shuklatushar226 (approved):

PR #2138, opened 2026-08-17
feat(connector): Worldpay Raft Connector Integration

### LINKED ISSUE #2117: Worldpay Raft Connector Integration
Cards and wallets with recurring

Shubhodip900:
## Description

Implements the Worldpay Native RAFT connector (worldpayraft) for card payments against Worldpay's core authorization platform. USA-only, enterprise merchants, API v1.35.1.

Auth: Custom Authorization: VANTIV license="<license>" header. Credentials: license (header) + merchant_id (request body as WorldPayMerchantID).

Amount format: StringMajorUnit — decimal string (e.g. "100.00").

Flows implemented: Authorize, Capture, Refund, SetupMandate, RepeatPayment

Not implemented: Void, PSync, RSync — RAFT has no card void, transaction inquiry, or refund inquiry endpoints. These correctly return not_implemented rather than synthesizing responses.

SetupMandate tokenizes the card via POST /tokenization/token, storing the returned TokenizedPAN as connector_mandate_id. RepeatPayment uses that token as PAN on a standard auth call with MerchantInitiatedTransaction: "Y".

Credit + Debit card support — both card types share inner structs; the outer JSON wrapper key and URL differ:
- Credit: { "creditauth": {...} } → /credit/authorization
- Debit: { "debitpreauth": {...} } → /debit/preauth
- Same pattern for Capture (completion) and Refund (refund).

Card type is detected from card.card_type on Authorize and encoded as a C| / D| prefix in connector_transaction_id so Capture and Refund can route correctly without re-reading card data.

New files: connectors/worldpayraft.rs, connectors/worldpayraft/transformers.rs

Touched files: connectors.rs, default_implementations.rs, router_data.rs, payment.proto, field-probe/auth.rs, config TOMLs.
### Additional Changes

## How did you test it?

Alpha connector integration, no creds available, check build using `cargo build` and `cargo run --bin grpc-server"


XyneSpaces:
Automated review findings for the Worldpay RAFT connector integration:

**[S2] `card_type` matching is case-sensitive and silently defaults to credit**
Added line:
```rust
let is_debit = matches!(card.card_type.as_deref(), Some("debit") | Some("Debit"));
```
Any value other than exactly `"debit"` / `"Debit"` (e.g. `"DEBIT"`, `"debit_card"`, or `None`) is routed to the credit endpoint. Debit cards and integrations that omit `card_type` will hit the wrong RAFT endpoint and fail at the processor.

```suggestion
let is_debit = matches!(
    c.card_type.as_deref(),
    Some(t) if t.eq_ignore_ascii_case("debit")
);
```

**[S2] Empty `ExpirationDate` sent for repeat payments**
```rust
expiration_date: Secret::new(String::new()),
```
If RAFT validates the `YYMM` format on `/credit/authorization`, an empty string will reject all merchant-initiated/repeat payments. Confirm the spec; if expiry is not required, omit the field with `#[serde(skip_serializing_if = "Option::is_none")]` and use `Option<Secret<String>>`.

**[S2] Error response parsing assumes a flat body**
```rust
let response: worldpayraft::WorldpayraftErrorResponse = res
    .response
    .parse_struct("WorldpayraftErrorResponse")
    .change_context(errors::ConnectorError::ResponseDeserializationFailed { ... })?;
```
RAFT errors may be wrapped in the same `creditauthresponse`/`debitpreauthresponse` envelopes as successes. In that case deserialization fails and the real decline reason is lost. Parse into an untagged enum mirroring the success wrappers, or use a catch-all `serde_json::Value` first to inspect the shape.

gopikrishna000 (dismissed):

gopikrishna000 (dismissed):

crates/integrations/connector-integration/src/connectors/worldpayraft/transformers.rs
```diff
@@ -0,0 +1,1475 @@
...
+#[derive(Debug, Serialize)]
+pub struct WorldpayraftAddressVerificationData {
+    #[serde(skip_serializing_if = "Option::is_none", rename = "AVSZIPCode")]
+    pub avs_zip_code: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "AVSAddress")]
+    pub avs_address: Option<Secret<String>>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct WorldpayraftTerminalData {
```
Vani-1107: we can use `#[serde(rename_all = "PascalCase")]` wherever applicable

crates/integrations/connector-integration/src/connectors/worldpayraft/transformers.rs
```diff
@@ -0,0 +1,1475 @@
...
+    pub system_trace_number: String,
+}
+
+// =============================================================================
+// AUTHORIZE REQUEST
+// =============================================================================
+
+/// Inner fields shared by both creditauth and debitpreauth requests.
+#[derive(Debug, Serialize)]
+pub struct WorldpayraftCardAuthInner<
```
Vani-1107, line 236: lets use `#[serde(rename_all = "PascalCase")]` instead of renaming individual fields wherever applicable.

Vani-1107 (approved):

JeevaRamu0104 (approved):

gopikrishna000 (approved):

PR #2142, opened 2026-08-17
feat : Integrate deja(ART) for hyperswitch-prism

iemyashasvi:
### Additional Changes


XyneSpaces:
Automated review findings for the Déjà (ART) integration:

**[S2] Corrupted tape body is silently replaced with an empty response**
`crates/common/external-services/src/deja_codec.rs`:
```rust
let body = base64::engine::general_purpose::STANDARD
    .decode(&self.body_b64)
    .unwrap_or_default();
```
During replay, a corrupt or truncated `body_b64` becomes an empty body with the recorded status code. Results become silently wrong and data corruption is hidden. Propagate the decode failure instead of using `unwrap_or_default`.

**[S2] Schema drift in the taped envelope can fall back to a live call**
`crates/common/external-services/src/deja_codec.rs` (also `KafkaOutcomeCodec` and `InjectorOutcomeCodec`):
```rust
match serde_json::from_value::<HttpTapeOutcome>(recorded).ok()? {
```
Returning `None` when the tape envelope fails to deserialize violates the module’s "always-substitute" policy: a future format/schema change will cause the replay runtime to issue real connector/injector calls. Treat deserialization failure as a fatal replay error rather than `None`.

XyneSpaces:
:rotating_light: `deja_codec::http_args` and `deja_codec::kafka_args` reveal masked secrets via `secret.peek().clone()` and serialize full request bodies/payloads into the recording tape. This stores plaintext credentials, auth headers, and potentially card/PII data on the Kafka recording topic.

Either redact/deny-list sensitive headers and fields before capture, hash identifying values for matching, or encrypt the recording topic with the same ACL posture as other sensitive audit streams. Do not persist unmasked secrets to Déjà tapes.

crates/common/external-services/Cargo.toml
```diff
@@ -60,15 +59,12 @@ injector = { git = "https://github.com/juspay/hyperswitch", tag = "2026.06.18.0"
...
-log_utils = { git = "https://github.com/juspay/framework-libs-rs", rev = "3cdcc24b461917b29641942b04f3aef8110b1adf", package = "log_utils", features = [
-    "tracing",
-    "tracing-storage-api",
-] }
 
 # OpenTelemetry (OTLP metrics push) — optional, enabled via the `otel` feature.
 opentelemetry = { version = "0.27", optional = true, default-features = false, features = [
     "metrics",
 ] }
+deja = { git = "https://github.com/iemyashasvi/deja", rev = "55fb014ed57f4a1d2d827b8990e4235cc10da0b7", optional = true, default-features = false }
```
Aishwariyaa-Anand, line 67: why are we using forked deja

crates/common/ucs_env/src/deja_config.rs
```diff
@@ -0,0 +1,210 @@
...
+impl DejaConfig {
+    /// Fail-loud validation, run at boot. Replay must never be reachable in
+    /// production, and recording needs somewhere to send events.
+    pub fn validate(&self, environment: &consts::Env) -> Result<(), config::ConfigError> {
+        match self.mode {
+            DejaMode::Disabled => Ok(()),
+            DejaMode::Replay => match environment {
+                consts::Env::Production => Err(config::ConfigError::Message(
+                    "deja.mode = \"replay\" is not permitted in the production environment"
+                        .to_string(),
```
Aishwariyaa-Anand, line 51: The `test_mode` field in `domain_types/src/types.rs:11143` is set based on `consts::Env` and drives connector-level decisions for connectors like Adyen. When replaying production recordings in sandbox, this environment-dependent logic will cause mismatches. is there any mechanism to handle this?

crates/common/external-services/Cargo.toml
iemyashasvi: we are using forked deja till the deja side changes are not merged in the main deja branch , before this PR , will have the main deja pr merged and have the rev and the git updated accordingly

PR #2144, opened 2026-08-17
Docs: remove php sdk references

nfarah86:
## Description

Removed all PHP references because PHP isn't supported

### Additional Changes


nfarah86:
@JeevaRamu0104 @Sakilmostak I rebased on top of 2102, so I think this PR is now good for a review. Please let me know. I removed all PHP references since PHP is not supported.

One thing worth checking before this merges: if someone regenerates the SDK docs, do the php pages come back? I couldn't find a supported-language list in rules.md.... so it's my understanding that nothing there marks php as out of scope. If generation would recreate these, then deleting them is a temporary fix and the real fix is excluding php wherever the language set is decided. If it is generated, can we apply the fix on top of this PR, so all the PHP references and changes are in 1 scope.

JeevaRamu0104 (approved):

PR #2168, opened 2026-08-21
Ci: docs drift check

nfarah86:
Screenshot 2026-08-21 at 4 06 34 PM

## Description
When a PR changes the proto, this reports which documentation pages no longer match it. It runs on every proto PR, prints findings ad annotations on the changed files, and does not block the build. 

Right now, nothing in PRISM connects a proto change to the docs that are suppose to reflect/describe it.  The motivation behind this PR was after PR #1812 . IT ADDED `order_details` to `PaymentServiceCreateOrderRequest`. The connector docs got regenerated, but the the generator watches the field probes. Yet, the API-REF docs did not.  There were a few other instances where something similar happened. 

In the image, I made a mock field in the proto file, and as you can see, the script reported it undocumented. The SDKs references were commented out because we don't have a generator for them right now. This will be built soon. As it stands today, the SDKs are manually generated in claude... So there's nothing updating it or connecting the updates to the SDKs. 

## Motivation and Context
Why is this change required? What problem does it solve?
If it fixes an open issue, please link to the issue here. See PR #1812 

I'm building tooling that keeps the docs current with the codebase automatically. It has to do three things: notice the contract changed, work out whether the docs still describe it, and update them. This PR is the first two. The third needs the api-reference pipeline @Jeeva is writing, and once that exists this becomes the check on its output.

### Additional Changes

### DIFF
--- a/.github/workflows/proto-checks.yml
+++ b/.github/workflows/proto-checks.yml
@@ -61,3 +61,17 @@ jobs:
           echo "       Link the consumer migration PR in the PR description."
           echo "       Ensure removed field numbers appear under 'reserved'."
           exit 1
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Install proto tooling
+        run: pip install grpcio-tools
+
+      - name: Docs drift check
+        run: |
+          python3 scripts/ci/docs_drift_check.py \
+            --base-ref "origin/${GITHUB_BASE_REF}" \
+            --format github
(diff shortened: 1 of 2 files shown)


XyneSpaces:
[should-fix] The docs drift check can miss normal proto field edits because `changed_symbols()` only looks for `message` / `service` in the `git diff -U0` hunk header. With the repo’s default `.gitattributes`, adding a field inside an existing message produces a hunk for the field line, so `touched_messages` stays empty and the script exits before checking docs.

Please derive the enclosing message/service by parsing the changed proto file around each changed line, or configure/use a protobuf-aware diff driver before relying on hunk headers.

PR #2172, opened 2026-08-24
feat(pre-authenticate): resolve connector variant in preAuthN, support FRM (Kount) via x-frm-connector

Aishwariyaa-Anand:
## Description

`PRE_AUTHENTICATE` only resolved connectors tagged `ConnectorVariant::Payment`. Due to which Kount is dual-registered as both a payment connector and an FRM connector, but requests tagged `ConnectorVariant::Frm` (i.e. `x-frm-connector: kount`) were invisible to these flows — the composite `/composite/pre_authenticate` endpoint even 400'd outright with "missing x-connector metadata" since it hard-required `x-connector`. This adds FRM support for PreAuthenticate end-to-end, without hand-writing a `match ConnectorVariant { ... }` per call site, and without special-casing Kount anywhere — connector-family resolution is now declarative and reusable.

## Changes

`FrmServiceTrait` (`interfaces/connector_types.rs`) gains `PaymentPreAuthenticateV2<DefaultPCIHolder>` as a supertrait, so FRM connectors' `PreAuthenticate` implementation is reachable through the FRM path, not just Payment. Kount already implements this trait — no behavior change on its side, just newly visible through `FrmConnectorData`. In `kount.rs`, the `FrmServiceTrait` impl was narrowed from generic-over-`T` to `Kount<DefaultPCIHolder>`, matching the only monomorphization `FrmConnectorData` ever constructs (FRM requests never carry payment-method data, so this can't be generic the way the payment-side impl is).

In `utils.rs`, a shared `resolve_connector_integration!` macro was added — it tries a list of `ConnectorDataProvider` families against a request's `ConnectorVariant` in order and returns whichever resolves. This is the one place connector-family resolution lives now; both macro-generated code and hand-written resolvers use it. `implement_connector_operation!` was extended so a single flow can list multiple connector families (`connector_data_types: [Primary, Extra, ...]`) instead of requiring one macro invocation plus a hand-written dispatcher per family.

In `payments.rs`, `internal_pre_authenticate` now resolves via `[ConnectorData, FrmConnectorData]` — transitional, so Kount keeps working through `ConnectorVariant::Payment` until callers fully cut over to `ConnectorVariant::Frm`, at which point dropping `ConnectorData` from the list is the entire migration. Two other instances of the same "N functions + hand-written `match`/`matches!` dispatcher" pattern were also collapsed into single multi-family calls: `internal_sdk_session_token` (was `..._payment` + `..._authenticator` + a dispatcher) and `internal_get_payment_method` (same shape). `handle_access_token` (backs `ServerAuthenticationToken`) now resolves via `resolve_connector_integration!` instead of a hand-written match across Payment/Frm/Payout.

In `composite-service/payments.rs`, `/composite/pre_authenticate` now resolves the full `ConnectorVariant` (`connector_variant_from_composite_metadata`, priority `x-connector` → `x-surcharge-connector` → `x-frm-connector` → `x-auth-connector`) instead of hard-requiring `x-connector`. `create_server_authentication_token` was widened to accept `&ConnectorVariant` and resolves `should_do_access_token` across both Payment and FRM. The other six composite flows that share this helper (authorize, get, refund, refund-get, void, capture) are unaffected — they still resolve Payment-only, now just explicitly wrapped as `ConnectorVariant::Payment(...)` at the call site.

### Additional Changes


crates/grpc-server/grpc-server/src/utils.rs
```diff
@@ -1168,4 +1273,179 @@ macro_rules! implement_connector_operation {
         result
     }
 };
+
+    // Same as the pattern above, but resolving across multiple connector families
+    // via `resolve_connector_integration!` — no hand-written match on
+    // `ConnectorVariant` and no separate `$fn_name`-per-family + dispatcher needed;
+    // list every family this flow serves and one function handles all of them.
+    (
+        fn_name: $fn_name:ident,
```
gopikrishna000, line 1086: when existing vs this macro will be used?
why divergence

Aishwariyaa-Anand: the singular arm is still used by ~15 call sites that is used against one connector variant (payouts, frm, refunds, disputes, surcharges). The new list-based arm is only used where a single flow now needs to try multiple variant at runtime (GetPaymentMethod → Payment or Authenticator; your new PreAuthenticate → Payment or Frm/Kount)

gopikrishna000: Then at runtime it is decided which variant to call right, frm or payment for the same flow

Why don't we update existing macro to support generic instead of creating another macro

gopikrishna000:
The `$()*` extras block is `run_holder_flow`'s body copy-pasted with `T` pinned to `DefaultPCIHolder` — same `RouterDataV2`, same `execute_connector_processing_step`, same `generate_response_fn`. It only exists because `run_holder_flow` resolves the connector itself, hardcoded to `[$connector_data<T>]`.

Two ways to drop it:

**1. Pass the resolved integration in.** `get_connector_integration_v2()` returns the same `BoxedConnectorIntegrationV2<'_, Flow, CommonData, Req, Resp>` for every family, so once the macro fixes those four params it's one concrete type. Take that instead of `&ConnectorVariant` and both the `connector:` param and the `where $connector_data<T>: ConnectorDataProvider` bound go away — extras become a resolve + one call.

**2. Or drop it outright.** Extras are `DefaultPCIHolder` + `(payload, None)`, which is what the `None =>` arm already does, and FRM requests carry no payment-method data so they land there anyway. Just widen that arm's list:

```rust
None => {
    let request = $request_data_constructor((payload.clone(), None)).to_grpc_error()?;
    run_holder_flow::<DefaultPCIHolder>(
        resolve_connector_integration!(
            &metadata_payload.connector,
            [$($extra_connector_data_type,)* $connector_data<DefaultPCIHolder>]
        ).ok_or_else(...)?,
        request, common_flow_data, connector_config, None, &config.proxy, ...
    ).await?
}
```

Also fixes a silent drop: an FRM request that does carry a card currently hits the extras block and gets `payment_method_data: None` with no error (the `foreign_try_from` accepts it). In the `None` arm it'd fail resolution and error instead.

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2180, opened 2026-08-25
ci(docker): publish a latest image tag and document running the server image

gopikrishna000:
## Description

The gRPC server is already published to GHCR on every nightly tag, but the README never said so — the only container instructions live in `setup.md` and describe a local `docker build`. This adds a **Running the gRPC Server (Docker)** section to the README, and publishes a `latest` tag so the first command in it is the one people actually type.

**CI** (`docker-to-ghcr-publish.yml`): `create-manifest` now also points `latest` at the freshly created manifest. It is guarded so only nightly CalVer tags move it — hotfix tags are cut from older release branches and a `workflow_dispatch` builds an unreleased branch, so neither may drag `latest` backwards.

**README**: pull/run, health check, configuration, metrics, and a local build.

- `ghcr.io/juspay/hyperswitch-prism`, multi-arch (`linux/amd64`, `linux/arm64`); `latest` for a quick start, dated tags to pin
- ports: gRPC `8000`; Prometheus metrics `8080`, called out separately rather than in the headline command since it is only needed if you scrape it
- OTLP push documented alongside it: the pipeline is compiled into the image but disabled in every config the image ships with, so it needs `CS__METRICS__OTEL__ENABLED=true` plus an endpoint
- `grpcurl` health check and `list` (server reflection is on, so no local `.proto` files are needed)
- config selection via `CS__COMMON__ENVIRONMENT`, per-key overrides via `CS__<SECTION>__<KEY>`, and mounting a custom config file

Open question for reviewers: `latest` currently follows nightly CalVer tags. Now that stable `vX.Y.Z` tags build images too, you may prefer `latest` to follow those instead — happy to switch, though `latest` would then not exist until the next stable tag.

## How did you test it?

Every command in the new section was run against the published image `ghcr.io/juspay/hyperswitch-prism:2026.08.25.0`:

- anonymous `docker pull` (no `docker login` needed — the package is public)
- `docker run -p 8000:8000 -p 8080:8080` → `grpc.health.v1.Health/Check` returns `{"status": "SERVING"}`; `/metrics` returns `200` with populated `GRPC_SERVER_REQUESTS_TOTAL` / `GRPC_SERVER_REQUEST_LATENCY` series
- `-e CS__COMMON__ENVIRONMENT=sandbox` → boots and serves
- `-e CS__SERVER__PORT=9001` → serves on the overridden port
- `-e CS__METRICS__OTEL__ENABLED=true -e CS__METRICS__OTEL__OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4317` → logs `OTLP metrics push pipeline initialised, endpoint: Some("http://otel-collector:4317")`
- config mounted over `/app/config/development.toml` → serves on the port set in the mounted file
- the platform claim was checked against the live OCI index, not just the build matrix: `linux/amd64` + `linux/arm64` (the two `unknown/unknown` entries are provenance attestations)

The `latest` step itself cannot run until this lands on a tag build. Note that `latest` will therefore not resolve until the next nightly tag after merge; the README links the packages page for dated tags in the meantime.

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all` (n/a — no Rust changes)
- [x] I addressed lints thrown by `cargo clippy` (n/a — no Rust changes)
- [x] I reviewed the submitted code

### DIFF
--- a/.github/workflows/docker-to-ghcr-publish.yml
+++ b/.github/workflows/docker-to-ghcr-publish.yml
@@ -182,6 +182,7 @@ jobs:
         env:
           REF_NAME: ${{ github.ref_name }}
         run: |
+          set -euo pipefail
           if [[ "${REF_NAME}" =~ ^20[0-9]{2}\.[0-9]{2}\.[0-9]{2}\.[0-9]+$ ]]; then
             docker buildx imagetools create --tag "ghcr.io/${REPO}:latest" "ghcr.io/${REPO}:${VERSION}"
             echo "Moved latest to ghcr.io/${REPO}:${VERSION}"
--- a/.github/workflows/docker-to-ghcr-publish.yml
+++ b/.github/workflows/docker-to-ghcr-publish.yml
@@ -159,21 +159,6 @@ jobs:
             "ghcr.io/${REPO}:${VERSION}-linux-arm64"
           echo "Successfully created manifest: ghcr.io/${REPO}:${VERSION}"
 
-      # `latest` follows nightly CalVer tags only. Hotfix tags are cut from
-      # older release branches and branch dispatches build unreleased code, so
-      # neither may move it backwards.
-      - name: Tag latest
-        if: ${{ github.ref_type == 'tag' }}
-        env:
-          REF_NAME: ${{ github.ref_name }}
-        run: |
-          if [[ "${REF_NAME}" =~ ^20[0-9]{2}\.[0-9]{2}\.[0-9]{2}\.[0-9]+$ ]]; then
-            docker buildx imagetools create --tag "ghcr.io/${REPO}:latest" "ghcr.io/${REPO}:${VERSION}"
-            echo "Moved latest to ghcr.io/${REPO}:${VERSION}"
-          else
-            echo "Tag '${REF_NAME}' is not a nightly tag - leaving latest unchanged."
-          fi
-
       - name: Get manifest digest
         id: digest
         run: |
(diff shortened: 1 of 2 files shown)


Shubhodip900:
**Minor:** Missing set -euo pipefail

If docker buildx imagetools create fails (auth expiry, registry hiccup), the step exits 0 because the failure is inside an if branch in bash — GitHub Actions sees a green step while latest silently wasn't updated. Fix:
```

run: |
  set -euo pipefail
  if [[ "${REF_NAME}" =~ ^20[0-9]{2}\.[0-9]{2}\.[0-9]{2}\.[0-9]+$ ]]; then
    ...
```

Source tag is correct

ghcr.io/${REPO}:${VERSION} is the multi-arch manifest already created by the preceding "Create manifest" step in the same job. Re-tagging it as latest via imagetools create copies the manifest pointer without re-pulling layers. Correct approach.

**README:-**

Content: good overall

- --rm on all examples prevents orphaned containers.
- Omitting -p 8080:8080 from the headline command is intentional and explained in the Metrics subsection. Good.
- Server reflection usage (grpcurl list, describe) means users never need local .proto files — clearly documented.
- Config override pattern (CS__<SECTION>__<KEY>) is documented correctly.

Gap: OTLP not documented

The PR description and test notes cover `CS__METRICS__OTEL__ENABLED=true / CS__METRICS__OTEL__OTEL_EXPORTER_OTLP_ENDPOINT,` but it's absent from the README. Since the pipeline is compiled-in and disabled by default, users wanting push-based metrics have no docs. Suggest adding under the Metrics subsection:

# Push metrics via OTLP (disabled by default)
```
docker run --rm -p 8000:8000 \
  -e CS__METRICS__OTEL__ENABLED=true \
  -e CS__METRICS__OTEL__OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4317 \
  ghcr.io/juspay/hyperswitch-prism:latest
```

AmitsinghTanwar007 (dismissed):

gopikrishna000:
Thanks — both applied in 4b7d483af.

**OTLP:** added under the Metrics subsection, roughly as suggested. Verified against the published image (`2026.08.25.0`, since `latest` doesn't exist until this merges):

```
INFO ucs_env::logger::metrics: OTLP metrics push pipeline initialised, endpoint: Some("http://otel-collector:4317")
```

gRPC and the Prometheus endpoint both keep working with it on, so the README says so explicitly.

While testing I also found that `/metrics` returns 200 with an **empty body** until the server handles its first request (true with OTLP on or off — 0 bytes before traffic, 3841 after). Someone following the README literally would think it was broken, so I noted it in one line.

**`set -euo pipefail`:** added. One correction on the reasoning, though — the step wasn't exiting 0. GitHub Actions' default shell on Linux is `bash -e {0}`, so `-e` was already on, and `set -e` is exempt only for the `if` *condition*, not the then-body:

```
$ cat sim.sh
if [[ "x" == "x" ]]; then
  false
  echo "REACHED"
else
  echo "else"
fi
$ bash -e sim.sh; echo "exit: $?"
exit: 1          # "REACHED" never printed
```

So a failing `imagetools create` already turned the step red. Worth having the explicit guard regardless, and `-u` is safe here since `REF_NAME` is step-level env and `VERSION`/`REPO` are job-level.

Unrelated, but the pre-existing **Get manifest digest** step does pipe `imagetools inspect` into `tr` with no `pipefail` — that's the one where a swallowed failure is actually reachable. Happy to fix it here or leave it for a separate PR.

Shubhodip900 (approved):

AmitsinghTanwar007 (approved):

PR #2183, opened 2026-08-26
feat(connector): [SAFERPAY] add Saferpay connector — Authorize, PSync, Capture, Void, Refund, RSync

shuklatushar226:
## Summary

Adds the Saferpay connector for Authorize, PSync, PreAuthenticate, Capture, Void, Refund, and RSync. This update incorporates the requested Jeeva review fixes.

## Review fixes in 2a1d7f116

- RSync now carries a refund settlement stage in refund metadata; it no longer relies on the always-Pending `RefundSyncData::refund_status`.
- Saferpay `RequestId` is stable across retries, using merchant request ID with connector reference fallback.
- Refund errors use the refund-specific terminal error mapper.
- Capture `CANCELED` maps to terminal Failure.
- Partial capture remains `PartialCharged` when the authorized total is unavailable, and uses captured metadata when available.
- Saferpay integration error contexts include suggested action and official documentation URL.
- `RefundResponse.refund_metadata` round-trips connector metadata needed by the settlement stage.
- Alpha connector verification now has a truthful non-empty Saferpay reason because CI has no Saferpay credentials.

## Validation

- Current `prism/main` merged before the fix commit.
- `cargo fmt --all -- --check`: PASS
- `RUSTFLAGS="-D warnings" cargo clippy --all-features --all-targets --profile release-fast`: PASS
- Existing CI nextest command: 486 passed, 22 failed, 11 skipped. The 22 failures are live connector-flow tests that stop at missing local connector-config files (`No such file or directory`); no assertion failures occurred. No new tests were added per requested scope.
- New-connector verifier for Saferpay: PASS (warns that CI has no sandbox credentials).

## Live verification

Re-run against **this** branch head (build of `67abeb19e`), so it reflects the current flow
rather than the pre-`dd0b4c33` PSync design the earlier description described.

Driven with `grpcurl` straight at the UCS server on `:8000`. Credentials are passed in the
`x-connector-config` header and are **not** reproduced here — export your own first:

```bash
export SFP_CFG='{"config":{"Saferpay":{"api_key":"<api-username>","key1":"<api-password>","api_secret":"<customer-id>","key2":"<terminal-id>"}}}'
H=(-H "x-connector: saferpay" -H "x-connector-config: $SFP_CFG"
   -H "x-merchant-id: sfp01" -H "x-tenant-id: public")
```

The Secure PayGate terminal. **USD only** — EUR/CHF/GBP return HTTP 402 `NO_CONTRACT`
on this account. Card `4111111111111111`, exp `12/2030`, CVC `123`.

### 1. Authorize — non-3DS

```bash
grpcurl -plaintext "${H[@]}" -H "x-request-id: r1" -d '{
  "merchant_transaction_id":"pr-nothreeds-1",
  "amount":{"minor_amount":1000,"currency":"USD"},
  "capture_method":"MANUAL","auth_type":"NO_THREE_DS",
  "payment_method":{"card":{
     "card_number":{"value":"4111111111111111"},
     "card_exp_month":{"value":"12"},"card_exp_year":{"value":"2030"},
     "card_cvc":{"value":"123"},"card_holder_name":{"value":"Test User"}}},
  "return_url":"https://example.com/return",
  "address":{"billing_address":{"first_name":{"value":"Test"},"last_name":{"value":"User"},
     "line1":{"value":"1 Main St"},"city":{"value":"Zurich"},"zip_code":{"value":"8000"},
     "country_alpha2_code":"CH","email":{"value":"test@example.com"}}}
}' localhost:8000 types.PaymentService/Authorize
```
```json
{"status":"AUTHORIZED","connectorTransactionId":"jUjCvMb2G93hSAYzQUUUAt2vG1Sb","statusCode":200,
 "connectorFeatureData":{"value":"{\"authorized_amount\":\"1000\"}"}}
```

### 2. PSync

```bash
grpcurl -plaintext "${H[@]}" -H "x-request-id: r2" -d '{
  "merchant_transaction_id":"pr-nothreeds-1",
  "connector_transaction_id":"jUjCvMb2G93hSAYzQUUUAt2vG1Sb",
  "amount":{"minor_amount":1000,"currency":"USD"}
}' localhost:8000 types.PaymentService/Get
```
```json
{"status":"AUTHORIZED","statusCode":200}
```

### 3. Capture

`connector_feature_data` carries the `authorized_amount` published by Authorize; without it the
authorized total is unknown and the capture is conservatively reported `PARTIAL_CHARGED`.

```bash
grpcurl -plaintext "${H[@]}" -H "x-request-id: r3" -d '{
  "merchant_capture_id":"pr-nothreeds-2",
  "connector_transaction_id":"MWEdtObGv3fWSA1hUz7rA3nb82zA",
  "amount_to_capture":{"minor_amount":1000,"currency":"USD"},
  "connector_feature_data":{"value":"{\"authorized_amount\":\"1000\"}"}
}' localhost:8000 types.PaymentService/Capture
```
```json
{"status":"CHARGED","statusCode":200,
 "connectorFeatureData":{"value":"{\"capture_id\":\"77G9xrbjK3vAtAfhYMbnbjp80IdA_c\"}"}}
```

### 4. Refund → RSync ×2 — the settle-once behaviour

A Saferpay refund is two calls: `Transaction/Refund` only authorizes it, and RSync issues the
settling `Transaction/Capture`. The second RSync must **read back**, not re-capture.

```bash
# Refund — refund_metadata carries the capture_id from step 3
grpcurl -plaintext "${H[@]}" -H "x-request-id: r4" -d '{
  "merchant_refund_id":"pr-rf-2","connector_transaction_id":"<txn>",
  "payment_amount":1000,"refund_amount":{"minor_amount":500,"currency":"USD"},
  "reason":"e2e","refund_metadata":{"value":"{\"capture_id\":\"<capture-id>\"}"}
}' localhost:8000 types.PaymentService/Refund
# -> {"status":"REFUND_PENDING","connectorRefundId":"..."}

# RSync #1 — settles, and returns the stage marker
grpcurl -plaintext "${H[@]}" -H "x-request-id: r5" -d '{
  "merchant_refund_id":"pr-rf-2","connector_transaction_id":"<txn>",
  "refund_id":"pr-rf-2","connector_refund_id":"<refund-txn>",
  "refund_amount":{"minor_amount":500,"currency":"USD"},
  "refund_metadata":{"value":"{\"capture_id\":\"<capture-id>\"}"}
}' localhost:8000 types.RefundService/Get
# -> {"status":"REFUND_SUCCESS",
#     "refundMetadata":{"value":"{\"capture_id\":\"...\",\"stage\":\"refund_settled\"}"}}

# RSync #2 — feed that refundMetadata back
# -> {"status":"REFUND_SUCCESS","refundMetadata":{"value":"{...,\"stage\":\"refund_settled\"}"}}
```

Endpoints actually hit, in order: `Refund` → `Capture` (settle) → **`Inquire`**. The second sync
does not re-POST Capture, so the earlier `TRANSACTION_ALREADY_CAPTURED` failure mode is gone.

### 5. 3DS — PreAuthenticate, real ACS challenge, settle Authorize

```bash
# Leg 1 — Initialize, returns the redirect and the session token
grpcurl -plaintext "${H[@]}" -H "x-request-id: r6" -d '{
  "merchant_order_id":"pr-3ds-1","merchant_transaction_id":"pr-3ds-1",
  "amount":{"minor_amount":1000,"currency":"USD"},
  "capture_method":"MANUAL","enrolled_for_3ds":true,
  "payment_method":{"card":{ ...same card... }},
  "return_url":"https://example.com/return",
  "continue_redirection_url":"https://example.com/complete",
  "address":{ ...same billing... }
}' localhost:8000 types.PaymentMethodAuthenticationService/PreAuthenticate
```
```json
{"status":"AUTHENTICATION_PENDING","connectorTransactionId":"afkklcx465xobrhwc9jnwd403",
 "redirectionData":{"form":{
   "endpoint":"https://test.saferpay.com/vt2/Api/Dcc/<customer-id>/<terminal-id>/afkklcx465xobrhwc9jnwd403",
   "method":"HTTP_METHOD_GET"}},
 "connectorFeatureData":{"value":"{\"saferpay_token\":\"afkklcx465xobrhwc9jnwd403\",\"stage\":\"initialized\"}"}}
```

The browser leg was completed for real, not stubbed — three Saferpay-hosted pages: DCC currency
choice → 3DS device-data collection (a JS SPA, needs a real browser) → ACS challenge. Saferpay's
test ACS takes no password; it returns a `cres` with `transStatus: "Y"` behind a Process button.

```
ACS: AnyBank TestCard Secure (3D-Secure Simulator)
cres transStatus: "Y"
returned to: https://example.com/complete
```

```bash
# Leg 2 — settle. redirection_response present + the token in connector_feature_data
grpcurl -plaintext "${H[@]}" -H "x-request-id: r7" -d '{
  "merchant_transaction_id":"pr-3ds-1","auth_type":"THREE_DS",
  "amount":{"minor_amount":1000,"currency":"USD"},"capture_method":"MANUAL",
  "payment_method":{"card":{ ...same card... }},
  "connector_feature_data":{"value":"{\"saferpay_token\":\"afkklcx465xobrhwc9jnwd403\",\"stage\":\"initialized\"}"},
  "redirection_response":{"params":"","payload":{}},
  "return_url":"https://example.com/return","address":{ ...same billing... }
}' localhost:8000 types.PaymentService/Authorize
```
```json
{"status":"AUTHORIZED","connectorTransactionId":"8Ejj4MAC7EUSSA38b9b4AndCY1Gb","statusCode":200,
 "connectorFeatureData":{"value":"{\"authorized_amount\":\"1000\"}"}}
```

Capture on that 3DS payment → `{"status":"CHARGED","statusCode":200}`.

### 6. Void

```bash
grpcurl -plaintext "${H[@]}" -H "x-request-id: r8" -d '{
  "merchant_void_id":"pr-void-2","connector_transaction_id":"GStG0zbWA7b0SACG7lbSA4nnAfKb",
  "cancellation_reason":"e2e"
}' localhost:8000 types.PaymentService/Void
```
```json
{"status":"VOIDED","statusCode":200}
```

### Call trace

Distinct Saferpay calls, keyed by the `RequestId` actually sent:

```
Transaction/AuthorizeDirect  RequestId=pr-rf-2
Transaction/Capture          RequestId=pr-rf-2
Transaction/Refund           RequestId=pr-rf-2
Transaction/Inquire          RequestId=pr-rf-2      <- 2nd RSync reads back
Transaction/Initialize       RequestId=pr-3ds-1     <- 3DS leg 1
Transaction/Authorize        RequestId=pr-3ds-1     <- 3DS leg 2 (settle)
Transaction/Capture          RequestId=pr-3ds-1
Transaction/AuthorizeDirect  RequestId=pr-void-2
Transaction/Cancel           RequestId=pr-void-2
```

`RequestId` is the stable merchant reference rather than a per-call UUID, so a retried call
after a client timeout is deduplicated by Saferpay instead of authorizing twice.

A 3DS payment is exactly **two** Saferpay calls to authorize — `Initialize` then
`Authorize {Token}` — plus `Capture` to settle. No `Inquire` between the legs.

### Not covered

- **Automatic capture** is rejected at authorize time (`CaptureMethodNotSupported`): Saferpay has
  no sale mode — no capture field on its authorize calls, no combined endpoint, and a terminal's
  only "auto" switch is `AutoCloseDailyStatement`, which closes the daily batch of *already
  captured* transactions.
- **Webhooks** are not implemented — Saferpay's Transaction interface offers only unauthenticated,
  bodyless `NotifyUrl` pings, so reconciliation goes through PSync/RSync.
- A declined-card path and a failed/abandoned 3DS challenge (`transStatus: N`) were not exercised.

No credentials are included in source or in this description.


shuklatushar226:
Follow-up commit `9d661756c` fixes a defect in the 3DS second leg that the original implementation would have hit in production.

`pending_three_ds_token` gated the token-based `POST /Transaction/Authorize` on `PaymentsSyncData::encoded_data`. That carrier is never populated for Saferpay: hyperswitch fills `encoded_data` with the raw query string of the redirect return URL (`payments.rs`: `param: Some(req.query_string().to_string())`), and Saferpay invokes the return URL with **no query parameters** — the session `Token` is only ever handed back in the `Initialize` response body.

Verified against the live test API. With `encoded_data` empty, PSync falls through to `Inquire` using the session token as the `TransactionId`:

```
POST /Payment/v1/Transaction/Inquire
{"TransactionReference":{"TransactionId":"eauznhl1au6w1frev3h4jo0ku"}}

{"Behavior":"DO_NOT_RETRY","ErrorName":"TRANSACTION_NOT_FOUND","ErrorMessage":"Transaction not found"}
```

so every 3DS payment would have been stuck in `AuthenticationPending`.

The token now comes from `connector_feature_data` — the `connector_metadata` this connector already publishes from `Initialize`, which round-trips as `payment_attempt.connector_metadata` → `PaymentsSyncData.connector_meta` → gRPC `connector_feature_data`. Once the second leg succeeds the sync response replaces that metadata with the capture-id blob, so the token disappears and later syncs correctly fall through to `Inquire`.

Hyperswitch-side registration PR: https://github.com/juspay/hyperswitch/pull/13873

XyneSpaces:
[should-fix] The generated Saferpay support table marks non-card payment methods as partially supported, but the connector rejects non-card `PaymentMethodData` with `NotImplemented("Only card payments are supported by saferpay")`. Please regenerate the connector docs so unsupported methods render as unsupported instead of `⚠`.

Location: `docs-generated/connectors/saferpay.md` lines 159-257.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1493 @@
...
+
+/// `Transaction.Type` for a payment leg.
+const TRANSACTION_TYPE_PAYMENT: &str = "PAYMENT";
+
+/// `Transaction.Status` values. Saferpay spells the cancelled state `CANCELED`
+/// with a single `L`.
+const STATUS_AUTHORIZED: &str = "AUTHORIZED";
+const STATUS_CAPTURED: &str = "CAPTURED";
+const STATUS_PENDING: &str = "PENDING";
+const STATUS_CANCELED: &str = "CANCELED";
```
AmitsinghTanwar007: can't we have a enum for this

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1493 @@
...
+}
+
+/// `ExpYear` and `ExpMonth` are JSON **numbers**, not strings — the single most
+/// common Saferpay integration trap.
+#[derive(Debug, Clone, Serialize)]
+pub struct SaferpayCardDetails<T: PaymentMethodDataTypes> {
+    #[serde(rename = "Number")]
+    pub number: RawCardNumber<T>,
+    #[serde(rename = "ExpYear")]
+    pub exp_year: u16,
```
swangi-kumari: this should be secret

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1493 @@
...
+    #[serde(rename = "ExpMonth")]
+    pub exp_month: u8,
```
swangi-kumari: this should be secret

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1493 @@
...
+#[serde(untagged)]
+pub enum SaferpayAuthorizeRequest<T: PaymentMethodDataTypes> {
+    /// `POST /Payment/v1/Transaction/AuthorizeDirect`
+    Direct(Box<SaferpayCardAuthorizationRequest<T>>),
+    /// `POST /Payment/v1/Transaction/Authorize`
+    Settle {
+        #[serde(rename = "RequestHeader")]
+        request_header: SaferpayRequestHeader,
+        #[serde(rename = "Token")]
+        token: String,
```
swangi-kumari: What is this token used for? Should this be treated as a secret/sensitive value?

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Done — both string sets are enums now.

`Transaction.Type` → `SaferpayTransactionType` and `Transaction.Status` → `SaferpayTransactionStatus`, deserialized by serde and matched exhaustively instead of `&str` comparisons against four consts (where a typo would have fallen through to the catch-all).

Both carry `#[serde(other)] Unknown`, so a value Saferpay introduces later degrades to "still in flight" rather than failing the whole response parse.

Verified live after the change: Authorize → `AUTHORIZED`, Capture → `CAPTURED`, PSync → `Type: PAYMENT` / `Status: CAPTURED`.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Done — now `Secret<u16>` / `Secret<u8>`, matching the shape novalnet and silverflow use for numeric expiry.

They were bare `u16`/`u8`, so they rendered in full anywhere the request was logged. The Saferpay wire format is unchanged — still JSON numbers, which is what Saferpay requires; masking only rewrites Debug/logging output.

Confirmed live: the request log now shows `"ExpYear":"*** u16 ***","ExpMonth":"*** u8 ***"` while Saferpay still returns `AUTHORIZED`, so it is masked in logs and intact on the wire.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Done — now `Secret<u16>` / `Secret<u8>`, matching the shape novalnet and silverflow use for numeric expiry.

They were bare `u16`/`u8`, so they rendered in full anywhere the request was logged. The Saferpay wire format is unchanged — still JSON numbers, which is what Saferpay requires; masking only rewrites Debug/logging output.

Confirmed live: the request log now shows `"ExpYear":"*** u16 ***","ExpMonth":"*** u8 ***"` while Saferpay still returns `AUTHORIZED`, so it is masked in logs and intact on the wire.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Good catch — yes, it is sensitive, and it is `Secret<String>` now.

That is the Saferpay session token returned by `Transaction/Initialize` at the start of a 3DS payment. Presenting it to `Transaction/Authorize` is what finalises the authentication **and** authorizes the payment, so anyone holding it can complete the charge — a bearer credential, not just a correlation id.

Its other copy, in `connector_feature_data` (how it survives the redirect), was already `Secret`-wrapped; this was the one remaining place it sat in plaintext.

crates/integrations/connector-integration/src/connectors/saferpay.rs
```diff
@@ -0,0 +1,572 @@
...
+
+        fn get_url(
+            &self,
+            req: &RouterDataV2<RSync, RefundFlowData, RefundSyncData, RefundsResponseData>,
+        ) -> CustomResult<String, IntegrationError> {
+            // A Saferpay refund is created `AUTHORIZED` and never settles by itself;
+            // it needs an explicit Capture against the refund's own transaction id.
+            // There is no "capture a refund" flow in UCS, so RSync issues it while
+            // the refund is still pending, then falls back to Inquire once settled.
+            let path = if saferpay::refund_needs_settlement(&req.request) {
```
JeevaRamu0104, line 479: `RefundSyncData::refund_status` is hardcoded to `Pending` for every `RefundService/Get` (`domain_types/src/types.rs:8744`), so this always picks `PATH_CAPTURE` and `Inquire` is unreachable. The first sync settles the refund; every later one re-POSTs Capture, gets 402 `TRANSACTION_IN_WRONG_STATE`, and a settled refund can never be read back. Can this key off a stage marker in `refund_connector_metadata` instead, the way the 3DS token round-trips?

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1524 @@
...
+    #[serde(rename = "RetryIndicator")]
+    pub retry_indicator: u8,
+}
+
+impl SaferpayRequestHeader {
+    fn new(auth: &SaferpayAuthType) -> Self {
+        Self {
+            spec_version: SAFERPAY_SPEC_VERSION.to_string(),
+            customer_id: auth.customer_id.clone(),
+            request_id: uuid::Uuid::new_v4().to_string(),
```
JeevaRamu0104: Saferpay dedupes on `RequestId` + `RetryIndicator`, and a fresh UUID per call opts out of it — a retried `AuthorizeDirect` after a client timeout authorizes twice.

`PaymentFlowData`/`RefundFlowData` carry `merchant_request_id` for exactly this (`resource_common_data.get_merchant_request_id()`, used in `twoc_twop_paco/transformers.rs:889`); falling back to `connector_request_reference_id` keeps it stable across retries.

crates/integrations/connector-integration/src/connectors/saferpay.rs
```diff
@@ -0,0 +1,572 @@
...
+        }
+
+        fn get_url(
+            &self,
+            req: &RouterDataV2<PreAuthenticate, PaymentFlowData, PaymentsPreAuthenticateData<T>, PaymentsResponseData>,
+        ) -> CustomResult<String, IntegrationError> {
+            Ok(format!(
+                "{}{}",
+                self.connector_base_url_payments(req),
+                PATH_INITIALIZE
```
JeevaRamu0104, line 330: The grpcurl block in the PR description predates this flow — it was written when the 3DS second leg ran from PSync on `encoded_data`, before `dd0b4c33` moved it onto PreAuthenticate. So `Initialize` here and the settle `Authorize` have no live verification, and the description notes the challenge was never completed. Can you re-run the 3DS legs against the current code?

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1524 @@
...
+            network_advice_code: None,
+            network_error_message: self.processor_message.clone(),
+            raw_connector_response: None,
+            raw_connector_request: None,
+            typed_connector_response: None,
+            typed_connector_request: None,
+        }
+    }
+
+    pub fn to_refund_error_response(&self, status_code: u16) -> ErrorResponse {
```
JeevaRamu0104, line 763: `to_refund_error_response` is never called — `ConnectorCommon::build_error_response` is flow-agnostic and always uses `to_error_response`, which leaves `attempt_status: None`. A hard-declined refund stays `Pending` and keeps getting retried.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1524 @@
...
+}
+
+impl SaferpayCaptureResponse {
+    fn attempt_status(&self, is_partial: bool) -> AttemptStatus {
+        match self.status.unwrap_or(SaferpayTransactionStatus::Unknown) {
+            SaferpayTransactionStatus::Captured if is_partial => AttemptStatus::PartialCharged,
+            SaferpayTransactionStatus::Captured => AttemptStatus::Charged,
+            SaferpayTransactionStatus::Authorized
+            | SaferpayTransactionStatus::Pending
+            | SaferpayTransactionStatus::Canceled
```
JeevaRamu0104: A capture Saferpay reports as `CANCELED` is terminal, but it maps to `Pending` here, so the attempt polls forever. Needs its own arm mapping to `AttemptStatus::Failure`.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1524 @@
...
+    fn try_from(
+        item: ResponseRouterData<SaferpayCaptureResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let response = item.response;
+        let request = &item.router_data.request;
+
+        // `minor_amount_authorized` is a response-reporting field and is `None` on
+        // every request path, so a capture can only be classified as partial when
+        // the caller told us the payment total.
+        let is_partial = item
```
JeevaRamu0104: When `resource_common_data.amount` is `None`, a genuine partial capture reports `Charged` rather than `PartialCharged`, so the payment looks fully settled. Is there another carrier for the authorized total here, or should an unknown total report `PartialCharged` whenever `minor_amount_to_capture` was explicitly supplied?

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
```diff
@@ -0,0 +1,1524 @@
...
+pub const CAPTURE_ID_METADATA_KEY: &str = "capture_id";
+/// Marks which leg of the flow last wrote the metadata blob.
+pub const SAFERPAY_STAGE_METADATA_KEY: &str = "stage";
+/// Stage written by the 3DS `Initialize` leg, while the session token is live.
+pub const STAGE_INITIALIZED: &str = "initialized";
+/// Key under which the 3DS `Initialize` response publishes the Saferpay session
+/// token in `connector_metadata`.
+pub const SAFERPAY_TOKEN_METADATA_KEY: &str = "saferpay_token";
+
+fn context() -> IntegrationErrorContext {
```
JeevaRamu0104, line 92: nit: `IntegrationErrorContext::default()` everywhere means these errors carry no `suggested_action` or `doc_url` — `capture_method_not_supported` below shows what the filled-in version buys.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/saferpay.rs
shuklatushar226: Fixed in 2a1d7f116. RSync now uses a refund settlement stage marker in refund_connector_metadata. The first successful Capture response marks the stage, and subsequent RSync calls use Inquire, so settled refunds no longer replay Capture.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Fixed in 2a1d7f116. Saferpay RequestId is now derived from merchant_request_id, falling back to connector_request_reference_id when empty. Retries therefore reuse the same RequestId; RetryIndicator remains 0 because this flow does not expose a retry counter.

crates/integrations/connector-integration/src/connectors/saferpay.rs
shuklatushar226: Per the requested scope, I did not rerun the 3DS browser flow in this update. The workspace has no Saferpay sandbox credentials, so I cannot honestly claim fresh live evidence. The PR body now marks this explicitly instead of treating the stale evidence as current.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Fixed in 2a1d7f116. Refund and RSync now override the generic payment error mapper and call to_refund_error_response, so hard refund declines produce terminal refund failure metadata instead of remaining Pending.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Fixed in 2a1d7f116. Saferpay CANCELED capture responses now map to AttemptStatus::Failure, preventing terminal canceled captures from being polled forever.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Fixed in 2a1d7f116. Capture reads the authorized total from the common amount or connector_feature_data.authorized_amount. If it is unavailable, the response conservatively stays PartialCharged; it only reports Charged when the captured amount is known to cover the authorized total.

crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs
shuklatushar226: Fixed in 2a1d7f116. Saferpay integration errors now include a suggested action and the official Transaction API documentation URL, including the capture-method-not-supported path.

crates/integrations/connector-integration/src/connectors/saferpay.rs
shuklatushar226: Re-run against this branch head (`67abeb19e`) and the PR description now carries the real transcripts — you were right that the old block predated the flow.

3DS is verified end to end with a genuinely completed challenge, not a stub:

- `PaymentMethodAuthenticationService/PreAuthenticate` → `AUTHENTICATION_PENDING`, redirect form, and `connectorFeatureData` = `{"saferpay_token":"...","stage":"initialized"}`
- browser leg driven for real through all three Saferpay-hosted pages (DCC → device-data collection, which is a JS SPA and needs a real browser → ACS). The test ACS takes no password; it returns a `cres` with `transStatus: "Y"`
- settle `PaymentService/Authorize` with `redirection_response` + the token → `AUTHORIZED`, real transaction id
- `Capture` on it → `CHARGED`

Endpoint trace for a 3DS payment is exactly `Initialize` → `Authorize` → `Capture` — two calls to authorize, no `Inquire` between the legs.

The refund thread is covered there too: `Refund` → `Capture` (settle, from RSync #1) → **`Inquire`** (RSync #2), so a settled refund reads back instead of re-capturing.

swangi-kumari (approved):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2184, opened 2026-08-27
fix(ucs): record res_code as integer and populate response body/headers on error path in outgoing golden log

kanikac199:
## Summary

Fixes three gaps in UCS outgoing golden log lines so they match the euler `OutgoingLogEntry` schema:

- **`res_code` as integer**: Euler expects `res_code :: Int`. Previously only `response.status_code` was recorded, via `tracing::field::display()` which stores as a `String`. Added a dedicated `res_code` span field recorded via `u64::from(status_code)` which routes through `record_u64` in the tracing subscriber → stored as `serde_json::Number` → integer in log output. Applied on both success and error paths.

- **`response.body` on error path**: Was only emitted on success. Changed the event borrow from a consuming move to `as_deref_mut()` so the event is accessible after `set_error_response`, then recorded `evt.error` (which `set_error_response` populates) as `response.body`.

- **`response.headers` on error path**: Was only emitted on success. Now recorded from `body.headers: Option<http::HeaderMap>` available in the `Err(body)` arm, inline-converted to `serde_json::Value::Object`.

## Test plan

- [ ] Deploy to staging, trigger connector calls that succeed (2xx) and fail (4xx/5xx)
- [ ] In Kibana, filter for `message_: "Golden Log Line (outgoing)"` and verify `res_code` is an integer (not a string)
- [ ] Verify `response.body` and `response.headers` are present on error-path golden log lines
- [ ] Confirm Unified LogViewer / LP ingests the updated format correctly

### DIFF
--- a/crates/common/external-services/src/service.rs
+++ b/crates/common/external-services/src/service.rs
@@ -476,25 +476,17 @@ where
 
                     if let Some(evt) = event {
                         evt.set_error_response(&error_response);
+                        let mut json_fields: Vec<(&'static str, serde_json::Value)> = Vec::new();
                         if let Some(error_data) = &evt.error {
-                            record_json_fields_on_span(vec![(
-                                "response.body",
-                                error_data.inner().clone(),
-                            )]);
+                            json_fields.push(("response.body", error_data.inner().clone()));
                         }
-                    }
-                    if let Some(response_headers) = &body.headers {
-                        let headers_json: serde_json::Value = response_headers
-                            .iter()
-                            .map(|(k, v)| {
-                                (
-                                    k.as_str().to_string(),
-                                    serde_json::Value::String(v.to_str().unwrap_or("").to_string()),
-                                )
-                            })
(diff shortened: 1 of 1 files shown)


crates/common/external-services/src/service.rs
```diff
@@ -475,6 +476,25 @@ where
...
+                            .iter()
+                            .map(|(k, v)| {
+                                (
+                                    k.as_str().to_string(),
+                                    serde_json::Value::String(v.to_str().unwrap_or("").to_string()),
+                                )
+                            })
+                            .collect::<serde_json::Map<_, _>>()
+                            .into();
+                        record_json_fields_on_span(vec![("response.headers", headers_json)]);
```
AmitsinghTanwar007: can we make sure the headers value are masked

AmitsinghTanwar007 (approved):

Aishwariyaa-Anand (approved):

PR #2187, opened 2026-08-27
feat(connector): [JPMORGAN ORBITAL] add JP Morgan Orbital connector with Card Authorize (3DS + non-3DS)

shuklatushar226:
## Summary

Adds a new connector `jpmorganorbital` / `ConnectorEnum::JpmorganOrbital` / display name **"JP Morgan Orbital"**, implementing **Card Authorize** (non-3DS and 3DS) plus **PSync**.

This is the JPMorgan **Orbital Gateway** JSON API v4 (`orbitalvar1.chasepaymentech.com/gwapi/v4/gateway`) — a **different product** from the existing `jpmorgan` connector (Payments API v2, `api-ms.payments.jpmorgan.com`). The existing `jpmorgan` connector is **untouched**. The naming mirrors the existing `worldpay` / `worldpayvantiv` sibling-suffix convention.

## Scope

Card one-time payments only. **Out of scope for this PR:** mandates / MIT / recurring, wallets, bank transfers, bank debits, and BNPL.

**Auto-capture only.** `CaptureMethod::Manual` is refused with `NotSupported`. Manual capture would send `transType: "A"`, holding funds at the issuer with no way to settle or release them, because Capture, Void and Refund are all `not_implemented` here — so it is refused rather than stranding an authorization the caller cannot act on. `Automatic` and `SequentialAutomatic` both map to `"AC"`.

## 3DS model

Orbital is **external-3DS passthrough only**. There is no Orbital-hosted challenge, no ACS redirect, and no redirect URL in the response. Authorize is a **single synchronous call** for both non-3DS and 3DS; externally-produced authentication data is mapped onto the Orbital `cryptogram` and `additionalAuthInfo` objects. No `RedirectForm` flow exists.

## Auth

Three plain HTTP headers — `orbitalConnectionUsername`, `orbitalConnectionPassword`, `merchantID` — plus `merchant.bin` and `merchant.terminalID` in the body, supplied via `ConnectorSpecificConfig::JpmorganOrbital`.

## Notable correctness details

These are non-obvious and worth reviewer attention:

- **`order.amount` uses two implied decimals for EVERY currency**, including zero-exponent ones. ¥100 is sent as `10000`, not `100`. A plain `StringMinorUnit` would have been a **100x under-charge** for JPY/KRW. Currency is never sent — it is implied by the MID. The encoding now lives in the shared `common_utils::types::StringTwoDecimalUnit` / `StringTwoDecimalUnitForConnector`; Orbital's own unsigned and 12-character rules stay at the call site in `build_amount`, each with its own error reason so the three ways an amount can be rejected stay distinguishable at the gRPC boundary.
- **A wrong-currency charge is guarded, request field first.** Orbital has no currency field, so a EUR request routed to a USD MID would be authorized as USD at the same numeric amount. `validate_currency` reads `PaymentsAuthorizeData::merchant_config_currency` first and the connector-config copy only as a fallback — the precedence Braintree already uses. Configured in neither place, it is a no-op.
- **Mastercard ECI scale inversion.** UCS ECI `02` maps to Orbital `authenticationECIInd` `5`, not `2` (Orbital's Mastercard scale is 5/6/7). Sending `2` would be read by Orbital as "recurring CDPT".
- **`order.retryTrace` is both Orbital's idempotency key and the PSync lookup key** (`/inquiry` → `order.inquiryRetryNumber`). It is numeric and ≤16 chars, so a UUID does not fit. It is derived **deterministically** — SHA-256 of the attempt reference, the leading 8 bytes folded into `[1e15, 1e16)`. SHA-256 rather than a short non-cryptographic hash because the fold keeps only the low bits, which is exactly where a hash like FNV-1a is weakest, and a collision inside Orbital's 48-hour window makes the gateway echo another payment's approval back as this one's result.
- **PSync keys `/inquiry` off the persisted trace, not a re-derivation.** Authorize persists `retry_trace` and `order_id` into `connector_metadata`; PSync reads them back off `connector_feature_data` and only re-derives when the caller echoes nothing. Re-deriving unconditionally meant a sync that passed a different reference derived different keys, found no record, and reported a charged payment as failed.
- **An incomplete `/inquiry` is `Unresolved`, never `Failure`.** On `/inquiry` a non-zero `procStatus` means the *lookup* did not complete — no record for this `retryTrace`, past the 48-hour window, or a mismatched key — and says nothing about the payment. A found-and-declined payment answers `procStatus "0"` with `approvalStatus "0"` instead, and only that is recorded as terminal. Authorize keeps the old reading, where a gateway rejection does mean the payment never processed.
- **`txRefIdx` is persisted alongside `txRefNum`** for future Capture/Void.
- **`procStatus` / `respCode` / `avsRespCode` / `cvvRespCode` are treated as opaque strings.** The exhaustive tables are not published in any machine-readable JPM asset, so no enum was invented.
- **The production base URL is UNCONFIRMED.** JPM does not publish it. It is committed with an explicit comment to confirm with JPM Merchant Services during onboarding.

## Files

**Added**
- `crates/integrations/connector-integration/src/connectors/jpmorganorbital.rs`
- `crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs`
- `crates/internal/integration-tests/src/connector_specs/jpmorganorbital/specs.json`
- `data/field_probe/jpmorganorbital.json`
- `docs-generated/connectors/jpmorganorbital.md`, `examples/jpmorganorbital/*`

**Modified**
- `crates/common/common_utils/src/types.rs`, `crates/common/common_utils/src/lib.rs` — the shared two-implied-decimal amount type
- `crates/types-traits/grpc-api-types/proto/payment.proto`
- `crates/internal/integration-tests/src/connector_specs/alpha_connectors.json`
- `config/development.toml`, `config/sandbox.toml`, `config/production.toml`, `config/superposition.toml`, `.typos.toml`
- `scripts/generators/snippet_examples/generate.py` — the connector-id-to-proto-identifier map, which also regenerates `docs-generated/` and `examples/` output for `absasanlam`, `pinelabsonline` and `tsystransit`
- `sdk/javascript/src/payments/_generated_grpc_client.ts`, `docs-generated/all_connector.md`, `docs-generated/llms.txt`

## Testing

> **⚠️ Please read: no real sandbox credentials exist.**
> **NO real JPMorgan Orbital sandbox credentials were available for this integration.** Validation was therefore performed against a **LOCAL MOCK Orbital server** that captured the exact outbound HTTP request the connector produces. The captured outbound requests below are the real evidence of correctness — they show byte-for-byte what would go on the wire to Orbital. This has **not** been executed against JPMorgan's live VAR environment.

`cargo build` passes on the full workspace with **zero warnings** from the new code, and `cargo clippy -p connector-integration --all-targets` is clean.

<details>
<summary><b>grpcurl calls and responses (credentials redacted)</b></summary>

> Captured against the mock at the revision noted on each run.

```
--- (1) Card Authorize, non-3DS (Mastercard, AUTOMATIC -> transType "AC") ---
grpcurl -plaintext \
  -H 'x-connector: jpmorgan_orbital' \
  -H 'x-merchant-id: <REDACTED>' \
  -H 'x-request-id: req-jpmorganorbital-non3ds-001' \
  -H 'x-connector-config: {"config":{"JpmorganOrbital":{"username":"<REDACTED>","password":"<REDACTED>","merchant_id":"<REDACTED>","bin":"<REDACTED>","terminal_id":"<REDACTED>"}}}' \
  -d '{
    "merchant_transaction_id": "hs_orbital_non3ds_001",
    "amount": { "minor_amount": 10000, "currency": "USD" },
    "payment_method": { "card": {
        "card_number": { "value": "<REDACTED>" },
        "card_exp_month": { "value": "06" },
        "card_exp_year": { "value": "2030" },
        "card_cvc": { "value": "<REDACTED>" },
        "card_holder_name": { "value": "John Doe" },
        "card_network": "MASTERCARD" } },
    "auth_type": "NO_THREE_DS",
    "enrolled_for_3ds": false
  }' \
  localhost:8000 types.PaymentService/Authorize

Response:
{
  "merchantTransactionId": "MOCKORDER",
  "connectorTransactionId": "625EC0E49LK88FCB0906BB8AAD435F42A2FC553E0",
  "status": "CHARGED",
  "statusCode": 200,
  "connectorReferenceId": "MOCKORDER"
}

--- (2) Card Authorize, 3DS ---
REMOVED. This run was captured with "capture_method": "MANUAL" and reported
transType "A" / AUTHORIZED. Manual capture is now refused with NotSupported
(see Scope), so that response is no longer reachable and has been deleted
rather than restated. It has NOT been re-captured — there are no credentials
and the mock was ad hoc. The outbound 3DS request body below is retained with
its own provenance note.

--- (3) HTTP 200 issuer decline (approvalStatus "0") ---
{
  "connectorTransactionId": "625EC0E49LK88FCB0906BB8AAD435F42A2FC553E1",
  "status": "FAILURE",
  "error": { "connectorDetails": { "code": "05", "message": "Do Not Honor", "reason": "05" } },
  "statusCode": 200
}
```

</details>

<details>
<summary><b>Captured outbound HTTP requests — the key evidence (credentials redacted)</b></summary>

```
--- (1) NON-3DS ---
POST /gwapi/v4/gateway/payments HTTP/1.1
Host: <orbital-host>
merchantid: <REDACTED>
orbitalconnectionusername: <REDACTED>
orbitalconnectionpassword: <REDACTED>
via: HyperSwitch

{
  "version": "5.2",
  "transType": "AC",
  "merchant": { "bin": "<REDACTED>", "merchantID": "<REDACTED>", "terminalID": "<REDACTED>" },
  "paymentInstrument": { "card": { "ccAccountNum": "<REDACTED>", "ccExp": "203006" } },
  "order": { "orderID": "hs-orbital-non3ds-001", "amount": "10000", "industryType": "EC", "retryTrace": "5965648070021058" },
  "cardholderVerification": { "ccCardVerifyNum": "<REDACTED>", "ccCardVerifyPresenceInd": "1" }
}

--- (2) 3DS (Visa Secure) ---
PROVENANCE: captured before Manual capture was refused, so this body reads
transType "A". `capture_method` feeds exactly one wire field — `transType`,
via `trans_type_for` — so the same payment sent today with AUTOMATIC produces
"AC" and every other field below is unchanged. The cryptogram /
additionalAuthInfo mapping this capture exists to show is unaffected.

POST /gwapi/v4/gateway/payments HTTP/1.1
Host: <orbital-host>
merchantid: <REDACTED>
orbitalconnectionusername: <REDACTED>
orbitalconnectionpassword: <REDACTED>
via: HyperSwitch

{
  "version": "5.2",
  "transType": "A",
  "merchant": { "bin": "<REDACTED>", "merchantID": "<REDACTED>", "terminalID": "<REDACTED>" },
  "paymentInstrument": { "card": { "ccAccountNum": "<REDACTED>", "ccExp": "203006" } },
  "order": { "orderID": "hs-orbital-3ds-001", "amount": "10000", "industryType": "EC", "retryTrace": "9294143268873549" },
  "cardholderVerification": { "ccCardVerifyNum": "<REDACTED>", "ccCardVerifyPresenceInd": "1" },
  "cryptogram": { "verifyByVisaCAVV": "<REDACTED>", "verifyByVisaXID": "<REDACTED>" },
  "additionalAuthInfo": { "authenticationECIInd": "5" }
}

--- (3) JPY 100 (zero-exponent) - proves the amount rule ---
"order": { "orderID": "hs-orbital-jpy-001", "amount": "10000", "industryType": "EC", "retryTrace": "3872083475311052" }
(minor_amount 100 + currency JPY -> "10000"; plain StringMinorUnit would have emitted "100")

--- (4) Mastercard 3DS with UCS eci "02" - proves the ECI scale inversion ---
"cryptogram": { "mcSecureCodeAAV": "<REDACTED>" },
"additionalAuthInfo": { "authenticationECIInd": "5", "ucafInd": "2", "mcProgramProtocol": "2",
                        "mcDirectoryTransID": "f38e6948-5388-41a6-bca4-b49723c19437" }
(UCS ECI "02" -> Orbital "5", not "2")
```

</details>

## Validation checklist

- [x] `cargo build` passes on the full workspace with zero warnings from the new code
- [x] `cargo clippy -p connector-integration --all-targets` clean; `cargo fmt --check` clean
- [x] Authorize (non-3DS and 3DS) and PSync exercised end-to-end against a local mock Orbital server, with each capture's provenance recorded above
- [x] No credentials, no `localhost` base URLs and no card numbers in the committed diff
- [x] The existing `jpmorgan` connector is untouched
- [ ] Validated against JPMorgan's real VAR/sandbox environment — **blocked on credentials**
- [ ] Production base URL confirmed with JPM Merchant Services


crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1200 @@
...
+pub struct JpmorganOrbitalPaymentInstrument<T: PaymentMethodDataTypes> {
+    pub card: JpmorganOrbitalCard<T>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct JpmorganOrbitalOrder {
+    #[serde(rename = "orderID")]
+    pub order_id: String,
+    /// Two implied decimals for every currency. See [`orbital_amount`].
+    pub amount: String,
```
swangi-kumari: amount should not be string.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1200 @@
...
+    minor_amount: MinorUnit,
+    currency: Currency,
+) -> Result<String, error_stack::Report<IntegrationError>> {
+    let major = JpmorganOrbitalAmountConvertor::convert(minor_amount, currency)?;
+    shift_two_decimal_places(&major.get_amount_as_string())
+}
+
+/// Multiply a decimal string by 100 without floating point, and without a decimal
+/// point, sign, separator or currency symbol in the result.
+fn shift_two_decimal_places(major: &str) -> Result<String, error_stack::Report<IntegrationError>> {
```
swangi-kumari: why is this required?

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1200 @@
...
+
+/// Sanitize a UCS reference into an `order.orderID`.
+///
+/// Orbital allows `a-z A-Z 0-9 - , $ @ &` and space, forbids a leading space, caps
+/// the field at 22 characters and requires the **first 8** to be unique per
+/// transaction. UCS reference ids routinely contain `_` and share a long common
+/// prefix, so disallowed characters are mapped to `-` and an over-long reference is
+/// truncated from the **left** — the tail is where UCS ids actually differ, so
+/// keeping it preserves the first-8-unique property that keeping the head destroys.
+fn build_order_id(reference: &str) -> Result<String, error_stack::Report<IntegrationError>> {
```
swangi-kumari, line 218: why is this required?

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1367 @@
...
+    fn convert_back(
+        &self,
+        amount: Self::Output,
+        currency: Currency,
+    ) -> Result<MinorUnit, error_stack::Report<ParsingError>> {
+        let wire = amount.get_amount_as_string().parse::<i128>().map_err(|_| {
+            amount_parse_error(format!(
+                "`{}` is not an integer",
+                amount.get_amount_as_string()
+            ))
```
AmitsinghTanwar007: This amount conversion can be moved to a reusable common amount type because the
transformation itself is generic: "major amount with exactly two decimal places,
serialized without the decimal point".
Formula:
wire = minor * 100 / 10^currency_exponent
Example: USD minor 1234 -> "1234", JPY minor 1234 -> "123400".
Connector-specific constraints such as "amount must be unsigned" or "amount must
fit within N characters" should stay outside the converter as validations on the
converted amount. We can expose small helper methods on the common amount wrapper,
 e.g. `validate_unsigned(field_name)` and `validate_max_len(max_len, field_name)`,
 so connectors can reuse the checks while keeping their own field constraints
 explicit at the call site.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1367 @@
...
+        Ok(Self {
+            version: ORBITAL_VERSION.to_string(),
+            merchant: JpmorganOrbitalMerchant {
+                bin: auth.bin.clone(),
+                merchant_id: auth.merchant_id.clone(),
+                terminal_id: auth.terminal_id.clone(),
+            },
+            order: JpmorganOrbitalInquiryOrder {
+                order_id: Some(build_order_id(reference)?),
+                inquiry_retry_number: build_retry_trace(reference),
```
JeevaRamu0104: `inquiryRetryNumber` is re-derived from `merchant_transaction_id`, so PSync only finds the payment if the sync caller resends the identical value. A different one makes `/inquiry` return no record, which maps to `Failure` — a charged payment reported as failed, on the exact path meant to prevent a double charge. Authorize already persists `retry_trace` into `connector_metadata`; can PSync read it back off `encoded_data` instead of re-deriving?

The generated example shows the mismatch is reachable — it authorizes with `probe_txn_001` and syncs with `probe_merchant_txn_001`.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1367 @@
...
+        if request.mandate_id.is_some() || request.setup_mandate_details.is_some() {
+            return Err(error_stack::report!(IntegrationError::NotSupported {
+                message: "Mandates / merchant-initiated transactions".to_string(),
+                connector: "jpmorganorbital",
+                context: IntegrationErrorContext::default(),
+            }));
+        }
+
+        let auth = JpmorganOrbitalAuthType::try_from(&router_data.connector_config)?;
+        validate_currency(request.currency, auth.merchant_config_currency.as_deref())?;
```
JeevaRamu0104: This reads only the connector-config copy, but `PaymentsAuthorizeData` already carries `merchant_config_currency`, and Braintree prefers the per-request value with config as fallback (`braintree/transformers.rs:437`). A caller that sets it on the request gets no currency guard at all — the thing this field exists to prevent.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1367 @@
...
+        // concerned; the framework groups it with `Automatic` in
+        // `PaymentsSyncData::is_auto_capture`.
+        None | Some(CaptureMethod::Automatic) | Some(CaptureMethod::SequentialAutomatic) => {
+            Ok(TRANS_TYPE_AUTH_CAPTURE)
+        }
+        // Manual capture would send `transType: "A"`, holding funds at the issuer with
+        // no way to settle or release them: Capture, Void and Refund are all
+        // `not_implemented` for this connector. Refuse it rather than stranding an
+        // authorization the caller cannot act on.
+        Some(other) => Err(error_stack::report!(IntegrationError::NotSupported {
```
JeevaRamu0104, line 543: This rejects `CaptureMethod::Manual`, but the PR description's captured run (2) shows MANUAL producing `transType "A"` and `AUTHORIZED`, and `success_status`'s non-`AC` branch is unreachable as a result. Since the local mock is this PR's only validation, was that evidence captured from a different revision?

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1367 @@
...
+}
+
+fn map_eci(eci: &str, brand: OrbitalBrand) -> Option<&'static str> {
+    let trimmed = eci.trim();
+    let unpadded = trimmed.trim_start_matches('0');
+    let normalized = if unpadded.is_empty() { "0" } else { unpadded };
+
+    match brand {
+        OrbitalBrand::Mastercard => match normalized {
+            "2" => Some("5"), // full authentication
```
JeevaRamu0104: nit: Mastercard accepts only ECI 0/1/2 here, so a 3DS server that normalizes Mastercard onto the Visa scale (05/06/07) gets `NotSupported`. It fails closed, so just confirming that's intended.

config/production.toml
```diff
@@ -68,6 +68,10 @@ interpayments.base_url = "https://api.interpayments.com/v1"
 twoc_twop_paco.base_url = "https://core.paco.2c2p.com"
 jpmorgan.base_url = "https://api-ms.payments.jpmorgan.com/api/v2"
 jpmorgan.secondary_base_url = "https://id.payments.jpmorgan.com"
+# NOTE: UNCONFIRMED - JPMorgan does not publish the production Orbital hostname
+# anywhere. Inferred from the Chase Paymentech naming convention and corroborated
+# only by public DNS. Confirm with JPMorgan Merchant Services during onboarding.
+jpmorganorbital.base_url = "https://orbital1.chasepaymentech.com/gwapi/v4/gateway"
```
JeevaRamu0104, line 75: This hostname is a guess, so the first real production payment routes somewhere nobody has confirmed. Can it ship commented out until JPM Merchant Services confirms it? A missing key fails at config load, which is a much better place to find out than at payment time.

scripts/generators/snippet_examples/generate.py
```diff
@@ -1049,18 +1049,45 @@ def _to_camel_ts(snake: str) -> str:
     "razorpayv2": "RAZORPAY",  # razorpayv2 uses the same RAZORPAY proto enum
 }
 
+# Connectors whose proto identifiers are multi-word while the connector id runs the words
+# together. Both derivations below assume `id -> IDENTIFIER` and `id -> Identifier`, which
+# silently produces names that do not exist (e.g. `JpmorganorbitalConfig`) and drops the
+# generated snippet into its "no proto metadata" placeholder branch. Values are the
+# identifiers as they actually appear in payment.proto: (enum member, PascalCase stem).
+_CONN_PROTO_IDENTIFIERS: dict[str, tuple[str, str, str]] = {
```
JeevaRamu0104, line 1057: This fix is correct, but it also regenerates docs and examples for absasanlam, pinelabsonline and tsystransit — 7 files of unrelated output riding along with a new connector. Worth splitting out so those diffs get reviewed on their own.

JeevaRamu0104 (changes requested):

swangi-kumari (dismissed):

AmitsinghTanwar007 (dismissed):

swangi-kumari (dismissed):

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1297 @@
...
+                    .resource_common_data
+                    .connector_request_reference_id,
+            )
+        });
+
+        if !response.is_success() {
+            return Ok(Self {
+                response: Err(response.to_error_response(item.http_code)),
+                resource_common_data: PaymentFlowData {
+                    status: response.failure_status(),
```
JeevaRamu0104: On `/inquiry`, `procStatus != "0"` means the inquiry didn't complete — no record for the derived `retryTrace`, past the 48h window, or a different `merchant_transaction_id` on the sync. A found-and-declined payment returns `procStatus "0"` with `approvalStatus "0"` instead. So this stamps `Failure` on exactly the charged-but-response-lost payment PSync exists to recover.

```rust
let status = if response.gateway_accepted() {
    response.failure_status()
} else {
    AttemptStatus::Unresolved
};
```

Authorize can keep `failure_status()` as-is — there `!gateway_accepted` does mean the payment never processed.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1297 @@
...
+        if request.mandate_id.is_some() || request.setup_mandate_details.is_some() {
+            return Err(error_stack::report!(IntegrationError::NotSupported {
+                message: "Mandates / merchant-initiated transactions".to_string(),
+                connector: "jpmorganorbital",
+                context: IntegrationErrorContext::default(),
+            }));
+        }
+
+        let auth = JpmorganOrbitalAuthType::try_from(&router_data.connector_config)?;
+        validate_currency(request.currency, auth.merchant_config_currency.as_deref())?;
```
JeevaRamu0104: `PaymentsAuthorizeData` already carries `merchant_config_currency: Option<Currency>` — Braintree reads it at `braintree/transformers.rs:440`. Reading only the config copy means a merchant who sets the standard request field gets no guard, and the wrong-currency charge this exists to prevent still happens. Can it read the request field first and fall back to config?

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1297 @@
...
+                // Merchant ID setup and has no field to receive it.
+                // The encoding is shared; the unsigned and 12-character rules are
+                // Orbital's own field constraints, so they stay visible here.
+                amount: StringTwoDecimalUnitForConnector
+                    .convert(request.minor_amount, request.currency)
+                    .and_then(|amount| amount.validate_unsigned("order.amount"))
+                    .and_then(|amount| amount.validate_max_len(MAX_AMOUNT_LEN, "order.amount"))
+                    .change_context(IntegrationError::InvalidDataFormat {
+                        field_name: "order.amount",
+                        context: IntegrationErrorContext::default(),
```
JeevaRamu0104: `IntegrationErrorContext::default()` collapses three causes — negative amount, over 12 characters, and a 3-decimal currency that can't be expressed with two — into one opaque `InvalidDataFormat` at the gRPC boundary. The `attach_printable` from `validate_unsigned`/`validate_max_len` never reaches `error_message`. Pass the reason through `additional_context`.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
```diff
@@ -0,0 +1,1297 @@
...
+///
+/// `retryTrace` is both the idempotency key and the only `/inquiry` lookup key, so a
+/// collision inside Orbital's 48-hour window makes the gateway echo a different
+/// payment's approval back as this payment's result. SHA-256 is used rather than a
+/// short non-cryptographic hash because the modulo fold below keeps only the low bits,
+/// which is exactly where a hash like FNV-1a is weakest.
+///
+/// The derivation must stay deterministic: PSync recomputes it from the same reference.
+pub fn build_retry_trace(reference: &str) -> String {
+    let digest = Sha256::digest(reference.as_bytes());
```
JeevaRamu0104, line 360: The PR description says `retryTrace` is FNV-1a/64 folded; this is SHA-256. It also shows validation run #2 with `"capture_method": "MANUAL"` returning `AUTHORIZED`, which `trans_type_for` now rejects. Those mock captures are the only evidence offered for a connector with no sandbox credentials — worth re-syncing the description before merge.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
shuklatushar226: Good catch — fixed in 75ca63a.

Added `inquiry_failure_status`, which is your snippet: `gateway_accepted()` keeps `failure_status()`, everything else is `Unresolved`. Authorize is untouched, as you said it could be.

One thing your snippet did not reach, which I had to fix alongside it: `to_error_response` computes its own `attempt_status` from `self.failure_status()`, so setting only `PaymentFlowData.status` would have left the `ErrorResponse` still carrying `Failure` and the two disagreeing. `to_error_response` is now a thin delegate over `error_response_with_status(http_code, attempt_status)`, and PSync passes the same value into both.

Knowingly **not** fixed: `build_error_response` (`jpmorganorbital.rs:173`) is flow-agnostic, so a *non-2xx* `/inquiry` still routes through `failure_status()` and can be stamped `Failure` by the same reasoning. Differentiating there needs a per-flow `get_error_response_v2` override rather than the default, which is a wider change on a path you have not looked at — happy to do it here or in a follow-up, whichever you prefer.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
shuklatushar226: Agreed — fixed in 75ca63a, with Braintree's precedence.

`validate_currency` now takes `request_config_currency: Option<Currency>` ahead of the config string and uses it when present; the config copy is the fallback, with the existing blank-string-means-unconfigured rule and the same `InvalidConnectorConfig` for an unparseable value. Configured in neither place it stays a no-op, so no existing config changes behaviour.

```rust
validate_currency(
    request.currency,
    request.merchant_config_currency,
    auth.merchant_config_currency.as_deref(),
)?;
```

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
shuklatushar226: Right — fixed in 75ca63a.

The chain is now a `build_amount` free function with one `map_err` per rule, each building its own `additional_context`, so the three causes stay distinguishable at the gRPC boundary:

- 3-decimal currency → `"BHD has more than two decimal places, so 1234 minor units cannot be expressed in the two decimals order.amount implies"`
- negative → `"order.amount is unsigned at Orbital, but -1 minor units of USD is negative"`
- over 12 characters → `"order.amount holds at most 12 characters, but 99999999999999 minor units of USD encodes to \`99999999999999\` (14 characters)"`

I wrote the reasons out rather than forwarding the `attach_printable` from `common_utils`, to match the convention elsewhere in the crate (e.g. `bluesnap.rs:679`) — an `error_stack::Report` `{:?}` dump would have carried location and backtrace noise into `error_message`. The `attach_printable` is still there for the logs. `ResultExt` is dropped from the imports since that was its only use.

crates/integrations/connector-integration/src/connectors/jpmorganorbital/transformers.rs
shuklatushar226: Both correct, and thank you for cross-reading the description against the code. Fixed in 75ca63a.

**SHA-256, not FNV-1a.** The description was wrong, the code is right — `build_retry_trace` is SHA-256 with the leading 8 bytes folded into `[1e15, 1e16)`, precisely *because* the fold keeps only the low bits, where FNV-1a is weakest. Description corrected.

**The MANUAL capture.** That run predates `trans_type_for` refusing `CaptureMethod::Manual`, so it is stale evidence. There are no credentials and the mock was ad hoc, so rather than restate a capture I cannot reproduce, I have deleted the grpcurl run and its `AUTHORIZED` response outright and said so in place. The captured *outbound* 3DS body is kept, because it is the only byte-level evidence of the cryptogram / `additionalAuthInfo` mapping, but now carries a provenance note: `capture_method` feeds exactly one wire field — `transType`, via `trans_type_for` — so the same payment sent today with AUTOMATIC produces `"AC"` and nothing else in that body changes. This also answers your `:480` thread: auto-capture-only is deliberate (Capture/Void/Refund are `not_implemented`, so `"A"` would strand an authorization the caller cannot act on) and is now stated under **Scope** instead of being contradicted by a capture.

I also re-synced the rest of the drift you would have hit next: the amount type is now the shared `common_utils::types::StringTwoDecimalUnit`, PSync keys `/inquiry` off the persisted `retry_trace`, and the Files list was missing `common_utils`, `generate.py`, `examples/`, `docs-generated/`, `data/field_probe/` and the integration-test specs.

On "the mock captures are the only evidence offered" — that stands, and the description now records the provenance of each capture rather than presenting them undated. Real sandbox validation is still blocked on credentials.

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #2188, opened 2026-08-27
feat(framework): add account_funded_transactions support in payment request

Shubhodip900:
## Description

Add account_funded_transactions support in payment request.

Proto fields added:

- date_of_birth (SecretString, field 13) on Address — carries the recipient/cardholder's date of birth (ISO 8601 yyyy-MM-dd) as a sensitive secret field; needed by connectors that require DOB for identity verification or compliance (e.g. BNPL, regulated payouts).
- is_account_funded_transaction (bool) and recipient_details (RecipientDetails) on PaymentServiceAuthorizeRequest, PaymentServiceSetupRecurringRequest, RecurringPaymentServiceChargeRequest, CompositeAuthorizeRequest, and CompositeVerifyRedirectResponseRequest — together these support Account Funded Transactions (AFT), where funds are pushed from a pre-funded account to a recipient rather than pulled from a card; RecipientDetails carries who receives the money (bank account via IBAN/routing/BIC/bare number/truncated PAN, card, wallet, email, phone, or social network handle) along with their phone, tax ID, and address.
### Additional Changes

## How did you test it?

used `cargo build` and `cargo run --bin grpc-server` to verify the build.
Connector changes and complete test will be added in separate PR.


Vani-1107 (dismissed):

gopikrishna000:
A few notes on the proto contract before this lands. `proto-checks.yml` gates rename/removal behind `proto-breaking-approved` plus a merged-and-deployed consumer migration, so the names and shapes added here are effectively permanent from merge — worth a close look now rather than later.

**1. The AFT definition is inverted.** The proto comment says funds are "pulled from a pre-funded account rather than a card". An Account Funding Transaction pulls *from* the card in order *to fund* an account (wallet top-up, P2P send, funding a disbursement); the outbound push to the recipient is a separate OCT. The PR description states the opposite of the proto comment, so the two can't both be right. This one ships to SDK consumers, so worth getting exact.

Related: the scheme term is Account **Funding** Transaction, which makes `is_account_funded_transaction` an odd permanent field name.

**2. `is_account_funded_transaction` and `recipient_details` can contradict each other.** `flag = true, details = absent` and `details = present, flag = absent` are both representable and neither is defined. Folding the flag into the payload:

```proto
// Present when this payment is an Account Funding Transaction.
optional AccountFundingDetails account_funding_details = 53;

message AccountFundingDetails {
  RecipientAccount recipient_account = 1;   // required by all schemes
  optional SecretString recipient_tax_id = 2;
  optional Address recipient_address = 3;   // name, phone, DOB
}
```

makes presence the signal, makes "AFT with no recipient" unrepresentable, and removes a good part of the hand-written `ok_or_else` validation in `types.rs`.

**3. `RecipientBankAccount` discriminates by identifier shape rather than by scheme.** As written, a UK sort code has nowhere to go except `routing_number`, and SEPA is indistinguishable from Bacs that happens to carry a BIC. `payouts.proto` already models this space by scheme — `AchBankTransferPayout` / `BacsBankTransferPayout` / `SepaBankTransferPayout` / `PixBankTransferPayout` — and carries `bank_name`, `bank_country_code`, `bank_city`, `ispb`, `bank_branch`, all of which the new tree drops and several connectors need. Suggest mirroring that split so the connector knows which identifier semantics apply.

**4. `RecipientAccount.card_number` is a bare `SecretString`.** Every other card number in the contract is `CardNumberType` (8 sites, including `CardPayout`), so whatever hangs off that wrapper won't apply to recipient PANs.

**5. Two homes for the recipient phone number** — `RecipientDetails.phone_number` and `RecipientDetails.address.phone_number` (+ `phone_country_code`), with no stated precedence.

**6. `social_network_id`** has no precedent elsewhere in the contract and no connector consuming it. Suggest deferring until something needs it.

**7. Scope.** The pair lands on five request messages, including `PaymentServiceSetupRecurringRequest` and `RecurringPaymentServiceChargeRequest`. AFT on a zero-amount mandate setup isn't motivated in the description. Adding a field later is cheap; removing one is not — `PaymentServiceAuthorizeRequest` + `CompositeAuthorizeRequest` only would be the safer starting scope.

**8. No consumer yet.** All the Rust here is `None`-plumbing and the connector work is deferred to a follow-up. Given the shape can't be revised cheaply post-merge, landing this alongside the first connector that actually populates it would let the processor's real requirements settle the open questions — e.g. whether `bank_country_code` is needed per bank variant.

`Address.date_of_birth` looks good as-is: consistent with `first_name` / `last_name` / `email` already living there, and it covers both payer DOB (billing address) and recipient DOB in one field.

Field numbering, `buf lint` and `buf breaking` all check out.

Vani-1107 (dismissed):

Vani-1107 (dismissed):

gopikrishna000 (approved):

JeevaRamu0104 (approved):

Vani-1107 (approved):

PR #2190, opened 2026-08-28
feat(core): add customer.date_of_birth and source Ilixium's dateOfBirth from it

shuklatushar226:
## What

Ilixium requires `customer.dateOfBirth` on every card authorisation. The UCS payment model had nowhere to put one, so `connectors/ilixium/transformers.rs` read it out of the merchant's free-form metadata blob:

```rust
fn extract_date_of_birth(metadata: Option<&SecretSerdeValue>) -> Option<Secret<String>> {
    let value = metadata?.clone().expose();
    ["ilixium_date_of_birth", "date_of_birth"]
        .iter()
        .find_map(|key| value.get(key).and_then(|v| v.as_str()).map(String::from))
        .map(Secret::new)
}
```

It was a deliberate stop-gap — the tech spec logged it as an open question — but the string was passed through untouched, so a merchant sending the obvious `1970-04-03` got a `VC8` rejection with no hint why. It was also invisible: nothing upstream could know the field was needed.

This gives it a first-class home.

## The contract

Verified against `docs.ilixium.com`:

| | |
|---|---|
| Field | `customer.dateOfBirth` on `POST /direct/auth` |
| Format | `ddmmyyyy` — eight digits, no separators. `"03042000"` is 3 April 2000. |
| Rejection | `VA8` missing · `VB8` invalid value · `VC8` invalid format, each as `status.code: "REJECTED"` with the code in `status.reasons[]` |

It appears in every documented customer object — Direct API 3DS, merchant-managed 3DS, hosted payment page initialisation, card validation. The strongest evidence it is genuinely mandatory is the *Optional Address* page: that feature strips the customer address down to `country` alone, and the example still carries `dateOfBirth`. It survives the trim, so it is not address-conditional.

One gap worth stating: the request-schema table on `/docs/api/authorisation` renders client-side from an OpenAPI spec whose JS chunk 404s on their CDN, so there is no machine-readable `required: true` to quote. A negative sandbox call settles it empirically — see *Testing*.

## Changes

**Proto.** `Customer.date_of_birth = 11`, carried as ISO-8601 `YYYY-MM-DD`. One field covers Authorize, PreAuthenticate, SetupMandate and RepeatPayment, since all four requests embed `Customer`.

`RecurringPaymentServiceChargeRequest` gets a flat `customer_date_of_birth = 40` instead — that request carries no `Customer` message, so without it the field on `RepeatPaymentData` would have been structurally always-`None`, which is worse than not having it.

**Domain types.** `customer_date_of_birth: Option<Secret<Date>>` on the four request structs that already carry `customer_document_details`, plus the matching `get_optional_…` / `get_…` pair beside the existing ones.

**Conversions.** A new `map_customer_date_of_birth` parses at the boundary, wired into the three sites that already map `customer_document_details` and into the PreAuthenticate builder. Unlike `map_customer_document_details`, a value that is present but unparseable is an **error** rather than a silent `None` — the caller asked for a date of birth to be sent, and dropping it would resurface as an opaque `VB8`/`VC8` from the connector instead of a field-level error here.

**Connector.** `format_date_of_birth` produces `ddmmyyyy`; `resolve_date_of_birth` prefers the structured field. This is the only place in the stack that knows Ilixium's shape. Both legs are covered — Ilixium drives 3DS through PreAuthenticate, so Authorize alone would not be.

Absent from both sources the field is omitted rather than filled with a placeholder: a fabricated date would reach the issuer and could distort Ilixium's own fraud checks.

## Backwards compatibility

The metadata path stays for one release behind a deprecation warning. Merchants are passing `ilixium_date_of_birth` in production today, and removing it in the same PR would break them silently on the next deploy.

Resolution order: `customer.date_of_birth` → `metadata.ilixium_date_of_birth` (deprecated) → omitted. Every layer is `Option`; no existing merchant is affected.

## Testing

`cargo check --workspace --all-targets`, clippy and fmt are clean. Four unit tests cover the formatting and resolution:

- `1970-04-03` → `"03041970"`, including the single-digit day and month case where the zero-padding matters (Ilixium reads the string positionally, so `"341970"` would be a `VC8` rather than a short date)
- the structured field wins over the deprecated metadata key
- the fallback resolves under either legacy key
- absent from both sources, the field is omitted

**Not yet run, and worth doing against sandbox before merge:** omit the field on a real authorisation and confirm Ilixium answers `REJECTED` with `VA8`. That is the empirical answer to the schema gap noted above, and belongs in the tech spec once observed.

## Follow-up

Hyperswitch pins this client to a git tag, so a tag has to be cut after this merges before the HS side can populate the field. The corresponding Hyperswitch PR is open as a draft and links back here.

https://claude.ai/code/session_01AHpWDF3Aro7EiXWKjWWg4L


XyneSpaces:
[blocking] `Secret::<time::Date>::foreign_try_from(dob.expose())` in `map_customer_date_of_birth` does not have a matching `ForeignTryFrom` implementation for `Secret<time::Date>` / `time::Date`, so this helper will fail to compile. Please parse the ISO-8601 string explicitly into `time::Date` and wrap it in `Secret::new`, or add the required conversion impl before calling it here.

XyneSpaces:
:warning: **[should-fix]** `crates/types-traits/domain_types/src/types.rs` (around the new `map_customer_date_of_birth` function) — the call `Secret::<time::Date>::foreign_try_from(dob.expose())` assumes a `ForeignTryFrom` implementation exists that can parse an ISO-8601 string into `Secret<time::Date>`, but no such impl could be located in this repo (searches for `Secret<time::Date>` / `Secret<Date>` + `ForeignTryFrom`/`foreign_try_from` returned zero matches).

Please either confirm the impl exists elsewhere and compiles, or parse explicitly and wrap the result:
```rust
let date = time::Date::parse(
    dob.peek(),
    &time::format_description::well_known::Iso8601::DATE,
).change_context(...)?;
Secret::new(date)
```

This avoids a potential compile-time trait-resolution failure for a brand-new conversion path.

PR #2193, opened 2026-08-31
feat(payouts): Add merchant_request_id field to payout gRPC requests and use it in Gotyme connector

Saptak88:
## Description

- Add a new optional `merchant_request_id` field to all payout gRPC request and PayoutFlowData so caller systems can supply reference ID generated on their end for payout gateway requests.

### Additional Changes


Aishwariyaa-Anand (approved):

JeevaRamu0104 (approved):

XyneSpaces:
[blocking] Preserve compatibility for Gotyme payout transfer/get requests that still provide `merchant_payout_id` but omit the new optional `merchant_request_id`. The new transformer paths currently return `MissingRequiredField("merchant_request_id")` for those existing requests; either fall back to `merchant_payout_id` where its semantics match, or make and version this as an explicit required contract change.

PR #2199, opened 2026-09-01
ci: exclude macros.rs from connector name detection

pixincreate:
## Description

The `names_in_dir` function in `.github/workflows/ci.yml` treats every `.rs` file under `connectors/` as a connector name. Shared utility files like `macros.rs` get extracted as connector "macros", causing `verify-new-connectors.sh` to fail:

    Error: New connector 'macros' has no specs.json

Fix: `names_in_dir` now accepts an optional `specs_root` parameter. When given, `.rs` files without a matching `connector_specs/<name>/specs.json` are excluded. This correctly handles:
- `macros.rs` — excluded (no specs.json)
- `absa_sanlam.rs` — included (has `connector_specs/absa_sanlam/specs.json`)
- All directory-based connectors — unaffected (directories always pass)

## Motivation and Context

PR #2086 introduced connector certification as a merge-blocking gate. The `connector-names` step extracts names from `.rs` files to detect changed/new connectors. It was designed for directories (each connector is a directory under `connectors/`), but also picks up top-level `.rs` files. Any shared utility file in that directory breaks the pipeline.

### Additional Changes

## How did you test it?

Verified against the live connectors directory:

| File | Has specs.json | Result |
|------|---------------|--------|
| `macros.rs` | No | Excluded |
| `absa_sanlam.rs` | Yes | Included |
| `adyen.rs` + `adyen/` | Yes | Included |
| All 107 real connectors | Yes | Included |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,9 +154,18 @@ jobs:
           # name can only ever be one that actually exists.
           names_in_dir() {
             local dir="$1"
+            local specs_root="${2:-}"
             [ -d "${dir}" ] || return 0
             find "${dir}" -mindepth 1 -maxdepth 1 \( -type d -o -name '*.rs' \) -exec basename {} \; | \
-              sed 's|\.rs$||' | sort -u
+              sed 's|\.rs$||' | while IFS= read -r name; do
+                # If a specs root is given, only include names that have a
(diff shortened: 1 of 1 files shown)


pixincreate:
closing in favor of #2201

PR #2204, opened 2026-09-01
feat(connector): add GlobalpaymentsRealex card flows (Authorize, PSync, Capture, Void, Refund, RSync) and merchant-driven 3DS2

### LINKED ISSUE #2331: Connector Reference ID Normalization Framework
### 🧩 What is the ID Framework?

Every payment connector imposes its own rules on the merchant reference ID that
accompanies a payment request — maximum length, allowed characters,
card-network-specific overrides, and what to do when the incoming ID doesn't fit.

The **ID Framework** is a centralized, declarative abstraction that each connector
configures once to express these rules, replacing the current pattern of ad-hoc,
inline transformation logic scattered across individual transformer files.

At the framework level, a connector declares a `ConnectorReferenceIdPolicy`
(max length, character restrictions, overflow strategy, per-network overrides).
The framework applies those rules uniformly before the ID reaches the connector's
request builder — so transformers never contain bespoke truncation, hashing, or
UUID-generation code again.

---

### 🔍 Current State — No Framework Exists

The only framework-level function today is in
`crates/types-traits/domain_types/src/utils.rs:386`:

```rust
pub(crate) fn extract_connector_request_reference_id(identifier: &Option<String>) -> String {
    identifier.clone().unwrap_or_default()
}
```
This is a bare unwrap_or_default(). No length enforcement, no character
normalization, no overflow strategy. Every connector that needs anything beyond
the raw string invents its own solution inline.

---
❓ Why We Need This

1. Silent truncation breaks reconciliation

Several connectors silently truncate the reference ID before sending it to the
connector API. The ID stored in Hyperswitch no longer matches the ID the connector
received. Chargebacks, refunds, and dispute lookups that rely on the reference ID
break silently.

Affected connectors using .chars().take(N).collect() with no warning or fallback:
Saferpay, Etisalat, WorldPay Raft, Razorpay
```rust
2. Five different overflow strategies, no common abstraction

┌─────────────────────────────────┬─────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│            Strategy             │                 Connectors                  │                                                     Problem                                                      │
├─────────────────────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Silent truncate                 │ Saferpay, Etisalat, WorldPay Raft, Razorpay │ Breaks reconciliation silently                                                                                   │
├─────────────────────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ UUID / random fallback          │ Checkout (Amex), Authorize.net              │ Breaks correlation with the original payment intent; Authorize.net uses a random string — not even deterministic │
├─────────────────────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ SHA-256 hash fallback           │ JPMorgan Orbital, Qwikcilver                │ Deterministic but invisible — behavior is undocumented at the connector config level                             │
├─────────────────────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Error + metadata escape hatch   │ Redsys                                      │ Correct but forces merchants to discover and pre-format IDs manually                                             │
├─────────────────────────────────┼─────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Dynamic prefix-aware truncation │ WorldPayVantiv                              │ Correct but connector-specific arithmetic with no reusable abstraction                                           │
└─────────────────────────────────┴─────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```
3. Card-network-specific overrides are buried in transformer functions

The get_checkout_reference function in checkout/transformers.rs:642 — which
prompted this issue — implements a card-network-specific length override
(Amex: max 30 chars, all other networks: no limit). This logic belongs at the
framework/config level, not buried in a transformer function.

4. Growing maintenance surface

There are already 15 connectors with bespoke reference ID handling. Without a
framework, this list grows unbounded and each new implementation is independently
wrong in a different way.

---
```rust
📋 Connectors That Need Framework Support

┌─────────────────┬───────────────────┬───────────────────────┬───────────────────────────────────────────────────┬────────────────────────────────────────────┐
│    Connector    │    Max Length     │      Constraint       │                 Current Strategy                  │                    File                    │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ redsys          │ 12                │ Hard limit            │ Errors; requires metadata override                │ redsys/transformers.rs:42,65–73            │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ nexixpay        │ 18                │ Alphanumeric          │ Truncate or error                                 │ nexixpay/transformers.rs:38,41–52          │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ jpmorganorbital │ 22                │ Alphanumeric          │ SHA-256 → hex → take 22                           │ jpmorganorbital/transformers.rs:57,238,360 │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ authorizedotnet │ 20                │ —                     │ Random string fallback                            │ authorizedotnet/transformers.rs:42,146–157 │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ ilixium         │ 5–20              │ Alphanumeric, min+max │ Truncate + filter non-alphanumeric                │ ilixium/transformers.rs:38,132,190         │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ worldpayraft    │ 16                │ —                     │ Silent truncate                                   │ worldpayraft/transformers.rs:124           │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ worldpayvantiv  │ 36 minus prefix   │ —                     │ 36 - prefix.len() then byte-slice                 │ worldpayvantiv/transformers.rs:2680–2682   │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ worldpayvantiv  │ 17 (customer ref) │ —                     │ Hard const CUSTOMER_REFERENCE_MAX_LENGTH          │ worldpayvantiv/transformers.rs:242         │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ checkout        │ 30 (Amex only)    │ —                     │ UUID v4 hex (no dashes) truncated to 30           │ checkout/transformers.rs:88,643–647        │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ saferpay        │ 80                │ —                     │ Silent truncate                                   │ saferpay/transformers.rs:42,139            │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ etisalat        │ 25                │ —                     │ Silent truncate                                   │ etisalat/transformers.rs:46,62             │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ razorpay        │ 40                │ —                     │ Silent truncate                                   │ razorpay/transformers.rs:1273              │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ qwikcilver      │ derived           │ —                     │ SHA-256 hash to derive transaction ID             │ qwikcilver/transformers.rs:1449            │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ tsys_transit    │ various per field │ Alphanumeric + space  │ Local sanitize_alphanumeric_space(value, max_len) │ tsys_transit/transformers.rs:1234–1238     │
├─────────────────┼───────────────────┼───────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────────┤
│ merchante       │ —                 │ —                     │ Inline comment: "truncate defensively"            │ merchante/transformers.rs:88               │
└─────────────────┴───────────────────┴───────────────────────┴───────────────────────────────────────────────────┴────────────────────────────────────────────┘
```
Additional constraint types not yet expressible by any connector:
- Minimum length — ilixium requires ≥ 5 chars
- Extended character allowlist — twoc_twop_paco allows ., -, _ in addition to alphanumeric (twoc_twop_paco/transformers.rs:383)
- Prefix-aware max — worldpayvantiv's available length depends on a connector-assigned prefix that is only known at runtime

---
🏗️ How We Will Implement It

Step 1 — Define the policy types

Add to crates/types-traits/interfaces/src/reference_id.rs (new file):
```rust
/// Declared once per connector; applied by the framework before the request builder runs.
pub struct ConnectorReferenceIdPolicy {
    /// Absolute maximum length after character filtering. `None` = no limit.
    pub max_length: Option<usize>,
    /// Minimum required length. `None` = no minimum.
    pub min_length: Option<usize>,
    /// Strip characters not in the allowlist before applying length checks.
    pub char_filter: CharFilter,
    /// What to do when the ID still exceeds `max_length` after filtering.
    pub overflow: OverflowStrategy,
    /// Per-card-network max length overrides (e.g. Amex = 30 on Checkout.com).
    /// Overrides `max_length` for the matching network.
    pub network_overrides: &'static [(CardNetwork, usize)],
}
```
```rust
pub enum CharFilter {
    /// Allow any character. No filtering.
    None,
    /// Keep only ASCII alphanumeric characters [a-zA-Z0-9].
    AlphanumericOnly,
    /// Keep ASCII alphanumeric plus specific extra characters.
    AlphanumericPlus(&'static [char]),
}

pub enum OverflowStrategy {
    /// Truncate to `max_length` characters.
    /// Use only when the connector echoes the truncated ID back in responses
    /// for reconciliation. Emits a structured log warning at runtime.
    Truncate,
    /// Generate a UUID v4, strip dashes, truncate to `max_length`.
    /// Not deterministic across retries — breaks cross-attempt correlation.
    UuidFallback,
    /// SHA-256-hash the raw ID, hex-encode, truncate to `max_length`.
    /// Deterministic: the same input always produces the same output.
    HashFallback,
    /// Return `IntegrationError::InvalidDataFormat` and surface `metadata_key`
    /// as the escape hatch for merchants who pre-format a short ID.
    ErrorRequireMetadata { metadata_key: &'static str },
}
```
---
Step 2 — Add reference_id_policy() to ConnectorServiceTrait
```rust
pub trait ConnectorServiceTrait {
    // ... existing methods ...

    /// Returns the reference ID policy for this connector.
    /// Default: no constraints — pass through unchanged.
    fn reference_id_policy() -> ConnectorReferenceIdPolicy {
        ConnectorReferenceIdPolicy::default() // no-op passthrough
    }
}
```
---
Step 3 — Implement the normalization function
```rust
pub fn normalize_reference_id(
    raw: &str,
    policy: &ConnectorReferenceIdPolicy,
    card_network: Option<&CardNetwork>,
    metadata: Option<&serde_json::Value>,
) -> Result<String, IntegrationError> {
    // 1. Apply character filter
    let filtered = apply_char_filter(raw, &policy.char_filter);

    // 2. Resolve effective max_length (network override wins if present)
    let effective_max = card_network
        .and_then(|n| policy.network_overrides.iter().find(|(cn, _)| cn == n))
        .map(|(_, len)| *len)
        .or(policy.max_length);

    // 3. Check min length
    if let Some(min) = policy.min_length {
        if filtered.len() < min {
            return Err(/* InvalidDataFormat */);
        }
    }

    // 4. Apply overflow strategy if needed
    match effective_max {
        Some(max) if filtered.len() > max => {
            apply_overflow(&filtered, max, &policy.overflow, metadata)
        }
        _ => Ok(filtered),
    }
}
```
---
Step 4 — Call normalize_reference_id in the dispatch path

Apply the policy in the payment flow dispatch, before build_request is called,
replacing the current call to extract_connector_request_reference_id.

---
Step 5 — Migrate existing connectors (priority order)
```rust
For each connector:
1. Declare fn reference_id_policy() with the appropriate ConnectorReferenceIdPolicy
2. Delete the inline truncation / hashing / UUID function from the transformer
3. Remove the local length constant (AMEX_PAYMENT_REFERENCE_LENGTH, MAX_ORDER_ID_LEN, etc.)

┌──────────┬────────────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────┐
│ Priority │                 Connector                  │                                    Reason                                    │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P0       │ redsys                                     │ Hard limit with no escape hatch — merchant-facing failures today             │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P0       │ checkout                                   │ Triggering example; Amex network override is the canonical use case          │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P1       │ jpmorganorbital                            │ SHA-256 strategy is correct but invisible at config level                    │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P1       │ authorizedotnet                            │ Random fallback is non-deterministic — should be HashFallback                │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P1       │ nexixpay                                   │ Error path is correct but should go through framework                        │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P2       │ ilixium                                    │ Only connector with a min_length constraint — validates framework handles it │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P2       │ worldpayvantiv                             │ Prefix-aware max may require a framework extension                           │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P3       │ saferpay, etisalat, worldpayraft, razorpay │ Migrate silent truncation to explicit Truncate with log warning              │
├──────────┼────────────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────┤
│ P3       │ tsys_transit, merchante, qwikcilver        │ Remaining cleanup                                                            │
└──────────┴────────────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────┘
```
---
✅ Acceptance Criteria

- [ ] ConnectorReferenceIdPolicy, CharFilter, and OverflowStrategy types defined in a new reference_id.rs module under crates/types-traits/interfaces/src/
- [ ] reference_id_policy() method added to ConnectorServiceTrait with a no-op default implementation
- [ ] normalize_reference_id function implemented and unit-tested, covering:
  - [ ] Truncate strategy
  - [ ] UUID fallback strategy
  - [ ] Hash fallback strategy
  - [ ] Error + metadata escape hatch strategy
  - [ ] Per-card-network length override
  - [ ] Min-length rejection
  - [ ] Character filtering (AlphanumericOnly, AlphanumericPlus)
- [ ] All 15 connectors listed above migrated — their inline length constants and transformation functions deleted
- [ ] No connector transformer file contains a standalone reference-ID truncation or UUID-generation function after migration
- [ ] For Truncate strategy, the framework emits a structured log warning so operators know the stored ID differs from the connector ID
- [ ] extract_connector_request_reference_id in domain_types/src/utils.rs updated to call normalize_reference_id as its implementation

shuklatushar226:
## What

Adds **Global Payments Ecommerce XML API** (legacy *Realex Payments* "XML API") as a new connector for **Card**, with all six core flows — **Authorize, PSync, Capture, Void, Refund, RSync** — plus **merchant-driven 3DS2** over the separate JSON API.

> This is **not** the GP-API JSON product. The existing `globalpay` connector integrates `apis.sandbox.globalpay.com/ucp` (JSON, OAuth bearer). This one is the older single-CGI endpoint that speaks XML both ways and authenticates with a two-stage SHA-1 digest. The two share no types, transformers, status maps or configuration, and `globalpay` is untouched.

**This description is a summary.** Every flow was landed as its own commit with a full write-up — protocol findings, digest blueprints, the reasoning behind each mapping and the live evidence. Read the commit messages on this branch for the long form; this body exists to be readable in one sitting and to stay under GitHub's 65 KB limit.

## The connector at a glance

| | |
|---|---|
| XML endpoint | `https://api[.sandbox].realexpayments.com/epage-remote.cgi` — one URL for every operation |
| Operation selection | the `type` attribute on the `<request>` root |
| Wire format | XML request, XML response |
| Auth | `<sha1hash>` **in the body** — no auth headers |
| HTTP status | **always 200**, including for declines and validation errors |
| 3DS2 endpoint | `https://api[.sandbox].globalpay-ecommerce.com` — **separate host, JSON, `Authorization: securehash` header, real HTTP statuses** |

Request digest: `SHA1(SHA1("f1.f2.…") + "." + sharedSecret)`. The digest inputs are built from the values actually being sent, never re-derived, so body and hash cannot drift.

## Flows

| Flow | `type` | Reference it keys on | Headline behaviour | Live scenarios |
|---|---|---|---|---|
| **Authorize** | `auth` | — | `autosettle flag="1"` → `Charged`, `flag="0"` → `Authorized`. Publishes `{orderid, authcode}` as `connector_metadata`. | 9 |
| **PSync** | `query` | **`<orderid>` only** — `<pasref>` is inert | `result != 00` → `Failure`; void marker → `Voided`; `batchid > 0` → `Charged`; else `Authorized`, in that order | 12 |
| **Capture** | `settle` | original `orderid` + `authcode` | Mints a **new** `pasref`, returns `authcode 000000` | 11 |
| **Void** | `void` | **original `auth` `pasref`** | Returns `batchid 0`, mints a new `pasref`; republishes `void_pasref` into metadata so PSync can see the void | 13 |
| **Refund** | `rebate` | **original `auth` `pasref`** | Carries a **second, independent** digest (`<refundhash>`) keyed on a different credential | 14 |
| **RSync** | `query` | synthetic `_rebate_<orderid>` leg | 8 rules; refuses rather than mis-reports when the leg's `pasref` does not match the refund being synced | 14 + 7 regression |
| **3DS2** | JSON API | `server_trans_id` decoded from the ACS's own browser POST | `PreAuthenticate` / `Authenticate` / `PostAuthenticate` drive the authentication; `Authorize` consumes the result via the existing `<mpi>` block | see below |

## Merchant-driven 3DS2

Before this, 3DS was **consume-only**: `build_mpi()` forwarded an `<mpi>` block built from `authentication_data`, `redirection_data` was hardcoded `None`, and driving `authentication_type: three_ds` returned `succeeded` with `next_action: null` and outbound XML carrying no `<mpi>` at all — a plain non-3DS auth wearing a 3DS label. That passthrough still works (the HS shim skips pre-authentication when `authentication_data` is already present); this adds the missing half.

3DS2 for this product is a **separate JSON API on a separate host**, with a different auth scheme, a different timestamp format, a different card-scheme vocabulary, and real HTTP error statuses where the XML gateway always answers 200. It therefore lives in its own module, `connectors/globalpayments_realex/three_ds_two.rs`.

| # | Call | UCS flow | Side |
|---|---|---|---|
| 1 | `POST /3ds2/protocol-versions` | `PreAuthenticate` | server |
| 2 | `POST {method_url}` — `threeDSMethodData` | *(redirect emitted by 1)* | browser |
| 3 | `POST /3ds2/authentications` | `Authenticate` | server |
| 4 | `POST {challenge_request_url}` — `creq` | *(redirect emitted by 3)* | browser |
| 5 | `GET /3ds2/authentications/{sid}` | `PostAuthenticate` | server |
| 6 | XML `auth` with `<mpi>` | `Authorize` — already built | server |

A frictionless authentication skips 4 and 5. Routing through the dedicated authentication RPCs rather than driving 3DS inside `Authorize` is what Cybersource, Barclaycard, Redsys and Moneris do, and it needs **no UCS core change, no proto change and no `authorize_flow.rs` change** — `PreAuthenticate` always returns a redirect, so the existing `_ => false` default there is already correct. Please don't add an arm for it.

### Two things that would break this silently

**1. The `securehash` timestamp must be read once per request.** The digest covers the `request_timestamp` that the body — or, for call 5, the query string — also carries. The default `build_request_v2` calls `get_url` and `get_headers` as *separate* calls, so a timestamp read independently in each hashes a different microsecond than the payload contains: **HTTP 401 on every call, with code that reads as correct**. All three flows therefore build the whole request themselves from one clock read, with a blocking comment saying so, and `get_url` deliberately returns a self-describing error so that deleting an override fails loudly instead of quietly 401ing.

**2. Two card-scheme mappers must coexist and disagree.** The JSON API spells Mastercard `MASTERCARD`; the XML API spells it `MC`. **One 3DS payment sends both, one per leg.** `map_3ds2_scheme` and `map_card_type` are kept a file apart and a unit test asserts they differ on `CardNetwork::Mastercard`, so the tempting "unify the card-type mapper" refactor fails the build.

### Digest blueprints

| Call | Stage-1 string |
|---|---|
| `POST /3ds2/protocol-versions` | `request_timestamp.merchant_id.cardnumber` |
| `POST /3ds2/authentications` | `request_timestamp.merchant_id.cardnumber.server_trans_id` |
| `GET /3ds2/authentications/{sid}` | `request_timestamp.merchant_id.server_trans_id` |

Same two-stage SHA-1 primitive and same Shared Secret as the XML API — that is the *only* thing the two APIs share. `request_timestamp` is `yyyy-MM-ddTHH:mm:ss.SSSSSS` (UTC, no `Z`), not the XML API's `YYYYMMDDHHMMSS`.

### Telling the two browser returns apart

Both ACS returns POST to the same hyperswitch completion endpoint. Hyperswitch puts the query string into `redirect_response.params` and the form body into `.payload`, so:

- Method Notification URL = `{continue_redirection_url}?gp3ds=method` → `params` non-empty → `Authenticate`
- Challenge Notification URL = `{continue_redirection_url}` (bare) → `params` empty → `PostAuthenticate`

Byte-identical to the discriminator Cybersource and Barclaycard already use.

**No-DDC normalisation.** When the ACS offers no Method URL, `PreAuthenticate` still returns a redirect: a one-hop self-POST back to the marked URL carrying a **synthesised** `threeDSMethodData` plus `threeDSMethodCompletion=UNAVAILABLE`. Both branches then land on the same URL with the same field names, so `Authenticate` has exactly one code path instead of two that can drift apart.

### Cross-leg state

`server_trans_id` **cannot** travel via `authentication_data`: hyperswitch builds `PostAuthenticateRequest` with `authentication_data: None`, and sources the `Authenticate` leg's from its own authentication table, which a connector-driven 3DS never writes. It is decoded from the ACS's own echo instead — self-contained, no cross-repo change, and authoritative because it comes from the ACS.

| Value | Carrier | Read from |
|---|---|---|
| `server_trans_id` → `Authenticate` | ACS echo of `threeDSMethodData` | `payload["threeDSMethodData"]`, base64url-decode → `threeDSServerTransID` |
| `server_trans_id` → `PostAuthenticate` | ACS `cres` | `payload["cres"]`, base64-decode → `threeDSServerTransID` |
| `method_url_completion` | our own synthesised form | `payload["threeDSMethodCompletion"]` present ⇒ `UNAVAILABLE`, else `YES` |
| eci / cavv / ds_trans_id / message_version → `Authorize` | `{Authenticate,PostAuthenticate}Response.authentication_data` | existing `build_mpi()` — **unchanged** |
| `<orderid>` | `connector_request_reference_id` | `sanitize_order_id` — the 3DS2 `order.id` uses the **same** value as the XML `<orderid>`, so the two systems reconcile in the GP portal |

Base64 decoding is tolerant of all four alphabets: GP emits `threeDSMethodData` base64url-unpadded, while an ACS `cres` is typically standard and sometimes padded.

### Status mapping

Wire values carry an `AUTHENTICATION_` prefix the EMVCo names do not.

| GP `status` | EMVCo | `TransactionStatus` | Proceeds to the XML `auth`? |
|---|---|---|---|
| `AUTHENTICATION_SUCCESSFUL` | `Y` | `Success` | **yes** |
| `AUTHENTICATION_ATTEMPTED_BUT_NOT_SUCCESSFUL` | `A` | `NotVerified` | **yes** — liability still shifts (ECI 06 Visa / 01 MC) |
| `CHALLENGE_REQUIRED` | `C` | `ChallengeRequired` | no — redirect to the ACS |
| `DECOUPLED_AUTHENTICATION_CONFIRMED` | `D` | `ChallengeRequiredDecoupledAuthentication` | no — redirect |
| `AUTHENTICATION_FAILED` | `N` | `Failure` | no |
| `AUTHENTICATION_ISSUER_REJECTED` | `R` | `Rejected` | no |
| `AUTHENTICATION_COULD_NOT_BE_PERFORMED` | `U` | `VerificationNotPerformed` | no |
| `CHALLENGE_PREFERENCE_ACKNOWLEDGED_INFORMATIONAL_ONLY` | `I` | `InformationOnly` | no |
| *(unrecognised)* | — | `VerificationNotPerformed` | no |

The challenge branch keys on **`status`**, never on the presence of `challenge_request_url` / `encoded_creq`: an ACS can request a challenge without setting `challenge_mandated`, and a field-presence test would silently treat that payment as frictionless.

**A card not enrolled in 3DS2 fails the payment; it does not fall through to a non-3DS auth.** `TransactionStatus` has no "not enrolled" variant, and silently dropping the liability shift on a payment the merchant marked `three_ds` would change the liability model without telling anyone. `enrolled: false` surfaces the distinguishable `GP3DS2_NOT_ENROLLED` so the caller can decide to retry as `no_three_ds`.

### Error model

`ConnectorCommon::build_error_response` cannot be widened — one method, no flow parameter, shared by all nine flows, and the two APIs disagree on both wire format and the meaning of an HTTP status. Each 3DS2 flow overrides `get_error_response_v2` instead.

| Status | Body | Handling |
|---|---|---|
| `400` | structured EMVCo error | parsed into `Gp3ds2ErrorResponse` |
| `401` | **empty** (verified live) | plain-text fallback |
| `403` / `404` | plain text (verified live) | plain-text fallback, message surfaced in `reason` |

Every 4xx sets `attempt_status: AuthenticationFailed` on **both** branches, so hyperswitch's `should_continue` guards suppress the XML auth that would otherwise follow.

### Configuration

`globalpayments_realex.secondary_base_url` in `config/{development,sandbox,production}.toml`, and `connector_secondary_base_url` in both `globalpayments_realex` contexts in `superposition.toml`. It is a **hard requirement, not a fallback to `base_url`** — 3DS2 traffic sent to the CGI endpoint returns an opaque XML/HTML error instead of a clear config failure.

Redirects are `RedirectForm::Form`: only `Form`, `Html`, `Braintree`, `Mifinity` and `Nmi` survive the domain→proto→domain round trip.

## Cross-cutting protocol facts

- **Errors are not HTTP errors on the XML API.** Every outcome is HTTP 200, so failures are classified from `<result>` in the response transformer.
- **`5xx` results are a smaller document** carrying only `timestamp` / `result` / `message` / `orderid` — no `sha1hash`, `pasref` or `merchantid`. Every response field except `result` is `Option`, and hash verification is **skipped, not failed**, when the digest is absent.
- **The response digest uses the response's own `timestamp`**, which is server-local and routinely differs from the request's by hours. A mismatch maps to `IntegrityFailure`, not to a decline.
- **`<batchid>` is an empty element on declines** but a number on success, so it is read through a deserializer that maps blank text to `None`.
- **No `deny_unknown_fields`** — the live response carries an undocumented `<fraudresponse>` and a `<cardissuer>` whose children differ from the published docs.
- **The orderid fed to a digest is the sanitized reference actually sent, never the gateway echo** — error documents echo it back prefixed `_settle_` / `_void_` / `_rebate_`, which would silently break a follow-up call.
- **`<autosettle>` is mandatory** (`502 Mandatory Fields missing` otherwise). `MULTI` is not enabled on this account (`503`), so `ManualMultiple` is rejected with `NotImplemented` before any call rather than claimed as supported.
- **`<exempt_status>` is omitted by default** — this account rejects it with `508 Exemption is not configured for this merchant`.
- **JPY amounts are multiplied by 100** per the currency-codes rule, unlike other zero-exponent currencies. From the documentation; not exercisable on this sandbox account.

## Live-gateway findings that contradict the vendor docs

Each is recorded in the connector's technical specification and evidenced in the relevant commit message.

**Capture.** The spec left the settle digest's amount slot undecided. Resolved empirically: **the amount slot must mirror the body exactly.** `<amount>` present → `ts.mid.oid.<amount>..`; omitted → `ts.mid.oid...`; filling the currency slot on a settle → `505 sha1hash incorrect`. A full-amount settle *with* an explicit `<amount>` is accepted, so one code path serves both full and partial capture.

**Void.** `513 Can't void a settled transaction` is **not reachable** on this sandbox — a `settle` only queues into the *open* batch, so the authorization stays voidable until end-of-day. A void must reference the **ORIGINAL `auth` `pasref`**, not the settle-minted one (`508` otherwise). `<reasoncode>` is not part of the digest.

**Refund.** `rebate` also requires the **ORIGINAL `auth` `pasref`** — it behaves like Void despite running *after* capture. Two documented response values are wrong live: a successful rebate returns a **real issuer `<authcode>`** (not `000000`) and `<message>` is `AUTH CODE: nnnnnn` (not "Rebated Successfully"); both are hashed verbatim, so they are read from the response rather than assumed. The over-refund ceiling on this account is **100%**, not the documented 115%/105% — vindicating the decision not to hard-code a ceiling.

**Refund's two independent digests.**

| | `<sha1hash>` | `<refundhash>` |
|---|---|---|
| Construction | two-stage `SHA1(SHA1(blueprint) + "." + secret)` | **one stage**, `SHA1_hex(plaintext)` |
| Keyed on | Shared Secret (`api_key`) | rebate password (`api_secret`) |
| Transaction fields | yes | **none** — constant for the account |

Verified independently at the header level by varying one credential at a time: wrong rebate password → `505 The refund password you entered was incorrect`; wrong shared secret → `505 sha1hash incorrect`. Rather than adding a second near-identical builder, the shared `build_reference_request_hash` gained an `Option<&str>` currency parameter — `None` reproduces the settle/void blueprint byte-for-byte, `Some(cur)` fills the rebate slot.

**PSync / RSync.** `type="query"` is **absent from the entire vendor documentation set** — the request shape, digest blueprint and lifecycle mapping were established empirically. `query` keys on **`<orderid>` only**; `<pasref>` is inert. The **response** check-hash blueprint that verifies for `auth` / `settle` / `void` / `rebate` **does not hold for `query`**: it survived a ~5 × 10⁸-candidate search without matching, so for `query` **only** the check-hash is computed, logged at debug, and does **not** decide status. `IntegrityFailure` is unchanged on the four money-moving flows, which showed **zero** mismatches across the whole run.

**3DS2.** Server-side validation order, established live: `401` (bad hash) → `400/203` (timestamp format) → `400/201` (missing required field) → `404` (merchant lookup) → `403` (account 3DS entitlement) → field-value validation.

## Two problems worth calling out

**A void is invisible to a `query` on the original order id.** A voided payment keeps reporting `00 … AUTHORISED` with its original `batchid`. `PaymentServiceGetRequest` carries no attempt status and the domain conversion hard-codes `Pending`, so the usual "don't downgrade the known status" guard cannot fire. Fix: Void republishes `connector_metadata` as `{orderid, authcode, void_pasref}` — **additive and backward-compatible**, verified live in both directions — and PSync returns `Voided` when that marker is present, outranking the `batchid > 0` → `Charged` rule.

**RSync refuses rather than mis-reports.** The gateway mints exactly one `_rebate_<orderid>` leg per payment. On one live chain it **refused** the second rebate with `512 already been rebated`, while `_rebate_<ORIG>` went on answering `result 00` carrying the **first** rebate's pasref. Without the pasref identity check, syncing that rejected refund would have reported `REFUND_SUCCESS`. The gateway said `00`, and the connector refused to call it success.

## Known limits

- **Out-of-band voids stay undetectable to PSync.** The only gateway-side evidence lives on the `_void_<orderid>` leg, and reading it needs a second HTTP request, which the one-request-per-flow contract does not allow. The proper fix is a platform change carrying attempt status on `PaymentServiceGetRequest`.
- **Multiple partial refunds are not individually identifiable by RSync.** Eight alternative leg spellings were tried live; every one returned `508`. Moot on an account that refuses a second rebate; **anyone enabling multi-refund must re-run that probe first.** Non-matching refunds land on the "refuse" rule rather than mis-reporting.
- **A partial refund is not distinguishable from a full one** — no `query` response on this API carries an amount or currency. No amount-based integrity object is populated, and none is faked.
- **RSync's order id resolves through a ladder** (`connector_feature_data.orderid` → `connector_order_id` → actionable `MissingRequiredField`), because `RefundServiceGetRequest` hard-codes `connector_feature_data: None` in hyperswitch. Step 2 was proven live in exactly the HS shape, so **no hyperswitch-side change is required for RSync**.
- **3DS2 DDC has no client-side timeout.** EMVCo expects abandonment after ~10 s with `method_url_completion: NO`; hyperswitch renders `RedirectForm::Form` as a top-level auto-submitting page rather than a hidden iframe with a timer. Same limitation Cybersource, Barclaycard and Netcetera DDC already have. `RedirectForm::Html` with a hand-written hidden iframe is the fallback if it bites.
- **3DS2 `merchant_contact_url`** is mandatory and has no home in the UCS payment model; it is derived from the origin of `continue_redirection_url`. A merchant-configurable MCA field is the correct long-term fix.
- **No 3DS2 response integrity check.** Unlike the XML API's `<sha1hash>`, the 3DS2 API signs nothing on the way back, so `verify_response_hash` has no analogue there.
- **Production 3DS2 host is unconfirmed.** Two Global Payments documents disagree (`api.globalpay-ecommerce.com` vs `api.globalpay.com`); the majority spelling is configured and must be confirmed with GP before the production cut-over.
- **SCA exemptions are out of scope.** `challenge_request_indicator` is fixed at `NO_PREFERENCE`.

## Test evidence — live sandbox

All XML calls against `https://api.sandbox.realexpayments.com/epage-remote.cgi`; all 3DS2 calls against `https://api.sandbox.globalpay-ecommerce.com`. Credentials never appear in source, logs, commits or this description. Per-scenario transcripts are in the commit messages.

| Flow | Scenarios | Outcome |
|---|---|---|
| Authorize | 9 | all pass |
| Capture | 11 | all pass |
| Void | 13 | all pass |
| Refund | 14 | all pass |
| PSync | 12 | all pass |
| RSync | 14 + 7 regression | all pass |

**3DS2, verified live:**

- A digest built to the documented blueprint is **accepted** — the request proceeds past authentication to the account-entitlement check. Corrupting the Shared Secret returns **HTTP 401 with a zero-length body**, and `get_error_response_v2` degrades to the plain-text handler without panicking.
- The `yyyy-MM-ddTHH:mm:ss.SSSSSS` timestamp is **accepted**; sending the XML API's `YYYYMMDDHHMMSS` is rejected with `400` / `error_code 203`.
- An empty `account_id` returns exactly the structured EMVCo `400` body the error parser expects (`error_code 201`, `error_component S`, `error_detail accountId`).
- A missing `continue_redirection_url` **fails fast** with `MISSING_REQUIRED_FIELD` before any network call, rather than sending a notification URL the ACS could never return through.
- Driven end to end through `grpcurl` → UCS → Global Payments for both a Visa and a Mastercard PAN.

**3DS2, NOT verified — needs a 3DS-enabled account:** the sandbox account is **not entitled for 3D Secure**. `/3ds2/protocol-versions` answers `403 That Account ID is not configured for 3D Secure` for every card and every sub-account spelling tried. The enrolment, frictionless, challenge and results legs therefore still need either a 3DS-enabled account or the browser legs. The `MASTERCARD` scheme value is likewise unconfirmed on the wire, because Global Payments validates field *values* only after the entitlement check.

**Regression after the 3DS2 commit:** all six XML flows re-run clean — Authorize `AUTHORIZED`, PSync `CHARGED`, Capture `CHARGED`, Refund `REFUND_SUCCESS`, RSync `REFUND_SUCCESS`, Void `VOIDED`, every one `result=00`. The digest helpers and `build_mpi` are shared surface and are untouched.

## Out of scope

Wallets, APMs, mandates / MIT / recurring, card storage and tokenisation, payouts, disputes, webhooks, DCC, installments, SCA exemptions.

## Checks

- `cargo build` — `connector-integration`, `grpc-server` ✅
- `cargo fmt --all --check` ✅
- `cargo clippy --all-targets` — `connector-integration`, `grpc-server` ✅ zero warnings
- `cargo test -p connector-integration globalpayments_realex` — 7 unit tests ✅

---

## Follow-up fix — the reversal pasref after a manual capture (`0291baeb5`)

**Bug, reproduced on the sandbox.** A RealEx `settle` mints a **new** `<pasref>`, and UCS surfaces that one as the Capture response's `connector_transaction_id`. Refund (`rebate`) and Void (`void`) both read `<pasref>` straight off `request.connector_transaction_id`, so after a *manual* capture they addressed the settle leg and the gateway answered `508 Original transaction not found.` — even though the `<orderid>` and the `<authcode>` on the same request were correct. Auto-captured payments were never affected: with no separate settle the transaction id *is* the auth pasref, which is why the earlier refund scenarios all passed.

**Fix.** The metadata block Authorize publishes now also carries the authorization's own pasref, following the `void_pasref` pattern exactly — additive, `Option`, `skip_serializing_if`, so metadata written by the already-shipped flows keeps deserializing:

- `GlobalpaymentsRealexPaymentMetadata::auth_pasref`, populated on the Authorize success path from the `auth` response's `<pasref>`.
- Capture republishes the metadata block instead of dropping it, backfilling `auth_pasref` from the capture request's own `connector_transaction_id` (still the auth pasref at that point) so payments authorized before this change are healed on their way through a settle. Void carries the field forward rather than republishing the void-minted pasref.
- Refund and Void resolve `<pasref>` through a ladder — `metadata.auth_pasref`, then `connector_transaction_id`, then an actionable `MissingRequiredField` — and log a `warn` when the two sources disagree, matching the divergence warning RSync already emits for the order id.
- Capture's own request is unchanged (it already sent the auth pasref), and so is what the Capture response reports as `connector_transaction_id`.

**Verification — live sandbox over `grpcurl`.** Every `<pasref>` below is read back out of the request UCS actually put on the wire.

| # | Scenario | `<pasref>` sent on the reversal | Gateway |
|---|---|---|---|
| 1 | manual auth → settle → **rebate**, transaction id = settle pasref | the **auth** pasref (`…76091308`, not the settle `…82346954`) | `00` `REFUND_SUCCESS` |
| 2 | manual auth → settle → **void**, transaction id = settle pasref | the **auth** pasref (`…04826559`, not the settle `…10861735`) | `00` `VOIDED` |
| 3 | auto-capture auth → rebate, metadata stripped of `auth_pasref` | auth pasref via the `connector_transaction_id` fallback | `00` `REFUND_SUCCESS` |
| 4 | old-shape metadata (no `auth_pasref`) through capture → rebate | deserializes unchanged; `<orderid>` and `<authcode>` still carried; auth pasref via fallback, and via the capture backfill when the transaction id is the settle one | `00` `REFUND_SUCCESS` |
| 5 | `auth_pasref` present, transaction id deliberately different | the metadata value wins; divergence logged at `warn` with both values and the request type | `00` `REFUND_SUCCESS` |
| 6 | regression smoke — Authorize (MANUAL + AUTOMATIC), Capture, PSync (authorized / captured / auto / voided), Void, Refund, RSync | unchanged | every call `result=00` |

**Checks:** `cargo build` (workspace) ✅ · `cargo fmt --all --check` ✅ · `cargo clippy --all-targets` (workspace) ✅ zero warnings.

---

### Two files, not three (`1bc693d3f`)

`three_ds_two.rs` is gone; the connector is now `globalpayments_realex.rs` plus
`globalpayments_realex/transformers.rs`, matching every other connector in the tree. The move was
mechanical: the file held no connector-layer code — all four `impl` blocks were transformers — and
the two type sets share no names (62 exported symbols against 44, empty intersection), so nothing
needed renaming and nothing moved into the connector file. Beyond the move the diff is the merged
`use` block and six `common_utils::consts::…` paths shortened, which `-W unused-qualifications`
flags once both halves live together.

### CI: `check_connector_specs` was failing (`d26757870`)

This connector was the only one of 109 with no `connector_specs/` directory, so Phase 1 of the
check exited non-zero and — because `ci.yml` only emits a connector name when `specs.json` exists —
the harness ran **zero** scenarios against it. `specs.json` now declares all nine implemented
suites, `alpha_connectors.json` records why CI cannot run it live, and the check reports
`109 / 109 — sets match`.

### Fail-open behaviours, now closed (`fdcb040b0`)

| Finding | Before | After |
|---|---|---|
| 3DS bypass | `build_mpi` returns `None` when no authentication result is attached, and an absent `<mpi>` is not an error — a failed or abandoned challenge authorized as plain non-3DS | `ensure_liability_shift` refuses the `auth` unless the result carries `Y` or `A`; `N`/`U`/`R`/`C`/`D`/`I`, a missing result, and one with neither cryptogram nor ECI are all rejected |
| 3DS2 5xx | only `4xx` stamped `AuthenticationFailed`, so a GP outage left a non-terminal status and the XML auth followed unauthenticated | any non-2xx is terminal |
| `threeDSServerTransID` | browser-supplied value interpolated raw into the results URL and hashed into the `securehash` | must be UUID-shaped or the request is refused |
| Challenge return misrouted | a merchant whose `continue_redirection_url` already carries a query string breaks the params/no-params discriminator, and a `cres` was processed as device profiling | a `cres` reaching `Authenticate` is refused with an error naming the cause |
| Refund integrity | a digest mismatch on `result=00` reported `RefundStatus::Failure` — the money moved, and the caller is invited to refund twice | `ManualReview`, mirroring the payment flows' `IntegrityFailure` |
| PSync lookup fault | `502/503/505/506/508` mapped to `AttemptStatus::Failure`, flipping a healthy Charged payment to failed on a sync | `Unresolved` plus the structured connector error; `map_rsync_outcome` already refused this trap |

Live proof of the two behavioural ones, against the sandbox over gRPC:

- a `THREE_DS` authorize with no authentication result is refused **before any network call** — no
  `<auth>` reaches the gateway;
- a captured payment synced without its `connector_feature_data` returns `UNRESOLVED` + `508
  Original transaction not found.`, where it previously reported `FAILURE`.

Five unit tests come with them — the first automated coverage on the XML side — for the pasref
ladder, the query-fault classification, the refund-status distinction, the transaction-id validator
and the challenge-return misroute. All 12 connector tests pass.

Re-running the full six-flow harness after each commit: Authorize (both capture modes), Capture,
PSync (authorized / captured / auto / voided), Void, Refund and RSync are all `result=00`, and the
manual-capture → settle → rebate / void pasref assertions still hold.

### Known residual, deliberately not papered over

The 3DS server transaction id is validated for **shape** but not **bound to the payment**:
`PaymentsPostAuthenticateData` carries no server-side copy to compare against and the results
document echoes no merchant reference, so a well-formed id belonging to another payment is still
accepted. Closing it needs the caller to carry the id from the `Authenticate` leg into the
`PostAuthenticate` request. It is documented at the validator rather than left implied.

Two further items are shared-code problems this PR deliberately does not widen into:
`connector_key!` lowercases `GlobalpaymentsRealex` to `globalpaymentsrealex` while the `Connectors`
field is `globalpayments_realex`, and `ConnectorsPatch` is `deny_unknown_fields`, so a per-request
`base_url` override is rejected (also affects `AbsaSanlam`, `TsysTransit`, `PinelabsOnline`); and
the composite authentication loop's `PostAuthenticate` arm is the only one with no break guard,
which is what let a failed challenge reach Authorize in the first place. The connector-side guard
above makes this connector safe regardless; both deserve their own PRs.

**Checks:** `cargo build` (workspace) ✅ · `cargo fmt --all --check` ✅ · `cargo clippy --all-targets`
✅ (the one remaining warning is a pre-existing `nonminimal_bool` in `connectors/saferpay`, arrived
from main byte-identical) · `cargo run --bin check_connector_specs` ✅ · 12 unit tests ✅.

### Spell check and the merge chain

CI's `Spell check` job runs `crate-ci/typos` against **`./.typos.toml`** — the dotfile, not the
`typos.toml` the tool auto-discovers, and the two configs differ. That is why a plain local `typos`
run looked clean while CI was red. Three doc-comment occurrences of "unparseable" in this connector
were the only tracked hits; they are now "unparsable". Verified by running
`typos --config ./.typos.toml` against exactly the tree CI checks out (untracked working files
moved aside): exit 0.

While this PR has been open, main has landed Saferpay, JpmorganOrbital and TravelHub, each of which
claimed the connector enum value and `oneof config` tag this branch was using. Every merge kept both
sides and moved ours up, re-checking both number spaces for reuse each time:

| Merge | Theirs | Ours becomes |
|---|---|---|
| Saferpay | `SAFERPAY = 140`, tag `151` | `141` / tag `152` |
| JpmorganOrbital | `JPMORGAN_ORBITAL = 141`, tag `152` | `142` / tag `153` |
| TravelHub | `TRAVELHUB = 142`, tag `153` | **`143` / tag `154`** |

No existing number changed meaning in any of them, and the enum and oneof spaces are duplicate-free
at max 143 and 154.

The TravelHub merge also fixed something it merely exposed: `Connectors::apply`'s "Supported
connectors: …" error message never listed `globalpayments_realex`, so a superposition URL patch for
this connector was reported as unsupported *by name* while the match arm above it handled the
connector correctly.

Full six-flow live harness re-run after each merge — Authorize (both capture modes), Capture, PSync
(authorized / captured / auto / voided), Void, Refund, RSync — every call `result=00`, with the
manual-capture → settle → rebate / void pasref assertions still holding.

### Unit tests for the pure logic (`3ac9bd714`)

The last blocking review finding: 4,400 lines of transformer had seven tests, all of them 3DS2
helpers, and none at all on the XML side. **7 → 20 tests.**

Covered: the four digest builders; the response check-hash and its verified / case-insensitive /
mismatched / skipped tri-state; `format_amount` including the JPY branch; `sanitize_order_id`; all
five status maps; `map_rsync_outcome`'s seven rules; and the two fixes made earlier in this PR,
`resolve_reference_pasref` and `ensure_liability_shift`.

Two choices worth calling out:

- **The digest tests spell the dotted wire string out literally** rather than rebuilding it from a
  field array. `build_reference_request_hash` encodes three different positional slot fillings —
  settle (amount, no currency), void and query (neither), rebate (both) — and the
  "an empty slot still emits its separator" rule was established empirically against the sandbox. A
  refactor that tidies the builder by dropping empty entries would still produce a plausible hash;
  it cannot satisfy a test that spells the separators out.
- **`build_refund_password_hash` is pinned against the published SHA-1 of the ASCII string
  `"password"`** — an independent oracle, not a value copied from this connector's own output — and
  asserted *not* to equal the two-stage digest, which is the mistake that returns `505 The refund
  password you entered was incorrect.` No credential appears in any test.

`format_amount`'s JPY case is named `..._documented_but_never_verified_live` and says so in the body
too: the ×100 rule comes from the vendor documentation, could not be exercised against the sandbox
account, and the test pins the implemented behaviour rather than evidencing the gateway's.

**Every test was mutation-checked rather than assumed.** Nine deliberate defects were injected one
at a time and each was caught by exactly the intended test and no other:

| Injected defect | Caught by |
|---|---|
| reference digest loses its trailing empty slot | `every_positional_slot_of_the_request_digest_is_load_bearing` |
| rsync stops checking pasref identity | `the_rsync_decision_table_never_guesses` |
| 3DS guard accepts `transStatus N` | `a_three_ds_authorize_needs_a_liability_shifting_result` |
| JPY multiplier becomes ×10 | `format_amount_applies_the_jpy_rule_documented_but_never_verified_live` |
| order-id cap off by one | `sanitize_order_id_rewrites_illegal_characters_and_caps_the_length` |
| delayed settle reported as `Charged` | `the_payment_status_maps_are_settlement_aware_and_never_optimistic` |
| `refundhash` built with the two-stage digest | `the_rebate_password_hash_is_a_single_sha1_not_the_two_stage_digest` |
| response blueprint drops `authcode` | `the_response_hash_is_verified_skipped_or_mismatched_and_never_assumed` |
| **pasref ladder inverted** | `the_pasref_ladder_prefers_the_authorization_over_the_transaction_id` |

The last one is the production bug this PR opened with. It can no longer regress silently.

`ensure_liability_shift` now takes `Option<&AuthenticationData>` rather than the whole 57-field
`PaymentsAuthorizeData` — narrower, and testable without constructing a payment. No behaviour change.


crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5168 @@
...
+// 3DS2 JSON API digests (separate API — see `super::three_ds_two`)
+// -----------------------------------------------------------------------------
+
+/// The `Authorization: securehash <digest>` blueprints of the Global Payments **3DS2 JSON API**.
+///
+/// The JSON API reuses the XML API's two-stage SHA-1 primitive ([`realex_digest`]) and the same
+/// Shared Secret, but **nothing else**: different field lists, a different timestamp format
+/// ([`current_3ds2_timestamp`]) and a different card-scheme vocabulary
+/// (`MASTERCARD`, not the XML `MC` — see [`map_card_type`] and
+/// [`super::three_ds_two::map_3ds2_scheme`]).
```
JeevaRamu0104: `super::three_ds_two` doesn't exist — the 3DS2 code is in this file, so this intra-doc link is broken (rustdoc warns), as is the plain reference on line 370. The same stale path is in all three config TOMLs.

config/production.toml
```diff
@@ -154,6 +154,14 @@ easebuzz.base_url = "https://pay.easebuzz.in"
...
 axisbank.base_url = "https://upisdk.axisbank.co.in/api/b2/"
 tsys_transit.base_url = "https://gateway.transit-pass.com/servlets/TransNox_API_Server"
+globalpayments_realex.base_url = "https://api.realexpayments.com/"
+# 3DS2 lives on a different host, a different protocol (JSON, not XML) and a different
+# auth scheme (`Authorization: securehash`) from the XML gateway above. See
+# connectors/globalpayments_realex/three_ds_two.rs.
+# NOTE: two Global Payments documents disagree on the production 3DS2 host
+# (`api.globalpay-ecommerce.com` vs `api.globalpay.com`); the majority spelling is
+# configured here and must be confirmed with GP before the production cut-over.
+globalpayments_realex.secondary_base_url = "https://api.globalpay-ecommerce.com/"
```
JeevaRamu0104: This host is a guess — the comment above says two GP documents disagree and it needs confirming. `connector_3ds2_base_url` already hard-errors with a clear message when `secondary_base_url` is unset, so leaving production unset until GP confirms fails better than pointing 3DS2 traffic at a possibly-wrong host.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5168 @@
...
+/// | `508` | `You may only rebate up to 115% of the original amount.` | the 115% variant of the ceiling on accounts configured that way |
+/// | `508` | `Original transaction not found.` | wrong `<pasref>` / `<orderid>` pair. **Verified live** with the `settle`-minted pasref |
+/// | `512` | `This transaction has already been rebated and cannot be rebated again.` | second partial refund on an account not enabled for multiple refunds |
+/// | `512` | `You can't refund a delayed transaction that has not been sent for settlement …` | the authorization was never captured — void it instead |
+/// | `506` | `The line number 2 which contains '…' does not conform to the schema` | unparsable `<pasref>`. **Verified live** |
+///
+/// Every one of these arrives as HTTP 200 in the small error document, so they are classified from
+/// `<result>` here rather than from a transport status.
+fn map_refund_status(result: &str) -> RefundStatus {
```
JeevaRamu0104, line 2199: `505 The refund password you entered was incorrect.` is a credential misconfiguration, and it lands on `RefundStatus::Failure` — terminal, so every refund on a misconfigured account becomes a permanently-failed record nobody can retry. `is_query_lookup_fault` already treats `502|503|505|506|508` as saying nothing about the outcome; `ManualReview` fits them here too, and in the capture/void maps.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5168 @@
...
+                ),
+                ..Default::default()
+            },
+        })
+}
+
+/// `<orderid>` accepts `[a-zA-Z0-9_-]` only, 1–50 characters, and must be unique per attempt
+/// (a reuse returns `501 … already been processed`). Anything outside the charset is replaced
+/// with `-`, then the value is truncated.
+pub fn sanitize_order_id(reference: &str) -> Result<String, error_stack::Report<IntegrationError>> {
```
JeevaRamu0104, line 532: The cap here is 50, but RSync refuses any original order id over 42 (`RSYNC_ORIGINAL_ORDER_ID_MAX_LEN`, line 2813) because `_rebate_` eats 8. A 43-50 character reference authorizes and refunds fine, then every RSync on it fails. Worth capping at 42 where the order id is minted rather than only at sync time.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5168 @@
...
+            response,
+            router_data,
+            http_code,
+        } = value;
+
+        // The flag we sent is what decides whether `00` means Charged or Authorized, so recover it
+        // from the request rather than guessing from the response.
+        let auto_settle =
+            GlobalpaymentsRealexAutoSettleFlag::try_from(router_data.request.capture_method)
+                .change_context(ConnectorError::ResponseHandlingFailed {
```
JeevaRamu0104, line 1122: `Default::default()` here (and at 14 other sites in this file) means a response we can't map gives us nothing to debug with — and on a gateway that answers HTTP 200 for everything, that's the only signal there is. Can the status code and the `<result>`/`<message>` go in, at least on the `ResponseHandlingFailed` paths with no `attach_printable`?

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5168 @@
...
+                ),
+                ..Default::default()
+            },
+        }));
+    }
+
+    let encoded = payload_field(payload, FIELD_THREE_DS_METHOD_DATA)
+        .ok_or_else(|| missing("redirect_response.payload.threeDSMethodData"))?;
+
+    let server_trans_id = server_trans_id_from_encoded(&encoded).ok_or_else(|| {
```
JeevaRamu0104, line 3989: nit: `read_challenge_return` validates this with `validate_server_trans_id`; the method return doesn't. No injection path here (JSON body, not a URL), but the asymmetry reads as an oversight.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
shuklatushar226: Fixed in 23bff0b5e — and thank you, this was my own refactor's litter. When `three_ds_two.rs` was folded into `transformers.rs` I retargeted the code references but missed the two doc ones and all four config comments. `[`super::three_ds_two::map_3ds2_scheme`]` is now `[`map_3ds2_scheme`]`, the banner on line 370 points at the 3DS2 section, and `development.toml` / `sandbox.toml` / `production.toml` / `superposition.toml` (x2) name the section of `transformers.rs` instead of the deleted file. Confirmed with `cargo doc -p connector-integration --no-deps`: no unresolved link for this connector.

config/production.toml
shuklatushar226: Agreed, and removed in 23bff0b5e — both here and in the superposition production override. Failing loud beats guessing: an unset value gets `InvalidConnectorConfig` naming the key, whereas a wrong host means a real PAN and a real `securehash` going somewhere nobody has vetted. The comment now says DELIBERATELY UNSET with the reason, so the next person does not "helpfully" fill it in.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
shuklatushar226: Right, and fixed in 23bff0b5e. `505 The refund password you entered was incorrect.` is just the wrong `api_secret`, and burning that to a terminal `Failure` leaves an operator with a refund record they cannot retry once the credential is fixed. The `502/503/505/506/508` family now maps to `ManualReview` — the gateway refused to act on the request, so no money moved, but the refund was never *declined* either. `512 already been rebated`, the over-refund ceiling and declines stay `Failure`, since those are its actual answer about this refund. Helper renamed `is_integration_fault` now that PSync and Refund share it.

I did **not** extend it to the capture/void maps, and want to flag that rather than let it pass silently. `CaptureFailed` / `VoidFailed` leave the payment itself intact and are already distinct from a payment failure, so the "permanently unusable record" argument that carries the refund case does not obviously carry there — whether a caller can retry a capture after `CaptureFailed` is a caller-side question I could not answer from inside the connector. Happy to change it if you know the answer.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
shuklatushar226: Good catch — this was a real trap and it is fixed in 23bff0b5e. `mint_order_id` now pays for the `_rebate_` prefix up front at the three sites that derive an id from `connector_request_reference_id` (Authorize request, the metadata the Authorize response publishes, and the 3DS2 AReq), and `RSYNC_ORIGINAL_ORDER_ID_MAX_LEN` is now *defined as* `MINTED_ORDER_ID_MAX_LEN` so the two cannot drift apart again.

Ids arriving from stored metadata still round-trip at full 50 via `sanitize_order_id` — truncating one the gateway has already accepted would break every follow-up on an in-flight payment.

Verified live: a 48-character reference now mints a 42-character order id, and authorize -> settle -> rebate -> RSync all return `00`, with the rebate leg addressed as `_rebate_` + 42 = exactly 50. Before the fix that payment refunded fine and then failed every sync.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
shuklatushar226: Fixed in 23bff0b5e. The eight `ResponseHandlingFailed` sites with no `attach_printable` now carry `response_error_context(&response, http_code)`: the HTTP status goes in `ResponseTransformationErrorContext::http_status_code` (the type has a dedicated field for it), and `<result>`, `<message>` and the order id go in `additional_context`. Your framing is the right one — on a gateway that answers HTTP 200 for everything, `<result>`/`<message>` are the only signal there is.

Deliberately left out: `<pasref>`, `<authcode>` and the digest. This string is appended to an error message that travels, and the order id is enough to tie a failure back to a payment.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
shuklatushar226: Fixed in 23bff0b5e — `read_method_return` now applies the same `validate_server_trans_id`. You are right that there is no injection path here (the id goes into the AReq body and its digest, never a URL), but a value that is not a 3DS Server transaction id cannot identify an authentication either way.

One consequence worth knowing, since it is not free: this also validates our own synthesised no-DDC payload, whose id comes from the protocol-versions response. If GP ever emitted a non-UUID `serverTransactionId` the method leg would now fail — though the challenge leg already would, so the asymmetry was hiding the risk rather than avoiding it. Two test fixtures used `"sid-1"` and had to become real UUIDs, which is how I noticed.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
```diff
@@ -0,0 +1,5316 @@
...
+}
+
+/// `<amount currency="EUR">1001</amount>` — text content plus a currency attribute.
+#[derive(Debug, Serialize)]
+pub struct GlobalpaymentsRealexAmount {
+    #[serde(rename = "@currency")]
+    pub currency: String,
+    #[serde(rename = "$text")]
+    pub value: String,
+}
```
deepanshu-iiitu, line 729: Can we use currency enum and amount struct here?

shuklatushar226: Half of this is a clear win and is now done; the other half would be a money bug, so I've left it and written the reason into the struct doc.

**Currency: done.** `currency` is now `common_enums::Currency` instead of `String`. It serialises `UPPERCASE` via serde's `rename_all` and Displays `UPPERCASE` via strum's `serialize_all`, so the bytes on the wire stay identical to the `to_string()` that feeds the SHA-1 digest, and the field can no longer be handed an arbitrary string.

**Amount: deliberately still a `String`.** `value` is produced by `format_amount`, which isn't a plain integer rendering — it applies the documented JPY x100 rule (tech spec 5.4) and rejects zero/negative amounts. Serialising a `MinorUnit` directly would silently drop both: a JPY payment would go out at a hundredth of its intended value. It would also break the rule that every digest input is byte-identical to what ends up on the wire — the digest is built from these exact strings, so a second, independent rendering path is exactly what we don't want here.

Those two paths agreeing was load-bearing and unguarded, so I added `amount_element_currency_matches_the_digest_input_byte_for_byte`, which pins it across EUR/USD/GBP/JPY. If either the serde or the strum attribute is ever changed, that test fails instead of the gateway rejecting every request with a hash mismatch.

In `0e865abcb`.

crates/integrations/connector-integration/src/connectors/globalpayments_realex/transformers.rs
JeevaRamu0104, line 538: these need to be moved to Id framework 
https://github.com/juspay/hyperswitch-prism/issues/2331. Please add this PR in the issue comment

JeevaRamu0104 (dismissed):

deepanshu-iiitu (dismissed):

JeevaRamu0104 (approved):

deepanshu-iiitu (approved):

PR #2207, opened 2026-09-01
feat(connector): implement Authorize, PSync, Capture, Void, Refund, RSync for globalpaymentsheartland

shuklatushar226:
## Summary

Implement **Authorize, PSync, Capture, Void, Refund, RSync** for the new **GlobalpaymentsHeartland** connector.

This is a brand-new connector for Global Payments' **Heartland Portico** gateway (`Hps.Exchange.PosGateway`). It is a **SOAP 1.1 / XML** connector, not JSON, and it is **distinct from the pre-existing `globalpay`** (GP-API / Realex) connector — the two should not be conflated.

**Scope:** Card only, one-time payments, 3DS and non-3DS. No mandates/MIT, no tokenization, no wallets, no webhooks, no payouts.

## Flow → Portico transaction mapping

| Flow | Portico transaction |
|---|---|
| Authorize (manual capture) | `CreditAuth` |
| Authorize (auto capture) | `CreditSale` |
| Capture | `CreditAddToBatch` |
| Void | `CreditVoid` |
| Refund | `CreditReturn` |
| PSync / RSync | `ReportTxnDetail` |

## Implementation notes

- **SOAP transport.** Uses the existing `curl_request: SoapXml(...)` macro path, with `Content-Type: text/xml; charset=utf-8` plus a `SOAPAction` header carrying literal surrounding quotes. The body is plain nested XML with **no CDATA wrapper** — `bamboraapac`'s CDATA wrapping is deliberately *not* copied, because Portico rejects it.
- **Response preprocessing.** `preprocess_response: true` strips the `soap:` prefix and all four `xmlns` declarations, because `quick-xml` does not strip namespace prefixes.
- **Amounts are major-unit decimal strings** (`StringMajorUnitForConnector` / `CurrencyUnit::Base`). `minor_amount: 3000` goes on the wire as `Amt` = `"30.00"`.
- **Every Portico response is HTTP 200.** Status is derived from the body through a two-level gate: first `Header/GatewayRspCode`, then the issuer `RspCode` inside the transaction element.
- **`Transaction` is `Option<...>`.** A gateway-level rejection omits the element entirely; making it optional is what turns that into a readable connector error instead of a deserialization failure. Proven end to end by negative test #7 below.
- **Capture, Void and Refund return *empty* XML elements** — success is `GatewayRspCode == 0` only.
- **3DS is external pass-through** via a `<Secure3D>` block inside `Block1`. No redirect, no ACS URL, no `RedirectForm`.
- **Auth is `ConnectorAuthType::HeaderKey`.** The key travels in the SOAP body at `Ver1.0/Header/SecretAPIKey`, never in an HTTP header, so `get_auth_header()` returns an empty vec.
- **Enum spelling.** The variant is `ConnectorEnum::Globalpaymentsheartland`, not `GlobalpaymentsHeartland`, because that enum carries `#[strum(serialize_all = "snake_case")]` and the camel-case spelling would produce the wrong wire name `globalpayments_heartland`.

## Changes

**Connector implementation**
- `crates/integrations/connector-integration/src/connectors/globalpaymentsheartland.rs`
- `crates/integrations/connector-integration/src/connectors/globalpaymentsheartland/transformers.rs`
- `crates/internal/integration-tests/src/connector_specs/globalpaymentsheartland/specs.json`

**Registration wiring**
- `crates/types-traits/grpc-api-types/proto/payment.proto` — `GLOBALPAYMENTSHEARTLAND = 140` (**appended; nothing renumbered**), new `GlobalpaymentsHeartlandConfig`, oneof field `globalpaymentsheartland = 151`
- `crates/types-traits/domain_types/src/connector_types.rs` — `ConnectorEnum::Globalpaymentsheartland`
- `crates/types-traits/domain_types/src/router_data.rs` — `ConnectorSpecificConfig`, masking lists, proto config conversion, `HeaderKey` legacy-auth arm
- `crates/types-traits/domain_types/src/types.rs` — `Connectors.globalpaymentsheartland`

**Config**
- `config/development.toml`, `config/sandbox.toml` — CERT `.asmx` endpoint
- `config/production.toml` — production `.asmx` endpoint
- `config/superposition.toml` — connector enum entry + override block

## gRPC Test Results

**Status: PASS** — all six flows tested live against the Portico **CERT** gateway on 2026-09-01, plus a negative gateway-rejection case. `cargo build` was green on the first iteration (the only rework was `cargo fmt`).

<details>
<summary>Full grpcurl log — all six flows + negative case (credentials redacted)</summary>

# GRPCURL_OUTPUT — GlobalpaymentsHeartland (Portico CERT gateway, live, 2026-09-01)

Build: `cargo build` green on the first iteration (only rework was `cargo fmt`). All six flows PASS.

All calls used the same three headers (server on `localhost:8055`):
```
-H 'x-connector: globalpaymentsheartland'
-H 'x-auth: header-key'
-H 'x-api-key: <REDACTED_SECRET_API_KEY>'
```

Shared card payload (`gph_auth_manual.json`; the other payloads differ only in `merchant_transaction_id`, `amount`, `capture_method` and the 3DS block):
```json
{
  "merchant_transaction_id": "gph_auth_manual",
  "amount": {"minor_amount": 3000, "currency": "USD"},
  "payment_method": {"card": {
    "card_number": {"value": "4111111111111111"},
    "card_exp_month": {"value": "12"},
    "card_exp_year": {"value": "2030"},
    "card_cvc": {"value": "123"},
    "card_holder_name": {"value": "John Doe"}}},
  "capture_method": "MANUAL",
  "address": {"billing_address": {
    "first_name": {"value": "John"}, "last_name": {"value": "Doe"},
    "line1": {"value": "1 Heartland Way"}, "city": {"value": "Jeffersonville"},
    "state": {"value": "IN"}, "zip_code": {"value": "47130"},
    "country_alpha2_code": "US", "email": {"value": "test@example.com"},
    "phone_number": {"value": "4155552671"}, "phone_country_code": "+1"}},
  "auth_type": "NO_THREE_DS",
  "return_url": "https://example.com/return",
  "browser_info": {"ip_address": "1.2.3.4"}
}
```

## 1. Authorize — Card, non-3DS, MANUAL capture (`CreditAuth`)
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d @ localhost:8055 types.PaymentService/Authorize < gph_auth_manual.json
```
```json
{
  "merchantTransactionId": "624419121582",
  "connectorTransactionId": "200139844518",
  "status": "AUTHORIZED",
  "statusCode": 200,
  "networkTransactionId": "006244411372140",
  "connectorReferenceId": "624419121582",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139844518\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"CreditAuth\":{\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"AuthCode\":\"02006A\",\"AVSRsltCode\":\"0\",\"CVVRsltCode\":\"M\",\"RefNbr\":\"624419121582\",\"CardType\":\"Visa\",\"CardBrandTxnId\":\"006244411372140\"},\"CreditSale\":null}}}}}"}
}
```

## 2. Capture (`CreditAddToBatch`) of `200139844518`
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_capture_id":"gph_capture","connector_transaction_id":"200139844518","amount_to_capture":{"minor_amount":3000,"currency":"USD"}}' \
  localhost:8055 types.PaymentService/Capture
```
```json
{
  "connectorTransactionId": "200139844518",
  "status": "CHARGED",
  "statusCode": 200,
  "merchantCaptureId": "200139845257",
  "connectorReferenceId": "200139845257",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139845257\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"}}}}}"}
}
```
Note the empty `<CreditAddToBatch />`: no `Transaction` key in the parsed response, `resource_id` still the authorization, batch id in `connectorReferenceId`.

## 3. PSync (`ReportTxnDetail`) of the authorization
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_transaction_id":"gph_auth_manual","connector_transaction_id":"200139844518","capture_method":"MANUAL","amount":{"minor_amount":3000,"currency":"USD"}}' \
  localhost:8055 types.PaymentService/Get
```
```json
{
  "connectorTransactionId": "200139844518",
  "status": "AUTHORIZED",
  "statusCode": 200,
  "amount": {"minorAmount": "3000", "currency": "USD"},
  "merchantTransactionId": "624419121582",
  "connectorReferenceId": "624419121582",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139845259\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"ReportTxnDetail\":{\"GatewayTxnId\":\"200139844518\",\"ServiceName\":\"CreditAuth\",\"OriginalGatewayTxnId\":\"0\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\",\"Data\":{\"TxnStatus\":\"A\",\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"Amt\":\"30.00\",\"AuthAmt\":\"30.00\",\"SettlementAmt\":\"30.00\",\"AuthCode\":\"02006A\",\"SaleReturnInd\":\"S\",\"ReturnAmtInfo\":null,\"ReversalAmtInfo\":\"0.00\",\"MaskedCardNbr\":\"411111******1111\",\"CardType\":\"Visa\",\"RefNbr\":\"624419121582\"}}}}}}}"}
}
```
`ServiceName=CreditAuth` x `TxnStatus=A` -> `AUTHORIZED` (not `CHARGED`).

## 4a. Authorize — non-3DS, AUTOMATIC capture (`CreditSale`), 31.00
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d @ localhost:8055 types.PaymentService/Authorize < gph_sale_void.json
```
```json
{
  "merchantTransactionId": "624419121789",
  "connectorTransactionId": "200139845260",
  "status": "CHARGED",
  "statusCode": 200,
  "networkTransactionId": "006244411542228",
  "connectorReferenceId": "624419121789",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139845260\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"CreditAuth\":null,\"CreditSale\":{\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"AuthCode\":\"02077A\",\"AVSRsltCode\":\"0\",\"CVVRsltCode\":\"M\",\"RefNbr\":\"624419121789\",\"CardType\":\"Visa\",\"CardBrandTxnId\":\"006244411542228\"}}}}}}"}
}
```

## 4b. Void (`CreditVoid`) of `200139845260`
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_void_id":"gph_void","connector_transaction_id":"200139845260","cancellation_reason":"requested_by_customer","amount":{"minor_amount":3100,"currency":"USD"}}' \
  localhost:8055 types.PaymentService/Void
```
```json
{
  "connectorTransactionId": "200139845260",
  "status": "VOIDED",
  "statusCode": 200,
  "merchantVoidId": "200139847021",
  "connectorReferenceId": "200139847021",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139847021\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"}}}}}"}
}
```

## 5a. Authorize — AUTOMATIC capture (`CreditSale`), 33.00
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d @ localhost:8055 types.PaymentService/Authorize < gph_sale_refund.json
```
```json
{
  "merchantTransactionId": "624419121791",
  "connectorTransactionId": "200139845261",
  "status": "CHARGED",
  "statusCode": 200,
  "networkTransactionId": "006244411554698",
  "connectorReferenceId": "624419121791",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139845261\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"CreditAuth\":null,\"CreditSale\":{\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"AuthCode\":\"02082A\",\"AVSRsltCode\":\"0\",\"CVVRsltCode\":\"M\",\"RefNbr\":\"624419121791\",\"CardType\":\"Visa\",\"CardBrandTxnId\":\"006244411554698\"}}}}}}"}
}
```

## 5b. Refund 5.00 (`CreditReturn`, partial) of `200139845261`
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_refund_id":"gph_refund","connector_transaction_id":"200139845261","payment_amount":3300,"refund_amount":{"minor_amount":500,"currency":"USD"},"reason":"partial refund test"}' \
  localhost:8055 types.PaymentService/Refund
```
```json
{
  "merchantRefundId": "gph_refund",
  "connectorRefundId": "200139844519",
  "status": "REFUND_SUCCESS",
  "statusCode": 200,
  "connectorTransactionId": "200139845261",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139844519\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"}}}}}"}
}
```

## 5c. RSync (`ReportTxnDetail` on the refund's own id `200139844519`)
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_refund_id":"gph_refund","connector_transaction_id":"200139845261","refund_id":"gph_refund","connector_refund_id":"200139844519","refund_amount":{"minor_amount":500,"currency":"USD"}}' \
  localhost:8055 types.RefundService/Get
```
```json
{
  "merchantRefundId": "gph_refund",
  "connectorRefundId": "200139844519",
  "status": "REFUND_SUCCESS",
  "statusCode": 200,
  "connectorTransactionId": "200139845261",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139842841\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"ReportTxnDetail\":{\"GatewayTxnId\":\"200139844519\",\"ServiceName\":\"CreditReturn\",\"OriginalGatewayTxnId\":\"200139845261\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\",\"Data\":{\"TxnStatus\":\"A\",\"RspCode\":\"\",\"RspText\":\"\",\"Amt\":\"-5.00\",\"AuthAmt\":\"-5.00\",\"SettlementAmt\":\"-5.00\",\"AuthCode\":\"\",\"SaleReturnInd\":\"R\",\"ReturnAmtInfo\":null,\"ReversalAmtInfo\":null,\"MaskedCardNbr\":\"411111******1111\",\"CardType\":\"Visa\",\"RefNbr\":\"\"}}}}}}}"}
}
```
The RSync trap in action: `RspCode` is empty and `Amt` is `-5.00`, and it still reads `REFUND_SUCCESS` because the gate is `ServiceName == CreditReturn` first.

## 6a. Authorize — Card, external 3DS pass-through (`CreditSale` + `<Secure3D>`), 34.00
Extra fields over the base payload:
```json
"auth_type": "THREE_DS",
"enrolled_for_3ds": true,
"authentication_data": {
  "eci": "5",
  "cavv": "AAACAWQyc5UyIzUwNTJlNTQAAAAA",
  "message_version": "2.2.0",
  "ds_transaction_id": "0dc42c8b-3e07-4c8b-9a5f-1c1d0d0f2a11",
  "threeds_server_transaction_id": "9f8d2a1b-77c3-4d1a-9a2e-1b6c5e3f4a02"
}
```
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d @ localhost:8055 types.PaymentService/Authorize < gph_sale_3ds.json
```
```json
{
  "merchantTransactionId": "624419122011",
  "connectorTransactionId": "200139844524",
  "status": "CHARGED",
  "statusCode": 200,
  "networkTransactionId": "006244411692961",
  "connectorReferenceId": "624419122011",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139844524\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"CreditAuth\":null,\"CreditSale\":{\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"AuthCode\":\"02140A\",\"AVSRsltCode\":\"0\",\"CVVRsltCode\":\"M\",\"RefNbr\":\"624419122011\",\"CardType\":\"Visa\",\"CardBrandTxnId\":\"006244411692961\"}}}}}}"}
}
```
Outbound request body as logged by the server (secrets masked by the framework), showing `<Secure3D>` in the correct `Block1` position:
```json
{"header":{"SecretAPIKey":"***","DeveloperID":"002914","VersionNbr":"3333"},
 "transaction":{"CreditSale":{"Block1":{"AllowDup":"Y","Amt":"34.00",
   "CardData":{"ManualEntry":{"CardNbr":"***","ExpMonth":"***","ExpYear":"***","CVV2":"***","CardPresent":"N","ReaderPresent":"N"}},
   "CardHolderData":{"...":"..."},
   "Secure3D":{"Version":"2","AuthenticationValue":"***","ECI":"5","DirectoryServerTxnId":"0dc42c8b-3e07-4c8b-9a5f-1c1d0d0f2a11"}}}}}
```
No `redirectionData` in the response — external pass-through only.

## 6b. PSync of the 3DS sale `200139844524`
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_transaction_id":"gph_sale_3ds","connector_transaction_id":"200139844524","capture_method":"AUTOMATIC","amount":{"minor_amount":3400,"currency":"USD"}}' \
  localhost:8055 types.PaymentService/Get
```
```json
{
  "connectorTransactionId": "200139844524",
  "status": "CHARGED",
  "statusCode": 200,
  "amount": {"minorAmount": "3400", "currency": "USD"},
  "merchantTransactionId": "624419122011",
  "connectorReferenceId": "624419122011",
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139842843\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\"},\"Transaction\":{\"ReportTxnDetail\":{\"GatewayTxnId\":\"200139844524\",\"ServiceName\":\"CreditSale\",\"OriginalGatewayTxnId\":\"0\",\"GatewayRspCode\":\"0\",\"GatewayRspMsg\":\"Success\",\"Data\":{\"TxnStatus\":\"A\",\"RspCode\":\"00\",\"RspText\":\"APPROVAL\",\"Amt\":\"34.00\",\"AuthAmt\":\"34.00\",\"SettlementAmt\":\"34.00\",\"AuthCode\":\"02140A\",\"SaleReturnInd\":\"S\",\"ReturnAmtInfo\":null,\"ReversalAmtInfo\":null,\"MaskedCardNbr\":\"411111******1111\",\"CardType\":\"Visa\",\"RefNbr\":\"624419122011\"}}}}}}}"}
}
```
`ServiceName=CreditSale` x `TxnStatus=A` -> `CHARGED`, the counterpart to call 3.

## 7. Negative — Capture referencing an unknown transaction (gateway-level reject, `<Transaction>` absent)
```bash
grpcurl -plaintext -H 'x-connector: globalpaymentsheartland' -H 'x-auth: header-key' \
  -H 'x-api-key: <REDACTED_SECRET_API_KEY>' \
  -d '{"merchant_capture_id":"gph_capture_bad","connector_transaction_id":"1","amount_to_capture":{"minor_amount":100,"currency":"USD"}}' \
  localhost:8055 types.PaymentService/Capture
```
```json
{
  "connectorTransactionId": "200139847024",
  "status": "CAPTURE_FAILED",
  "error": {
    "issuerDetails": {"networkDetails": {}},
    "connectorDetails": {
      "code": "3",
      "message": "Transaction rejected because the referenced original transaction is invalid. Subject '1'.  Original transaction not found.",
      "reason": "Transaction rejected because the referenced original transaction is invalid. Subject '1'.  Original transaction not found.",
      "connectorTransactionId": "200139847024"
    }
  },
  "statusCode": 200,
  "typedConnectorResponse": {"value": "{\"Body\":{\"PosResponse\":{\"Ver1.0\":{\"Header\":{\"GatewayTxnId\":\"200139847024\",\"GatewayRspCode\":\"3\",\"GatewayRspMsg\":\"Transaction rejected because the referenced original transaction is invalid. Subject '1'.  Original transaction not found.\"}}}}}"}
}
```
HTTP 200 with a readable gateway error rather than a deserialization failure — the `Option<Transaction>` requirement proven end to end.

</details>

## ~~Known follow-up (note, not a blocker)~~ — RESOLVED in this PR

~~PSync on a **voided `CreditSale`** returns `TxnStatus = "R"`, which maps to `PENDING` per the spec's explicit rule to *default unrecognised `TxnStatus` values to Pending and not invent mappings*. Hardening `R -> Voided`, gated on `ServiceName == CreditSale`, is a deliberate follow-up decision and is intentionally not done in this PR.~~

**Superseded.** This was originally deferred, but it is a real defect — a PSync after a successful Void would report `PENDING` forever — so it is now fixed in commit `f11b86be2`. `R` was confirmed live on both voided `CreditSale` *and* voided `CreditAuth`, so the mapping is gated on each rather than on `CreditSale` alone. See the *Follow-up fix: PSync after Void* section at the end of this description for the evidence.

## Validation Checklist

- [x] grpcurl Authorize returned success status (2xx) — `AUTHORIZED`, `statusCode: 200`
- [x] All six flows (Authorize, PSync, Capture, Void, Refund, RSync) verified live against Portico CERT
- [x] Negative gateway-rejection case verified (surfaces a readable error, not a deserialization failure)
- [x] No credentials in committed source code, config, or `specs.json`
- [x] Only connector-specific files and registration wiring modified
- [x] Proto enum appended — no existing field numbers renumbered

---

### Follow-up fix: PSync after Void (`TxnStatus = R`)

A defect found after the initial round of testing: `map_psync_status` matched only
`TxnStatus == "A"`, so a PSync issued **after** a successful Void fell through to the
conservative `_ => Pending` arm and reported `PENDING` forever. The original test sequence
never paired a PSync with a Void, so it could not surface this.

Re-probing `ReportTxnDetail` on the CERT gateway showed `A` is not the only value:

| GatewayTxnId | ServiceName | TxnStatus | ReversalAmtInfo | Meaning |
|---|---|---|---|---|
| `200139845260` | `CreditSale` | `R` | `31.00` | sale that was voided |
| `200139841418` | `CreditAuth` | `R` | `16.00` | auth that was voided |
| `200139837899` | `CreditSale` | `R` | `12.00` | sale that was voided |
| `200139837897` | `CreditAuth` | `A` | `0.00`  | auth captured, **not** voided |
| `200139847021` | `CreditVoid` | `I` | `31.00` | the void record itself |

So `R` on the **original** transaction means reversed/voided, and a `CreditVoid` record
reports `I`, not `A` — which made the old `(CreditVoid, "A")` arm unreachable in practice.

Fix: added `(CreditAuth, "R") -> Voided`, `(CreditSale, "R") -> Voided`, and replaced the
unreachable arm with `(CreditVoid, _) -> Voided` (a `ReportTxnDetail` fetched by a
`CreditVoid` id *is* the void record, so its existence proves the void). The conservative
`_ => Pending` fallback is kept for genuinely unobserved values, and the `TXN_STATUS_ACTIVE`
doc comment claiming `A` was the only value ever observed has been corrected.

Re-verified live after rebuilding (`types.PaymentService/Get`):

| Case | GatewayTxnId | Before | After |
|---|---|---|---|
| voided sale (`CreditSale`/`R`) | `200139845260` | `PENDING` | **`VOIDED`** |
| voided auth (`CreditAuth`/`R`) | `200139841418` | `PENDING` | **`VOIDED`** |
| void record (`CreditVoid`/`I`) | `200139847021` | `PENDING` | **`VOIDED`** |
| open auth (`CreditAuth`/`A`) | `200139844518` | `AUTHORIZED` | `AUTHORIZED` (no regression) |
| 3DS sale (`CreditSale`/`A`) | `200139844524` | `CHARGED` | `CHARGED` (no regression) |
| captured auth (`CreditAuth`/`A`) | `200139837897` | `AUTHORIZED` | `AUTHORIZED` (no regression) |
| RSync (`CreditReturn`) | `200139844519` | `REFUND_SUCCESS` | `REFUND_SUCCESS` (no regression) |

`cargo build`, `cargo clippy` and `cargo fmt --check` are all clean for the touched crate.


shuklatushar226:
### CI: `Run Tests` failure was a workflow permissions bug, now fixed in this PR

The red `Run Tests` job on `f11b86be2` was **not** a test or connector failure. Every substantive step passed:

| Step | Result |
|---|---|
| 11 Run Tests | pass |
| 12 Build certification binaries | pass |
| 13 Install grpcurl | pass |
| 14 Verify newly added connectors' declared scenarios | pass |
| 15 Certify selected connectors | pass |
| 16 Upload certification reports | pass |
| **17 Comment unproven connector declarations on the PR** | **fail** |

Step 17 died on:

```
HttpError: Resource not accessible by integration  (403)
POST /repos/juspay/hyperswitch-prism/issues/2207/comments
```

**Root cause.** The `test` job declares no `permissions:` block, so it inherits the workflow-level default `permissions: contents: read` from the top of `ci.yml`. `github.rest.issues.createComment` needs `pull-requests: write`, so the POST is rejected and the whole job fails after all the real work has already succeeded.

**This is latent, not a regression.** The step only fires when a brand-new connector directory is declared unproven (`steps.verify-new-connectors.outputs.unproven == 'true'`). `globalpaymentsheartland` is the first entry in `alpha_connectors.json` to carry a `reason` — every other entry is `{}` — so this is the first PR ever to take that path. The step has, as far as I can tell, never run successfully.

**Fix** (965f41b27): grant the `test` job `pull-requests: write`, keeping `contents: read` explicit since naming a `permissions` block drops inherited defaults. Fork PRs are unaffected: `GITHUB_TOKEN` is read-only for them regardless, and the job already skips forks.

Happy to split this into its own PR if you'd rather not carry a CI change alongside a connector — say the word. It's here because without it this PR cannot go green.

docs-generated/llms.txt
```diff
@@ -342,6 +342,22 @@ payment_methods: Card, Eps, Ideal
...
 
+## Globalpayments Heartland
+connector_id: globalpayments_heartland
+doc: docs/connectors/globalpayments_heartland.md
+scenarios: checkout_autocapture, checkout_card, refund, void_payment, get_payment
+payment_methods: Card
+flows: authorize, capture, get, proxy_authorize, refund, refund_get, void
+examples_python: examples/globalpayments_heartland/globalpayments_heartland.py
+
+## Globalpaymentsheartland
```
JeevaRamu0104, line 353: `globalpaymentsheartland` isn't a real connector id — `x-connector` resolves through `ConnectorEnum::from_str`, and the enum carries `#[strum(serialize_all = "snake_case")]`, so only `globalpayments_heartland` parses. This block and its siblings (`docs-generated/connectors/globalpaymentsheartland.md`, `data/field_probe/globalpaymentsheartland.json`, `examples/globalpaymentsheartland/`) are left over from the pre-rename spelling. Deleting those nine files and regenerating also fixes `total_connectors: 114` -> 113.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
```diff
@@ -0,0 +1,1628 @@
...
+        // and `SettlementAmt` are all byte-identical. `SettlementAmt` is populated *before*
+        // capture — it is the amount that would settle, not the amount that has. The only
+        // delta is `ReversalAmtInfo` appearing as `0.00`, a zero-valued reversal field that
+        // says nothing about capture.
+        //
+        // The framework cannot help either: `PaymentFlowData::status` is hard-coded to
+        // `Pending` for every sync and `PaymentServiceGetRequest` carries no prior status, so
+        // there is nothing to compare against. Closing this needs either a Portico batch query
+        // (a second round trip per sync) or an optional prior-status field on the proto.
+        (Some(SERVICE_NAME_CREDIT_AUTH), Some(TXN_STATUS_ACTIVE)) => AttemptStatus::Authorized,
```
JeevaRamu0104, line 1377: This arm matches a captured `CreditAuth` as well as an open one, so a PSync after Capture walks the payment back from `Charged` to `Authorized` — a merchant polling status can then re-capture or void a payment that already settled. The evidence in the comment is solid, but the constraint is visible only here. Can it go in the connector spec/docs too, and was `ReportBatchDetail`/`BatchSeqNbr` checked before settling on this?

crates/types-traits/domain_types/src/router_data.rs
```diff
@@ -1035,6 +1035,39 @@ pub enum ConnectorSpecificConfig {
...
+/// `InvalidDataFormat { field_name: "config_override" }`, not a silent drop, so a merchant
+/// supplying `base_url` in `x-connector-config` has the whole request rejected.
+///
+/// Lowercasing the variant name is right for the ~105 single-word connectors. The variants
+/// below are the ones whose field name is not simply the lowercased variant.
+///
+/// Deliberately NOT listed, and deliberately not snake_cased: `RazorpayV2` -> `razorpayv2` and
+/// `BankOfAmerica` -> `bankofamerica`. Those really are the field names, so the default is
+/// correct and a blanket snake_case conversion would break two connectors that work today.
+fn connector_patch_key(variant: &str) -> String {
```
JeevaRamu0104, line 1060: This changes the patch key for five existing connectors — `absa_sanlam`, `gotyme_sanlam`, `tsys_transit`, `pinelabs_online`, `twoc_twop_paco`. It's correct (I checked each against the field names at `types.rs:407-415`), but it's unrelated to the new connector and nothing pins it. A unit test asserting each variant's key equals its `Connectors` field name would stop the next multi-word connector from silently regressing it.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
```diff
@@ -0,0 +1,1628 @@
...
+///
+/// Portico takes **plain nested XML** — unlike `bamboraapac`, there is no `<![CDATA[…]]>`
+/// wrapper, and adding one produces a body the gateway rejects. Both the header and the
+/// transaction element go through `quick_xml::se` so every value is XML-escaped; only the
+/// fixed envelope scaffolding is formatted in.
+fn to_pos_request_envelope<TXN: Serialize>(
+    header: &GlobalpaymentsHeartlandRequestHeader,
+    transaction: &TXN,
+) -> String {
+    let header_xml = quick_xml::se::to_string(header).unwrap_or_else(|_| String::from("<Header/>"));
```
JeevaRamu0104, line 250: A serialization failure here sends `<Header/>` — a request with no `SecretAPIKey` — and `validate_xml_structure` in the `SoapXml` macro passes it, because it is structurally valid XML. Line 251 does the same for the transaction body. `bamboraapac` has this pattern but doesn't put the credential in the fallback path. Could the serialization move into the `TryFrom`, where it can return `Err`?

crates/integrations/connector-integration/src/connectors/globalpayments_heartland.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// would never bind to a `Body` field and the response structs would silently fail.
+    /// Drop the `soap:` prefix and the four namespace declarations Portico emits before
+    /// the body reaches the parser. The structure is otherwise left untouched.
+    pub fn preprocess_response_bytes<F, FCD, Req, Res>(
+        &self,
+        _req: &RouterDataV2<F, FCD, Req, Res>,
+        bytes: bytes::Bytes,
+        _status_code: u16,
+    ) -> CustomResult<bytes::Bytes, IntegrationError> {
+        let response_str = String::from_utf8(bytes.to_vec()).change_context(
```
JeevaRamu0104, line 182: `RequestEncodingFailed` with an empty context, for a UTF-8 failure decoding a *response*. The transformers in this connector already use the better helper — `crate::utils::response_deserialization_fail(status_code, "...")` — which carries the HTTP status and a message.

data/field_probe/globalpaymentsheartland.json
```diff
@@ -0,0 +1,780 @@
+{
```
JeevaRamu0104, line 1: nit: pre-rename duplicate of `globalpayments_heartland.json` — see the `llms.txt` comment.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
```diff
@@ -0,0 +1,1650 @@
...
+/// Anything that is not a 1- or 2-digit number is surfaced as an error rather than
+/// truncated: silently reshaping an unrecognised ECI would assert an authentication outcome
+/// the 3DS server never reported.
+fn normalize_eci(eci: &str) -> Result<String, Report<IntegrationError>> {
+    let trimmed = eci.trim();
+    let normalized = match trimmed.len() {
+        1 => trimmed,
+        2 if trimmed.starts_with('0') => &trimmed[1..],
+        _ => "",
+    };
```
deepanshu-iiitu, line 118: Check this

deepanshu-iiitu (dismissed):

data/field_probe/globalpaymentsheartland.json
```diff
@@ -0,0 +1,784 @@
...
```
JeevaRamu0104: Root of the phantom `globalpaymentsheartland` connector: `scripts/generators/docs/generate.py:577` globs `data/field_probe/*.json`, so this file regenerates a second doc page, four examples, an `llms.txt` block, and `total_connectors: 117` — one too many. The id never parses (`ConnectorEnum` is snake_case) and the examples name `GlobalpaymentsheartlandConfig`, absent from the proto. Delete it and regenerate.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
```diff
@@ -0,0 +1,1654 @@
...
+        // and `SettlementAmt` are all byte-identical. `SettlementAmt` is populated *before*
+        // capture — it is the amount that would settle, not the amount that has. The only
+        // delta is `ReversalAmtInfo` appearing as `0.00`, a zero-valued reversal field that
+        // says nothing about capture.
+        //
+        // The framework cannot help either: `PaymentFlowData::status` is hard-coded to
+        // `Pending` for every sync and `PaymentServiceGetRequest` carries no prior status, so
+        // there is nothing to compare against. Closing this needs either a Portico batch query
+        // (a second round trip per sync) or an optional prior-status field on the proto.
+        (Some(SERVICE_NAME_CREDIT_AUTH), Some(TXN_STATUS_ACTIVE)) => AttemptStatus::Authorized,
```
JeevaRamu0104, line 1413: Still open from the last round. This arm reports `Authorized` for a captured auth as well as an open one, so a PSync after Capture walks the payment back from `Charged` and a merchant polling status can void a payment already in the batch. Was `ReportBatchDetail`/`BatchSeqNbr` checked before settling on a report-only derivation?

crates/types-traits/domain_types/src/router_data.rs
```diff
@@ -1067,6 +1067,39 @@ pub enum ConnectorSpecificConfig {
...
```
JeevaRamu0104, line 1102: Still no test pinning this. I checked all 123 variants against the 119 `Connectors` fields: the six listed are correct, but five others resolve to keys matching no field at all — `Screenstream`, `Ebanx`, `Globepay`, `Coinbase`, `Coingate`. Pre-existing, and exactly what a test asserting key == field name would catch. Worth adding.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
```diff
@@ -0,0 +1,1654 @@
...
+// =============================================================================
+
+/// Sent on every verified pre-flight call. Optional per the Portico guide, pinned here so
+/// the wire format matches exactly what was verified.
+const DEVELOPER_ID: &str = "002914";
+/// Ditto — pinned alongside `DeveloperID`.
+const VERSION_NBR: &str = "3333";
+/// **Required.** Without it Portico rejects a repeated same-amount transaction as a
+/// duplicate instead of approving it.
+const ALLOW_DUP: &str = "Y";
```
JeevaRamu0104, line 60: `Y` unconditionally disables Portico's duplicate check for every live merchant, not just CERT. The justification holds for repeated fixed-amount test calls, but in production that check is the gateway-side double-charge guard, and it is pinned on for Refund too (`:1190`). Should this be request- or config-driven rather than a constant?

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
shuklatushar226: Agreed, and fixed for refunds in 4716f7e97 — `AllowDup` is no longer sent on `CreditReturn`.

Verified against the cert gateway rather than reasoned about:

```
refund without AllowDup      -> Success
identical refund repeated    -> "Transaction was rejected because it is a duplicate. Subject '200140119808'."
```

and end to end through UCS:

```
first refund      -> REFUND_SUCCESS
identical retry   -> REFUND_FAILURE
```

So omitting it genuinely restores the double-payout guard. Worth noting Portico exposes no `ClientTxnId` on this `Block1` (the schema fault lists only `AllowDup` among the idempotency-ish children), so the gateway duplicate check is the only mechanism the refund flow has — which is exactly why disabling it was wrong.

It stays on authorizations, where the same check would reject a legitimate repeat purchase of the same amount by the same card. Your request-/config-driven suggestion is the right end state and I have not done it: it needs a new field on `ConnectorSpecificConfig` plus the Hyperswitch side, so it is a bigger change than this PR should carry. I have recorded the trade-off on the constant so it is not silently pinned.

data/field_probe/globalpaymentsheartland.json
shuklatushar226: Fixed at the root in 4716f7e97, and thank you for pinning it to `generate.py` globbing `data/field_probe/*.json` — that is exactly the mechanism.

One correction on the remedy: deleting the file alone would not have held. `probe_connector` runs for every connector using a dummy `probe_key` (`orchestrator.rs`, no credentials needed), so the auto-fix job regenerates it — it already did once, in 67bedb99e, which is why it reappeared after I renamed the connector.

The cause is `orchestrator.rs:12`: `format!("{connector:?}").to_lowercase()` lowercases the Debug name and so drops the underscore for every multi-word connector. It now uses `connector.to_string()` — the strum snake_case `Display` the rest of the system already uses for `x-connector`, config keys and `connector_specs/`.

Stale artifacts for this connector are deleted and the `llms.txt` block removed (`total_connectors` corrected). This also fixes six pre-existing cases whose probe files the next auto-fix will rename: `absasanlam`, `pinelabsonline`, `tsystransit`, `twoctwoppaco`, `jpmorganorbital`, `razorpayv2`. `twoc_twop_paco` already carried the same duplicate pair on main.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
shuklatushar226: Checked, and you were right to push on it — I had only diffed `ReportTxnDetail` before. `ReportBatchDetail` does discriminate:

```
fresh CreditAuth 200140029684, uncaptured  -> 0 occurrences in the open batch
after CreditAddToBatch                     -> 1 occurrence          BatchStatus=O
```

So batch membership answers the question that `ReportTxnDetail` cannot. For completeness, the report-only derivation really is a dead end — diffing the same transaction either side of a capture (200139928251), `ServiceName`, `TxnStatus`, `Amt`, `AuthAmt` and `SettlementAmt` are byte-identical. `SettlementAmt` is populated *before* capture: it is the amount that would settle, not the amount that has.

I have still not implemented it, for two reasons I would rather state than bury:

1. It costs a second gateway round trip per PSync, returning ~100KB listing every transaction in the open batch (74 on the cert account) which then has to be scanned.
2. More seriously, it queries the **open** batch. Once the batch settles at end of day the transaction is no longer in it, so a captured-and-settled payment reads as uncaptured again — the same bug, now intermittent and time-dependent, which is worse than the current honest limitation.

Querying the specific historical `BatchId` would avoid that, but `ReportTxnDetail` does not expose which batch the transaction went into, so there is nothing to query with.

The framework cannot close it either: `PaymentFlowData::status` is hard-coded to `Pending` for every sync (`domain_types/src/types.rs`) and `PaymentServiceGetRequest` carries no prior status, so there is no previous value to compare against. Every connector in this repo that documents "retains the previous status" is in fact retaining `Pending`.

Happy to implement the batch lookup if you would rather have it despite the settlement window — your call. Otherwise the honest fix is an optional prior-status field on `PaymentServiceGetRequest`, which is a proto change and out of scope here.

crates/types-traits/domain_types/src/router_data.rs
shuklatushar226: Your five extra cases are correct and I confirmed them independently: `Screenstream`, `Ebanx`, `Globepay`, `Coinbase` and `Coingate` resolve to keys with no matching `Connectors` field. They are unreachable in practice — `ConnectorVariant::foreign_try_from` rejects all five before the patch path — but they are latent and a test asserting key == field name is exactly what would catch them, along with the next multi-word connector.

I have not added the test. That is not a disagreement with you: I am working under a standing instruction from the PR author not to add tests in this PR, and I have flagged your request to them twice rather than quietly overriding it. If they lift it, this is the first test I would write.

One implementation note in case it is useful whenever it does get written: the fix deliberately is not a blanket snake_case conversion. `RazorpayV2` and `BankOfAmerica` map to the fields `razorpayv2` and `bankofamerica`, so converting them would break two connectors that work today. The override list names only the variants whose field name is not the lowercased variant, and a test asserting key == field name would pin that too.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
shuklatushar226: I could not tell from "check this" what specifically you had in mind, so I exercised the function before changing anything, to be sure I was not papering over a real defect:

```
"05"->5  "02"->2  "07"->7  "00"->0  "5"->5  " 05 "->5     accepted
"5X"  "X"  ""  "10"  "12"  "099"  "0é"  "é"  "0\0"        rejected
```

No functional bug: `"10"`/`"12"` are correctly rejected because ECI is 0-7 so a two-digit form is always zero-padded, and the `&trimmed[1..]` byte-slice could not panic on multi-byte input because it was guarded by `starts_with('0')`, making index 1 always a char boundary.

What I did change (6c6909e22) is the thing I think you were pointing at: the match keyed off `trimmed.len()` and used an empty string as an "invalid" sentinel, then re-validated it below — you had to carry the sentinel in your head to see that the `1 => trimmed` arm was still digit-checked. It now matches on the characters, so each arm states its own condition and there is no sentinel. Behaviour is unchanged, re-verified live: ECI `05`/`02`/`07`/`5` all CHARGED, `5X`/`12` rejected before any gateway call.

If you meant something else, say the word and I will look again — I would rather you correct me than have me guess.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland.rs
shuklatushar226: Fixed in e64381ec4, with one deviation worth flagging.

`crate::utils::response_deserialization_fail` returns `ConnectorError`, but `preprocess_response_bytes` returns `CustomResult<_, IntegrationError>`, so it does not type-check at this call site — and `IntegrationError` has no `ResponseDeserializationFailed` variant to mirror it. I used `InvalidDataFormat { field_name: "response_body", .. }` instead, carrying the HTTP status and a message.

Your underlying point stands and is fixed: it was the wrong category (a request-encoding error for a response decode) with an empty context. If you would rather these two error enums converge so the helper is usable here, that seems worth doing, but it is broader than this PR.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
shuklatushar226: Fixed in e64381ec4 — and your detail about `validate_xml_structure` is what made it dangerous: `<Header/>` is structurally valid XML, so validation passed it and the credential-less request went out, with Portico's auth rejection as the only symptom.

Moving serialization into the `TryFrom` is the right shape and I did not do it: `GetSoapXml::to_soap_xml` returns `String`, not `Result`, and is implemented by other connectors, so returning an error from there is a shared-trait change across six request structs and all six flows.

What I did instead makes the failure impossible to send: both fallbacks are now deliberately malformed, so `validate_xml_structure` rejects them and the request never leaves, and each failure is logged rather than swallowed. That closes the credential-leak-shaped hole without a trait change, but it is a smaller fix than yours and I would support doing the `TryFrom` version separately.

docs-generated/llms.txt
shuklatushar226: Fixed at the root — see the reply on `data/field_probe/globalpaymentsheartland.json`. Short version: the file regenerates because field-probe named its output from the lowercased Debug name, dropping the underscore; that is now `connector.to_string()` (strum snake_case). The stale doc page, the four examples and this `llms.txt` block are removed and `total_connectors` corrected.

data/field_probe/globalpaymentsheartland.json
shuklatushar226: Deleted, and the generator that recreates it is fixed — full explanation in the reply on your 9 Sep comment on this same file.

crates/integrations/connector-integration/src/connectors/globalpayments_heartland/transformers.rs
shuklatushar226: Answered in full on the 9 Sep thread on this same line. Short version: I did check `ReportBatchDetail`, and it does discriminate — an uncaptured auth appears 0 times in the open batch, 1 time after `CreditAddToBatch`. I have not implemented it because it costs a second ~100KB round trip per PSync and, more importantly, it queries the **open** batch, so a captured-and-settled payment would read as uncaptured again after end of day — the same bug, but intermittent. The report-only route is genuinely a dead end: `SettlementAmt` is populated before capture, so it is byte-identical either side.

crates/types-traits/domain_types/src/router_data.rs
shuklatushar226: Answered on the 9 Sep thread on this same line, including your five extra cases (`Screenstream`, `Ebanx`, `Globepay`, `Coinbase`, `Coingate`) which I confirmed independently. The test is not added because I am under a standing instruction from the PR author not to add tests in this PR; I have raised your request with them rather than overriding it silently.

JeevaRamu0104 (approved):

swangi-kumari (approved):

PR #2205, opened 2026-09-01
feat(connector): implement Authorize for d24

shuklatushar226:
## Summary

Implement **Authorize** flow for the **D24** (Directa24) connector.

New connector integrating the Directa24 **Deposits v3 "webpaycard"** card deposit API — Authorize flow, Card payment method, one-time only, covering both no-3DS and D24-managed 3DS.

Tracking issue: juspay/hyperswitch-cloud#23005 — `feat(connector): add D24 - webpaycard`

This implementation was generated and validated by **GRACE** (automated connector integration pipeline).

## Changes

- Added new connector `d24.rs` with the `create_all_prerequisites!` macro and `macro_connector_implementation!` for Authorize
- Added D24 Authorize request/response types and `TryFrom` implementations in `d24/transformers.rs`
- Registered D24 in the connector enum / dispatch (`connectors.rs`, `default_implementations.rs`, `types.rs`, `domain_types`, `payment.proto`, `field-probe/src/auth.rs`)
- Added `d24.base_url` to `config/development.toml`, `config/sandbox.toml`, `config/production.toml` and the connector enum in `config/superposition.toml`

### Implementation highlights

- **Auth (signature-key)**: `Authorization: "D24 " + hex(HMAC_SHA256(api_secret, X-Date || X-Login || body))`, signed over the **byte-identical transmitted body** (the serialized JSON string is reused for both signing and transmission, so no re-serialization drift).
- **`X-Date`** formatted without sub-seconds (`%Y-%m-%dT%H:%M:%SZ`), as required by D24.
- **Amounts**: `CurrencyUnit::Base` with a `FloatMajorUnit` converter — D24 amounts are major-unit JSON numbers (e.g. `1000.0`), not minor-unit integers.
- **`expiration_year`** truncated to 2 digits (`2030` → `30`).
- **`authentication_url`** parsed from both the response root and the nested `payment_info` object, so either shape resolves to redirection data.
- **Errors**: flat `{code, description, details, type}` error envelope.

## Files Modified

- `crates/integrations/connector-integration/src/connectors/d24.rs` *(new)*
- `crates/integrations/connector-integration/src/connectors/d24/transformers.rs` *(new)*
- `crates/integrations/connector-integration/src/connectors.rs`
- `crates/integrations/connector-integration/src/default_implementations.rs`
- `crates/integrations/connector-integration/src/types.rs`
- `crates/internal/field-probe/src/auth.rs`
- `crates/types-traits/domain_types/src/connector_types.rs`
[...]

PR #2212, opened 2026-09-01
feat(connector): implement D24 (Directa24) WebPay Authorize, PSync, Refund and RSync

shuklatushar226:
> **Merge and deploy order: this PR must land and deploy before juspay/hyperswitch#13976.**
> `d24` is listed in the router's `ucs_only_connectors`, so Hyperswitch always delegates D24 to UCS and builds a `ConnectorSpecificConfig::D24` for it. A router that knows D24 talking to a UCS that does not will fail at the config mapping. The reverse order is inert — nothing references the connector until the router side lands.

## What

Implements **Directa24 (`D24`) WebPay** — `payment_method: "WP"`, Transbank's Chilean redirect method — as **Authorize + PSync + Refund + RSync**.

The customer is redirected to Transbank's own page and enters card details there, so **no card data passes through UCS**. The request body carries no card fields at all.

- **Endpoints:** `POST /v3/deposits`, `GET /v3/deposits/{deposit_id}`, `POST /v3/refunds`, `GET /v3/refunds/{refund_id}` — all on the **non-PCI** host (`api-stg.directa24.com` / `api.directa24.com`)
- **Scope:** Chile (`CL`) only, currencies `CLP` and `USD`
- **Payment method:** `PaymentMethodData::CardRedirect(CardRedirectData::CardRedirect {})`
- **Out of scope:** mandates, webhooks, 3DS modelling (Transbank runs its own challenge behind the redirect)
- **Capture and Void are not implemented because Directa24 does not support them** — see below. This is a documented absence, not a gap.

Paired HS PR: juspay/hyperswitch#13976 (D24 wired UCS-only).
Tracking issue: juspay/hyperswitch-cloud#23005.

## Supersedes #2205

PR #2205 implemented the **wrong Directa24 product** — the PCI direct-card flow on the `cc-api` host with a raw `credit_card` body. It has been closed and its branch deleted. This PR is a fresh implementation; none of the card request/response modelling was carried over.

## Capture and Void: why they are absent

Neither exists at Directa24, so neither is implementable:

- **Capture** — there is no capture endpoint. `POST /v3/deposits` carries no `capture`, `auto_capture`, `manual_capture` or `capture_method` field anywhere in its OpenAPI, and Directa24's own *Authorization and capture* documentation page is an empty stub. Deposits are sale-only. The connector rejects a manual-capture request locally rather than letting it fail at D24.
- **Void** — there is no cancel/void/annul/reverse endpoint. The deposit status `CANCELLED` is reachable only by the customer abandoning the flow, by the automatic transition 7 days after expiry, or by a click in the Merchant Panel — never through the API. `back_url` is a redirect target for abandonment, not an operation.

The only merchant-initiated reversal Directa24 offers over the API is a refund of a `COMPLETED` deposit, which this PR implements.

## Refund and RSync

`POST /v3/refunds` and `GET /v3/refunds/{refund_id}`, on the same non-PCI host.

**`POST /v3/refunds` requires both `deposit_id` (integer) and `invoice_id` (string)** — the endpoint identifies the target deposit by the pair, not by either alone. `deposit_id` comes from `connector_transaction_id`; `invoice_id` comes from `RefundsData.connector_order_id` passed back through the same `sanitize_invoice_id()` that Authorize used. Both sides bottom out in `payment_attempt.connector_request_reference_id`, and the sanitiser is idempotent on its own output, so the bytes match what D24 stored. `connector_order_id` is `None` on the relay-refund path and on older attempts — those are refused locally rather than sent with a wrong identifier.

**`amount` is always sent, never omitted.** Directa24 reads an absent `amount` as "refund the full deposit" and supports multiple cumulative partial refunds, so omitting it after a first partial would re-attempt the full amount and fail with `802 INVALID_AMOUNT_TO_REFUND`. No field on the refund request carries the *remaining* refundable balance, so the requested amount is the only correct value.

**`bank_account` is deliberately not sent.** Directa24 documents it as required for bank-account refunds but "not for credit card refunds", and WebPay's `payment_type` is `CREDIT_CARD`. See the open questions — this is the one unverified assumption in the flow.

**One create-response struct covers both documented shapes** (card, with `refund_info`; bank/APM, `refund_id` only). They differ by presence, not structure, so an all-optional struct is used rather than an untagged enum, which would let a merely-absent field select the wrong arm. `refund_id` is accepted both quoted and as a bare integer, since the docs emit it both ways.

**Refund status mapping — two deliberate choices:**

| D24 status | Mapped to | Why |
|---|---|---|
| `PENDING` | `Pending` | awaiting processing |
| `INCORRECT_DETAILS` | **`ManualReview`** | waits on a human supplying corrected details, so it does not self-resolve and polling is wrong. It is also the canary for the `bank_account` assumption: for a card refund it should be unreachable. |
| `DELIVERED` | **`Pending`** | "sent to the bank, no longer cancellable" describes *our* loss of control, not the outcome — the bank can still bounce it to `REJECTED`. Mapping this to `Success` would report money returned before it was. |
| `COMPLETED` | `Success` | terminal |
| `REJECTED` | `Failure` | terminal |
| `CANCELLED` | `Failure` | terminal, money did not move. `RefundStatus` has no `Cancelled`; not `TransactionFailure`, since the underlying transaction was fine. |

A `REJECTED` create response returns `Ok` with `refund_status: Failure`, not `Err` — HTTP is 200 and a `refund_id` exists, and returning `Err` would discard `connector_refund_id` and make the rejection unreconcilable.

## Design notes worth a reviewer's attention

**Payment-method representation.** `CardRedirect(CardRedirect {})` was chosen over `BankRedirectData::LocalBankRedirect {}` because WebPay is card-on-Transbank's-page, not account-to-account. Precedent: `boost` already implements this exact variant for the same semantics. Note that dLocal maps `CountryAlpha2::CL => "WP"` through `BankDebitData::SepaBankDebit` — that path requires an IBAN and is unreachable from HS merchant config, so it was deliberately **not** copied. Both candidate variants are empty structs and fully plumbed, so swapping is a three-line change if review prefers the APM bucket.

**Error envelope is not uniform across the two endpoints.** Deposit creation returns a flat `{code: integer, description, details: [string]}`; `GET /v3/deposits/{id}` returns a nested `{"error": {"code": string, "message": string, "details": string|null}}` — note `code` is an integer on one and a string on the other. Since `build_error_response` is a single shared `ConnectorCommon` method, `D24ErrorResponse` is modelled as an untagged `Nested | Flat` enum. A flat-only model would panic on any PSync error.

**`key1` is the read-only API Key and is the `X-Login` on PSync GETs.** `build_headers` takes a `read_only_login: bool` selecting which credential feeds both the header and the signed message, with a fallback to `api_key` when `key1` is empty. See open question 1 below.

**Two status-mapping traps, both deliberately mapped to `Pending`:**
- `EARLY_RELEASED` reads terminal but the docs are explicit that the customer has not paid and the money is not credited. Mapping it to `Charged` would book unpaid revenue.
- `FOR_REVIEW` is a transient anti-fraud hold that can resolve either way — not `Unresolved` (which HS reads as needing merchant action) and not `Failure`.

**`checkout_type` has no declared enum** in the OpenAPI, so it is `Option<...>` with `#[serde(other)] Unknown`, and Authorize status is driven off the presence of `redirect_url` rather than off `checkout_type`.

**`notification_url` is deliberately omitted** — webhooks are out of scope and `IncomingWebhook` is an empty impl; sending it would make D24 POST to an endpoint that cannot parse the body.

**Amount:** `FloatMajorUnit` + `CurrencyUnit::Base` is retained. `CLP` is already a zero-decimal currency in `common_enums`, so 1000 CLP minor → `1000.0` on the wire and USD 1050 minor → `10.5`. Both verified.

## Validation — mock server only

**There are no live Directa24 credentials.** Validation was done against a local mock that independently recomputes the HMAC and asserts a match. Two legs were exercised.

**UCS → mock via grpcurl:** 14 requests, all `signature_match: true`.

**HS → UCS → mock, end to end** (the paired HS PR running against a real router): 8 Authorize + 8 PSync + 4 negative cases.

| Check | Result |
|---|---|
| `signature_match` on every request | 16/16 |
| `X-Login` = write key on POST, read-only key on GET | 8/8 and 8/8 |
| GET signed payload length | `0` on every GET — confirms the empty-payload signing rule |
| Authorize body | `payment_method: "WP"`, `country: "CL"`, `payer.document_type: "RUT"`, `amount: 1000.0` (not `100000`) |
| Forbidden fields | no `credit_card`, no `three_domain_secure`, no `notification_url` |
| All 8 PSync status mappings | 8/8 match |
| Refund requests (11 grpcurl + 11 REST) | 22/22 with `signature_match`, `login_match`, `invoice_id_match`, `deposit_id_is_int`, `amount_present`, `bank_account_absent`, `notification_url_absent` all true |
| All 6 RSync status mappings | 6/6 match, incl. `DELIVERED` -> pending and `INCORRECT_DETAILS` -> review |
| 4 negative cases | rejected inside UCS with **zero** outbound HTTP |

All 8 PSync statuses mapped as designed: `COMPLETED→succeeded`, `PENDING→processing`, `CREATED→requires_customer_action`, `DECLINED→failed`, `CANCELLED→cancelled`, `EXPIRED→expired`, `EARLY_RELEASED→processing`, `FOR_REVIEW→processing`.

Negative guards (rejected before any HTTP call): currency `BRL`, billing country `BR`, manual capture, malformed RUT.

### Refund path, proven through the Hyperswitch REST API

A before/after control confirms the HS-side gate is load-bearing: the identical refund request against a router built at the previous commit (`refunds: NotSupported`) returns **HTTP 422 `IR_06`** and never reaches the connector; against this branch it returns **HTTP 200** and reaches D24.

Full refund plus multiple partials were driven through `POST /refunds`, and all six D24 refund statuses through `GET /refunds/{id}?force_sync=true`. The `invoice_id` on every refund matched the value the mock recorded at deposit creation for that `deposit_id`. `amount` was transmitted on all 22 refund requests. The 800-series errors surface with D24's own code and sentence — e.g. `error_code: "800"`, `error_message: "Refunds are not enabled for this merchant account"` — rather than as an opaque transport error.

Two observations, neither introduced by this PR:
- All 800-series refund errors map to `unified_code: UE_9000` / "Something went wrong", so a merchant reading only the unified fields learns nothing. Pre-existing gap in the unified error mapping.
- A `REJECTED` refund carries no `error_message` on the *create* response — D24's `reason` / `reason_code` do not reach the merchant there, though the refund is correctly `failed` and reconcilable by `connector_refund_id`.

## CI registration

`crates/internal/integration-tests/src/connector_specs/d24/specs.json` is included, covering Authorize / PSync / Refund / RSync, and `d24` is registered in `alpha_connectors.json` (no live Directa24 credentials in CI, so it cannot be certified against a sandbox yet).

This matters because `names_in_dir()` in `.github/workflows/ci.yml` emits a connector name **only** when that spec file exists — a constraint tightened by #2201. Without it `d24` would never enter `payment_names`, and would be silently excluded from regression and nextest scoping with nothing going red.

**Known limitation:** `docs-generated/connectors/d24.md` lists only `PaymentService.Get` and `RefundService.Get`, omitting Authorize and Refund. That is a consequence of this connector's deliberately strict guards meeting field-probe's synthetic inputs — the probe drives refund with a non-numeric `connector_transaction_id`, which the transformer rejects with `InvalidDataFormat` rather than sending a malformed `deposit_id`. Correcting the doc would mean adding a flow-specific rule to the shared `crates/internal/field-probe/patch-config.toml`, which is consumed by every connector; that is left for the probe-harness owner rather than changed unilaterally here.

## Open questions for Directa24 — please confirm before go-live

1. **Read-only key on GET.** The docs state two credential sets exist but never say whether the read-only key *must* be used on GET, whether the POST key also works, or whether the read-only key has its own signature (unrepresentable — `D24Config` carries one `api_secret`). This PR assumes the first. **If wrong, PSync 401s on every call.**
2. **`WP` is absent from the fetched documentation corpus.** Every WP-specific fact — the method code, `payment_type: CREDIT_CARD`, Chile-only, CLP+USD, RUT 8–9 chars — comes from the coverage page, not the API reference. Worth confirming.
3. **`checkout_type` for WP** — `HOSTED` or `ONE_SHOT`? Undeclared; both plus unknown are handled.
4. **Chilean document type.** Docs list ID / RUN / RUT; `DocumentKind` can only express RUT (via `Other`). If RUN is required, that is a cross-repo enum + proto change.
5. **RUT wire format** — dotted, hyphenated, or bare? This PR strips separators (`12.345.678-K` → `12345678K`).
6. **`language` accepted values** — no enum documented; `es-CL` is sent.
7. **Does a WebPay refund require `bank_account`?** This is the one unverified assumption in the refund flow. Directa24's refund docs branch on "credit card deposits" versus "other payment methods" and never mention WebPay; WP's `payment_type` is `CREDIT_CARD`, which is why we take the credit-card branch and omit `bank_account`. If that is wrong, D24 answers `804 MISSING_BANK_ACCOUNT` — cheap and unambiguous. **Please confirm.**
8. **Is `POST /v3/refunds` authenticated with the deposit (write) API Key?** We send the write key by analogy with `POST /v3/deposits`, and the read-only key on the RSync GET. If refunds are gated on a separate credential, `X-Login` needs a third case.
9. **Is `DELIVERED` genuinely non-final** — can a `DELIVERED` refund still become `REJECTED`? We map it to `Pending` on that assumption. If it is irreversible, `Success` would settle refunds days earlier.
10. **What is the refund-cancellation endpoint?** The docs say `PENDING` and `INCORRECT_DETAILS` refunds are "still cancellable", which implies one exists but is not in the published pages.
11. **Is the refund `amount` interpreted in the deposit's currency or in USD?** We send the payment currency, which is correct only under the first reading.

## Not verified

Mock-only — nothing is confirmed against the real Directa24 host. Real cumulative-refund balance accounting is not exercised (the mock holds no balance); what is proven is that `amount` is always transmitted. The return-from-Transbank leg was not driven (payments were resolved by explicit `force_sync`). The `USD` branch was not exercised end to end. Refunds, capture, void and webhooks are not implemented by design.

https://claude.ai/code/session_01QW85baiHcjDH9usBH1WZEZ


XyneSpaces:
[should-fix] Add a D24 integration-test connector spec under `crates/internal/integration-tests/src/connector_specs/d24/specs.json`. The connector implements Authorize and Get, but without this registration the integration harness will skip those flows; include both supported suites in the new spec.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1103 @@
...
+    // log. Status is resolved by PSync instead. Add this field together with a
+    // real `IncomingWebhook` implementation, not before.
+}
+
+#[derive(Debug, Serialize)]
+pub struct D24Payer {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    pub document: Secret<String>,
+    pub document_type: &'static str,
```
deepanshu-iiitu: Can we make document_type an enum?

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1103 @@
...
+    }
+}
+
+/// A Chilean RUT is 7-8 body digits plus one check character (`0-9` or `K`),
+/// i.e. 8-9 characters once dots and the hyphen are stripped.
+///
+/// `DocumentKind::Other` deliberately skips the checksum validation that
+/// `Cpf`/`Cnpj` get, so this is the only place a malformed RUT is caught before
+/// it reaches Directa24 as an opaque `BEAN_VALIDATION_ERROR`.
+fn validate_and_normalize_rut(
```
deepanshu-iiitu: Can we move this function to validate chilean rut to connector utils file?

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1103 @@
...
+    pub invoice_id: Option<String>,
+    pub user_id: Option<String>,
+    pub country: Option<String>,
+    pub currency: Option<String>,
+    pub local_amount: Option<f64>,
+    pub usd_amount: Option<f64>,
+    pub amount: Option<f64>,
+    pub payment_method: Option<String>,
+    pub payment_type: Option<String>,
+}
```
deepanshu-iiitu, line 630: Make PII fields secret and use amount structs for amount related fields

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1103 @@
...
+    #[serde(rename = "type")]
+    pub refund_type: Option<String>,
+    pub result: Option<D24RefundResult>,
+    pub reason: Option<String>,
+    /// Documented inconsistently (integer in one example, string in another).
+    pub reason_code: Option<serde_json::Value>,
+    pub payment_method: Option<String>,
+    pub payment_method_name: Option<String>,
+    pub amount: Option<f64>,
+    pub currency: Option<String>,
```
deepanshu-iiitu: use floatmajorunit for amount and currency enum for currency field

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1103 @@
...
+            Self::Pending => "PENDING",
+            Self::IncorrectDetails => "INCORRECT_DETAILS",
+            Self::Delivered => "DELIVERED",
+            Self::Completed => "COMPLETED",
+            Self::Rejected => "REJECTED",
+            Self::Cancelled => "CANCELLED",
+            Self::Unknown => "UNKNOWN",
+        }
+    }
+}
```
deepanshu-iiitu: Why is this impl needed?

deepanshu-iiitu (changes requested):

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Done in 8795ec018 — `document_type` is now the `D24DocumentType` enum (`Rut`/`Cpf`/`Cnpj`/`Psn`) instead of `&'static str`. It serialises with `rename_all = "SCREAMING_SNAKE_CASE"`, so the values on the wire (`RUT`/`CPF`/`CNPJ`/`PSN`) are unchanged; `d24_document_type` now returns the enum, so the mapping is exhaustive at the type level.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Done in 8795ec018 — moved to `crate::utils::validate_and_normalize_chilean_rut`. Agreed it does not belong here: a RUT is a general Chilean document rule rather than a D24 one, so other LATAM connectors can reuse it. Behaviour is unchanged (7-8 body digits + `0-9`/`K` check char, separators stripped, upper-cased).

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Done in 8795ec018 — in `D24SyncResponse`, `user_id` is now `Secret<String>`, and `local_amount` / `usd_amount` / `amount` are `FloatMajorUnit` instead of `f64`. `currency` is now `common_enums::Currency` too.

Worth noting for safety: none of these fields are read anywhere in the connector (PSync drives status off `status` and the id off `deposit_id`), and the Authorize guard restricts currency to CLP/USD, so the stricter types cannot change behaviour. I deliberately left `country` as `String` — it is neither PII nor an amount, and typing an unread field to `CountryAlpha2` would only add a way for PSync deserialization to fail on an unexpected value.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Done in 8795ec018 — `D24RefundInfo.amount` is now `FloatMajorUnit` and `currency` is `common_enums::Currency`. Same note as above: neither field is read (the refund status comes from `result`), so this is type-safety rather than a behaviour change.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Good catch — it was carrying its weight, but not enough to justify being hand-written.

`as_str` fed `raw_connector_status`, which surfaces Directa24's own status string to the merchant (`response.status.as_str().to_string()` in PSync/RSync, and the refund `result` in Refund). So the string was needed, but the hand-rolled match was redundant with the `SCREAMING_SNAKE_CASE` serde convention already on the enum.

Replaced in 8795ec018 by `#[derive(strum::Display)]` + `#[strum(serialize_all = "SCREAMING_SNAKE_CASE")]` on all three status enums (`D24DepositStatus`, `D24RefundResult`, `D24RefundSyncStatus`), matching the pattern used by helcim/novalnet/worldpayvantiv. Same output strings (`EARLY_RELEASED`, `IN_PROGRESS`, `INCORRECT_DETAILS`, …), ~40 lines less code, and the strings can no longer drift from the serde names.

shuklatushar226:
> [should-fix] Add a D24 integration-test connector spec under `crates/internal/integration-tests/src/connector_specs/d24/specs.json`

Already addressed on the branch. `connector_specs/d24/specs.json` exists and declares all four implemented suites:

```json
{
  "connector": "d24",
  "supported_suites": [
    "PaymentService/Authorize",
    "PaymentService/Get",
    "PaymentService/Refund",
    "RefundService/Get"
  ]
}
```

This mattered more than it looks: since #2201, `names_in_dir` in `ci.yml` only emits a connector name if that `specs.json` exists, so without it `d24` never entered `payment_names` and CI silently skipped the connector entirely rather than failing.

Registering it then surfaced a second gate — `verify-new-connectors.sh` requires a non-empty `reason` for any new connector on the alpha list, which was failing "Run Tests" with *"listed in alpha_connectors.json with no reason"*. Fixed in 32e06f1c0; I confirmed by running the script locally with `NEW_CONNECTORS=d24`, which now exits 0 with the unproven-connector warning instead of an error.

deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1036 @@
...
+///
+/// `refund_id` stays required. `RefundsResponseData::connector_refund_id` is a
+/// non-optional `String`, so a missing id could only be papered over by writing
+/// the merchant refund id there — producing a refund that 404s on every
+/// subsequent poll, for ever. A hard deserialisation failure is the better
+/// outcome, and the raw body is still preserved on the error path.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct D24RefundResponse {
+    pub refund_id: D24RefundId,
+    pub deposit_id: Option<i64>,
```
JeevaRamu0104, line 930: `refund_id` is required, but the doc above lists the credit-card shape as `deposit_id` + `merchant_invoice_id` + `refund_info`, with no `refund_id`. WebPay is `payment_type: CREDIT_CARD`, so that's the shape production hits — and if that list is right, every WebPay refund fails to deserialize after D24 already accepted it, so `connector_refund_id` is never stored and the refund can never be polled. The `D24RefundId` comment says the opposite. Which one holds? The mock can't settle it.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1036 @@
...
+            item.connector.amount_converter,
+            request.minor_amount,
+            request.currency,
+        )?;
+
+        let document = request.get_customer_document_details()?;
+        let document_type = d24_document_type(document.document_type, country)?;
+        let document_number =
+            crate::utils::validate_and_normalize_chilean_rut(&document.document_number)?;
+
```
JeevaRamu0104, line 412: This runs on every document type, but `d24_document_type` above also accepts `Cpf`, `Cnpj` and `Psn`. A Chilean payer with a passport gets `document_type: PSN` and is then rejected here for not being 8–9 characters, so those three arms are unreachable. Gate the RUT check on `D24DocumentType::Rut`, or drop the arms.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1036 @@
...
+                })
+            } else {
+                None
+            }
+        };
+
+        // UCS carries a single return URL; D24 wants three. All three land back
+        // on the same HS redirect-sync endpoint, which is what triggers PSync.
+        let return_url = request.router_return_url.clone();
+
```
JeevaRamu0104, line 438: `router_return_url` is optional, so when it's absent all three of `success_url`/`back_url`/`error_url` are dropped. WebPay is a mandatory redirect — the customer then has nowhere to return and HS never gets the redirect-sync that kicks off PSync. `request.get_router_return_url()?` fails locally instead.

crates/integrations/connector-integration/src/connectors/d24.rs
```diff
@@ -0,0 +1,542 @@
...
+            req: &RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>,
+        ) -> CustomResult<Vec<(String, Maskable<String>)>, IntegrationError> {
+            // Deposit status is a read-only endpoint — the read-only API Key.
+            self.build_headers(req, true)
+        }
+        fn get_url(
+            &self,
+            req: &RouterDataV2<PSync, PaymentFlowData, PaymentsSyncData, PaymentsResponseData>,
+        ) -> CustomResult<String, IntegrationError> {
+            let deposit_id = req
```
JeevaRamu0104, line 350: nit: the Refund transformer parses `connector_transaction_id` as `i64` with a suggested action; here it goes into the URL raw, so a non-numeric id 404s on every poll instead of failing locally.

crates/integrations/connector-integration/src/utils.rs
```diff
@@ -754,3 +754,45 @@ pub fn build_error_response(
...
+/// Validates and normalizes a Chilean RUT (Rol Único Tributario).
+///
+/// A RUT is 7-8 body digits plus one check character (`0-9` or `K`), i.e. 8-9
+/// characters once dots and the hyphen are stripped. Returns the normalized
+/// form: separators removed and the check character upper-cased.
+///
+/// `DocumentKind::Other` — the only spelling a RUT can arrive as, since
+/// `DocumentKind` has no `Rut` variant — skips the checksum validation that
+/// `Cpf`/`Cnpj` get, so this is the only place a malformed RUT is caught
+/// before it reaches the connector as an opaque validation error.
```
JeevaRamu0104: nit: this only checks shape — a RUT with a wrong mod-11 check digit still reaches D24 as the opaque error this line says it prevents.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -0,0 +1,1036 @@
...
+/// required; the documentation states the response is extensible, so unknown
+/// members are ignored. Note there is **no `refund_id` in the body** — the id
+/// the poll was issued against is echoed back from the request.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct D24RefundSyncResponse {
+    pub status: D24RefundSyncStatus,
+    pub deposit_id: Option<i64>,
+    pub merchant_invoice_id: Option<String>,
+    pub amount: Option<f64>,
+}
```
JeevaRamu0104, line 1043: nit: every other amount field in this file is `FloatMajorUnit`.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: You were right that the comment and the code contradicted each other, and the comment was the wrong one. Settled against the published OpenAPI rather than the mock.

`RefundSuccessResponse` is a **single** schema, not two. `refund_id` sits at the top of it, and it is `deposit_id`, `merchant_invoice_id` and `refund_info` that carry *"Returned for synchronously processed credit card refunds"*. So the two "shapes" differ only by the presence of those three, and `refund_id` is common to both — WebPay does return it, and the failure mode you describe does not occur.

Worth noting the schema declares **no `required` array at all**, so strictly nothing is guaranteed. Keeping `refund_id` required is therefore deliberate strictness, not schema-following, and the reason stands: `RefundsResponseData::connector_refund_id` is a non-optional `String`, so an absent id could only be papered over by writing the merchant refund id there, producing a refund that 404s on every poll for ever. Rewrote the comment in a09867cd0 to say all of this instead of the wrong thing.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Real bug, thanks. `validate_and_normalize_chilean_rut` ran unconditionally, so a payer with `DocumentKind::Psn` got `document_type: PSN` and was then rejected for not being 8-9 characters — the Cpf/Cnpj/Psn arms were unreachable in practice.

Fixed in a09867cd0 by taking your second option: dropped the arms rather than gating the check. Guard 2 already restricts WebPay to Chile and Directa24 accepts only a RUT for it, so those types were never reachable for this connector — the function was advertising support it could not honour. `D24DocumentType` now has a single `Rut` variant, so the mapping is exhaustive and the RUT check is unconditional by construction.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Agreed, real bug. Fixed in a09867cd0 — now `request.get_router_return_url()?`, and `success_url`/`back_url`/`error_url` are plain `String` rather than `Option<String>` so the type says they are always sent.

Silently dropping them was the worst available outcome: the deposit would still be created at Directa24, the customer would land on Transbank with nowhere to return, and HS would never get the redirect-sync that starts PSync — so the payment would sit at `requires_customer_action` while the money moved. Failing locally before the deposit exists is the only honest behaviour.

crates/integrations/connector-integration/src/connectors/d24.rs
shuklatushar226: Fixed in a09867cd0 — PSync now parses `connector_transaction_id` as `i64` with the same `suggested_action` message the Refund transformer uses, so the two flows behave identically. A non-numeric id fails locally with something actionable instead of building a URL that 404s on every poll for ever.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Fixed in a09867cd0 — `D24RefundSyncResponse.amount` is now `FloatMajorUnit`. It was the one I missed when the other amount fields were converted; nothing reads it, so this is consistency rather than a behaviour change.

crates/integrations/connector-integration/src/utils.rs
shuklatushar226: Fair — the doc claimed more than the code did. Implemented the check rather than softening the claim, in a09867cd0:

```rust
// Mod-11 check digit. Weights cycle 2,3,4,5,6,7 from the rightmost body
// digit; 11 - (sum mod 11) yields 11 => '0' and 10 => 'K'.
```

Shape alone genuinely was not enough: a transposed digit keeps both the length and the character classes intact, so it passed and then came back from Directa24 as exactly the opaque error the function exists to prevent.

**One consequence worth flagging:** `12345678K` — the document used in the mock-based e2e run recorded on this PR — is not a valid RUT (the correct check digit for body `12345678` is `5`), so it is now correctly rejected. Anyone re-running that harness should use `12345678-5`. I would rather surface that than have it look like a regression later.

JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

kanikac199 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

JeevaRamu0104 (approved):

kanikac199 (approved):

PR #2213, opened 2026-09-02
feat(Feature): [Connector] Implementing NSure FRM Connector

THE-CORE-BUG:
## Description
 Nsure FRM integration for Pre-auth flow

  Adds nSure.ai as an FRM connector, executed entirely by the connector-service.

  Flows implemented

  - PreRiskCheck — POST /transactions/{transactionId} with mode: preAuthorization
  - FrmPaymentOutcome — PUT /transactions/{transactionId}/status
  - FrmRefundProcessed — PUT /transactions/{transactionId}/status with refunded
  - FrmChargebackReceived — POST /transactions/{transactionId}/disputes

  Implementation notes

  - Auth is a static key — no token exchange. Authorization is sent verbatim with no scheme prefix, along with
    x-nsure-api-version (defaults to 2.0.0) and an optional x-nsure-app-id.
  - Amounts use FloatMajorUnit: {"valueInCurrency": 90, "currency": "USD"} is 90 dollars, not 90 cents.
  - merchant_transaction_id is used three ways — the URL path segment, metadata.uniqueRequestId, and the returned
    frm_transaction_id. nSure mints no id of its own, so the same value has to address the risk check and every later callback.
  - Decision mapping is fail-safe: Approved and SoftApproved map to Approve, Rejected to Reject, and Review / Not Reviewed / any
    unrecognised value to Review. Nothing unknown can approve a transaction.
  - The lifecycle callbacks are included because nSure's docs make the transition to fundsCaptured the formal handshake for
    their liability shift.

  Placement

  - nSure lives in frm_connectors/, matching the existing surcharge_connectors / payout_connectors / authenticator_connectors
    split.
  - Kount stays in connectors/ since it is dual-registered as a payment connector (PreAuthenticate DDC and
    ServerAuthenticationToken); the dispatch arm carries a comment explaining that exception.
  - Removes two stale // FrmConnectorEnum is currently empty comments and their #[allow(unreachable_code)] in composite-service,
    since that code is live now.

## How did you test it?
 No creds yet, Alpha integration


hrithikesh026 (dismissed):

crates/internal/integration-tests/src/connector_specs/nsure/specs.json
```diff
@@ -0,0 +1,4 @@
+{
+  "connector": "nsure",
```
gopikrishna000, line 2: nit: i think no need to put non payment connectors into spec json, they anyway dont have test scenerios yet

gopikrishna000 (dismissed):

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NsureSessionInfo {
+    /// Minted by the nSure browser/mobile SDK; optional within sessionInfo.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub device_id: Option<String>,
+    /// Required by nSure whenever `sessionInfo` is present.
+    pub user_agent: String,
+    /// Required by nSure whenever `sessionInfo` is present.
+    pub end_user_ip: String,
```
deepanshu-iiitu: Make this secret

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub language: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub country: Option<String>,
```
deepanshu-iiitu: Make this secret

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub account_type: Option<NsureAccountType>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NsureSessionInfo {
+    /// Minted by the nSure browser/mobile SDK; optional within sessionInfo.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub device_id: Option<String>,
```
deepanshu-iiitu: Make device_id secret

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+/// variants prism can fully populate are emitted; anything else uses the
+/// documented `other` variant (which requires only `name`) so the body stays
+/// valid rather than failing the variant's required fields.
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NsurePaymentMethod {
+    #[serde(rename = "type")]
+    pub payment_method_type: NsurePaymentMethodType,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub payment_processor: Option<String>,
```
deepanshu-iiitu, line 250: Should we make payment_processor an enum?

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration_year: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub card_holder_name: Option<Secret<String>>,
+    // ── shared ──
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub billing_info: Option<NsureBillingInfo>,
+    /// Required by the `other` variant.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
```
deepanshu-iiitu: Make this secret

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+    pub address: Option<NsureAddress>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub phone_info: Option<NsurePhoneInfo>,
+}
+
+#[derive(Debug, Clone, Serialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct NsureAddress {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub country: Option<String>,
```
deepanshu-iiitu: Can we use the country enum for country field instead of string

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+        user_first_seen_timestamp: feature_data.user_first_seen_timestamp,
+        user_first_successful_tx_timestamp: feature_data.user_first_successful_tx_timestamp,
+        ..info
+    };
+    let Some(customer) = customer else {
+        return tenure(NsureEndUserInfo::default());
+    };
+    // `first_name`/`last_name` are preferred; fall back to splitting the
+    // single `customer_name` field on the first space when they are absent.
+    let (first_name, last_name) = match (&customer.first_name, &customer.last_name) {
```
deepanshu-iiitu, line 459: Why are we using customer name and not card holder name?

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+        email: customer
+            .customer_email
+            .as_ref()
+            .map(|email| Secret::new(email.peek().to_string())),
+        first_name,
+        last_name,
+        phone_info: nsure_phone_info(
+            customer.customer_phone_number.as_ref(),
+            customer.customer_phone_country_code.as_ref(),
+        ),
```
deepanshu-iiitu, line 487: Why are we not using billing phone?

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+            .as_ref()
+            .and_then(|phone| nsure_phone_info(phone.number.as_ref(), phone.country_code.as_ref())),
+    };
+    // Don't send an object where every field is empty.
+    let is_empty = info.first_name.is_none()
+        && info.last_name.is_none()
+        && info.address.is_none()
+        && info.phone_info.is_none();
+    (!is_empty).then_some(info)
+}
```
deepanshu-iiitu, line 520: Can we use pre-exising utils wherever possible?

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+
+/// Split a PAN into (bin, last4). nSure's card variant requires `last4`; `bin`
+/// is the first six digits, which it uses for issuer lookup.
+fn card_bin_last4(pan: &str) -> (Option<Secret<String>>, Option<Secret<String>>) {
+    let digits: String = pan.chars().filter(|c| c.is_ascii_digit()).collect();
+    let bin = (digits.len() >= 6).then(|| Secret::new(digits[..6].to_string()));
+    let last4 = (digits.len() >= 4).then(|| Secret::new(digits[digits.len() - 4..].to_string()));
+    (bin, last4)
+}
```
deepanshu-iiitu: Use/Create a util for this

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+
+/// Normalise a 2- or 4-digit expiry year to the 4-digit form nSure documents.
+fn four_digit_year(year: &Secret<String>) -> Secret<String> {
+    let raw = year.peek().trim();
+    if raw.len() == 2 {
+        Secret::new(format!("20{raw}"))
+    } else {
+        Secret::new(raw.to_string())
+    }
+}
```
deepanshu-iiitu: Use/Create a util for this

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1225 @@
...
+                | domain_types::payment_method_data::WalletData::PaypalSdk(_) => {
+                    NsurePaymentMethod {
+                        payment_method_type: NsurePaymentMethodType::Paypal,
+                        ..base
+                    }
+                }
+                domain_types::payment_method_data::WalletData::GooglePay(_)
+                | domain_types::payment_method_data::WalletData::GooglePayRedirect(_) => {
+                    NsurePaymentMethod {
+                        name: Some("googlePay".to_string()),
```
deepanshu-iiitu: Can we use an enum for the name field

deepanshu-iiitu (changes requested):

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: NsurePaymentMethod.name never holds user data. It's a fixed label we generate ourselves, and wrapping it in Secret would mask a constant giving no benefit

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: No specific list is present yet of their accepted values, neither in docs, will raise to the team for the request

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: Can be a case the endUser might not be the one using the card, then it might be different

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: We do -> inside paymentMethod.billingInfo.phoneInfo. 
Here it is buyer's phone

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
deepanshu-iiitu: What is the value stored in this field?

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1233 @@
...
+
+#[derive(Debug, Clone, Serialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct NsureEndUserInfo {
+    /// Stable merchant-side user key. nSure's `endUserInfo` oneOf requires it and
+    /// it is what lets them build cross-transaction history for the buyer.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub email: Option<Secret<String>>,
```
deepanshu-iiitu: lets use the email struct for this field

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1233 @@
...
+        // `AddressDetails::get_combined_address_line` is deliberately not used:
+        // it joins with `,` rather than a space and errors unless *both* lines
+        // are present, whereas most addresses carry only line1.
+        street: match (details.get_optional_line1(), details.get_optional_line2()) {
+            (Some(line1), Some(line2)) => {
+                Some(Secret::new(format!("{} {}", line1.peek(), line2.peek())))
+            }
+            (Some(line), None) | (None, Some(line)) => Some(line),
+            (None, None) => None,
+        },
```
deepanshu-iiitu: use/create a util for this

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+fn nsure_billing_info(address: Option<&Address>) -> Option<NsureBillingInfo> {
+    let address = address?;
+    let details = address.address.as_ref();
+    let nsure_address = details.map(|details| NsureAddress {
+        country: details.get_optional_country(),
+        // No `get_optional_state` exists; `get_state` returns a `Result` and this
+        // field is optional to nSure, so the field is read directly.
+        state: details.state.clone(),
```
deepanshu-iiitu: create the get_optional_state util

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1233 @@
...
+            .phone
+            .as_ref()
+            .and_then(|phone| nsure_phone_info(phone.number.as_ref(), phone.country_code.as_ref())),
+    };
+    // Don't send an object where every field is empty.
+    let is_empty = info.first_name.is_none()
+        && info.last_name.is_none()
+        && info.address.is_none()
+        && info.phone_info.is_none();
+    (!is_empty).then_some(info)
```
deepanshu-iiitu, line 519: use/create utils for this

deepanshu-iiitu (dismissed):
Connector changes look good

gopikrishna000:
core changes looks fine

gopikrishna000 (dismissed):

hrithikesh026 (dismissed):

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+    {
+        self.parsed_response.serialize(serializer)
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct NsureOrderDecision {
+    #[serde(default)]
+    pub decision: Option<NsureDecision>,
```
JeevaRamu0104, line 870: `decision` is optional-with-default under a `transparent` wrapper, so any 200 body parses. `{}` parses — `macros.rs:138` substitutes `{}` for an empty body — and so does a nested `{"orderDecision":{...}}`. Both yield `frm_decision: None`, which becomes `FrmDecision::Unspecified` on a success response with no error. The `Unknown -> Review` guard below only catches unrecognised strings, not a missing field. Make `decision` required, or map `None` to `Review`.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+        let session_info = match (user_agent, end_user_ip) {
+            (Some(user_agent), Some(end_user_ip)) => Some(NsureSessionInfo {
+                device_id: feature_data.device_id.clone().map(Secret::new),
+                user_agent: Secret::new(user_agent),
+                end_user_ip: Secret::new(end_user_ip),
+                language: browser
+                    .and_then(|info| info.language.as_deref())
+                    .and_then(nsure_language)
+                    .map(Secret::new),
+                country: req
```
JeevaRamu0104: This is the billing country, already sent as `paymentMethod.billingInfo.address.country`. `sessionInfo` is the device block (deviceId, userAgent, endUserIp), so its `country` is where the session is. Sourcing it from billing makes the two always agree and kills the mismatch signal — the same collapse you deliberately avoided for cardholder vs buyer name above. Omit it unless a session-derived country is available.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+        match status {
+            Some(
+                AttemptStatus::Charged
+                | AttemptStatus::PartialCharged
+                | AttemptStatus::PartialChargedAndChargeable,
+            ) => Some(Self::FundsCaptured),
+            Some(AttemptStatus::AuthorizationFailed) => Some(Self::ProcessorAuthorizationFailure),
+            Some(AttemptStatus::CaptureFailed) => Some(Self::FundsCaptureFailure),
+            Some(AttemptStatus::Failure) => Some(Self::ProcessorAuthorizationFailure),
+            Some(AttemptStatus::Voided | AttemptStatus::VoidedPostCapture) => Some(Self::Rejected),
```
JeevaRamu0104: The doc comment on `Rejected` says "Merchant followed a `Rejected` recommendation and stopped the payment." A void after an nSure *Approve* — buyer cancelled, stock-out, auth expiry — gets reported as their rejection being honoured, which feeds back into their model and the liability record. nSure has no "cancelled" value, so gating on `frm_decision == Some(Reject)` and returning `None` otherwise looks more honest than mislabelling it.

crates/integrations/connector-integration/src/frm_connectors/nsure.rs
```diff
@@ -0,0 +1,519 @@
...
+/// was the merchant transaction id. Falls back to the ids nSure/the PSP echoed
+/// back so a late notification can still be addressed.
+fn nsure_transaction_id(
+    merchant_transaction_id: Option<&str>,
+    frm_transaction_id: Option<&str>,
+    connector_transaction_id: Option<&str>,
+) -> CustomResult<String, IntegrationError> {
+    merchant_transaction_id
+        .or(frm_transaction_id)
+        .or(connector_transaction_id)
```
JeevaRamu0104: nSure keys transactions by the id sent on `POST /transactions/{id}` — the merchant's. The PSP's `connector_transaction_id` was never sent to them, so this arm addresses an id they have no record of and 404s, hiding the real problem (missing `frm_transaction_id`) behind a connector error. Drop the fallback and let the `MissingRequiredField` below fire.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+    Rejected,
+    SoftApproved,
+    Review,
+    #[serde(rename = "Not Reviewed")]
+    NotReviewed,
+    #[serde(other)]
+    Unknown,
+}
+
+impl From<&NsureDecision> for FrmDecision {
```
JeevaRamu0104, line 817: This mapping and `from_attempt_status` are pure table lookups carrying the fraud decision and the liability handshake, and neither has a unit test. Most connectors here ship none, so it isn't unusual — but these two are cheap to pin and expensive to get wrong.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+    ) -> Option<Self> {
+        match status {
+            Some(
+                AttemptStatus::Charged
+                | AttemptStatus::PartialCharged
+                | AttemptStatus::PartialChargedAndChargeable,
+            ) => Some(Self::FundsCaptured),
+            Some(AttemptStatus::AuthorizationFailed) => Some(Self::ProcessorAuthorizationFailure),
+            Some(AttemptStatus::CaptureFailed) => Some(Self::FundsCaptureFailure),
+            Some(AttemptStatus::Failure) => Some(Self::ProcessorAuthorizationFailure),
```
JeevaRamu0104, line 990: nit: `Failure` is the generic terminal status — a capture-stage failure gets reported to nSure as an authorization failure.

crates/integrations/connector-integration/src/frm_connectors/nsure.rs
```diff
@@ -0,0 +1,519 @@
...
+
+    fn common_get_content_type(&self) -> &'static str {
+        "application/json;charset=UTF-8"
+    }
+
+    fn base_url<'a>(&self, connectors: &'a Connectors) -> &'a str {
+        &connectors.nsure.base_url
+    }
+
+    fn get_auth_header(
```
JeevaRamu0104, line 182: nit: this returns only `Authorization`, omitting `x-nsure-api-version` and `x-nsure-app-id`. Nothing calls it today, but any default implementation that reaches for it instead of `nsure_headers` sends a request nSure rejects.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+                        value_in_currency: super::NsureAmountConvertor::convert(
+                            detail.amount,
+                            total.currency,
+                        )?,
+                        currency: total.currency,
+                    },
+                    sku: detail.sku.clone().or_else(|| detail.product_id.clone()),
+                    categories: detail.category.as_ref().map(|c| vec![c.clone()]),
+                    product_name: Some(detail.product_name.clone()),
+                    product_type: feature_data.product_type.clone(),
```
JeevaRamu0104, line 662: nit: `OrderDetailsWithAmount.product_type` is a typed `ProductType` on each item; this discards it and stamps one `connector_feature_data` string onto every cart line.

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
```diff
@@ -0,0 +1,1239 @@
...
+                    item_class: feature_data.item_class.clone(),
+                    quantity: detail.quantity,
+                    // Unknown shipping requirement is treated as physical:
+                    // claiming `digital` for a shipped good would overstate risk.
+                    item_fulfillment: if detail.requires_shipping == Some(false) {
+                        NsureItemFulfillment::Digital
+                    } else {
+                        NsureItemFulfillment::Physical
+                    },
+                    selling_price: NsureAmount {
```
JeevaRamu0104, line 652: `detail.amount` is documented as "the amount per quantity of product" and it's sent alongside `quantity`. Is nSure's `sellingPrice` per-unit or the line total? If it's the line total, the cart won't reconcile against `paidAmount` whenever quantity > 1.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: Really good catch, thanks !

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: Keeping it as-is deliberately. The alternative is mapping Failure to None, which drops all generic failures — and most payment failures do occur at authorization, so we'd lose more signal than we'd gain in precision. For both nsure marks it as "payment didn't complete"

crates/integrations/connector-integration/src/frm_connectors/nsure/transformers.rs
THE-CORE-BUG: Added a comment for now as unconfirmed, once the creds work fine, will update it, for now will keep it as it is

JeevaRamu0104 (approved):

deepanshu-iiitu (approved):

gopikrishna000 (approved):

PR #2220, opened 2026-09-02
feat(deja): record/replay instrumentation for UCS — ART

iemyashasvi:
## What

The complete déjà (ART) record/replay series for UCS .

Everything is gated behind the `deja` cargo feature and inert by default: `deja.mode` defaults to `disabled`, every boundary is a pure passthrough, and the config surface is excluded from per-request `x-config-override` (`#[patch(ignore)]`, deny-unknown-fields). Feature-off builds compile unchanged.

The series adds: typed `[deja]` config; shared entropy seams (time/id/crypto); gRPC ingress boundary (tower layer) + correlation tracing layer; connector HTTP/Kafka/injector egress boundaries; Kafka record sink with boot-time hook install; execution-graph span capture (`ucs::*` pipeline + `connector::*` macro spans + `ucs::connector_call` essentials — url/headers/body digests, status, integrity outcome); the local record/replay rig (`scripts/deja-local/`); and the Dockerfile bakes the feature + ships the gRPC descriptor set for the replay driver.

## Verification

- `cargo check -p grpc-server --features kafka,connector-request-kafka,otel,log-transformations,deja` (the image's feature set) — clean
- `cargo check -p grpc-server` (feature-off) — clean
- `cargo test -p external-services --features deja,injector-client` and `cargo test -p grpc-server --features deja --lib` — green
- The series has been running in the sandbox from `deja-integration` images: record leg (Kafka → Vector → S3) and replay leg (scored span lattice, 180/180 matched on the latest run) are both proven end-to-end.


crates/common/external-services/Cargo.toml
```diff
@@ -69,6 +69,7 @@ log_utils = { git = "https://github.com/juspay/framework-libs-rs", rev = "3cdcc2
 opentelemetry = { version = "0.27", optional = true, default-features = false, features = [
     "metrics",
 ] }
+deja = { git = "https://github.com/iemyashasvi/deja", rev = "55fb014ed57f4a1d2d827b8990e4235cc10da0b7", optional = true, default-features = false }
```
AmitsinghTanwar007, line 72: is it some fork repo?

AmitsinghTanwar007 (dismissed):

crates/common/external-services/src/deja_fields.rs
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+    }
+    let mut out = Vec::new();
+    canonicalize(value, &mut out);
+    out
+}
+
+/// `(origin, path, sorted query KEY names)` — the URL split that never
+/// carries a credential. Unparseable URLs record verbatim under origin (a
```
gopikrishna000, line 84: ci spell check

crates/common/common_utils/Cargo.toml
```diff
@@ -46,6 +46,7 @@ regex = "1.11.1"
 semver = { version = "1.0.26", features = ["serde"] }
 openssl = { version = "0.10", features = ["vendored"] }
 josekit = "0.8.7"
+deja = { git = "https://github.com/iemyashasvi/deja", rev = "55fb014ed57f4a1d2d827b8990e4235cc10da0b7", optional = true, default-features = false }
```
gopikrishna000, line 49: why forked?

crates/common/external-services/src/service.rs
```diff
@@ -1012,6 +1082,8 @@ where
                     let status_code = response.as_ref().ok().map(|result| match result {
                         Ok(body) | Err(body) => i32::from(body.status_code),
                     });
+                    #[cfg(feature = "deja")]
+                    connector_call.record_status(status_code);
```
gopikrishna000, line 1086: should it be record_http_status_code
as status may be confused with attempt status

crates/common/external-services/src/service.rs
```diff
@@ -1135,6 +1209,8 @@ where
...
```
gopikrishna000, line 1213: same

crates/common/ucs_env/src/deja_config.rs
```diff
@@ -0,0 +1,210 @@
...
+            record_key: "deja_record".to_string(),
+            fail_closed: true,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(default)]
+pub struct IdentityConfig {
+    pub pod_name_env: String,
```
gopikrishna000, line 171: how these are set

crates/grpc-server/grpc-server/src/main.rs
```diff
@@ -46,6 +51,21 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     }
 
+    let _guard = logger::setup(
```
gopikrishna000, line 54: **`logger::setup` moved below the `superposition.toml` load — those logs are now silently dropped**

The `superposition.toml` block above still calls `tracing::info!`/`tracing::warn!`, but the tracing subscriber isn't installed until this `logger::setup` line. `tracing` events emitted before `registry()...init()` hit the no-op `NoSubscriber` and are discarded — not buffered or replayed — so they never reach the storage layer, console, or Kafka/Loki.

Net effect: `warn!("Failed to load superposition.toml ... Connector URLs will use defaults from sandbox.toml")` (and the success `info!`) previously landed in Loki and now vanish. That warning is the signal that connectors fell back to default URLs instead of Superposition-resolved ones — losing it makes wrong-connector-URL incidents harder to diagnose.

Only the `deja::boot::install` call genuinely needs to precede `logger::setup` (the hook cell latches when the logger's deja layers peek it, and deja deliberately uses `eprintln!` in that window). The `superposition.toml` block has no such constraint — suggest moving just that block to after `logger::setup`.

config/development.toml
```diff
@@ -252,3 +252,16 @@ keys = ["x-request-id","x-merchant-id","x-lineage-ids","x-reference-id","x-conne
...
+[deja]
+mode = "disabled"
+
+[deja.recording.kafka]
+topic = "ucs-deja-recording-events"
+brokers = []          # empty => inherit the [events] brokers at boot
+
+[deja.sampler]
+record_key = "deja_record"
+fail_closed = true
```
Aishwariyaa-Anand: Only recording is configured here. how do we enable replay using deja?

crates/common/external-services/Cargo.toml
iemyashasvi: It was the development fork used while the déjà-side changes were in flight — all of that work is now merged into juspay/deja main, and the pins point at the canonical repo as of d6b5792ce (a pure fast-forward: the old fork rev is an ancestor of juspay/deja main).

crates/common/common_utils/Cargo.toml
iemyashasvi: Same as the thread on external-services: the fork was the in-flight development repo; everything is merged upstream now and all seven pins moved to juspay/deja main in d6b5792ce.

crates/common/external-services/src/deja_fields.rs
iemyashasvi: Fixed — the failure was a single word, `Unparseable` → `Unparsable` (8d1782740).

crates/common/external-services/src/service.rs
iemyashasvi: Agreed — renamed to `record_http_status_code` in 8d1782740. The span field itself stays `status_code`: it sits beside `url.*`/`headers.*`/`body.*` on the HTTP-call span and matches the golden log's `response.status_code` convention.

crates/common/external-services/src/service.rs
iemyashasvi: Renamed along with the other call site in 8d1782740.

crates/common/ucs_env/src/deja_config.rs
iemyashasvi: Same convention as hyperswitch's config files: `[deja.identity]` is now declared in development/sandbox/production.toml with the defaults visible (fd2b403eb) — `pod_name_env`/`git_sha_env` name env vars read at boot (defaults `POD_NAME`, the k8s Downward API convention, and `VERGEN_GIT_SHA`), while `instance_id`/`code_sha` accept literal overrides. Everything is env-overridable as `CS__DEJA__IDENTITY__*`. With nothing set, boot falls back to `pi-{pid}-{boot_nanos}` for the instance and the sha baked into the binary at compile time. Struct-level doc comments added in 8d1782740.

crates/grpc-server/grpc-server/src/main.rs
iemyashasvi: fixed in 8d1782740: the superposition block is back below `logger::setup`, with a comment stating the ordering constraint so it can't silently regress. Only the déjà hook install genuinely needs to precede the logger (the hook cell latches when the logger's layers first peek it, which is why that window deliberately uses `eprintln!`).

config/development.toml
iemyashasvi: Replay isn't a mode a serving deployment flips in this file — it's a sealed-pod mode the déjà orchestrator drives: the replay job boots this same image with `CS__DEJA__MODE=replay` plus a rendered lookup table, re-drives the recorded requests, and substitutes every side effect from the tape. Locally, `scripts/deja-local/replay-dev.sh` wires it end to end. The knobs are now declared inline: fd2b403eb adds the full inert `[deja]` surface — including `[deja.replay]` (`source`/`lookup_dir`/`observed_sink`) — to development, sandbox, and production TOMLs, mirroring hyperswitch's config convention, so the env overrides bind to visible keys.

config/development.toml
```diff
@@ -256,3 +256,44 @@ keys = ["x-request-id","x-merchant-id","x-lineage-ids","x-reference-id","x-conne
...
+lookup_dir = ""       # with no `source`: uses <lookup_dir>/<run_id>.jsonl
+observed_sink = ""    # optional: where replay records what the candidate DID
+
+[deja.sampler]
+record_key = "deja_record"
+fail_closed = true
+
+[deja.identity]
+pod_name_env = "POD_NAME"       # env var read at boot for the instance name
+git_sha_env = "VERGEN_GIT_SHA"  # env var read at boot for the code sha
```
gopikrishna000: git version is already there is run time get function being used in codebase

plus can you once confirm with @kanikac199 regarding how these infra informations are taken for ucs for euler ART as well

AmitsinghTanwar007:
hey @iemyashasvi after review you may need to remove this env fields (related to deja) from toml files and have them in example only.

config/development.toml
iemyashasvi: Right — and that baked git version is already the terminal default: `resolved_code_sha` falls back to the compile-time `VERGEN_GIT_SHA` (the same vergen source the existing `git_describe!()`/`version!()` getters use), so with nothing configured, recordings carry the binary's real sha. `git_sha_env` is only an optional infra override on top; these TOML keys declare defaults that previously applied invisibly via the struct's serde `Default` — documentation, not new behavior.

Confirmed with @kanikac199: the pod hostname already carries the infra identity (application, release tag, replicaset hash, pod suffix — e.g. `connector-service-grpc-20xxxxxhotfix1-5xxxxd-lr9xx6`), the same way euler ART sees it. Adopted in beae69852: instance resolution now falls back to the hostname (k8s sets it to the pod name) after the explicit override and `POD_NAME` env — so identity is self-derived with zero infra injection, used as one opaque string and deliberately never parsed into segments (the format is a naming convention, not an API). Code provenance stays on the baked sha, which is more precise than the release tag — and the tag rides along inside the hostname string anyway.

iemyashasvi: Went one step further on the single-source-of-truth direction: dropped the `[deja.identity]` declarations from all three config files in e9fb6b9a4 — the resolution chain lives only in code now (pod hostname for the instance, baked `VERGEN_GIT_SHA` for provenance, `CS__DEJA__IDENTITY__*` still binds through serde defaults). A pointer comment marks the section as deliberately absent so the next reader knows where the truth lives.

crates/common/common_enums/Cargo.toml
```diff
@@ -9,5 +9,11 @@ utoipa = { version = "4.2.0", features = ["preserve_order", "preserve_path_order
 strum = { version = "0.26", features = ["derive"] }
 thiserror = { workspace = true }
 
+[features]
+default = []
+# Déjà record/replay: adds serde derives on the error enums so a recorded connector
+# failure replays as the same typed error. Dependency-free (only gates derives).
+deja = []
+
```
maverox, line 17: check this please why do we need this

gopikrishna000 (dismissed):

crates/common/external-services/src/service.rs
```diff
@@ -824,6 +892,8 @@ where
                 Some(request) => {
                     let url = request.url.clone();
                     let method = request.method;
+                    #[cfg(feature = "deja")]
+                    connector_call.record_request(&request);
```
maverox: why do we need this way of span? why can't we rely on tracing ??

iemyashasvi: Removed in 889247de7 — you were right. The deja boundary attributes already record the full request/response on the tape; this span only carried derived projections of that data (digests, header names, parsed URL). Any detail/diff view can be computed from the boundary records offline. service.rs instrumentation now matches hyperswitch one-to-one: deja boundary attribute for capture, plain #[instrument] for structure.

crates/common/common_enums/Cargo.toml
iemyashasvi: It gates exactly two serde derives — KafkaClientError and ApiClientError (enums.rs:2489/2507) — the error types of the recorded egress boundaries (service.rs:275/301/1409). Deja records a boundary's full Result including the Err arm, so a recorded connector failure replays as the same typed error; that needs Serialize/Deserialize. Gated rather than unconditional because derives are public API surface, and `deja = []` is dependency-free — it only flips derives on.

crates/common/external-services/src/deja_codec.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    fn into_response(self) -> Response {
+        // `Response.headers` uses reqwest's (http 0.2) HeaderMap, not the http 1.x in scope.
+        let mut header_map = reqwest::header::HeaderMap::new();
+        for (name, value) in &self.headers {
+            if let (Ok(name), Ok(value)) = (
+                reqwest::header::HeaderName::from_bytes(name.as_bytes()),
+                reqwest::header::HeaderValue::from_str(value),
+            ) {
+                header_map.insert(name, value);
```
maverox: **High — multi-valued headers are silently collapsed here. This is the defect we observed live in a sandbox self-replay yesterday.**

`HeaderMap::insert` *removes every previously-associated value for that name*. Capture is fine — `HeaderMap::iter()` at L36 yields one pair per **value**, so all three `set-cookie` entries reach the tape — but this loop puts them back one at a time with `insert`, so only the last survives.

Combined with `headers.sort()` at L47 (which discards wire order for alphabetical), this reproduces the observed signature exactly: an outbound response captured with **15 entries in wire order including three `set-cookie`**, reconstructed as **13 entries in alphabetical order with one `set-cookie`**. The sort accounts for the ordering; this line accounts for the two lost values.

The existing test cannot catch it — the `response()` fixture builds a single `content-type` header, so the duplicate case is never exercised.

Worth noting the producer/consumer split is *inside this PR*: the HTTP ingress side gets it right. `deja/http_layer.rs:236` builds headers through `deja::http::headers(...)` — deja's multimap — and its test explicitly appends `x-multi: a` / `x-multi: b` and asserts both survive in order. Two halves of one change, each locally correct, disagreeing on the same contract.

Fix: `append` instead of `insert`; drop the `sort()` at L47 (or carry an explicit order index) so wire order survives; add a repeated-`set-cookie` fixture so the property is asserted rather than assumed.

Two smaller things in the same function:
- L59-62: `.to_str().ok()` drops non-UTF-8 header values with no counter and no name. The doc says they "never occur on the connector paths" — worth asserting rather than narrating.
- L70: `(!header_map.is_empty()).then_some(header_map)` turns a recorded `Some(empty)` into `None`, and turns a response whose headers all failed to parse into a headerless one, silently.

crates/common/external-services/src/deja_codec.rs
```diff
@@ -0,0 +1,379 @@
...
+
+#[cfg(test)]
+#[allow(clippy::expect_used, clippy::unwrap_used, clippy::panic)]
+mod tests {
+    use super::*;
+    use deja::codec::ReplayCodec;
+
+    fn response(status: u16, body: &[u8]) -> Response {
+        let mut headers = reqwest::header::HeaderMap::new();
+        headers.insert("content-type", "application/json".parse().unwrap());
```
maverox, line 346: **Note (supports the `into_response` finding):** this fixture is why the round-trip test passes despite the `insert` bug — one `content-type` header, so header multiplicity is never exercised. A fixture with two or three `set-cookie` values would fail today and would pin the property going forward.

crates/common/common_utils/src/fp_utils.rs
```diff
@@ -40,27 +40,51 @@ where
...
 }
 
 /// Generate a random (v4) UUID string. Connector code that needs a random
 /// idempotency key / client-request-id should call this helper rather than
-/// `uuid::Uuid::new_v4()` directly, so every entropy source in the codebase is
-/// a single, auditable function (enforced by the connector crate's clippy
-/// `disallowed-methods`).
+/// `uuid::Uuid::new_v4()` directly — every entropy source in the codebase is a
+/// single, auditable function, and under déjà the value is captured on tape
+/// and replayed deterministically.
```
maverox, line 81: **High — this removes the only trace of an entropy gate that never existed, rather than adding one.**

The replaced text claimed entropy discipline was *"enforced by the connector crate's clippy `disallowed-methods`"*. There is no `clippy.toml` anywhere in this repo — not on `main`, not on this branch — and `git grep disallowed` across the whole tree returns nothing but a binary artifact. That gate has never existed; this edit deletes the claim and the gap becomes invisible.

The sweep result today is genuinely clean, which is exactly why it's fragile: every production `Uuid::new_v4` / `Uuid::now_v7` / `nanoid!` site is inside a seamed helper here or in `lib.rs`, and `crates/integrations/connector-integration/src/**` has **zero** direct `OffsetDateTime::now*` / `SystemTime::now` / `Utc::now` / `Uuid::*` / `nanoid!` / `thread_rng` calls after the one fix in `paynearme/transformers.rs`. So ~100 connectors are correct by hand today, with nothing stopping the next connector PR from regressing it — which is precisely how the redis webhook lock tokens diverged on the hyperswitch side (seamed `id` matched 27/27 in the same run; the two unseamed UUIDs came back different).

Suggested: a workspace-root `clippy.toml` with `disallowed-methods` for `uuid::Uuid::new_v4`, `uuid::Uuid::now_v7`, `time::OffsetDateTime::now_utc`, `std::time::SystemTime::now`, `chrono::Utc::now`, `chrono::Local::now`, `rand::thread_rng`, plus `disallowed-macros` for `nanoid::nanoid`, and `-D clippy::disallowed_methods` in CI. Allow-list the seamed helpers and `deja/boot.rs` (which deliberately uses raw `SystemTime` to avoid recursing into the hook it is installing).

Residual unseamed production clock reads a gate would catch — all four are observability-event timestamps so they don't condition connector requests, but they're what the absence leaves behind: `external-services/src/service.rs:1335`, `grpc-server/src/utils.rs:572`, `interfaces/src/events/connector_api_logs.rs:64`, `interfaces/src/events/routing_api_logs.rs:92`.

crates/grpc-server/grpc-server/src/app.rs
```diff
@@ -315,10 +325,17 @@ impl Service {
...
             .layer(logging_layer)
             .layer(request_id_layer)
-            .layer(propagate_request_id_layer)
+            .layer(propagate_request_id_layer);
+        // gRPC ingress record/replay boundary. Spliced inside SetRequestId (the request
+        // always carries x-request-id) and outside config-override (a rejected request
+        // still yields one event). Inert until a boot hook is installed; feature-off this
+        // binding disappears entirely and the layer chain is unchanged.
+        #[cfg(feature = "deja")]
+        let server_builder = server_builder.layer(crate::deja::layer::DejaIngressLayer::new(None));
```
maverox: **High — the sampler is wired to nothing, so record mode captures 100% of traffic.**

Both ingress layers are constructed with `None` (here, and `http_layer` at L276), and both read `None` as record-everything (`layer.rs:118`, `http_layer.rs:117`: `None => true`). No type in the repo implements `RequestRecordingSampler`, and `SamplerConfig` is read by nothing except the config purity test.

Meanwhile `config/production.toml:226` ships `[deja.sampler] record_key = "deja_record"` / `fail_closed = true`, and `deja_config.rs:7` states *"Defaults are fail-closed throughout: … sampler fail-closed"*. An operator reading either one reasonably concludes recording is Superposition-gated and fail-closed. It is neither — `CS__DEJA__MODE=record` on a production pod captures every request.

This is what turns the `http_args` secrets finding from theoretical into operational: 100% of payment traffic, with unmasked connector API keys and plaintext card bodies, into Kafka and S3.

`deja/sampler.rs:5-7` says the concrete implementation *"land[s] with the boot/sink change"* — that change is in this PR, and it didn't, so the module doc is stale too.

Fix: wire the Superposition-backed sampler, or delete `[deja.sampler]` from all three config files along with the trait, so the absence is visible rather than advertised.

config/production.toml
```diff
@@ -200,3 +200,42 @@ connectors_with_webhook_source_verification_call = "paypal, truelayer"
...
+brokers = []          # empty => inherit the [events] brokers at boot
+
+[deja.replay]
+source = ""           # rendered LookupTable JSONL, or a raw-tape artifact dir
+lookup_dir = ""       # with no `source`: uses <lookup_dir>/<run_id>.jsonl
+observed_sink = ""    # optional: where replay records what the candidate DID
+
+[deja.sampler]
+record_key = "deja_record"
+fail_closed = true
```
maverox: **High — this block is inert.** Nothing reads `SamplerConfig` (see the comment on `app.rs:337`): both ingress layers are constructed with `None`, which means record-everything. `fail_closed = true` shipped in the production config is a promise nothing keeps. Either wire the sampler or remove this block — a config surface that silently no-ops is the same shape as a skip rendered in a UI.

crates/grpc-server/grpc-server/src/deja/layer.rs
```diff
@@ -0,0 +1,317 @@
...
+            } else {
+                false
+            };
+
+            // 2. Buffer the unary request body (needed to decode args and to rebuild the
+            //    body `inner` reads). Capture failures never fail the request.
+            let request_bytes = body
+                .collect()
+                .await
+                .map_err(|error| tonic::Status::internal(error.to_string()))?
```
maverox: **Medium — instrumentation can fail a request it has already decided not to record.**

The comment on L130 says "Capture failures never fail the request"; the next statement returns `tonic::Status::internal` out of the middleware.

The sharper problem is that this block isn't gated on `should_record`. Once a hook is installed and the path starts with `/types.`, the body is buffered unconditionally — so this fires for a request the sampler declined, and for **every** request in replay mode (where `should_record` is false by construction at L120). Instrumentation that has decided to record nothing still buffers the whole body and can still fail the call.

Precondition: any client stream reset, mid-body disconnect, or h2 body error on any `/types.*` rpc while deja is active. Without deja that body error reaches tonic's codec and gets tonic's own status mapping; with deja every such error becomes `Status::internal`.

The correct shape is already in the sibling file, ~100 lines away: `http_layer.rs:132-138` early-returns a plain passthrough with the original body when `!should_record` (no buffering at all), and `http_layer.rs:146-150` maps a buffering failure to a `body_capture_failed` flag that suppresses the event instead of `?`. Suggest mirroring both here.

crates/grpc-server/grpc-server/src/deja/boot.rs
```diff
@@ -0,0 +1,469 @@
...
+        })
+        .unwrap_or_else(|| format!("pi-{}-{}", std::process::id(), now_ns()))
+}
+
+fn resolved_code_sha(config: &DejaConfig) -> Option<String> {
+    configured_value(config.identity.code_sha.as_deref())
+        .or_else(|| env_value_named(&config.identity.git_sha_env))
+        .or_else(|| option_env!("VERGEN_GIT_SHA").map(str::to_owned))
+    // No "unknown" placeholder: absence is a fact worth being able to observe.
+}
```
maverox, line 172: **Medium — a runtime env var silently shadows the commit the binary was built from.**

`git_sha_env` defaults to `"VERGEN_GIT_SHA"` (`deja_config.rs:196`), so the middle arm is a **runtime** `std::env::var` read of a variable with the *same name* as the compile-time constant in the last arm. A deployment that exports `VERGEN_GIT_SHA` — to a release tag, an image tag, anything — shadows the commit the binary actually knows, and the two are never compared. Sharing the name is what makes it invisible: a mismatch looks like the same source.

The process then carries two identities that can disagree with no reconciliation: `ucs_env/src/logger/env.rs:21` uses `env!("VERGEN_GIT_SHA")` (compile-time) for the service's own build info, while `record_sink.rs` stamps `Code { sha: resolved_code_sha }` on every envelope. Logs say one commit; every tape says another.

Suggested: when both resolve and differ, stamp both (`code_sha` and `built_sha`) on the envelope, or at minimum warn loudly at boot. Absence is already treated as a fact worth observing (L171) — disagreement deserves the same.

Unrelated doc drift nearby: `deja_config.rs:19` says an absent run_id yields a `run-{now_ns}` id, but `fallback_run_id` mints `rec-<sha>-<MMDDhhmm>-<inst>` and only falls back to `run-{now_ns}` when no sha resolves at all.

crates/grpc-server/grpc-server/src/deja/record_sink.rs
```diff
@@ -0,0 +1,352 @@
...
+    fn write_batch(&mut self, records: &[deja::DejaRecord]) -> io::Result<()> {
+        for record in records {
+            match record {
+                deja::DejaRecord::BoundaryEvent(event) => self.write_boundary_event(event)?,
+                deja::DejaRecord::GraphNode(node) => self.write_graph_node(node)?,
+                // Record mode never produces observations; skip instead of failing the
+                // writer if the library contract is breached.
+                deja::DejaRecord::Observed(_) => {
+                    debug_assert!(false, "observed record reached the record-mode Kafka sink");
+                }
```
maverox: **Medium — this drop is silent in the profile that ships, and a panic in the one that doesn't.**

`debug_assert!` is compiled out in release, so an `Observed` record is discarded with no counter, no log, no marker. In debug it panics on a writer thread, which is the opposite failure and breaks "recording never fails the service". Neither profile does the right thing.

Every input should become an output or a **named** drop. The marker vocabulary already exists in this file (`MarkerBody`, `MARKER_ARTIFACT_TYPE`) — suggest incrementing a named drop counter and emitting a `dropped` marker, in both profiles.

Minor, same function: `write_batch` `?`-returns on the first failed record, so the rest of the batch is never attempted and the writer books the whole batch as dropped — including records that already enqueued successfully. Over-counting is the safe direction, but the loss ledger and the topic then disagree, which weakens the accounting this is here to provide.

crates/grpc-server/grpc-server/src/app.rs
maverox: this we need before sandbox recording right or at least we need a mechanism of trickling down the decision to record from hyperswitch to UCS

crates/grpc-server/grpc-server/src/deja/boot.rs
maverox: This I think we can change later together for both hyperswitch and UCS.

crates/common/external-services/src/deja_codec.rs
maverox: Following up on the multi-valued header thread above with the concrete fix, so this is one click rather than a re-derivation.

The minimal correction is `insert` → `append`. `HeaderMap::insert` removes every value already associated with that name; `append` adds alongside them, which is the whole reason the two methods exist:

```suggestion
                header_map.append(name, value);
```

That alone stops the loss — all three `set-cookie` values come back. It does not restore *wire order*, because `headers.sort()` at L47 has already reordered them alphabetically by the time they reach the tape. Worth deciding separately: dropping that sort preserves the order the connector actually sent, but `HeaderMap::iter()` walks internal table order rather than a guaranteed wire order, so if the sort is load-bearing for tape determinism it may be better to keep it and record an explicit index per entry. Your call — the loss is the bug, the ordering is a fidelity improvement.

The thing that makes it stick is a fixture that would fail today. The current `response()` helper builds a single `content-type` header, so header multiplicity is never exercised and the round-trip test passes regardless. Something like:

```rust
#[test]
fn repeated_headers_survive_the_round_trip() {
    let mut headers = reqwest::header::HeaderMap::new();
    headers.append("set-cookie", "a=1".parse().unwrap());
    headers.append("set-cookie", "b=2".parse().unwrap());
    headers.append("set-cookie", "c=3".parse().unwrap());
    let original = Response {
        headers: Some(headers),
        response: bytes::Bytes::from_static(b"{}"),
        status_code: 200,
    };

    let (json, _) = HttpOutcomeCodec::capture(&Ok(Ok(original)));
    let back = HttpOutcomeCodec::reconstruct(json)
        .expect("reconstructs")
        .expect("outer ok")
        .expect("inner ok");

    let cookies: Vec<&str> = back
        .headers
        .as_ref()
        .expect("headers present")
        .get_all("set-cookie")
        .iter()
        .map(|v| v.to_str().unwrap())
        .collect();
    assert_eq!(cookies.len(), 3, "all three set-cookie values must survive");
}
```

On `insert` today that fails with `cookies.len() == 1`, which is exactly the failure we want a test to be able to state.

Broader context worth having: this is not a slip specific to this PR. The identical collapse showed up in hyperswitch on a live same-image replay — a response recorded with 15 headers in wire order including three `set-cookie` came back as 13 alphabetical with one. Two codebases, two authors, same shape. The common factor is that déjà ships a correct multimap helper for the **capture** direction (`deja::http::headers`, which this PR uses correctly in `deja/http_layer.rs:236`) and nothing for the **reconstruct** direction, because reconstruct has to build a concrete `HeaderMap` and déjà deliberately names no host types. `ReplayCodec` hands you `capture` and `reconstruct` as a pair and states no round-trip invariant on them. So the missing guard rail is ours on the déjà side, and we're picking that up — a codec conformance property (`reconstruct(capture(x))` must re-capture identically) that hosts can run against their own codecs. The `append` fix here is still yours, but you should not have had to find this one unaided.

iemyashasvi: fixed in [8a6bf24](https://github.com/juspay/hyperswitch-prism/pull/2220/commits/8a6bf2420a75d12d46f4660a38c19a7eaee384ec)

crates/common/external-services/src/deja_codec.rs
iemyashasvi: fixed in [8a6bf24](https://github.com/juspay/hyperswitch-prism/pull/2220/commits/8a6bf2420a75d12d46f4660a38c19a7eaee384ec)

crates/common/external-services/src/deja_codec.rs
iemyashasvi: fixed in  [8a6bf24](https://github.com/juspay/hyperswitch-prism/pull/2220/commits/8a6bf2420a75d12d46f4660a38c19a7eaee384ec)

crates/common/common_utils/src/fp_utils.rs
iemyashasvi: this is fine , we will introduce a clippy.toml in a separate pr itself

crates/grpc-server/grpc-server/src/app.rs
iemyashasvi: Wired in 7963be213 + follow-ups. The ingress layers install a Superposition-backed sampler exactly when the process is in record mode: deja_record (default FALSE) evaluated in-process against the baked superposition.toml, dimensioned on environment x rpc_method x rpc_service, fail-closed on every failure path, logged no-source state instead of silent record-everything. Production is dark by default; development and sandbox keep recording via explicit overrides. 220b04745 adds per-class targeting (rpc_service derived structurally from the path — no maintained path map), 7b28f759b adds deterministic percentage sampling (deja_record_percent, request-id FNV bucket) — hyperswitch's live posture ("8% of the payments cohort", per the staging dashboard's deja-sampler-payments experiment) is one toml override here. The trickle-down channel was prototyped and deliberately dropped as one revertable commit (531505ad8) since nothing stamps x-deja-record upstream yet.

Verified end to end on the local rig (scripts/deja-local/sampler-e2e.sh, 13/13): dev wholesale records, health excluded, production dark, targeted rpc + class targeting discriminate correctly, percent-50 splits deterministically, no-source fail-closed stays dark with the install error logged (fail-open records). Plus a 20-request batch at percent 50: the tape contained exactly the 13 ids an independent offline FNV mirror predicted, id-for-id; a 10k-id unit test pins the rate at 50±3%. The stale module doc is gone with the implementation it described.

crates/grpc-server/grpc-server/src/deja/layer.rs
iemyashasvi: Fixed in 966c0d401, mirroring the http_layer shape as suggested: !should_record now early-returns a pure passthrough with the ORIGINAL streaming body — no buffering at all — covering sampled-out requests and all of replay mode, inside the kept deja::grpc_incoming span (kept because the correlation layer reads its request_id). On the record path a collect failure degrades to body_capture_failed: the event is suppressed and tonic's codec produces its own native error — this middleware never mints Status::internal anymore, so the comment finally tells the truth. Verified: 13/13 sampler E2E through the new paths (every dark case exercises the early return), and an A/B tape replay (same tape, old vs new layer) produced identical verdicts with body diffs 0.

crates/grpc-server/grpc-server/src/deja/record_sink.rs
iemyashasvi: Fixed in 10c7cfbc4. The Observed contract breach is now a NAMED drop in every profile: warn-logged, counted (contract_breach_drops on the sink), and stamped onto the tape as a MarkerKind::Dropped marker — best-effort, so a marker failure logs rather than compounding the batch error. No silent release skip, no writer-thread panic.

The batch point too: write_batch now attempts every record and returns the first error after the loop, so the topic carries everything that could enqueue. The writer still books the whole batch as dropped on Err — the binary Result is the library contract — so the ledger keeps over-counting (the safe direction), now with the gap minimized and documented at the site.

config/superposition.toml
```diff
@@ -1000,3 +1018,13 @@ connector_base_url = "https://api-stg.directa24.com"
...
+
+# ============================================================================
+# Sandbox Environment Overrides
+# ============================================================================
+# Deja: the sandbox custom pod keeps recording exactly as before the sampler
+# landed. Production inherits the dark default; sample-in there is always a
+# deliberate, targeted act (environment x rpc_method override).
+[[overrides]]
+_context_ = { environment = "sandbox" }
+deja_record = true
```
maverox: remove this

crates/common/common_utils/src/superposition_config.rs
```diff
@@ -81,16 +81,24 @@ impl SuperpositionConfig {
...
+    /// use this directly.
+    pub fn resolve_with(
+        &self,
+        dimensions: &[(&str, &str)],
     ) -> Result<HashMap<String, Value>, SuperpositionConfigError> {
         let mut dims: Map<String, Value> = Map::new();
-        dims.insert(
-            DIMENSION_CONNECTOR.to_string(),
-            Value::String(connector.to_string()),
-        );
```
maverox, line 89: unnecessary refactor

maverox (dismissed):

config/superposition.toml
iemyashasvi: Removed in cd927160d — sandbox inherits the dark default; recording is opt-in everywhere deployed, only the local-rig development override remains. Heads-up: the sandbox pod stops recording on its next deploy until a targeted override (rpc, class, or percent) lands.

crates/common/common_utils/src/superposition_config.rs
iemyashasvi: resolve_with is the sampler's only evaluation path (sampler.rs:150) — resolve()'s hard-coded connector x environment can't express the sampler's environment x rpc_method x rpc_service, so the function is required capability, not refactoring. The resolve() delegation is behavior-identical for every existing caller: it builds the same two-entry dims map and runs the same eval — same inputs, same outputs. The alternative was duplicating the evaluation tail inside the same impl.

maverox (approved):

gopikrishna000 (approved):

deepanshu-iiitu (approved):

PR #2221, opened 2026-09-02
feat(connector): [PayNearMe] add connector with Authorize, PSync, Void, Refund, RSync for cards

shuklatushar226:
## Summary

Brand-new **PayNearMe** connector (PayNearMe API v3.0, JSON API). Card payment method, one-time payments only.

- **Auth**: `ConnectorAuthType::BodyKey` — `api_key` = API Secret Key (HMAC-SHA256 signing key, **never transmitted**), `key1` = Site Identifier (sent as the `site_identifier` body field). Every request is signed with HMAC-SHA256 over the alphabetically-sorted key+value concatenation of the body, exempting `format` / `signature` / `call`.
- **Base URLs**: sandbox `https://api.paynearme-sandbox.com/json-api`, production `https://api.paynearme.com/json-api`.
- **Amounts**: `StringMajorUnit` / `CurrencyUnit::Base`.

## Flow mapping

| Flow | Endpoint | Notes |
|---|---|---|
| CreateOrder | `POST /create_order` | `should_do_order_create() -> true`; stores `pnm_order_identifier` in `connector_order_id` |
| Authorize | `POST /create_payment_method` | `send_payment=true` — tokenise + charge in one call |
| PSync | `POST /find_payment` | |
| Void | `POST /cancel_payment` | |
| Refund | `POST /refund_payment` | omits `refund_amount` / `refund_currency` on full refunds |
| RSync | `POST /find_payment` | reads nested `refund`; `connector_refund_id` = `pnm_payment_identifier`, since PayNearMe mints no refund id |
| Capture | — | **`NotImplemented`** (see below) |

### Capture is deliberately NotImplemented

PayNearMe has **no capture endpoint**; card payments are sale / auto-capture. Capture is intentionally **not** routed to `/make_payment`, which would create a second charge (double-charge). `CaptureMethod::Manual*` is rejected with `NotSupported`.

### No 3DS

The PayNearMe API has no 3DS / 3-D Secure support at all — an exhaustive search of the embedded OpenAPI 3.0.1 document and every guide page found zero references to 3DS / CAVV / ECI / ACS / liability shift. `AuthenticationType::ThreeDs` is therefore rejected with `NotSupported` rather than being silently downgraded to a non-3DS charge.

### Currency

USD only; any other currency is rejected with `NotSupported`.

## Documented API inconsistencies handled

- Identifiers and amounts arrive as **bare JSON numbers** in some responses and **quoted strings** in others → string-or-number serde helper.
- `/create_order` returns `"orders"` while `/create_payment_method` returns `"order"` → serde alias.
- `payments[]` appears under both `order.payments` and `order.electronic_payments.payments`.
- `/cancel_payment` returns `"cancelled"` while the enum declares `"canceled"` → serde alias.

## Known limitations

- **No 3DS support** — API limitation, not an implementation gap.
- **No manual capture / separate capture step** — API limitation.
- **Concurrent partial refunds cannot be disambiguated**, because PayNearMe returns no per-refund identifier.

## Files

**New**
- `crates/integrations/connector-integration/src/connectors/paynearme.rs`
- `crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs`
- `crates/internal/integration-tests/src/connector_specs/paynearme/specs.json`

**Modified**
- `config/development.toml`, `config/sandbox.toml`, `config/production.toml` (base URLs)
- `crates/internal/integration-tests/src/connector_specs/alpha_connectors.json`
- `crates/types-traits/grpc-api-types/proto/payment.proto` (`PAYNEARME = 141`, `PaynearmeConfig`, oneof field `152`)

## Testing

**No real PayNearMe sandbox credentials exist**, so verification was performed against a **local, spec-shaped mock PayNearMe server** that:

1. replays the technical spec's documented response bodies — including their documented inconsistencies (bare-number identifiers, singular `order` key, doubled-l `"cancelled"`),
2. logs every inbound HTTP request verbatim (headers + body), and
3. **independently recomputes the HMAC-SHA256 signature** from the received body and compares it against the one UCS sent, proving the outbound signing is correct.

UCS was pointed at the mock via a per-request `PaynearmeConfig.base_url` override. **The committed toml base URLs remain the real sandbox/prod hosts** — the mock URL appears nowhere in this diff.

### Build verification

- `cargo build --package connector-integration` — pass
- `cargo build --bin grpc-server` — pass
- `cargo check --workspace --all-targets` — pass
- `cargo clippy -p connector-integration -p domain_types` — pass
- `cargo fmt` — applied

<details>
<summary>Full grpcurl run against the mock server (6 flows + 4 negative assertions) — credentials redacted</summary>

```
Common config header used on every call:

CFG='{"config":{"Paynearme":{"api_key":"<REDACTED>","key1":"MOCK_PNM_SITE_IDENTIFIER","base_url":"http://127.0.0.1:9099/json-api"}}}'
-H "x-connector-config: $CFG"
-H 'x-merchant-id: merchant_paynearme'
-H 'x-tenant-id: default'
-H 'x-connector-request-reference-id: pnm_ref_001'

--- 1. CreateOrder -> POST /create_order ---

grpcurl -plaintext \
  -H "x-connector-config: $CFG" \
  -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' \
  -H 'x-connector-request-reference-id: pnm_ref_001' \
  -H 'x-request-id: req_pnm_create_order_001' \
  -d '{"merchant_order_id":"pnm_order_ref_001","amount":{"minor_amount":50499,"currency":"USD"},"webhook_url":"https://example.com/webhook"}' \
  localhost:8000 types.PaymentService/CreateOrder

Response:
{
  "connectorOrderId": "85237034088",
  "status": "PENDING",
  "statusCode": 200,
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":null,\"orders\":{\"pnm_order_identifier\":\"85237034088\",\"order_status\":\"open\",\"payments\":null,\"electronic_payments\":{\"payments\":null}},\"errors\":[]}"
  }
}

Mock server log - exact outbound HTTP request UCS sent:
>>> INBOUND REQUEST  POST /json-api/create_order HTTP/1.1
--- headers ---
via: HyperSwitch
accept: application/json
accept-encoding: gzip
host: 127.0.0.1:9099
content-length: 382
--- body ---
{
  "order_amount": "504.99",
  "order_currency": "USD",
  "order_is_standing": "false",
  "order_type": "exact",
  "return_minimal_info": "true",
  "signature": "cda0a2b435d5c57539d7626ef65e1d53e69014cccfc168ee88832ff75d026310",
  "site_customer_identifier": "pnm_order_ref_001",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "site_order_identifier": "pnm_order_ref_001",
  "timestamp": "1788369093",
  "version": "3.0"
}
--- signature verification ---
string_to_sign : order_amount504.99order_currencyUSDorder_is_standingfalseorder_typeexactreturn_minimal_infotruesite_customer_identifierpnm_order_ref_001site_identifierMOCK_PNM_SITE_IDENTIFIERsite_order_identifierpnm_order_ref_001timestamp1788369093version3.0
expected       : cda0a2b435d5c57539d7626ef65e1d53e69014cccfc168ee88832ff75d026310
received       : cda0a2b435d5c57539d7626ef65e1d53e69014cccfc168ee88832ff75d026310
MATCH          : True
--- response ---
HTTP 201
{"status":"ok","orders":{"site_name":"Mock PayNearMe Site","site_identifier":"MOCK_PNM_SITE_IDENTIFIER","pnm_order_crid":"cyu9Qg","pnm_order_identifier":85237034088,"pnm_order_short_identifier":"LSN19S","site_order_key":567060000,"order_created":"2026-09-02 21:36:15 -0700","order_status":"open","order_amount":"504.99","order_currency":"USD","order_type":"exact","order_is_standing":false,"electronic_payments":{"customer":{"pnm_customer_identifier":"U5506959413","site_customer_identifier":"pnm_order_ref_001"}}}}

Note: pnm_order_identifier came back as a bare JSON number and was correctly parsed to "85237034088".

--- 2. Authorize (Card) -> POST /create_payment_method ---

grpcurl -plaintext \
  -H "x-connector-config: $CFG" \
  -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' \
  -H 'x-connector-request-reference-id: pnm_ref_001' \
  -H 'x-request-id: req_pnm_authorize_001' \
  -d '{"merchant_transaction_id":"pnm_txn_001","amount":{"minor_amount":50499,"currency":"USD"},"connector_order_id":"85237034088","payment_method":{"card":{"card_number":{"value":"<REDACTED>"},"card_exp_month":{"value":"01"},"card_exp_year":{"value":"2027"},"card_cvc":{"value":"<REDACTED>"},"card_holder_name":{"value":"John Smith"}}},"capture_method":"AUTOMATIC","auth_type":"NO_THREE_DS","enrolled_for_3ds":false,"customer":{"id":"cust_pnm_001","email":{"value":"john.smith@example.com"}},"address":{"billing_address":{"first_name":{"value":"John"},"last_name":{"value":"Smith"},"line1":{"value":"123 Fake Street"},"city":{"value":"Plano"},"state":{"value":"TX"},"zip_code":{"value":"75013"},"country_alpha2_code":"US","phone_number":{"value":"<REDACTED>"},"phone_country_code":"+1","email":{"value":"john.smith@example.com"}}},"return_url":"https://example.com/return","webhook_url":"https://example.com/webhook"}' \
  localhost:8000 types.PaymentService/Authorize

Response:
{
  "merchantTransactionId": "85237034088",
  "connectorTransactionId": "798183461108",
  "status": "CHARGED",
  "statusCode": 201,
  "connectorFeatureData": {
    "value": "{\"payment_method_identifier\":\"a95ac4a03ef38\",\"pnm_order_identifier\":\"85237034088\"}"
  },
  "connectorReferenceId": "85237034088",
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":\"0\",\"orders\":{\"pnm_order_identifier\":\"85237034088\",\"order_status\":\"open\",\"payments\":[{\"payment_status\":\"approved\",\"pnm_payment_identifier\":\"798183461108\",\"payment_method_identifier\":\"a95ac4a03ef38\",\"payment_amount\":\"100\",\"net_payment_amount\":\"100\",\"payment_type\":\"debit\",\"site_payment_identifier\":\"pnm_txn_001\",\"refund\":null}],\"electronic_payments\":{\"payments\":null}},\"errors\":[]}"
  }
}

Mock server log - exact outbound HTTP request UCS sent:
>>> INBOUND REQUEST  POST /json-api/create_payment_method HTTP/1.1
--- headers ---
accept: application/json
via: HyperSwitch
accept-encoding: gzip
host: 127.0.0.1:9099
content-length: 701
--- body ---
{
  "last_pmt_only": "true",
  "payment_amount": "504.99",
  "payment_currency": "USD",
  "payment_method_billing_address": "123 Fake Street",
  "payment_method_billing_name": "John Smith",
  "payment_method_billing_phone": "<REDACTED>",
  "payment_method_billing_zipcode": "75013",
  "payment_method_card_expiry_pii": "01/2027",
  "payment_method_card_number_pii": "<REDACTED>",
  "payment_method_cvv_pii": "<REDACTED>",
  "payment_method_type": "card",
  "pnm_order_identifier": "85237034088",
  "send_payment": "true",
  "signature": "6ff096e3e72ffac72a4b5f101d73a6ba0e8c6a18a8c0f9385d5fa9091683531b",
  "site_channel": "consumer",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "site_payment_identifier": "pnm_txn_001",
  "timestamp": "1788369094",
  "version": "3.0"
}
--- signature verification ---
expected       : 6ff096e3e72ffac72a4b5f101d73a6ba0e8c6a18a8c0f9385d5fa9091683531b
received       : 6ff096e3e72ffac72a4b5f101d73a6ba0e8c6a18a8c0f9385d5fa9091683531b
MATCH          : True
--- response ---
HTTP 201
{"response_code":"0","status":"ok","order":{"site_name":"Mock PayNearMe Site","site_order_identifier":"470070000","type":"order","site_identifier":"MOCK_PNM_SITE_IDENTIFIER","pnm_order_identifier":"85237034088","order_status":"open","order_amount":"504.99","order_currency":"USD","electronic_payments":{"payment_methods":[{"type":"debit","fee_amount":"4.99","fee_currency":"USD","accounts":[{"payment_method_identifier":"a95ac4a03ef38","status":"active","name":"John Smith","description":"Debit Card","account_type":"Debit","number":"6651","fee_amount":"4.99","fee_currency":"USD","expiration_date":"01/2027","card_brand":"PULSE"}]}]},"customer":{"pnm_customer_identifier":"U4509315551","site_customer_identifier":"470070000"},"payments":[{"payment_made":"2026-09-02 14:44:35 -0800","payment_amount":100,"payment_currency":"USD","payment_status":"approved","payment_type":"debit","payment_account":"Debit Card 6651","payment_method_identifier":"a95ac4a03ef38","net_payment_amount":100,"net_payment_currency":"USD","payment_processing_fee":0,"settled_to_site":false,"pnm_payment_identifier":798183461108,"site_payment_identifier":"pnm_txn_001","pricing_schedule_name":"consumer","site_channel":"consumer"}]}}

Exercises three documented traps at once: singular "order" key, payments[] nested under order.*, and a bare-number pnm_payment_identifier.

--- 3. PSync -> POST /find_payment ---

grpcurl -plaintext -H "x-connector-config: $CFG" -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' -H 'x-connector-request-reference-id: pnm_ref_001' -H 'x-request-id: req_pnm_psync_001' \
  -d '{"connector_transaction_id":"798183461108","amount":{"minor_amount":50499,"currency":"USD"},"capture_method":"AUTOMATIC"}' \
  localhost:8000 types.PaymentService/Get

Response:
{
  "connectorTransactionId": "798183461108",
  "status": "CHARGED",
  "statusCode": 201,
  "amount": {"minorAmount": "50499", "currency": "USD"},
  "merchantTransactionId": "0123456789",
  "connectorFeatureData": {"value": "{\"payment_method_identifier\":\"a95ac4a03ef38\",\"pnm_order_identifier\":null}"},
  "connectorReferenceId": "0123456789",
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":null,\"payment\":{\"payment_status\":\"approved\",\"pnm_payment_identifier\":\"798183461108\",\"payment_method_identifier\":\"a95ac4a03ef38\",\"payment_amount\":\"504.99\",\"net_payment_amount\":\"500.00\",\"payment_type\":\"debit\",\"site_payment_identifier\":\"0123456789\",\"refund\":null},\"errors\":[]}"
  }
}

Mock server log:
>>> INBOUND REQUEST  POST /json-api/find_payment HTTP/1.1
--- body ---
{
  "pnm_payment_identifier": "798183461108",
  "signature": "073e0094ff040bff6b3cd98faf2c5bad340200e7aa16fc24cd1e0223dc74c160",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "timestamp": "1788369094",
  "version": "3.0"
}
--- signature verification ---
string_to_sign : pnm_payment_identifier798183461108site_identifierMOCK_PNM_SITE_IDENTIFIERtimestamp1788369094version3.0
expected       : 073e0094ff040bff6b3cd98faf2c5bad340200e7aa16fc24cd1e0223dc74c160
received       : 073e0094ff040bff6b3cd98faf2c5bad340200e7aa16fc24cd1e0223dc74c160
MATCH          : True
--- response ---
HTTP 201

--- 4. Void -> POST /cancel_payment ---

grpcurl -plaintext -H "x-connector-config: $CFG" -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' -H 'x-connector-request-reference-id: pnm_ref_001' -H 'x-request-id: req_pnm_void_001' \
  -d '{"connector_transaction_id":"798183461108","cancellation_reason":"requested_by_customer","amount":{"minor_amount":50499,"currency":"USD"}}' \
  localhost:8000 types.PaymentService/Void

Response:
{
  "connectorTransactionId": "798183461108",
  "status": "VOIDED",
  "statusCode": 201,
  "connectorFeatureData": {"value": "{\"payment_method_identifier\":\"405eb56f20202\",\"pnm_order_identifier\":null}"},
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":null,\"payment\":{\"payment_status\":\"canceled\",\"pnm_payment_identifier\":\"798183461108\",\"payment_method_identifier\":\"405eb56f20202\",\"payment_amount\":\"560.00\",\"net_payment_amount\":\"500.00\",\"payment_type\":\"debit\",\"site_payment_identifier\":null,\"refund\":null},\"errors\":[]}"
  }
}

Mock server log (mock deliberately returns the doubled-l "cancelled" spelling; the connector normalises it to canceled -> VOIDED):
>>> INBOUND REQUEST  POST /json-api/cancel_payment HTTP/1.1
--- body ---
{
  "pnm_payment_identifier": "798183461108",
  "signature": "073e0094ff040bff6b3cd98faf2c5bad340200e7aa16fc24cd1e0223dc74c160",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "timestamp": "1788369094",
  "version": "3.0"
}
--- signature verification ---
MATCH          : True
--- response ---
HTTP 201
{"status":"ok","payment":{"payment_made":"2026-09-02 14:02:05 -0700","payment_amount":"560.00","payment_currency":"USD","payment_status":"cancelled","payment_type":"debit","payment_method_identifier":"405eb56f20202","pnm_payment_identifier":"798183461108","net_payment_amount":"500.00","settled_to_site":"false"}}

--- 5. Refund (partial) -> POST /refund_payment ---

grpcurl -plaintext -H "x-connector-config: $CFG" -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' -H 'x-connector-request-reference-id: pnm_ref_001' -H 'x-request-id: req_pnm_refund_001' \
  -d '{"merchant_refund_id":"pnm_refund_001","connector_transaction_id":"798183461108","payment_amount":50499,"refund_amount":{"minor_amount":12500,"currency":"USD"},"reason":"partial refund"}' \
  localhost:8000 types.PaymentService/Refund

Response:
{
  "merchantRefundId": "pnm_refund_001",
  "connectorRefundId": "798183461108",
  "status": "REFUND_PENDING",
  "statusCode": 201,
  "connectorTransactionId": "798183461108",
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":null,\"payment\":{\"payment_status\":\"refunded\",\"pnm_payment_identifier\":\"798183461108\",\"payment_method_identifier\":\"198fd2a399aae\",\"payment_amount\":\"210.99\",\"net_payment_amount\":\"200.00\",\"payment_type\":\"credit\",\"site_payment_identifier\":null,\"refund\":{\"refund_status\":\"started\",\"refund_amount\":\"125.00\",\"refund_currency\":\"USD\"}},\"errors\":[]}"
  }
}

Mock server log (partial refund -> refund_amount/refund_currency present, as the spec requires; a full refund omits both):
>>> INBOUND REQUEST  POST /json-api/refund_payment HTTP/1.1
--- body ---
{
  "pnm_payment_identifier": "798183461108",
  "refund_amount": "125.00",
  "refund_currency": "USD",
  "signature": "91da0dcfd7104a0a6cd89923ad7e9880bf8cebada397595cf2dc8a5584c2eee8",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "timestamp": "1788369094",
  "version": "3.0"
}
--- signature verification ---
string_to_sign : pnm_payment_identifier798183461108refund_amount125.00refund_currencyUSDsite_identifierMOCK_PNM_SITE_IDENTIFIERtimestamp1788369094version3.0
expected       : 91da0dcfd7104a0a6cd89923ad7e9880bf8cebada397595cf2dc8a5584c2eee8
received       : 91da0dcfd7104a0a6cd89923ad7e9880bf8cebada397595cf2dc8a5584c2eee8
MATCH          : True
--- response ---
HTTP 201

connectorRefundId correctly falls back to pnm_payment_identifier (PayNearMe mints no refund id), and refund_status "started" maps to REFUND_PENDING.

--- 6. RSync -> POST /find_payment ---

grpcurl -plaintext -H "x-connector-config: $CFG" -H 'x-merchant-id: merchant_paynearme' \
  -H 'x-tenant-id: default' -H 'x-connector-request-reference-id: pnm_ref_001' -H 'x-request-id: req_pnm_rsync_001' \
  -d '{"merchant_refund_id":"pnm_refund_001","connector_transaction_id":"798183461108","refund_id":"pnm_refund_001","connector_refund_id":"798183461108","refund_amount":{"minor_amount":12500,"currency":"USD"}}' \
  localhost:8000 types.RefundService/Get

Response:
{
  "merchantRefundId": "pnm_refund_001",
  "connectorRefundId": "798183461108",
  "status": "REFUND_PENDING",
  "statusCode": 201,
  "connectorTransactionId": "798183461108",
  "typedConnectorResponse": {
    "value": "{\"status\":\"ok\",\"response_code\":null,\"payment\":{\"payment_status\":\"approved\",\"pnm_payment_identifier\":\"798183461108\",\"payment_method_identifier\":\"a95ac4a03ef38\",\"payment_amount\":\"504.99\",\"net_payment_amount\":\"500.00\",\"payment_type\":\"debit\",\"site_payment_identifier\":\"0123456789\",\"refund\":null},\"errors\":[]}"
  }
}

Mock server log (keyed on pnm_payment_identifier, not a refund id - exactly as the spec mandates):
>>> INBOUND REQUEST  POST /json-api/find_payment HTTP/1.1
--- body ---
{
  "pnm_payment_identifier": "798183461108",
  "signature": "073e0094ff040bff6b3cd98faf2c5bad340200e7aa16fc24cd1e0223dc74c160",
  "site_identifier": "MOCK_PNM_SITE_IDENTIFIER",
  "timestamp": "1788369094",
  "version": "3.0"
}
--- signature verification ---
MATCH          : True
--- response ---
HTTP 201

Absent payment.refund on an ok response correctly yields REFUND_PENDING rather than a guessed terminal state.

--- Negative assertions (spec-mandated refusals) ---

N1 - 3DS must be refused, never downgraded
grpcurl -plaintext -H "x-connector-config: $CFG" ... -d '{... "auth_type":"THREE_DS","enrolled_for_3ds":true}' localhost:8000 types.PaymentService/Authorize
ERROR: grpc-status=Unimplemented, grpc-message="Three DS payments is not supported by paynearme"
       grpc-status-details-bin=message:"NOT_SUPPORTED"

N2 - manual capture must be refused
grpcurl -plaintext -H "x-connector-config: $CFG" ... -d '{... "capture_method":"MANUAL"}' localhost:8000 types.PaymentService/Authorize
ERROR: grpc-status=Unimplemented, grpc-message="manual capture is not supported by paynearme"
       grpc-status-details-bin=message:"NOT_SUPPORTED"

N3 - Capture must be NotImplemented and must NOT reach the connector
grpcurl -plaintext -H "x-connector-config: $CFG" ... -d '{"merchant_capture_id":"pnm_cap_001","connector_transaction_id":"798183461108","amount_to_capture":{"minor_amount":50499,"currency":"USD"}}' localhost:8000 types.PaymentService/Capture
ERROR: grpc-status=Unimplemented, grpc-message="This feature is not implemented: capture flow for paynearme"
       grpc-status-details-bin=message:"NOT_IMPLEMENTED"

N4 - non-USD must be refused
grpcurl -plaintext -H "x-connector-config: $CFG" ... -d '{"merchant_order_id":"pnm_order_eur","amount":{"minor_amount":50499,"currency":"EUR"}}' localhost:8000 types.PaymentService/CreateOrder
ERROR: grpc-status=Unimplemented, grpc-message="Currency EUR is not supported by paynearme"
       grpc-status-details-bin=message:"NOT_SUPPORTED"

Mock server total inbound request count for the entire run: 6 - the four negative tests produced zero outbound HTTP traffic, proving in particular that Capture never reaches /make_payment and cannot double-charge.
```

</details>

## Validation checklist

- [x] `cargo build` / `cargo check --workspace --all-targets` passed with zero errors
- [x] All six core flows exercised end-to-end against a spec-shaped mock; outbound HMAC signature independently verified on every call
- [x] Spec-mandated refusals (3DS, manual capture, Capture flow, non-USD) verified to produce zero outbound traffic
- [x] No credentials in committed source code; `creds.json` is not in the diff
- [x] Committed toml base URLs point at the real sandbox/production hosts (the mock URL appears nowhere in the diff)

https://claude.ai/code/session_01NmBdXRuQ6UoWEQTWPmxAku


crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+/// `/create_order` request. "With PayNearMe, an order is required any time money
+/// moves or is scheduled to move", so this runs ahead of every Authorize.
+#[derive(Debug, Serialize)]
+pub struct PaynearmeCreateOrderRequest {
+    pub site_identifier: Secret<String>,
+    pub timestamp: String,
+    pub version: String,
+    pub signature: Secret<String>,
+    pub order_amount: StringMajorUnit,
+    pub order_currency: String,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+    /// show `MM/YY`; `accounts.expiration_date` in responses uses `MM/YYYY`.)
+    pub payment_method_card_expiry_pii: Secret<String>,
+    pub payment_method_cvv_pii: Secret<String>,
+    pub payment_method_billing_name: Secret<String>,
+    pub payment_method_billing_address: Secret<String>,
+    pub payment_method_billing_zipcode: Secret<String>,
+    pub payment_method_billing_phone: Secret<String>,
+    pub send_payment: String,
+    pub payment_amount: StringMajorUnit,
+    pub payment_currency: String,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+        // `mandate_reference: None` — the merchant would believe a credential
+        // was stored when none was. Refuse, for the same reason 3DS is refused
+        // above rather than silently downgraded.
+        if request.mandate_id.is_some()
+            || request.setup_mandate_details.is_some()
+            || request.setup_future_usage == Some(FutureUsage::OffSession)
+        {
+            return Err(not_supported("Mandates / stored credentials"));
+        }
+
```
deepanshu-iiitu, line 390: Can we use/create a util for this?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+        // There is no capture endpoint anywhere in the API (see `Capture` in
+        // `paynearme.rs`): card payments are sale / auto-capture only.
+        if let Some(
+            method @ (CaptureMethod::Manual
+            | CaptureMethod::ManualMultiple
+            | CaptureMethod::Scheduled),
+        ) = request.capture_method
+        {
+            return Err(not_supported(format!("{method} capture")));
+        }
```
deepanshu-iiitu: Can we use/create a util for this?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+                .clone()
+                .ok_or(IntegrationError::MissingRequiredField {
+                    field_name: "connector_order_id",
+                    context: context(),
+                })?;
+
+        let billing_name = match common.get_optional_billing_full_name() {
+            Some(name) => name,
+            None => card.get_cardholder_name()?,
+        };
```
deepanshu-iiitu: Why are we using cardholdername as backup? Lets try to make billing_full_name as mandatory field?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+pub struct PaynearmeRefundRequest {
+    pub site_identifier: Secret<String>,
+    pub timestamp: String,
+    pub version: String,
+    pub signature: Secret<String>,
+    pub pnm_payment_identifier: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub refund_amount: Option<StringMajorUnit>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub refund_currency: Option<String>,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+
+/// The nested refund object. PayNearMe returns no refund identifier, so this is
+/// the only refund state the API exposes and it hangs off the payment record.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct PaynearmeRefundObject {
+    pub refund_status: Option<PaynearmeRefundStatus>,
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub refund_amount: Option<String>,
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub refund_currency: Option<String>,
```
deepanshu-iiitu: Can we use StringMajorUnit for amount field and also can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+    /// The connector transaction id — the key for PSync, Void, Refund and RSync.
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub pnm_payment_identifier: Option<String>,
+    /// The card token, retained in `connector_metadata` so a later
+    /// `/make_payment` against the saved instrument stays possible.
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub payment_method_identifier: Option<String>,
+    /// Total charged, **including** PayNearMe's convenience fee.
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub payment_amount: Option<String>,
```
deepanshu-iiitu: Can we use StringMajorUnit for payment_amount field?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+    /// Merchant settlement amount (payment minus fees) — not the captured amount.
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub net_payment_amount: Option<String>,
```
deepanshu-iiitu: Can we use StringMajorUnit for net_payment_amount field?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+    #[serde(default, deserialize_with = "deserialize_string_or_number")]
+    pub payment_type: Option<String>,
```
deepanshu-iiitu: Can we create an enum for payment_type field?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1370 @@
...
+            .filter(|payments| !payments.is_empty())
+            .or_else(|| {
+                self.electronic_payments
+                    .as_ref()
+                    .and_then(|electronic| electronic.payments.as_ref())
+                    .filter(|payments| !payments.is_empty())
+            })
+            .and_then(|payments| payments.last())
+    }
+}
```
deepanshu-iiitu, line 956: Please add a detailed comment about the last_payment fn?

deepanshu-iiitu (changes requested):

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — `order_currency` is now `common_enums::Currency` (was `String`).

Worth noting for the reviewer since this field is signed: the wire format is unchanged. `Currency` carries `#[serde(rename_all = "UPPERCASE")]` (`crates/common/common_enums/src/enums.rs:30`), so it serialises to exactly `"USD"` — the same bytes the previous `Currency::USD.to_string()` produced. That matters because the HMAC is computed over the serialised request body, so a changed encoding would have failed authentication at the gateway rather than anywhere visible locally.

`require_usd()` now returns `Currency` instead of `String` so the type flows through; the USD-only rejection is unchanged at all three call sites.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — `payment_currency` is now `common_enums::Currency`.

Same note as on `order_currency`: this field is part of the HMAC preimage, and `Currency`'s `#[serde(rename_all = "UPPERCASE")]` means it still serialises to `"USD"`, so the signed bytes are identical.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — the request-side `refund_currency` is now `Option<Currency>`.

`#[serde(skip_serializing_if = "Option::is_none")]` is untouched, so a full refund still omits the field entirely (per spec §8.6.1, omitting `refund_amount`/`refund_currency` is what asks PayNearMe for a full refund).

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Both done — this field is now `Option<StringMajorUnit>` and `refund_currency` beside it is now `Option<Currency>`.

One caveat worth recording, because the obvious change here is unsafe: `StringMajorUnit` is `pub struct StringMajorUnit(String)` with a plain derived `Deserialize` (`crates/common/common_utils/src/types.rs:373`), so it **rejects a JSON number**. PayNearMe returns these amounts as *both* a number and a string — that is exactly why `deserialize_string_or_number` exists in this file (`payment_amount` appears as `100` in one documented example and `"504.99"` in another).

So a direct swap would deserialize fine in any test we write and then fail on real traffic. It is implemented via a string-or-number tolerant deserializer that yields `StringMajorUnit`, which gives the type you asked for without the regression.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — `payment_amount` is now `Option<StringMajorUnit>`, via the same string-or-number tolerant deserializer described in the `refund_amount` thread (PayNearMe sends this field as a bare JSON number in some responses, which plain `StringMajorUnit` would reject).

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — `net_payment_amount` is now `Option<StringMajorUnit>`, same tolerant deserializer.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — added `PaynearmePaymentType`, covering all 12 values the API documents (`ach`, `ach_push`, `cash`, `cash_app`, `credit`, `debit`, `paypal`, `paypal-push`, `pin4`, `push-debit`, `venmo`, `venmo-push`), with serde renames for the hyphenated ones.

It carries `#[serde(other)] Unknown` so an undocumented value degrades to `Unknown` rather than failing the whole response body — this is an informational, output-only field (PayNearMe classifies credit vs debit from the BIN), so it should never be able to break a payment response.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — extracted to `is_mandate_or_stored_credential_request` in `crates/integrations/connector-integration/src/utils.rs`.

It takes the `PaymentsAuthorizeData` and returns `bool` (rather than the error itself) so each connector keeps its own error type and message. The three-way test — `mandate_id`, `setup_mandate_details`, `setup_future_usage == OffSession` — is documented there, including *why* a connector without `SetupMandate`/`RepeatPayment` has to reject rather than charge: the payment would otherwise succeed with `mandate_reference: None` and the merchant would believe a credential had been stored when none was.

`saferpay/transformers.rs:481` carries the same check inline (first two conditions only) and is cited in the util's doc comment as motivation — I have not refactored saferpay in this PR, to keep the diff scoped.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — extracted to `get_capture_method_requiring_separate_capture` in `utils.rs`. It returns `Option<CaptureMethod>` rather than `bool` so the caller can still name the offending method in its error message.

I deliberately did **not** reuse the existing `is_manual_capture` (`utils.rs:527`): it covers only `Manual | ManualMultiple`, so reusing it here would have silently dropped the `Scheduled` rejection that PayNearMe needs. The new helper's doc comment explains that difference so the two are not conflated later. `is_manual_capture` itself is unmodified.

Both new helpers use exhaustive matches rather than a `_` catch-all, so adding a `CaptureMethod` variant upstream breaks the build instead of silently falling through to "supported".

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Agreed, and changed — `billing_full_name` is now mandatory. The `card.get_cardholder_name()` fallback is gone; it is `common.get_billing_full_name()?`, so a caller that sends neither now gets a precise missing-field error instead of a gateway 400.

Flagging one consequence explicitly since it is visible in live traffic: a request that sends a card holder name but **no** billing name previously succeeded via the fallback and will now fail. That is the intended stricter behaviour, but it is a behaviour change rather than a pure refactor.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Done — `last_payment` now carries a full doc comment covering:

- **which envelopes it reads** — `order.payments[]` for card responses, falling back to `order.electronic_payments.payments[]`, which is where the ACH-shaped responses nest it;
- **why `.last()`** — an order can accumulate multiple payment attempts and the most recent one is the result of the call we just made;
- **when it returns `None`** — no order, or an absent/empty `payments[]`, which the callers deliberately treat as `Pending` (card tokenised but not charged) rather than as a failure.

deepanshu-iiitu (dismissed):

crates/integrations/connector-integration/src/utils.rs
```diff
@@ -529,6 +529,58 @@ pub(crate) fn is_manual_capture(capture_method: Option<enums::CaptureMethod>) ->
...
+    match capture_method {
+        Some(
+            method @ (enums::CaptureMethod::Manual
+            | enums::CaptureMethod::ManualMultiple
+            | enums::CaptureMethod::Scheduled),
+        ) => Some(method),
+        Some(enums::CaptureMethod::Automatic | enums::CaptureMethod::SequentialAutomatic)
+        | None => None,
+    }
+}
```
AmitsinghTanwar007: you can remove this instead just use is manual capture

crates/integrations/connector-integration/src/utils.rs
```diff
@@ -529,6 +529,58 @@ pub(crate) fn is_manual_capture(capture_method: Option<enums::CaptureMethod>) ->
...
+pub fn is_mandate_or_stored_credential_request<
+    T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static,
+>(
+    request: &PaymentsAuthorizeData<T>,
+) -> bool {
+    request.mandate_id.is_some()
+        || request.setup_mandate_details.is_some()
+        || request.setup_future_usage == Some(enums::FutureUsage::OffSession)
+}
+
```
AmitsinghTanwar007: here also is setup mandate function can be used

crates/integrations/connector-integration/src/utils.rs
shuklatushar226: You're right that this shouldn't be a new util — removed it in d7678439c.

One correction on the specific suggestion though: `is_manual_capture` (`utils.rs:527`) covers only `Manual | ManualMultiple`, not `Scheduled`. PayNearMe has no capture endpoint at all, so a `Scheduled` request has to be refused too — using `is_manual_capture` here would have let it through and charged it immediately as a sale. `CaptureMethod::Scheduled` is reachable (`domain_types/src/types.rs:1051` maps it straight from the proto value), so that isn't theoretical.

The better existing helper turned out to be `PaymentsAuthorizeData::is_auto_capture()` (`connector_types.rs:1773`), which is false for exactly `Manual | ManualMultiple | Scheduled` — the same set, already in the framework. So the guard is now `if !request.is_auto_capture()`, which is also what `worldpayxml/transformers.rs:808` does. `capture_method` is still read, but only to name the offending method in the error message.

`is_manual_capture` itself is left untouched.

crates/integrations/connector-integration/src/utils.rs
shuklatushar226: Agreed — removed in d7678439c. The guard now calls `PaymentsAuthorizeData::is_mandate_payment()` (`domain_types/src/connector_types.rs:1846`) instead of the new util. `utils.rs` is back to no net change in this PR.

One semantic difference worth recording, since it is now inherited rather than chosen: `is_mandate_payment()` requires `setup_future_usage == OffSession` **together with** `customer_acceptance` or `setup_mandate_details`, whereas the helper I had written also flagged `setup_future_usage: OffSession` on its own. So an authorize carrying only `OffSession` is no longer refused by this guard. I've gone with the framework's canonical definition rather than a stricter connector-local one — flagging it so the difference is a deliberate choice on record rather than a silent one.

AmitsinghTanwar007 (dismissed):

deepanshu-iiitu (dismissed):

XyneSpaces:
[blocking] `crates/types-traits/grpc-api-types/proto/payment.proto:912` adds `PAYNEARME = 143`, which collides with `PAYDOTCOM = 143` in PR #2223.

[blocking] `crates/types-traits/grpc-api-types/proto/payment.proto:6273` adds `PaynearmeConfig paynearme = 154`, which collides with `PaydotcomConfig paydotcom = 154` in PR #2223.

Coordinate with #2223 to use distinct enum/field numbers before either PR merges.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+
+impl PaynearmePaymentStatus {
+    /// Never guess a terminal state: anything undocumented resolves to `Pending`
+    /// and is settled by PSync.
+    fn attempt_status(&self) -> AttemptStatus {
+        match self {
+            // Terminal success. PayNearMe card payments are sale / auto-capture.
+            Self::Approved => AttemptStatus::Charged,
+            // Emitted as-is; there is no capture endpoint to move it forward.
+            Self::Authorized => AttemptStatus::Authorized,
```
JeevaRamu0104: `authorized` maps to `AttemptStatus::Authorized`, which tells the caller to capture — but Capture is `not_implemented` and `supported_capture_methods` is `[Automatic]` only. A payment landing here can't be advanced: capture errors, and PSync keeps returning the same status. Should this be `Pending` so PSync can settle it to `approved`?

crates/integrations/connector-integration/src/connectors/paynearme.rs
```diff
@@ -0,0 +1,579 @@
...
+            .change_context(crate::utils::response_deserialization_fail(
+                res.status_code,
+                "paynearme: response body did not match the documented {status, errors[]} error envelope.",
+            ))?;
+
+        with_error_response_body!(event_builder, response);
+
+        Ok(
+            response
+                .to_error_response(res.status_code, FlowStatus::Payment(AttemptStatus::Failure)),
```
JeevaRamu0104: `FlowStatus::Payment(AttemptStatus::Failure)` is hardcoded here, but `get_error_response_v2` routes Refund and RSync through this too, and `ForeignFrom<FlowStatus> for RefundStatus` (`types.rs:7815`) maps `Payment(_)` → `RefundFailure`. So an HTTP error on a refund reports terminal failure — while the transformer's own in-band paths deliberately use `FlowStatus::Refund(RefundStatus::Pending)`. A merchant told the refund failed may re-issue it.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+                payment_method_type: PAYMENT_METHOD_TYPE_CARD.to_string(),
+                payment_method_card_number_pii: Secret::new(card.card_number.peek().to_string()),
+                payment_method_card_expiry_pii: card_expiry_mm_yyyy(card)?,
+                payment_method_cvv_pii: card.card_cvc.clone(),
+                payment_method_billing_name: billing_name,
+                payment_method_billing_address: common.get_billing_line1()?,
+                payment_method_billing_zipcode: common.get_billing_zip()?,
+                // Documented as required. It can be made optional per site, but
+                // surfacing the gap here beats a 400 from the gateway.
+                payment_method_billing_phone: common.get_billing_phone_number()?,
```
JeevaRamu0104: `get_billing_phone_number()` prefixes the country code including the `+` (`payment_address.rs:380`). The committed probe shows what goes on the wire:

```
"payment_method_billing_phone": "+14155552671"
```

If PayNearMe wants a bare 10-digit US number, every card authorize 400s on a field this connector marks required. Was the format confirmed against the spec?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+}
+
+/// Build the `string_to_sign` from an already-serialised request body.
+///
+/// 1. Drop the exempt parameters (`format`, `signature`, `call`).
+/// 2. Drop absent parameters — a field skipped by
+///    `skip_serializing_if = "Option::is_none"` is not on the wire either.
+/// 3. Sort what remains alphabetically by key (plain lexicographic sort).
+/// 4. Concatenate `key + value` with **no** separators.
+fn paynearme_string_to_sign(
```
JeevaRamu0104, line 141: Nothing tests this. It's the only novel logic in the PR, there's no live sandbox, and the mock-server run in the description isn't reproducible in CI — so the sort order, the exempt-field skip and the null skip stay unverified on every future change. A `#[cfg(test)]` case pinning the preimage for a fixed body would lock it in.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+        let auth = PaynearmeAuthType::try_from(&router_data.connector_config)?;
+        let common = &router_data.resource_common_data;
+
+        let order_currency = require_usd(router_data.request.currency)?;
+        let order_amount =
+            PaynearmeAmountConvertor::convert(router_data.request.amount, Currency::USD)?;
+
+        // `site_customer_identifier` is required and is a client-created unique
+        // string; fall back to the attempt reference when no customer is attached.
+        let site_customer_identifier = common
```
JeevaRamu0104, line 268: nit: this fallback is unique per order, so every payment without a customer mints a fresh PayNearMe customer record.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+
+        Ok(Self {
+            response: Ok(PaymentsResponseData::TransactionResponse {
+                resource_id,
+                redirection_data: None,
+                mandate_reference: None,
+                connector_metadata: payment.connector_metadata(),
+                network_txn_id: None,
+                network_txn_link_id: None,
+                connector_response_reference_id: payment.site_payment_identifier.clone(),
```
JeevaRamu0104: PSync returns `site_payment_identifier` here while Authorize returns the order id (line 1201), so the same payment reports two different reference ids depending on which call the caller made.

crates/integrations/connector-integration/src/connectors/paynearme.rs
JeevaRamu0104: Every non-2xx on every flow gets `Payment(AttemptStatus::Failure)` here, and PSync reads it straight through (`domain_types/src/types.rs:8673`) — so a 400 from clock skew, a rate limit, or a PayNearMe 5xx reports a charged payment as FAILURE. The body-level PSync path returns `Pending` (transformers.rs:1262) and Void returns `VoidFailed` (transformers.rs:1331) for exactly this reason. Passing `attempt_status: None` lets each flow's own status stand, as `ilixium.rs:190` does.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+
+impl TryFrom<ResponseRouterData<PaynearmeRefundResponse, Self>> for RefundRouterData {
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<PaynearmeRefundResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let response = item.response.0;
+
+        let payment = match (response.is_ok(), response.payment.as_ref()) {
```
JeevaRamu0104, line 1475: `(false, _)` — PayNearMe explicitly declining the refund — falls into the same arm as `(true, None)` "accepted but not visible yet", so both report `Pending`. RSync then reads `payment.refund`, which never exists for a refund that was never created, so it stays Pending forever and the merchant waits on money that never moves.

Split the arms and map a declared error envelope to `RefundStatus::Failure`; Authorize and Void in this file already distinguish the two.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
JeevaRamu0104: Capture is wired `not_implemented` and `supported_capture_methods` is `[Automatic]`, so a payment reported `authorized` is stuck: Hyperswitch sees an attempt awaiting capture and Capture returns NotImplemented. Should this map to `Pending` so PSync polls through to `approved` instead?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+    type Error = error_stack::Report<IntegrationError>;
+
+    fn try_from(item: PaynearmeRouterData<RefundRouterData, T>) -> Result<Self, Self::Error> {
+        let router_data = &item.router_data;
+        let request = &router_data.request;
+        let auth = PaynearmeAuthType::try_from(&router_data.connector_config)?;
+
+        let currency = require_usd(request.currency)?;
+        let (refund_amount, refund_currency) =
+            if request.minor_refund_amount == request.minor_payment_amount {
```
JeevaRamu0104, line 606: Omitting `refund_amount` asks PayNearMe to refund its whole `payment_amount`, which line 824 documents as including PayNearMe's convenience fee. A UCS-full refund of 504.99 would return 504.99 + fee to the cardholder while UCS records 504.99. Is the fee meant to be refunded, or should `refund_amount` always be sent explicitly?

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1499 @@
...
+) -> ErrorResponse {
+    let rendered: Vec<String> = errors.iter().map(PaynearmeErrorItem::render).collect();
+    let code = response_code
+        .filter(|code| *code != RESPONSE_CODE_SUCCESS)
+        .map(str::to_string)
+        .unwrap_or_else(|| common_utils::consts::NO_ERROR_CODE.to_string());
+    let message = rendered
+        .first()
+        .cloned()
+        .unwrap_or_else(|| "Payment declined by Paynearme".to_string());
```
JeevaRamu0104: nit: a 401 signature rejection or a PayNearMe 5xx surfaces to the merchant as "Payment declined by Paynearme".

JeevaRamu0104 (changes requested):

AmitsinghTanwar007 (dismissed):

crates/integrations/connector-integration/src/connectors/paynearme.rs
shuklatushar226: Fixed in f4a454731 — good catch, this was the worst bug in the PR.

`build_error_response` now takes `Option<FlowStatus>` and the generic `to_error_response` (the only path `get_error_response_v2` reaches) passes `None`, so each flow's own status stands. That follows `ilixium.rs:190` as you suggested, and the framework's own `get_5xx_error_response` default (`interfaces/src/connector_integration_v2.rs:232`) does the same. The six in-band call sites keep their explicit statuses.

To confirm the blast radius you identified: `get_error_response_v2` is in `connector_default_implementations` on all six flows, so the single hardcoded `Payment(AttemptStatus::Failure)` really did apply to PSync, Void, Refund and RSync as well as Authorize.

shuklatushar226: Same fix as the sibling thread — f4a454731. `to_error_response` now passes no forced status, so the `ForeignFrom<FlowStatus> for RefundStatus` mapping of `Payment(_)` → `RefundFailure` can no longer turn a transport error on a refund into a terminal failure.

You were right that this contradicted the transformer's own in-band choices: those deliberately return `Refund(RefundStatus::Pending)` for exactly this reason, and the HTTP path was overriding that intent.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Fixed in f4a454731 — and you caught an over-correction of mine, so thank you.

An earlier commit had moved both of these arms to `Pending` to stop a transient body failing a refund that might have succeeded. That was right for the ambiguous case and wrong for yours: `(false, _)` is PayNearMe *declaring* an error, and lumping it in with "accepted but not visible yet" left RSync polling `payment.refund` for an object that would never exist.

The arms are now split in both Refund and RSync:
- `(false, _)` explicit error envelope → `RefundStatus::Failure`
- `(true, None)` ok envelope, no `payment` object → `Pending`
- `(true, Some(payment))` with `payment.refund == None` stays `Pending`, per spec §8.6.3

Worth noting for anyone reading later: non-2xx responses never reach this transformer (`external-services/src/service.rs:429`, `:484-494` route them to `get_error_response_v2` / `get_5xx_error_response`), so the `(false, _)` arm only fires on an HTTP 2xx that still says `status: "error"` — a settled refusal rather than a transient blip, which is what makes `Failure` the right call here.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Agreed and changed in f4a454731 — `authorized` now maps to `AttemptStatus::Pending`.

Your reasoning is exactly right: with Capture `not_implemented` and `supported_capture_methods` = `[Automatic]`, an attempt reported `Authorized` is unreachable in both directions — capture errors out, and PSync keeps returning the same status.

This is a deliberate deviation from spec §10.1 line 1159 ("emit as-is"), so I marked it in the code comment rather than leaving it to look like a codegen slip. The spec's own note for that row says the payment "will settle on its own or be resolved by PSync", which only actually works if the status is pollable — which `Pending` is and `Authorized` is not.

shuklatushar226: Yes — changed to `Pending` in f4a454731. Full reasoning in the sibling thread on the same line, including why this is a deliberate deviation from spec §10.1 and how it is marked in the code.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Good catch — fixed in f4a454731.

To answer directly: no, the format had not been confirmed against the spec, and the spec disagrees with what we were sending. Spec line 609's worked example is `"payment_method_billing_phone": "469-555-5878"` — a bare national number, no country code and no `+`. Nothing in §8.2.1 asks for one, and line 583 marks the field required, so as you say this plausibly 400'd every card authorize.

`get_billing_phone_number()` prefixes `+<country code>` (`payment_address.rs:380`). It now uses `common.get_billing_phone()?.get_number()?` (`connector_types.rs:886`, `payment_address.rs:375`) to send the national number alone.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Agreed — added in f4a454731. You're right that this was the only novel logic in the PR with nothing covering it, and that the mock run in the description isn't reproducible in CI.

`#[cfg(test)] mod tests` now pins four cases:
- both worked vectors from spec §3.1, asserting the **preimage and the signature** (`65c694f4…4b3b82` and `7986b4e5…aadcc`) — example 2 is the one that exercises the `format` exemption and the alphabetical ordering
- the refusal when a required envelope field is missing
- the invariant that an omitted field never enters the signature, which the full-refund request shape depends on

Both vectors were independently recomputed with Python `hmac`/`hashlib` before being pinned, so the test asserts against the spec rather than against the implementation's current behaviour. `cargo test -p connector-integration --lib paynearme` → 4 passed.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Fixed in f4a454731. All payment flows now report `pnm_order_identifier` (spec §7.3) and none of them reports `site_payment_identifier`, which as you say was only ever the echo of our own attempt reference.

- Authorize: the order id from the response, falling back to `resource_common_data.connector_order_id`
- PSync: `connector_order_id`, which the sync request carries as `connector_order_reference_id` (`domain_types/src/types.rs:6004`)
- Void: the same field — today always `None` (`types.rs:6083`), so it reports no new reference rather than a contradictory second one

On that last point: `None` here does not erase what Authorize recorded. `PaymentAttemptUpdateInternal` derives `AsChangeset` with no `treat_none_as_null`, so diesel's default applies and `None` skips the column rather than writing NULL. I checked that specifically before choosing `None` over inventing a value.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Fixed in f4a454731. The fallback chain is now:

1. `errors[0]` when PayNearMe sent one
2. `"Payment declined by Paynearme (response_code N)"` when it sent a non-zero `response_code` — that genuinely is a decline per §11.3
3. `"Paynearme request failed with HTTP status {code}"` when it sent neither

So a 401 signature rejection or a 5xx no longer reaches the merchant as a card decline.

I deliberately did not echo the raw body into `reason`: a body that isn't the documented JSON envelope never gets this far (`parse_struct` fails first at `paynearme.rs:215-221`), so it would add little, and putting an arbitrary gateway body into a merchant-visible field is an avoidable leak risk.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: Confirmed, and left as-is — this fallback is what spec §7.3 (line 453) prescribes: when the caller sends no customer, `site_customer_identifier` falls back to `connector_request_reference_id`.

So yes, a customer-less payment does mint a fresh PayNearMe customer record each time. That is the spec'd behaviour rather than a defect, and it also means there is no customer-scoped dedupe available on that path — worth knowing, but not something this PR should change unilaterally.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
shuklatushar226: You've found a real problem, and I've deliberately **not** changed the code for it — I think this needs a decision rather than a patch. Laying out what I found.

**The spec is unambiguous that the field must be omitted**, in three separate places:
- line 955, quoting PayNearMe's own field description: *"This parameter should only be included for partial-amount refunds."*
- lines 958-966, an explicit implementation-rule block whose full-refund branch is literally `(None, None)` — which this code reproduces
- line 1562, checklist item 10: *"`Refund` → `/refund_payment`, omitting `refund_amount`/`refund_currency` for a full refund"*

**And I agree with you that following it produces the wrong outcome.** Spec lines 340-344 and `transformers.rs:824` both state `payment_amount` includes PayNearMe's convenience fee, so an omitted `refund_amount` returns 504.99 + fee to the cardholder while UCS records 504.99. The merchant eats the difference.

The reason I have not just started sending `refund_amount` explicitly: PayNearMe's own documentation says the field is for partial refunds only, so an explicit full-amount value may well be rejected — trading a silent over-refund for a hard failure on every full refund. Both branches are guesses without an answer to the actual question, which is: **does an omitted `refund_amount` refund `payment_amount` or `net_payment_amount`?**

That is a question for PayNearMe support / onboarding, not something to settle in code. Given there are no sandbox credentials on this integration yet, it cannot be settled empirically here either.

Happy to change it the moment we have an answer — or to send it explicitly now if you'd rather take the rejection risk over the over-refund risk. Flagging it as a known open item on the connector either way, since it is real money.

crates/integrations/connector-integration/src/connectors/paynearme/transformers.rs
```diff
@@ -0,0 +1,1767 @@
...
+    ) -> Result<Self, Self::Error> {
+        let response = item.response.0;
+
+        let payment = match (response.is_ok(), response.payment.as_ref()) {
+            (true, Some(payment)) => payment,
+            // As in the Refund handler above: a declared error on a 2xx body
+            // (an unknown `pnm_payment_identifier`, say) is PayNearMe saying
+            // there is nothing to sync, and no amount of further polling will
+            // change that. Transport failures do not land here.
+            (false, _) => {
```
JeevaRamu0104: The premise doesn't hold: `PaynearmeRefundSyncRequest` (line 559) looks up `connector_transaction_id`, the payment id — there's no refund identifier that could be unknown (§8.6.3). So this fires on an in-band error about the *payment* and terminally fails the refund. PSync maps the same envelope condition to `Pending` (line 1330). Is `Failure` right here, given your note below that a false failure invites a second refund?

shuklatushar226: You're right, and I've reverted it in 8e0b21a3c — RSync now reports `Pending` for the whole non-`(true, Some)` case, matching PSync.

Your premise checks out: `PaynearmeRefundSyncRequest` builds a `PaynearmePaymentLookupRequest` from `router_data.request.connector_transaction_id`, the payment id, and per §8.6.3 the payment id doubles as `connector_refund_id` — so there genuinely is no refund identifier that could come back unknown. My comment's justification was about the wrong entity.

The distinction I should have drawn is write vs read:
- **Refund** is a write. A declared error there is PayNearMe refusing to *create* the refund, so nothing will ever appear under `payment.refund` for RSync to poll. Terminal failure is correct, and that arm is unchanged.
- **RSync** is a read. A declared error says we could not read the payment; it says nothing about whether the refund exists. Failing on it is a false failure — and, as you note, that is precisely the hazard the split was meant to prevent, so I had reintroduced it one flow over.

Thanks for catching it. That is twice now on this file that the refund status handling needed a second pass.

JeevaRamu0104 (dismissed):

AmitsinghTanwar007 (dismissed):

deepanshu-iiitu (dismissed):

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #2223, opened 2026-09-02
feat(connector): [PAY.COM] add Pay.com connector — Authorize (3DS + no-3DS), PSync, Capture, Void, Refund, RSync

shuklatushar226:
## Summary

Adds the **Pay.com** connector (`Paydotcom`) to UCS with the core card payment flows: **Authorize** (non-3DS, external-MPI 3DS pass-through, and the full gateway 3DS challenge journey), **PSync**, **Capture**, **Void**, **Refund**, **RSync**, plus **PreAuthenticate** and **Authenticate** as the first two legs of the 3DS challenge.

## Scope

- **Cards only, one-time payments.** Both non-3DS and 3DS.
- **Not** in scope: wallets, mandates/MIT, recurring, tokenization for reuse, payouts, disputes, webhooks.
- Proto enum: `PAYDOTCOM = 141`.
- Auth: `HeaderKey`, sent as the `x-paycom-api-key` header.
- Base URLs: `https://api.sandbox.pay.com` (sandbox/development), `https://api.pay.com` (production).
- `idempotency-key` (a fresh UUID v4) is sent on every POST.

## Endpoint map

| Flow | Request |
|---|---|
| Authorize (auto-capture) | `POST /v1/charges` |
| Authorize (manual capture) | `POST /v1/holds` |
| PSync | `GET /v1/charges/{id}` or `GET /v1/holds/{id}` — routed off the `chrg_`/`hld_` id prefix |
| Capture | `POST /v1/holds/{id}/capture` |
| Void | `POST /v1/holds/{id}/cancel` |
| Refund | `POST /v1/refunds` |
| RSync | `GET /v1/refunds/{id}` |

**Capture returns a Charge with a NEW `chrg_` id.** The transformer rewrites `connector_transaction_id` to that new id — without this, refunds issued after a manual capture break.

### 3DS challenge (3 legs)

1. **PreAuthenticate** → `POST /v1/charges` or `POST /v1/holds` with `authentication_context` and `payment_method_options.card.request_threed_secure = "challenge"`. Mints the `chrg_`/`hld_` id.
2. **Authenticate** → `POST /v1/sessions/authentication/linked`, which returns the challenge URL, surfaced to UCS as a `RedirectForm` with `HTTP_METHOD_GET`.
3. **Authorize** → `POST /v1/{charges|holds}/{id}/confirm`.

The `chrg_`/`hld_` id rides **`AuthenticationData.transaction_id`** (proto `connector_transaction_id`), which is the channel an orchestrator already moves between these three RPCs — so Hyperswitch needs no connector-specific plumbing to drive the journey. Only `transaction_id` is ever populated, never `eci`/`cavv`: those would be indistinguishable from a merchant-supplied external-MPI authentication in `three_ds_mode()`.

A mirror of the id is also published on `connector_feature_data` as `{"paydotcom_resource": …}`. That is a **fallback** for grpcurl / SDK / composite-service callers, which have no orchestrator moving `authentication_data` for them; the Hyperswitch path does not use it.

`connector_transaction_id` deliberately stays the `chrg_`/`hld_` id and **never** becomes the `sca_` session id.

An earlier revision ran leg 2 on the Authorize flow instead, which forced a change to a shared Hyperswitch transformer. Driving it from Authenticate removes that entirely — see juspay/hyperswitch#14016.

### External-MPI 3DS pass-through

`source_data.card.three_ds { eci, cavv, ds_trans_id, version }` is supported as a **single** Authorize call (no extra legs).

## Testing / evidence

> **No real Pay.com sandbox credentials exist for this run** — `creds.json` holds placeholders only. The connector was therefore validated against a **local mock Pay.com HTTP server** that recorded the exact URL, method, headers and JSON body UCS emits on the wire.

**23 grpcurl scenarios passed through UCS against that mock**, covering:

- no-3DS authorize
- external-MPI 3DS pass-through
- frictionless gateway 3DS
- the full 3-leg 3DS challenge journey for **both** the charge and the hold variant
- PSync
- manual-capture authorize
- capture — with the connector-transaction-id rewrite proven by a subsequent successful refund
- void
- refund
- rsync
- prefix-mismatch rejections (`chrg_` id sent to a `hld_` endpoint and vice versa)
- an **in-band decline** (HTTP 200 with status `"failed"`)
- an **out-of-band decline** (HTTP 402 error envelope)

<details>
<summary>Verbatim no-3DS Authorize request recorded by the mock</summary>

```
POST /v1/charges
x-paycom-api-key: <api key>
idempotency-key: 8823ca14-1c51-4ffb-9b30-36defdffb9d7
via: HyperSwitch

{
  "amount": 10000,
  "currency": "usd",
  "reference": "paydotcom_txn_no3ds_001",
  "customer_reference_id": "cust_paydotcom_001",
  "payment_method_options": { "card": { "request_threed_secure": "none" } },
  "source_data": {
    "type": "card",
    "card": {
      "number": "4111111111111111", "expiry_month": "08", "expiry_year": "2030",
      "cvc": "999", "name": "John Doe",
      "billing_address": { "line1": "Georgiou Griva Digeni 43", "line2": "Suite 5",
        "city": "Limassol", "state": "CA", "postal_code": "3106", "country": "US" }
    },
    "billing_details": { "email": "shopper@example.com", "name": "John Doe", "phone": "35797123456",
      "address": { "line1": "Georgiou Griva Digeni 43", "line2": "Suite 5", "city": "Limassol",
        "state": "CA", "postal_code": "3106", "country": "US" } }
  }
}
```

</details>

<details>
<summary>3DS challenge leg 2 (Authenticate) request recorded by the mock</summary>

```
POST /v1/sessions/authentication/linked
{"resource":"chrg_146231656762572869","return_url":"https://example.com/payment/return","confirm":false}
```

The response challenge URL is surfaced to UCS as `redirectionData.form.endpoint`:

```
"https://sca.pay.com/authenticate?client_secret=mock.jwt.chrg_146231656762572869"  (HTTP_METHOD_GET)
```

</details>

## Gates

- [x] `cargo build --workspace` — green
- [x] `cargo +nightly fmt --all --check` — clean
- [x] `RUSTFLAGS="-D warnings" cargo clippy -p connector-integration --all-features --all-targets` — clean
- [x] `check_connector_specs` — passes
- [x] `make field-probe` — reports all in-scope flows supported
- [x] Only Pay.com-specific files and the required registration points modified

## Open items — unverified without real credentials

Flagging these explicitly so a reviewer with live Pay.com access can confirm them:

1. **`amount_to_capture` / `amount_to_refund` are typed `string`** in Pay.com's OpenAPI while every other amount field is an integer. Implemented as `StringMinorUnit`. Unverified against the real API.
2. **Whether `POST /v1/sessions/authentication/linked` accepts a `hld_` resource.** The docs only show `chrg_`. The hold variant of the challenge journey is implemented and passes against the mock, but the real API may reject it.
3. **`billing_details` placement.** Emitted **both** inside `source_data` (per the 3DS guide) **and** as `card.billing_address` (per the OpenAPI). Unclear which one the real API honours.
4. **Which `authentication_context` members are truly mandatory** — the spec is ambiguous; the implementation sends a conservative superset.
5. **3DS leg 3 sends `{}` as the request body**, because the macro-generated `get_request_body` always emits one, while Pay.com documents no request body on `/confirm`. Harmless against the mock; unverified live.

### Pre-existing failure — NOT introduced by this PR

`clippy::nonminimal_bool` fails at `crates/integrations/connector-integration/src/connectors/saferpay/transformers.rs:1501`, introduced by commit `b0d5b028c`. It reddens workspace-wide clippy regardless of this PR and needs a separate one-line fix. **Deliberately not fixed here.**

---

https://claude.ai/code/session_01Ars9THCqUzhuZyH4QqgFXT


crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+// ===== REQUEST: CREATE CHARGE / CREATE HOLD =====
+
+/// Body of `POST /v1/charges` **and** `POST /v1/holds` — the two endpoints take an
+/// identical payload in this scope, only the URL differs (see `paydotcom.rs`).
+#[derive(Debug, Serialize)]
+pub struct PaydotcomCreateResourceRequest<T: PaymentMethodDataTypes> {
+    /// Minor units, JSON integer.
+    pub amount: MinorUnit,
+    /// Lower-case ISO-4217.
+    pub currency: String,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub authentication_context: Option<PaydotcomAuthenticationContext>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct PaydotcomThreeDsRaw {
+    pub eci: String,
+    pub cavv: Secret<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ds_trans_id: Option<String>,
```
deepanshu-iiitu: Should this field be a secret?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+            source_type: PaydotcomSourceType::Card,
+            card: PaydotcomCardSourceDetails {
+                number: card.card_number.clone(),
+                expiry_month: card.get_card_expiry_month_2_digit()?,
+                expiry_year: card.get_expiry_year_4_digit(),
+                cvc: card.card_cvc.clone(),
+                name: card
+                    .card_holder_name
+                    .clone()
+                    .or_else(|| common.get_optional_billing_full_name()),
```
deepanshu-iiitu: Lets make card holder name a required field and lets not fallback to billing_full_name

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct PaydotcomChargeResponse {
+    pub id: String,
+    pub status: PaydotcomChargeStatus,
+    #[serde(default)]
+    pub amount: Option<MinorUnit>,
+    #[serde(default)]
+    pub amount_refunded: Option<MinorUnit>,
+    #[serde(default)]
+    pub currency: Option<String>,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct PaydotcomHoldResponse {
+    pub id: String,
+    pub status: PaydotcomHoldStatus,
+    #[serde(default)]
+    pub amount: Option<MinorUnit>,
+    #[serde(default)]
+    pub amount_capturable: Option<MinorUnit>,
+    #[serde(default)]
+    pub currency: Option<String>,
```
deepanshu-iiitu: Can we use currency enum here?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+        let pending = item.response.pending_metadata();
+        let connector_feature_data = pending.clone().map(Secret::new);
+        // The id the following Authenticate leg has to authenticate. This rides
+        // `authentication_data`, the channel an orchestrator already carries from the
+        // PreAuthenticate response into the Authenticate request — no connector-specific
+        // metadata plumbing needed. It is mirrored on `connector_feature_data` above for
+        // callers that drive the gRPC flows directly.
+        let authentication_data = pending
+            .is_some()
+            .then(|| resource_id_authentication_data(&resource_id));
```
deepanshu-iiitu: Can we refactor this line and add an explanation on whats happening here

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1812 @@
...
+// ===== RESPONSE: REFUND =====
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct PaydotcomRefundResponse {
+    pub id: String,
+    pub status: PaydotcomRefundStatus,
+    #[serde(default)]
+    pub amount: Option<MinorUnit>,
+    #[serde(default)]
+    pub currency: Option<String>,
```
deepanshu-iiitu: Can we use currency enum here?

deepanshu-iiitu (changes requested):

XyneSpaces:
[blocking] `crates/types-traits/grpc-api-types/proto/payment.proto:912` adds `PAYDOTCOM = 143`, which collides with `PAYNEARME = 143` in PR #2221.

[blocking] `crates/types-traits/grpc-api-types/proto/payment.proto:6271` adds `PaydotcomConfig paydotcom = 154`, which collides with `PaynearmeConfig paynearme = 154` in PR #2221.

Coordinate with #2221 to use distinct enum/field numbers before either PR merges.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done — typed as `common_enums::Currency` now (create request), along with the three response bodies below.

Pay.com spells currency lower-case on the wire and the enum serialises UPPERCASE, so there is a small `paydotcom_currency` serde adapter module that normalises the case in both directions, replacing the ad-hoc `to_lowercase()` that was at the call site. The emitted wire value is unchanged — re-ran the end-to-end suite and the mock still records `"currency": "usd"`.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done — `Option<common_enums::Currency>` with the `paydotcom_currency::option` adapter, which uppercases before parsing so Pay.com's lower-case value round-trips.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done — same treatment as the Charge response.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done — same treatment on the Refund response.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done, wrapped in `Secret`.

Worth noting for the record: the domain model keeps `AuthenticationData.ds_trans_id` as a plain `Option<String>` (only `cavv` is `Secret` there), so this is masked at the connector boundary rather than inherited. It is a directory-server correlation id rather than a credential, but it costs nothing to keep it out of a log line next to the cryptogram.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done. `PaydotcomCardSourceDetails.name` is now a required `Secret<String>` and the `or_else(billing_full_name)` fallback is gone — agreed that the two can legitimately differ and the card name is what Pay.com wants.

Behaviour change worth being aware of: a card with no holder name is now rejected instead of silently sending the billing name. Verified through the API — it surfaces as `IR_04 Missing required param: Missing required field: payment_method_data.card.card_holder_name` rather than reaching the connector.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Done — the `pending.is_some().then(..)` is now a plain `map`, with a comment above the block explaining what `pending` actually means and why the id goes onto two carriers:

`pending` is `Some(resource_id)` only while Pay.com is waiting on the shopper, i.e. when the Charge/Hold still needs the linked-authentication session the following Authenticate leg creates. When it is `None` the journey is already resolved and there is no id to forward, so both carriers stay empty and the orchestrator stops after this leg. The id travels on `authentication_data` because that is the channel an orchestrator already moves from a PreAuthenticate response into the next Authenticate request; `connector_feature_data` mirrors it for callers driving the gRPC flows directly, who have no orchestrator doing that for them.

shuklatushar226:
Proto collision addressed: **`PAYDOTCOM` 143 → 144** and **`PaydotcomConfig` 154 → 155** (`e5c0f610c`), clear of PayNearMe (#2221).

Some context, because picking a number is not a stable fix on its own: **six open connector PRs are currently sitting on 143/154** — #2221, #2213, #2207, #2206, #2204 and this one. Each branched when that was the next free number, and main has since taken 141 (JPMorgan Orbital) and 142 (TravelHub). This PR has already been renumbered twice by merges landing underneath it.

144/155 is unclaimed by main and by every open PR as of now, so this de-conflicts with #2221 and with the other four. But whichever of these PRs merges second will still have to renumber, so the number here should be re-checked at merge time rather than trusted from this comment.

All seven review comments are addressed in the same push; replies are on the individual threads. Re-ran the end-to-end suite (HS → UCS → mock) after the changes — all seven scenarios pass, and the currency retype leaves the wire value unchanged at `"currency": "usd"`.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1871 @@
...
+            raw_connector_request: None,
+            typed_connector_request: None,
+        }
+    }
+
+    pub fn to_refund_error_response(&self, status_code: u16) -> ErrorResponse {
+        let mut error_response = self.to_error_response(status_code);
+        // `attempt_status` is a payment concept; a failed refund must not rewrite the
+        // payment attempt.
+        error_response.attempt_status = None;
```
JeevaRamu0104: `attempt_status = None` means every refund error reports `REFUND_STATUS_UNSPECIFIED`. The refund error builder (`domain_types/src/types.rs:9600`) reads only `e.attempt_status` and defaults to unspecified — it has no `resource_common_data.status` fallback the way the payment path does. Saferpay sets it explicitly (`saferpay/transformers.rs:765`):

```suggestion
        error_response.attempt_status = Some(FlowStatus::Refund(RefundStatus::Failure));
```

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1871 @@
...
+        ErrorResponse {
+            status_code: http_code,
+            code: failure_code
+                .clone()
+                .unwrap_or_else(|| NO_ERROR_CODE.to_string()),
+            message: failure_message
+                .clone()
+                .unwrap_or_else(|| NO_ERROR_MESSAGE.to_string()),
+            reason: failure_message,
+            attempt_status: Some(FlowStatus::Payment(AttemptStatus::Failure)),
```
JeevaRamu0104: Hardcoding `Failure` discards the status the caller just computed: Capture (:1585) derives `CaptureFailed`, Authenticate (:1503) derives `AuthenticationFailed`. `get_attempt_status_for_grpc` prefers this field over the `resource_common_data.status` fallback, so a declined capture reports `FAILURE` instead of `CAPTURE_FAILED` — the merchant sees a failed payment while the hold is still intact and capturable. Can `in_band_error` take the status as an argument?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1871 @@
...
+
+/// Decides the leg from the **request alone**, never from a previous response, so
+/// `get_url` and `get_request_body` can never disagree about which call is being made.
+///
+/// An external-MPI Authorize also carries `authentication_data`, but with `eci`/`cavv`
+/// and no `transaction_id` — so it falls through to `Create`, as it must.
+pub fn authorize_leg<T: PaymentMethodDataTypes>(
+    request: &PaymentsAuthorizeData<T>,
+) -> PaydotcomAuthorizeLeg {
+    if authorize_pending_resource_id(request).is_some() {
```
JeevaRamu0104, line 608: This treats any `authentication_data.transaction_id` as a pending Pay.com resource id, but proto field 8 is documented as "transaction identifier generated by the 3DS system" — exactly what an external MPI supplies alongside `eci`/`cavv`. If one does, the request routes to `Confirm` and `payment_resource_path` then rejects the id. Gating on the `chrg_`/`hld_` prefix would enforce the invariant the comment above assumes.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1871 @@
...
+{
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<PaydotcomRefundResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        Ok(Self {
+            response: Ok(RefundsResponseData {
+                connector_refund_id: item.response.id.clone(),
+                refund_status: RefundStatus::from(item.response.status),
```
JeevaRamu0104: A 200 carrying `status: "failed"` returns `Ok` and drops `failure_code` / `failure_message` — both are parsed at :1653 and read nowhere, so a merchant gets a failed refund with no reason. The payments path turns the same shape into an `ErrorResponse` at :1396.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Good catch — fixed, and you are right about the asymmetry: the refund error builder reads `e.attempt_status` and `unwrap_or_default()`s to `REFUND_STATUS_UNSPECIFIED`, with no `resource_common_data.status` fallback like the payment path has. Now set to `Some(FlowStatus::Refund(RefundStatus::Failure))`, matching the Saferpay precedent you pointed at.

Verified through the API against a 402 from the connector: the refund lands as `failure` with `refund_failed` / `issuer_unavailable`, where before it would have come back unspecified.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Fixed — `in_band_error` now takes the status as an argument, and all six call sites pass the one they just derived.

You were right about the consequence, and I checked it rather than assuming: a capture declined in band used to persist attempt status `failure`; it now persists `capture_failed` with `capture_declined` / "The capture was declined.", while the hold stays capturable. That is the difference between the merchant seeing a dead payment and seeing a retryable capture.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Fixed — now gated on the `chrg_`/`hld_` prefix.

You have found the real hazard in this design. Field 8 is the natural carrier for the Pay.com resource id and it round-trips cleanly, but it is documented as the 3DS system's own identifier, so an external MPI populating it alongside eci/cavv would have been read as a pending resource, routed the Authorize to `Confirm`, and then failed in `payment_resource_path`. The prefix filter enforces the invariant the comment above it was only asserting.

Re-ran both gateway-3DS journeys (charges and holds) to confirm the gate does not affect the path that legitimately uses it.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
shuklatushar226: Fixed for both Refund and RSync — a 200 carrying `status: "failed"` now builds an `ErrorResponse` from `failure_code`/`failure_message`, the same way the payments path does at the site you referenced.

Verified: a refund declined in band now surfaces as `failed` with `refund_declined` / "The refund was declined by the issuer.", where before the merchant got a failed refund and no reason at all.

deepanshu-iiitu (dismissed):

deepanshu-iiitu (approved):

deepanshu-iiitu (dismissed):

JeevaRamu0104 (dismissed):

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -0,0 +1,1926 @@
...
+pub enum PaydotcomAuthorizeLeg {
+    Create,
+    Confirm,
+}
+
+/// Decides the leg from the **request alone**, never from a previous response, so
+/// `get_url` and `get_request_body` can never disagree about which call is being made.
+///
+/// An external-MPI Authorize also carries `authentication_data`, but with `eci`/`cavv`
+/// and no `transaction_id` — so it falls through to `Create`, as it must.
```
JeevaRamu0104: This still says an external-MPI Authorize carries no `transaction_id` — the exact case the prefix filter at `:543` was added to handle, and the comment there now says the opposite. As written it reads like the gate is redundant.

```suggestion
/// An external-MPI Authorize also carries `authentication_data`, and may populate
/// `transaction_id` with its own 3DS id — the `chrg_`/`hld_` filter in
/// `pending_resource_id_from_authentication_data` is what makes it fall through to
/// `Create`, as it must.
```

shuklatushar226: Good catch — you are right that it read as self-contradicting. Applied your wording verbatim in `43e9fad08`.

The comment was written before the prefix filter existed and I did not revisit it when adding the gate, so it was left asserting the very thing the gate exists to defend against. It now points at `pending_resource_id_from_authentication_data` as the reason an external-MPI Authorize falls through to `Create`, which matches the comment on the filter itself.

Comment only, no behaviour change; fmt and `clippy -p connector-integration --all-features --all-targets` under `-D warnings` are clean.

deepanshu-iiitu (approved):

JeevaRamu0104 (approved):

JeevaRamu0104 (approved):

PR #2224, opened 2026-09-03
fix: single sanctioned source for every entropy and clock read in connector code

iemyashasvi:
## What

An audit of the connector crate found ~62 raw entropy/clock call sites across 41 files: `uuid::Uuid::new_v4`, `OffsetDateTime::now_utc`, `SystemTime::now`, `thread_rng`, raw `openssl`/`ring` RNG draws, and a hand-rolled RSA-OAEP block. This PR routes every one of them through a single sanctioned helper per entropy kind, and fixes five connectors whose behavior was genuinely nondeterministic (arbitrary picks over `HashMap` iteration order).

| kind | helper |
|---|---|
| time | `common_utils::date_time::{now, now_unix_timestamp, now_unix_millis}` (`now_unix_millis` added, derived from `now()`) |
| uuid | `common_utils::fp_utils::generate_uuid_v4` (added) |
| random string | `common_utils::crypto::generate_cryptographically_secure_random_string` |
| random bytes | `domain_types::utils::generate_random_bytes` |
| RSA-OAEP | `common_utils::crypto::RsaOaepSha256::encrypt` (boost's local PkeyCtx block replaced; parameters verified identical) |

## Why

- **Auditability**: one place per entropy kind to reason about (and, under a record/replay harness, substitute). Time-in-signature families (cybersource ×5, fiserv ×4, JWS `iat` claims) and idempotency/nonce generation were unauditable when scattered.
- **Real bugs**: payu PSync/RSync picked `.values().next()` from a `HashMap` deserialized off the wire — **which transaction id and status a sync reports was per-process random** whenever PayU returns several entries for one txnid. Same class in stax (which validation error surfaces), bluesnap (map order became the request's `metaData` array order), payload (auth pick relies on an unenforced invariant), nexixpay (`additionalData` order leaked into a stringified response field).

## Behavior preservation

Output formats are byte-for-byte identical at every redirected site (signature inputs, header shapes, millis/nanos widths, `.simple()`/uppercase transforms, paybox's substring semantics). The two `ring` RNGs feeding RSA blinding for PKCS#1 v1.5 signing are deliberately untouched — v1.5 signatures do not depend on the RNG.

## Verification

- `cargo check`/`clippy` clean; `cargo test -p connector-integration --lib` 166 passed; `cargo test -p ucs_common_utils` 27 passed.
- Census: zero remaining raw entropy patterns in `crates/integrations/connector-integration/src`.


crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
```diff
@@ -104,7 +104,9 @@ macros::create_all_prerequisites!(
         ) -> CustomResult<Vec<(String, Maskable<String>)>, IntegrationError> {
             let correlation_id = format!(
                 "{correlation_prefix}{}",
-                uuid::Uuid::new_v4().simple().to_string().to_uppercase()
+                common_utils::fp_utils::generate_uuid_v4()
+                    .replace('-', "")
```
deepanshu-iiitu, line 108: Why are we replacing - with empty char?

crates/integrations/connector-integration/src/connectors/payload.rs
```diff
@@ -262,11 +262,14 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize> Conn
...
+        // The API key is documented-same for all currencies — but that is an
+        // unenforced invariant, so pick by smallest currency code rather than
+        // HashMap iteration's per-process arbitrary entry.
         let api_key = auth
             .auths
-            .values()
-            .next()
+            .iter()
+            .min_by_key(|(currency, _)| currency.to_string())
+            .map(|(_, value)| value)
```
deepanshu-iiitu, line 272: Why is this change done?

crates/integrations/connector-integration/src/payout_connectors/deutschebank.rs
iemyashasvi: The sanctioned helper generate_uuid_v4() returns the canonical hyphenated UUID string, but this site originally used .simple() — the uuid crate's 32-hex-chars-no-hyphens format (Deutsche Bank's Correlation-Id shape). .replace('-', "") recreates the .simple() representation exactly, so the wire output is byte-identical to before; only the entropy source changed. It's format preservation, not a format change.

crates/integrations/connector-integration/src/connectors/payload.rs
iemyashasvi: `HashMap` iteration order is randomized per process (the hasher is re-seeded on every run), so `.values().next()` returns an arbitrary entry — **which currency's `api_key` gets used can differ between two runs over the same config**. The "API key is the same for all currencies" assumption is documented but unenforced: the moment a merchant configures different keys per currency, authentication becomes nondeterministic — an intermittent, unreproducible auth failure that depends on hash seeding.

The change keeps behavior identical whenever the invariant holds (every entry yields the same key either way), but makes the pick deterministic when it doesn't: the smallest currency code wins, stably, every run. `.to_string()` is used as the sort key because `Currency` doesn't derive `Ord`, and the map holds a handful of entries at most.

deepanshu-iiitu (dismissed):

gopikrishna000 (dismissed):

maverox (dismissed):

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

deepanshu-iiitu (approved):

PR #2226, opened 2026-09-03
feat(eligibility): check multiple payment methods per connector in one call

kanikac199:
## Summary

The eligibility API was built around **one payment method per call**. When a connector supports many payment methods, callers had to loop and hit the same connector once per payment method — N handshakes, N processor round-trips.

This PR reshapes the eligibility flow so a connector can be checked for **all of its payment methods in a single call** — the request carries a list of payment methods and the response returns one verdict per payment method. The connector makes as few processor calls as its API allows and fans the result across the requested payment-method list.

Before this, TAMARA_REDIRECT only existed in payment_methods.proto's payment_method_data.PaymentMethod oneof — a different enum. The composite/direct eligibility endpoints deserialize payment_method_types into payment.proto's flat PaymentMethodType, which has AFTERPAY_CLEARPAY, ALMA, AFFIRM, QWIKCILVER_WALLET, etc. — but no Tamara value at all. So any Tamara eligibility request that passed a PM list got rejected at the HTTP layer (Failed to deserialize ... Unknown enum value for PaymentMethodType: TAMARA_REDIRECT) before the handler ran.

## What changed

**Proto** (`payment.proto`, `composite_payment.proto`)
- Request: added `repeated PaymentMethodType payment_method_types`; deprecated the scalar `payment_method_type` (still honored as a one-element list).
- Response: added `repeated PaymentMethodEligibilityResult results` (new message with per-PM `payment_method_type`, `eligibility`, `error_info`, `payment_method_details`); deprecated the scalar `eligibility` / `error_info` / `payment_method_details`, now mirrored from `results[0]` for back-compat.
- Composite request: added `repeated payment_method_types`.

**Domain** (`connector_types.rs`, `types.rs`)
- `PaymentMethodEligibilityData.payment_method_types: Vec<PaymentMethodType>`.
- `PaymentMethodEligibilityResponse { results: Vec<PerPmEligibility>, status_code }`; new `PerPmEligibility` + `EligibilityErrorInfo`.
- Request conversion reads the list, falling back to the deprecated scalar.
- The payment method type is carried through the flow as the wire (proto) enum and echoed back verbatim in each result (`i32::from(...)`, no conversion) — preserving distinctions the domain enum collapses (e.g. `Credit` vs `Debit`, both of which map to domain `Card`). Precedent: `PostRiskCheckRequest.payment_connector` in the FRM flow holds a proto enum directly.
- Omitting the payment method entirely (supported via the deprecated scalar) still returns a real verdict: the empty request defaults to a single `Unspecified` entry, so the PM-agnostic connectors produce one result and the deprecated scalar `eligibility` mirrors it.
- Response generator builds `results`, fans a flow-level connector error across every requested payment method, and mirrors `results[0]` into the deprecated top-level fields.

**Connectors**
- **Tamara** and **Qwikcilver** fan their single (payment-method-agnostic) verdict across the requested payment methods; Qwikcilver also fans the resolved wallet balance/items.

**Composite service** threads `payment_method_types` through.

## Back-compat

Old single-PM callers keep working: the deprecated scalar request field is accepted (treated as a one-element list) and the deprecated scalar response fields mirror `results[0]`.

## Testing

Built and ran the `grpc-server` locally, verified end-to-end against the live Tamara sandbox. All credentials below are redacted.

### 1. Multi-PM eligibility — live Tamara sandbox (200, one verdict per PM)

```bash
CONFIG='{"config":{"Tamara":{"api_key":"<REDACTED_API_KEY>"}}}'
grpcurl -plaintext -max-time 45 \
  -H "x-connector: tamara" \
  -H "x-connector-config: $CONFIG" \
  -H "x-tenant-id: default" \
  -d @ localhost:8000 types.PaymentMethodService/Eligibility <<'JSON'
{
  "amount": { "minor_amount": "100000", "currency": "SAR" },
  "customer": { "email": {"value":"<REDACTED_EMAIL>"}, "phone_number": {"value":"<REDACTED_PHONE>"} },
  "payment_method_types": ["AFTERPAY_CLEARPAY", "ALMA", "AFFIRM"]
}
JSON
```

**Response** — one `results[]` entry per requested payment method; deprecated scalar `eligibility` mirrors `results[0]`:

```json
{
  "eligibility": "ELIGIBLE",
  "statusCode": 200,
  "typedConnectorResponse": { "value": "{\"eligible\":true,\"message\":null}" },
  "results": [
    { "paymentMethodType": "AFTERPAY_CLEARPAY", "eligibility": "ELIGIBLE" },
    { "paymentMethodType": "ALMA",              "eligibility": "ELIGIBLE" },
    { "paymentMethodType": "AFFIRM",            "eligibility": "ELIGIBLE" }
  ]
}
```

### 2. New schema is live (server reflection)

```console
$ grpcurl -plaintext localhost:8000 describe types.PaymentMethodServiceEligibilityRequest
  optional .types.PaymentMethodType payment_method_type = 6 [deprecated = true];
  repeated .types.PaymentMethodType payment_method_types = 13;

$ grpcurl -plaintext localhost:8000 describe types.PaymentMethodEligibilityResult
message PaymentMethodEligibilityResult {
  .types.PaymentMethodType payment_method_type = 1;
  .types.EligibilityStatus eligibility = 2;
  optional .types.ErrorInfo error_info = 3;
  optional .types.PaymentMethodDetails payment_method_details = 4;
}
```

### 3. Connector auth failure surfaces as a connector error

With an invalid key the call reaches the live Tamara sandbox and the genuine error is surfaced:

```console
Code: Unauthenticated
Message: unauthorized
details: CONNECTOR_ERROR_RESPONSE 401 {"message":"unauthorized", ...}
```

> Note: `test_mode: true` is intentionally omitted — in the `development` profile it redirects outbound calls to the local mock connector.

## Not in scope

- Rich reason enum on `EligibilityStatus` (kept coarse; the "why" lives in per-PM `error_info`).
- New per-PM fields beyond the current UCS model (offers/installments/dueDate).
- Per-PM-in-one-response connectors (e.g. Razorpay/M2P style) — none exist in UCS yet; the transformer pattern (find each PM in the connector response) is documented for when one lands.


crates/integrations/connector-integration/src/connectors/tamara/transformers.rs
```diff
@@ -981,10 +981,22 @@ impl TryFrom<ResponseRouterData<TamaraEligibilityResponse, Self>>
...
+        // PM-agnostic verdict fanned across every requested payment method.
+        let results = item
+            .router_data
+            .request
+            .payment_method_types
+            .iter()
+            .map(|payment_method_type| PerPmEligibility {
+                payment_method_type: *payment_method_type,
                 eligibility,
+                error_info: None,
```
Aishwariyaa-Anand, line 993: do we need errorInfo for per PM eligibility?

kanikac199: yes. We might get different errors for different pm eligibility for example with Qwikcilver a single PM can return 10532 Wallet Not Found (is_eligible=false, gateway_error_code, gateway_error_message) while another PM on the same card is fine, that error has to exist on that PM's result, not on the top-level response.

Aishwariyaa-Anand (dismissed):

crates/types-traits/domain_types/src/types.rs
```diff
@@ -6742,6 +6745,35 @@ impl ForeignFrom<EligibilityStatus> for grpc_api_types::payments::EligibilitySta
...
 /// Used by PaymentMethodService.Eligibility (new route).
+/// Builds a proto per-PM eligibility result; the payment method type is echoed verbatim.
+fn eligibility_result_to_proto(
+    result: PerPmEligibility,
+) -> grpc_api_types::payments::PaymentMethodEligibilityResult {
+    grpc_api_types::payments::PaymentMethodEligibilityResult {
+        payment_method_type: i32::from(result.payment_method_type),
+        eligibility: i32::from(grpc_api_types::payments::EligibilityStatus::foreign_from(
+            result.eligibility,
+        )),
```
AmitsinghTanwar007, line 6756: use foreignfrom this.

crates/types-traits/domain_types/src/connector_types.rs
```diff
@@ -1652,12 +1652,36 @@ pub struct PaymentMethodEligibilityData {
 
 #[derive(Debug, Clone)]
 pub struct PaymentMethodEligibilityResponse {
+    /// Per-payment-method eligibility verdicts, one per requested payment
+    /// method. Connectors that make a single PM-agnostic processor call fan the
+    /// same verdict across every requested payment method.
+    pub results: Vec<PerPmEligibility>,
```
AmitsinghTanwar007, line 1658: this can be `Vec<PMEligibility>`.

crates/integrations/connector-integration/src/connectors/tamara/transformers.rs
AmitsinghTanwar007, line 993: when this will be populated

kanikac199: error_info is populated on the failure/ineligible path only — a successful eligible verdict has no error. Here Tamara's check is amount/customer-based (PM-agnostic), so the same verdict fans across requested PMs and there's nothing to put in error_info on success. When the connector returns a per-PM rejection (e.g. Qwikcilver 10532 Wallet Not Found) it's populated with the connector's code/message so each PM surfaces why it was rejected. The deprecated top-level errorInfo mirrors results[0] for back-compat.

AmitsinghTanwar007 (approved):

Aishwariyaa-Anand (approved):

JeevaRamu0104 (approved):

PR #2227, opened 2026-09-03
fix(connector): align authorizedotnet, cybersource and adyen response parsing with hyperswitch

shuklatushar226:
## What this fixes

Three parity gaps where prism is **stricter than hyperswitch** when deserializing a
connector response, so prism hard-fails on bodies hyperswitch handles cleanly.

In shadow mode this is guaranteed to produce a diff, because the two sides parse
**byte-identical** input: the validation service's `uscHandler` does not forward
prism's connector call upstream — it replays hyperswitch's stored connector response
out of Redis (`controller/ucs/index.js:191`). So any response-parsing divergence is a
real prism defect, not connector flakiness.

When prism fails deserialization it returns `RESPONSE_DESERIALIZATION_FAILED`, whose
proto error code is not `CONNECTOR_ERROR_RESPONSE`, so hyperswitch's decoder rejects it
and the shadow gateway records `Err(String)` instead of a router data. That is why the
affected issues show ~55 keys present on the hyperswitch side and only `{"error": ...}`
on the UCS side — the router data was never built.

| Commit | Connector | Gap vs hyperswitch |
|---|---|---|
| `7bb0db179` | authorizedotnet | `SyncStatus` / `RSyncStatus` had no `#[serde(other)]` catch-all |
| `4c218b56b` | authorizedotnet | psync reason-code fields typed `Option<u8>`, overflowing past 255 |
| `05ce7cc06` | authorizedotnet | psync `transaction: None` + `E00104`/`E00053` returned an unpopulated router data |
| `814c4edc6` | cybersource | `Reason` had no catch-all; no empty-body guard in `build_error_response` |
| `fbcfe5c1b` | adyen | `AdyenErrorResponse` required `message` and `error_type` |

### authorizedotnet — `7bb0db179`

Authorize.Net documents roughly two dozen `transactionStatus` values; this connector
modelled eleven. `underReview`, `expired`, `settlementError`, `communicationError`,
`approvedReview`, `failedReview`, `updatingSettlement`, `returnedItem` and `chargeback`
all aborted the parse with `unknown variant ...`.

Hyperswitch has `#[serde(other)] Unknown` and **retains the previous status**
(`authorizedotnet/transformers.rs:2115`, `:2135`). Mirrored here by replacing the
infallible `From` conversions with `get_sync_attempt_status` / `get_rsync_refund_status`,
which take the previous status.

There is a **second, independent** failure mode behind the same four issues, fixed in
`4c218b56b`: `SyncTransactionResponse::response_code` and `response_reason_code` were typed
`Option<u8>`, which caps at 255. Authorize.Net reason codes run past that — 315 (invalid card
number), 316 (invalid expiry), 318 (duplicate transaction) — so those bodies failed to
deserialize with `invalid value: integer ..., expected u8`, producing an identical
`RESPONSE_DESERIALIZATION_FAILED`. Hyperswitch models neither field, so it never fails on them;
neither is read here, they are carried only for `typed_connector_response`. Both fixes were
needed for these issues to close.

Closes juspay/hyperswitch-cloud#23192, #23193, #23195, #23196.

### cybersource — `814c4edc6`

- `Reason` modelled only `SYSTEM_ERROR` / `SERVER_TIMEOUT` / `SERVICE_TIMEOUT`. Hyperswitch
  adds `#[serde(other)] Unknown` mapped to `attempt_status: None` (`cybersource.rs:522`).
- `build_error_response` parsed unconditionally; hyperswitch short-circuits an empty body
  first (`cybersource.rs:181`).

Deliberately **not** changed: `CybersourceServerErrorResponse` stays permissive (all fields
optional, no `deny_unknown_fields`). Hyperswitch's struct is identically permissive, so
tightening it would introduce a new divergence rather than remove one.

Refs juspay/hyperswitch-cloud#23179, #23180, #23182, #23183 — of these only #23183 is
expected to close from this change; the other three are hyperswitch-side client timeouts
(see "Not addressed here").

### adyen — `fbcfe5c1b`

`AdyenErrorResponse` required both `message` and `error_type`. Adyen omits `errorType` on
gateway errors such as `{"status":403,"errorCode":"901","message":"Invalid Merchant Account"}`
and returns `title` instead of `message` on others. Hyperswitch models both as optional,
does not model `error_type` at all, and resolves `message` → `title` → `NO_ERROR_MESSAGE`
(`adyen.rs:189`).

Also corrects this handler's deserialization-failure path, which reported the connector as
`"mifinity"` from inside the adyen connector.

Refs juspay/hyperswitch-cloud#23165.

## Verification

No unit tests are included in this change set.

Verified end-to-end against the real shadow pipeline, without connector credentials, by
seeding the validation service's Redis with the reported response bytes and driving prism
over gRPC — the harness then replays those bytes to prism exactly as it does in production:

| scenario | pre-fix binary | post-fix binary |
|---|---|---|
| unmodelled `transactionStatus: underReview` | `RESPONSE_DESERIALIZATION_FAILED` (http 200) | `PENDING`, previous status retained |
| `responseReasonCode: 315` | `RESPONSE_DESERIALIZATION_FAILED` (http 200) | `CHARGED`, raw value preserved |

Both failed with the identical error signature before the fix, which is why the second one was
only found by replaying it.

`cargo build`, `cargo clippy --all-targets` and `cargo fmt` are clean on the touched crate.

## Not addressed here

Investigated and root-caused, but **not** fixable in prism:

- **12 issues** (#23075, #23164, #23166, #23172, #23181, #23184, #23185, #23190, #23191,
  #23194, #23197, #23198) are validation-harness artefacts. Either
  hyperswitch's own connector leg died (mitmproxy synthesises a 502 via `HttpErrorHook`,
  which `forward_code.py` does not implement, so no Redis row is ever written), or the
  validation service returned `VS_07` despite a healthy hyperswitch leg. In both cases the
  UCS leg is compared against a response prism was never given. Fixes belong in
  `ucs-shadow-validation-service`.

  (#23188 and #23189 were originally in this group but are **fixed here** by `05ce7cc06`,
  which found a real prism defect behind them: the psync maintenance-code branch mirrored
  hyperswitch's `Ok(item.data)` without accounting for prism being stateless.)
- **#23171** (stripe `payment_method_token` / google_pay) is left alone deliberately. Prism's
  and hyperswitch's encrypted-google-pay branches are now structurally identical, so there is
  no parity gap left to close, and production runs the `2026.08.25.0` build which predates
  `3466a0ffe` (the 2026-08-28 Stripe Google Pay rework). It may already be fixed on `main`.
- **#23167–#23170** (adyen setup-mandate for wallets) deferred at the reporter's request.

---

https://claude.ai/code/session_01QSDJ7gES5D56vGqLz2JS4f

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
@@ -259,11 +259,16 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         Ok(WebhookDetailsResponse {
             connector_returned_payment_method_details: None,
             resource_id: Some(ResponseId::ConnectorTransactionId(transaction_id.clone())),
-            // `status` is derived from the webhook `event_type`, whose mapping never yields
-            // `SyncStatus::Unknown`, so the retained-status arm is unreachable here.
+            // A webhook carries no prior attempt status. `Unspecified` is the wire value
+            // hyperswitch reads as "retain the stored status"
+            // (hyperswitch_interfaces/src/unified_connector_service/transformers.rs:697,
+            // `PaymentStatus::Unspecified => Ok(prev_status)`), whereas any concrete status —
+            // `Pending` included — is taken at face value and would overwrite it. The webhook
+            // event mapping never yields `SyncStatus::Unknown`, so this argument is unused
+            // today; it must stay a safe no-op in case a future event variant makes it live.
             status: transformers::get_sync_attempt_status(
                 status,
-                common_enums::AttemptStatus::Pending,
+                common_enums::AttemptStatus::Unspecified,
             ),
             status_code: 200,
             mandate_reference: None,
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
@@ -259,17 +259,7 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         Ok(WebhookDetailsResponse {
             connector_returned_payment_method_details: None,
             resource_id: Some(ResponseId::ConnectorTransactionId(transaction_id.clone())),
-            // A webhook carries no prior attempt status. `Unspecified` is the wire value
-            // hyperswitch reads as "retain the stored status"
-            // (hyperswitch_interfaces/src/unified_connector_service/transformers.rs:697,
-            // `PaymentStatus::Unspecified => Ok(prev_status)`), whereas any concrete status —
-            // `Pending` included — is taken at face value and would overwrite it. The webhook
-            // event mapping never yields `SyncStatus::Unknown`, so this argument is unused
-            // today; it must stay a safe no-op in case a future event variant makes it live.
-            status: transformers::get_sync_attempt_status(
-                status,
-                common_enums::AttemptStatus::Unspecified,
-            ),
+            status: common_enums::AttemptStatus::from(status),
             status_code: 200,
             mandate_reference: None,
             connector_response_reference_id: Some(transaction_id.clone()),
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
@@ -259,7 +259,12 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         Ok(WebhookDetailsResponse {
             connector_returned_payment_method_details: None,
             resource_id: Some(ResponseId::ConnectorTransactionId(transaction_id.clone())),
-            status: common_enums::AttemptStatus::from(status),
+            // `status` is derived from the webhook `event_type`, whose mapping never yields
+            // `SyncStatus::Unknown`, so the retained-status arm is unreachable here.
+            status: transformers::get_sync_attempt_status(
+                status,
+                common_enums::AttemptStatus::Pending,
+            ),
             status_code: 200,
             mandate_reference: None,
             connector_response_reference_id: Some(transaction_id.clone()),
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
@@ -2409,17 +2409,26 @@ impl<F> TryFrom<ResponseRouterData<AuthorizedotnetPSyncResponse, Self>>
                 Ok(new_router_data)
             }
             None => {
-                // A psync response with no transaction record. Transient server-side
-                // conditions (server busy / maintenance) are not a payment failure.
+                // A psync poll with no transaction record. `E00053` ("server too
+                // busy") and `E00104` ("server in maintenance") are not payment
+                // failures - Authorize.Net is asking us to retry later.
                 //
-                // Hyperswitch returns `Ok(item.data)` unchanged here, but its psync
-                // `item.data.response` is already an `Ok(TransactionResponse)` rebuilt
-                // from the stored attempt. This connector is stateless: `router_data`
-                // is constructed with `response: Err(ErrorResponse::default())`, so
-                // returning it unchanged surfaces as `HE_00 / 500 / status unspecified`.
-                // Rebuild the same shape hyperswitch presents downstream: keep the
-                // request's connector transaction id and the incoming attempt status
-                // untouched. Any other message code is a real connector error.
+                // Hyperswitch returns `Ok(item.data)` unchanged, keeping the prior
+                // attempt status, because its psync `item.data.response` is a
+                // `TransactionResponse` seeded from the stored attempt
+                // (`construct_payment_router_data`) and `item.data.status` is the
+                // stored `payment_attempt.status`. This connector is stateless:
+                // `router_data` is built with `response: Err(ErrorResponse::default())`
+                // and no prior status, so returning it unchanged surfaces as
+                // `HE_00 / 500 / PAYMENT_STATUS_UNSPECIFIED` (a spurious error).
+                //
+                // Emit instead an `Ok(TransactionResponse)` with `AttemptStatus::Unknown`
+                // ("polled, could not infer") and `status_code = http_code` (200). It
+                // serializes to `PaymentStatus::Unspecified`, which hyperswitch maps
+                // back to the previous attempt status on psync
+                // (`unified_connector_service/transformers.rs`,
+                // `PaymentStatus::Unspecified => Ok(prev_status)`) - matching the
+                // direct connector leg. Any other message code is a real error.
                 match response.messages.message.iter().find(|m| {
                     m.code == TRANSIENT_ERROR_SERVER_BUSY || m.code == TRANSIENT_ERROR_MAINTENANCE
                 }) {
@@ -2430,6 +2439,7 @@ impl<F> TryFrom<ResponseRouterData<AuthorizedotnetPSyncResponse, Self>>
                             .get_connector_transaction_id()
                             .ok();
                         let mut new_router_data = router_data;
+                        new_router_data.resource_common_data.status = AttemptStatus::Unknown;
                         new_router_data.response = Ok(PaymentsResponseData::TransactionResponse {
                             resource_id: connector_transaction_id
                                 .clone()
--- a/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
@@ -2410,14 +2410,44 @@ impl<F> TryFrom<ResponseRouterData<AuthorizedotnetPSyncResponse, Self>>
             }
             None => {
                 // A psync response with no transaction record. Transient server-side
-                // conditions (server busy / maintenance) are not a payment failure, so
-                // mirror hyperswitch and keep the existing router data unchanged;
-                // otherwise surface the connector error while preserving the prior
-                // attempt status (as hyperswitch's `get_err_response` does).
+                // conditions (server busy / maintenance) are not a payment failure.
+                //
+                // Hyperswitch returns `Ok(item.data)` unchanged here, but its psync
+                // `item.data.response` is already an `Ok(TransactionResponse)` rebuilt
+                // from the stored attempt. This connector is stateless: `router_data`
+                // is constructed with `response: Err(ErrorResponse::default())`, so
+                // returning it unchanged surfaces as `HE_00 / 500 / status unspecified`.
+                // Rebuild the same shape hyperswitch presents downstream: keep the
+                // request's connector transaction id and the incoming attempt status
+                // untouched. Any other message code is a real connector error.
                 match response.messages.message.iter().find(|m| {
                     m.code == TRANSIENT_ERROR_SERVER_BUSY || m.code == TRANSIENT_ERROR_MAINTENANCE
                 }) {
-                    Some(_) => Ok(router_data),
+                    Some(_) => {
+                        let connector_transaction_id = router_data
+                            .request
+                            .connector_transaction_id
+                            .get_connector_transaction_id()
+                            .ok();
+                        let mut new_router_data = router_data;
+                        new_router_data.response = Ok(PaymentsResponseData::TransactionResponse {
+                            resource_id: connector_transaction_id
+                                .clone()
+                                .map(ResponseId::ConnectorTransactionId)
+                                .unwrap_or(ResponseId::NoResponseId),
+                            redirection_data: None,
+                            mandate_reference: None,
+                            connector_metadata: None,
+                            network_txn_id: None,
+                            network_txn_link_id: None,
+                            connector_response_reference_id: connector_transaction_id,
+                            incremental_authorization_allowed: None,
+                            status_code: http_code,
+                            splits: None,
+                            payment_account_reference: None,
+                        });
+                        Ok(new_router_data)
+                    }
                     None => Ok(Self {
                         response: Err(get_err_response(http_code, response.messages)),
                         ..router_data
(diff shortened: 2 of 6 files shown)


crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -2871,23 +2876,33 @@ pub struct SyncTransactionResponse {
...
+            AttemptStatus::Charged
+        }
+        SyncStatus::AuthorizedPendingCapture => AttemptStatus::Authorized,
+        SyncStatus::Declined => AttemptStatus::AuthenticationFailed,
+        SyncStatus::Voided => AttemptStatus::Voided,
+        SyncStatus::CouldNotVoid => AttemptStatus::VoidFailed,
+        SyncStatus::GeneralError => AttemptStatus::Failure,
+        SyncStatus::RefundSettledSuccessfully | SyncStatus::RefundPendingSettlement => {
+            AttemptStatus::Charged
+        }
```
AmitsinghTanwar007: how refund settled can be charged?

shuklatushar226: Fair question, but this mapping is pre-existing and unchanged by this PR. The only diff on that arm is `Self::Charged` -> `AttemptStatus::Charged`, a mechanical rename from converting the `impl From<SyncStatus> for AttemptStatus` into a free function that can take `prev_status` (needed for the new `Unknown` arm).

Three reasons to leave it as is:

1. **It is verbatim hyperswitch parity.** `crates/hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs:2109-2111` maps the same two variants to `AttemptStatus::Charged`. This PR exists specifically to remove UCS/hyperswitch shadow diffs, so changing it here would create a new one.

2. **This is the payment attempt status, not the refund status.** A refund does not un-charge the payment - the attempt genuinely was charged, and the refund is a separate object with its own lifecycle. Authorize.Net reports the refund state on the underlying transaction, which is why it surfaces in psync at all.

3. **The refund side does map correctly**, in `get_rsync_refund_status` in the same file: `RSyncStatus::RefundSettledSuccessfully -> RefundStatus::Success` and `RefundPendingSettlement -> RefundStatus::Pending`. So the refund state is tracked accurately, just on the refund object rather than the attempt.

I did check whether `AttemptStatus::AutoRefunded` would be a better target - it exists, but it means an automatic system-initiated refund rather than a settled merchant refund, and adopting it here would diverge from hyperswitch. If we think both sides should change, happy to raise it as a coordinated change across the two repos rather than folding it into this parity PR.

shuklatushar226:
Re-verified mechanically rather than by eye. I extracted the `match` block from `git show origin/main:<file>` and from the working tree, normalised `Self::` -> `AttemptStatus::`, and compared arm by arm:

```
ARM-BY-ARM, origin/main vs this branch
SAME  SettledSuccessfully | CapturedPendingSettlement    main=Charged               branch=Charged
SAME  AuthorizedPendingCapture                          main=Authorized            branch=Authorized
SAME  Declined                                          main=AuthenticationFailed  branch=AuthenticationFailed
SAME  Voided                                            main=Voided                branch=Voided
SAME  CouldNotVoid                                      main=VoidFailed            branch=VoidFailed
SAME  GeneralError                                      main=Failure               branch=Failure
SAME  RefundPendingSettlement | RefundSettledSuccessfully  main=Charged            branch=Charged   <-- flagged arm
SAME  FDSAuthorizedPendingReview | FDSPendingReview      main=Unresolved            branch=Unresolved

Every pre-existing arm identical : True
branch minus main                : ['Unknown']
branch vs hyperswitch            : IDENTICAL variant set (12 each)
```

The flagged arm, all three sources side by side:

```rust
// origin/main - impl From<SyncStatus> for AttemptStatus
SyncStatus::RefundSettledSuccessfully | SyncStatus::RefundPendingSettlement => {
    Self::Charged
}

// this branch - fn get_sync_attempt_status, transformers.rs:2928
SyncStatus::RefundSettledSuccessfully | SyncStatus::RefundPendingSettlement => {
    AttemptStatus::Charged
}

// hyperswitch - hyperswitch_connectors/src/connectors/authorizedotnet/transformers.rs:2109
SyncStatus::RefundSettledSuccessfully | SyncStatus::RefundPendingSettlement => {
    enums::AttemptStatus::Charged
}
```

So the only difference this PR introduces on that arm is the type qualifier, a mechanical consequence of turning the `impl From` into a free function that can take `prev_status` (required by the new `Unknown` arm). `Unknown` is the only variant added.

And the refund state is not lost - it is tracked on the refund object, where it belongs:

```rust
// prism transformers.rs:2980  ==  hyperswitch transformers.rs:2130
RSyncStatus::RefundSettledSuccessfully => RefundStatus::Success,
RSyncStatus::RefundPendingSettlement   => RefundStatus::Pending,
```

On the `AutoRefunded` alternative: the variant exists in `AttemptStatus`, and other connectors do use it (powertranz, flywire for `Reversed`), but hyperswitch's authorizedotnet references it **zero times** - `Charged` is its deliberate choice here. Adopting it in prism alone would diverge from hyperswitch and open a fresh `router` diff on authorizedotnet psync, which is exactly what this PR is closing.

Happy to raise it as a coordinated change across both repos if we agree the shared mapping is wrong - it just should not land inside a parity PR.

crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
```diff
@@ -259,7 +259,12 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
         Ok(WebhookDetailsResponse {
             connector_returned_payment_method_details: None,
             resource_id: Some(ResponseId::ConnectorTransactionId(transaction_id.clone())),
-            status: common_enums::AttemptStatus::from(status),
+            // `status` is derived from the webhook `event_type`, whose mapping never yields
+            // `SyncStatus::Unknown`, so the retained-status arm is unreachable here.
+            status: transformers::get_sync_attempt_status(
+                status,
+                common_enums::AttemptStatus::Pending,
```
AmitsinghTanwar007: in hyperswitch is pending the prev status always

gopikrishna000: no if ucs gives unspecified on psync , then only it gets prev status., if ucs sends pending, then i believe it updates db also to pending (from watver prev status)

shuklatushar226: You're right, and this was a real latent bug in my change - fixed in `bb8bb465b`.

I verified @gopikrishna000's reading from the code, and it holds. Hyperswitch treats `Unspecified` as "retain the stored status" and takes every concrete status at face value:

```rust
// hyperswitch_interfaces/src/unified_connector_service/transformers.rs
:689  payments_grpc::PaymentStatus::Pending      => Ok(Self::Pending),     // written through
:697  payments_grpc::PaymentStatus::Unspecified  => Ok(prev_status),        // retained

// :531, error path
match response.status() {
    payments_grpc::PaymentStatus::Unspecified => None,
    _ => Some(AttemptStatus::foreign_try_from((response.status(), prev_status))?),
}
```

So `Pending` was exactly the wrong sentinel - had the arm ever been reached it would have downgraded a `Charged` attempt to `Pending`. Changed to `AttemptStatus::Unspecified`, which is wire-equivalent to the `AttemptStatus::Unknown` used in `449419886` (`domain_types/src/types.rs:7701-7702` maps both to `PaymentStatus::Unspecified`); I picked `Unspecified` at this call site because the accurate statement is "this parameter has no value", not "we polled and could not infer".

On the reachability, to be explicit about why this was latent rather than live: `From<AuthorizedotnetWebhookEvent> for SyncStatus` covers all eight event variants concretely and `AuthorizedotnetWebhookEvent` has no `#[serde(other)]`, so `SyncStatus::Unknown` is unreachable from the webhook path and the argument is unused today. It needed to be a safe no-op anyway, for whenever someone adds a ninth event.

And confirming no observable change - all 8 webhook events still resolve to the same `AttemptStatus` as `origin/main`:

```
webhook event                   SyncStatus                  main        branch
AuthorizationCreated            AuthorizedPendingCapture    Authorized  Authorized   SAME
CaptureCreated                  CapturedPendingSettlement   Charged     Charged      SAME
AuthCapCreated                  CapturedPendingSettlement   Charged     Charged      SAME
PriorAuthCapture                SettledSuccessfully         Charged     Charged      SAME
VoidCreated                     Voided                      Voided      Voided       SAME
RefundCreated                   RefundSettledSuccessfully   Charged     Charged      SAME
CustomerCreated                 SettledSuccessfully         Charged     Charged      SAME
CustomerPaymentProfileCreated   SettledSuccessfully         Charged     Charged      SAME
```

One thing worth flagging for anyone reading later: there is also an `impl From<AuthorizedotnetWebhookEvent> for AttemptStatus` at `transformers.rs:3419` which looks like it would let the webhook path skip the `SyncStatus` hop entirely. It is **not** interchangeable - it maps `RefundCreated -> PartialCharged` whereas the route actually in use gives `Charged`. Switching to it would be a silent behaviour change, so I left it alone.

AmitsinghTanwar007 (dismissed):

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -2864,30 +2909,44 @@ pub struct SyncTransactionResponse {
...
-                Self::Charged
-            }
-            SyncStatus::FDSPendingReview | SyncStatus::FDSAuthorizedPendingReview => {
-                Self::Unresolved
-            }
+/// An unrecognised `transactionStatus` must not fail the sync: hyperswitch keeps the
+/// previous attempt status and logs, so mirror that instead of erroring out.
+pub fn get_sync_attempt_status(
+    transaction_status: SyncStatus,
+    prev_status: AttemptStatus,
```
gopikrishna000: there is no prev status in ucs right, if so can we use same try from as befroe, jsut that on unkniown, we need to give unspecified, and the hs does the prev status handling

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
```diff
@@ -2919,14 +2980,24 @@ pub struct AuthorizedotnetRSyncResponse {
...
-            RSyncStatus::RefundSettledSuccessfully => Self::Success,
-            RSyncStatus::RefundPendingSettlement => Self::Pending,
-            RSyncStatus::Declined | RSyncStatus::GeneralError | RSyncStatus::Voided => {
-                Self::Failure
-            }
+/// Refund-sync analogue of [`get_sync_attempt_status`] — an unknown `transactionStatus`
+/// retains the previous refund status rather than failing deserialization.
+pub fn get_rsync_refund_status(
+    transaction_status: RSyncStatus,
+    prev_status: RefundStatus,
```
gopikrishna000: same here, no prev status sent for ucs req rigtht if so, ucs give unknown or unsepcified status and the caller handle the prev status
So use the prev From impl with unknown mapping
Confirm for refund sync if ucs sends unspecified, whether hs takes prev refund status or not

crates/integrations/connector-integration/src/connectors/authorizedotnet.rs
```diff
@@ -259,7 +259,17 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
             resource_id: Some(ResponseId::ConnectorTransactionId(transaction_id.clone())),
-            status: common_enums::AttemptStatus::from(status),
+            // A webhook carries no prior attempt status. `Unspecified` is the wire value
+            // hyperswitch reads as "retain the stored status"
+            // (hyperswitch_interfaces/src/unified_connector_service/transformers.rs:697,
+            // `PaymentStatus::Unspecified => Ok(prev_status)`), whereas any concrete status —
+            // `Pending` included — is taken at face value and would overwrite it. The webhook
+            // event mapping never yields `SyncStatus::Unknown`, so this argument is unused
+            // today; it must stay a safe no-op in case a future event variant makes it live.
+            status: transformers::get_sync_attempt_status(
```
gopikrishna000: same here

crates/integrations/connector-integration/src/connectors/authorizedotnet/transformers.rs
shuklatushar226: Yes - you are right on both counts, and this was a real bug, not just extra plumbing. Done in `a8e8ae0ac`.

I checked whether `prev_status` was meaningful in UCS and it is not: every `PaymentFlowData` is constructed with `status: AttemptStatus::Pending` - for the psync request specifically at `domain_types/src/types.rs:5969`. So the "previous status" I was retaining was always a fabricated `Pending`, which serializes to `PaymentStatus::Pending` and gets written through by hyperswitch (`transformers.rs:689`). On any unmodelled `transactionStatus` that would have **silently downgraded a Charged attempt to Pending** - worse than the deserialization failure it replaced, since at least that one was loud.

So, exactly as you described:

- restored the infallible `From<SyncStatus> for AttemptStatus` and `From<RSyncStatus> for RefundStatus`, `prev_status` parameter gone
- `SyncStatus::Unknown => AttemptStatus::Unspecified` -> `PaymentStatus::Unspecified` -> `Ok(prev_status)` at `unified_connector_service/transformers.rs:697`
- `RSyncStatus::Unknown => RefundStatus::Unknown` -> proto `RefundStatus::Unspecified` (`types.rs:7787`) -> `Ok(prev_status)` at `router/src/core/unified_connector_service/transformers.rs:7994`

Retention now happens on the side that actually holds the attempt, and this matches what you did for the maintenance-code branch in `449419886`. It also removes the webhook placeholder argument that @AmitsinghTanwar007 flagged, rather than just making it a safer value - the parameter no longer exists.

Verified on the replay rig (seeding the VS redis with an authorize.net psync body carrying `transactionStatus: underReview` and driving prism over gRPC):

```
before:  "status": "PENDING"      <- concrete status; hyperswitch writes it through
after:   <field absent>           <- PAYMENT_STATUS_UNSPECIFIED; proto3 omits the zero value,
                                     hyperswitch maps it back to the stored status
```

All pre-existing arms verified unchanged in both conversions; `Unknown` is the only variant added to either. Net 24 lines smaller. clippy/fmt clean, 166 tests pass.

AmitsinghTanwar007 (approved):

gopikrishna000 (approved):

PR #2228, opened 2026-09-03
feat(framework): per-flow terminal status enforcement framework

Aishwariyaa-Anand:
## Motivation

Today, connector status mapping has no flow-level guardrails. Every connector
implements a single `From<ConnectorStatus> for AttemptStatus` (or a shared helper
like `get_status()`) that is reused across all flows indiscriminately. This creates
two classes of silent bugs:

1. **Cross-flow status leaks** — nothing prevents a Capture response from producing
   `Voided`, or a Void response from producing `CaptureFailed`. The framework accepts
   whatever `AttemptStatus` the connector sets, with no validation that it makes sense
   for the flow that was actually executed.

2. **Missing terminal coverage** — a connector can fail to map its success status to
   a terminal like `Charged` for a given flow entirely. For example, if a connector's
   shared `From` impl maps its capture-success status to `Pending` (perhaps correct
   for Authorize), every capture for that connector silently never reaches `Charged`.
   There is no compile-time or runtime check to catch this.

This PR introduces the foundational framework to make both classes of bugs
impossible — at compile time for terminal correctness, and at runtime for
cross-flow leaks.

## Summary

Introduces the foundational types and macros for per-flow status enforcement — a
framework that prevents connectors from producing incorrect `AttemptStatus` values
for a given flow (e.g. a Capture returning `Voided`, or an Authorize that can never
produce `Charged`).

### What's in this PR

**`domain_types::flow_status`** (new module)
- `FlowStatusRules` trait — implemented on flow marker types (`Authorize`, `Capture`,
  `Void`, `PSync`, `IncrementalAuthorization`, `RepeatPayment`, `SetupMandate`,
  `VoidPC`, and refund/dispute/payout flows). Declares `TERMINAL_SUCCESS_SET`,
  `TERMINAL_FAILURE_SET`, and `ALLOWED` for each flow.
- `ConnectorTerminalMapping<Flow>` trait — per-connector per-flow mapping. Declares
  `success_connector_status()`, `failure_connector_status()`, and `map_attempt_status()`.
- `const_contains` / `const_all_in` helpers — compile-time slice membership checks.
- `assert_flow_rules!` — fires at compile time to verify each flow's terminal sets
  are subsets of its `ALLOWED` set.

**`domain_types::status_mapping`** (new module)
- `impl_flow_status_mapping!` — generates `ConnectorTerminalMapping<Flow>` with const
  assertions that the declared success/failure targets land in the correct terminal sets.
- `impl_flow_status_mapping_ctx!` — context-aware variant for connectors where the
  mapping depends on an extra response field (e.g. Nexinets `NexinetsTransactionType`
  disambiguates `Preauth` → `Authorized` vs `Debit` → `Charged`).
- `assert_terminal_mapping!` — test-time macro that calls `success_connector_status()`
  and `failure_connector_status()`, maps them through `map_attempt_status()`, and
  asserts they land in the correct terminal sets.

**`PaymentFlowData::set_status_for_flow<F>()`**
- Consuming builder that validates a given `AttemptStatus` is in `F::ALLOWED` before
  writing it. Returns `Err` with a descriptive message if the status is disallowed.
- Used in the Stripe generic `TryFrom<ResponseRouterData<PaymentIntentResponse>>` as
  a proof-of-concept that the framework integrates correctly with existing transformer
  patterns.

**Stripe** (`stripe.rs`)
- `impl_flow_status_mapping!` added for: `Authorize`, `PSync`, `Void`, `Capture`,
  `IncrementalAuthorization`, `RepeatPayment`.
- The generic TryFrom in `stripe/transformers.rs` uses `set_status_for_flow::<F>` to
  validate whatever status the existing `From<StripePaymentStatus>` impl computes.

**Nexinets** (`nexinets.rs`)
- `impl_flow_status_mapping_ctx!` added for the Authorize flow, covering the
  `NexinetsTransactionType` context dependency.

### Enforcement layers in this PR

| Layer | Mechanism | When it fires |
|---|---|---|
| Compile time | `assert_flow_rules!` — terminal sets ⊆ ALLOWED | Always |
| Compile time | `impl_flow_status_mapping!` const assertions | Always |
| Runtime | `set_status_for_flow` validates against ALLOWED | On every Stripe TryFrom |
| Test time | `assert_terminal_mapping!` | When test is written |

### What this PR does NOT do

- `PaymentFlowData::status` remains public — direct assignment is still possible.
  Full enforcement requires making it private (subsequent PR).
- `get_status()` / `From<ConnectorStatus> for AttemptStatus` are not deleted.
  They remain the actual mapping source; `set_status_for_flow` only validates their
  output. Migration to `map_attempt_status` as the sole mapping source is a subsequent PR.
- Only Stripe and Nexinets are piloted. The remaining ~100 connectors will be migrated
  flow by flow in subsequent PRs.

### Subsequent PRs

1. Make `PaymentFlowData::status` private — breaks direct assignment, forces all
   connectors through `set_status_for_flow`.
2. Replace `get_status()` / `From` with `map_attempt_status` in each connector's
   TryFrom, then delete the old mapping functions.
3. Add `ConnectorTerminalMapping<Flow>` as a supertrait of `PaymentCapture`,
   `PaymentVoidV2`, etc. — compile-time enforcement that every connector declares
   its terminal mapping for each implemented flow.
4. Migrate remaining connectors.


gopikrishna000 (dismissed):

Sakilmostak (dismissed):
payout changes looks good!

gopikrishna000 (dismissed):

AmitsinghTanwar007 (dismissed):

crates/types-traits/domain_types/src/connector_types.rs
```diff
@@ -851,6 +851,31 @@ impl PaymentFlowData {
         self.status = status;
     }
 
+    /// Set `status` for a specific flow, validating at runtime that the status is in
+    /// the flow's `ALLOWED` set.  Prefer this over `set_status` in connector response
+    /// handlers — it prevents cross-flow status leaks (e.g. Capture returning `Voided`).
+    ///
+    /// Returns `Err` if `status` is not allowed for `F`.
+    pub fn set_status_for_flow<F: crate::flow_status::FlowStatusRules>(
```
JeevaRamu0104, line 859: `set_status_for_flow` has no callers, and `stripe/transformers.rs`

crates/types-traits/domain_types/src/flow_status.rs
```diff
@@ -0,0 +1,518 @@
...
+        AttemptStatus::Pending,
+    ];
+}
+
+impl FlowStatusRules for connector_flow::SetupMandate {
+    const NAME: &'static str = "SetupMandate";
+    const TERMINAL_SUCCESS_SET: &'static [AttemptStatus] = &[AttemptStatus::Charged];
+    const TERMINAL_FAILURE_SET: &'static [AttemptStatus] =
+        &[AttemptStatus::Failure, AttemptStatus::AuthorizationFailed];
+    const ALLOWED: &'static [AttemptStatus] = &[
```
JeevaRamu0104, line 168: `nexixpay/transformers.rs:2084-2088` maps its SetupMandate response to `AuthenticationSuccessful` / `AuthenticationFailed` — neither is in this ALLOWED set, and `AuthenticationFailed` is missing from `TERMINAL_FAILURE_SET` as well. Once enforcement is on, a 3DS-failed NexiXPay mandate setup errors instead of reaching a terminal. Can these sets be derived from what connectors emit today? Capture and Void look the same way.

crates/integrations/connector-integration/src/connectors/stripe.rs
```diff
@@ -112,10 +197,50 @@ impl<T: PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize>
...
+    source:    stripe::StripePaymentStatus,
+    success:   Succeeded                => Charged,
+    failure:   Failed                   => Failure,
+    {
+        RequiresCapture        => Authorized,
+        RequiresPaymentMethod  => Failure,
+        Canceled               => Voided,
+        Processing             => Authorizing,
+        RequiresCustomerAction => AuthenticationPending,
+        RequiresConfirmation   => ConfirmationAwaited,
```
JeevaRamu0104, line 238: `Canceled => Voided` and `RequiresConfirmation => ConfirmationAwaited`, but `RepeatPayment::ALLOWED` contains neither. The macro const-asserts only the `success:`/`failure:` targets, so arms inside the block go unchecked — the pilot already carries the bug class the framework is meant to stop. Asserting every target is in `ALLOWED` would catch it here and across the ~100-connector migration.

JeevaRamu0104 (changes requested):

crates/types-traits/domain_types/src/connector_types.rs
Aishwariyaa-Anand: This PR establishes the framework (macro declarations + ConnectorTerminalMapping impls). The enforcement  making PaymentFlowData::status private is through set_status_for_flow - will be in the follow-up PR.

crates/types-traits/domain_types/src/flow_status.rs
Aishwariyaa-Anand: have added authN status in setupMandate

crates/integrations/connector-integration/src/connectors/stripe.rs
Aishwariyaa-Anand: fixed this and also added additional assert

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #2237, opened 2026-09-04
ci(sdk): parallelize SDK test suites into matrix jobs

pixincreate:
## Description
Splits the sequential `SDK Tests` job (~72 min) into a build job plus 9 parallel test jobs. Expected wall clock: ~40 min on a cold cache, ~20 min on a warm cache.

- `sdk-build` builds all SDK binaries once (same single cargo command as before) and shares them through the actions cache.
- `sdk-test` runs as a matrix: gRPC and FFI suites split per language (Rust, JavaScript, Python, Kotlin), plus the mock suite. Each job restores the prebuilt binaries and runs only its own tests, so no test job compiles Rust.

Supporting changes:
- `sdk/Makefile`: new `test-grpc-single GRPC_SDK=<lang>` target — starts the test server and runs one language's gRPC suite (the old target always ran all four).
- `sdk/python/Makefile`, `sdk/java/Makefile`: use the prebuilt `uniffi-bindgen` when `FFI_SKIP_BUILD=1` instead of always calling cargo. Without this, test jobs (which skip the Rust build cache) would rebuild it from source. Local workflows unchanged.
- Binaries cache now also stores `smoke-test-webhook`; key bumped to `sdk-v2-` so stale entries are not reused.
- `ci-gate` now checks both `sdk-build` and `sdk-test`.

Trade-off: total runner minutes go up (each job repeats checkout/setup), wall clock halves.

## Motivation and Context
`SDK Tests` is the long pole of every CI run: build (~22 min cold), gRPC (~22 min), and FFI (~18 min) run one after another in a single job. The suites are independent once the binaries exist, so they can run in parallel.

### Additional Changes

## How did you test it?
- Workflow YAML validated; Makefile recipes checked with dry runs and `bash -n`.
- Audited all make targets reachable from the test jobs to confirm none invoke cargo when the skip flags are set and binaries are present.
- Full validation comes from this PR's own CI run (`.github/**` changes trigger the SDK pipeline) once marked ready for review.


gopikrishna000 (dismissed):

XyneSpaces:
## Review findings (`comment`)

**Classification:** ci-cd / developer-experience

**Summary:** Splits the monolithic SDK Tests job into `sdk-build` plus a parallel `sdk-test` matrix. The design is sound and avoids the external-services feature-unification recompile by restoring prebuilt binaries. No security, secret, or workflow-injection issues found; `pull_request_target` is not introduced.

- **[should-fix]** `.github/workflows/ci.yml:1124` — `fail-on-cache-miss: true` in `sdk-test` means a cache eviction between `sdk-build` and the matrix jobs will fail every test job. Consider uploading the binaries as workflow artifacts or adding a lightweight rebuild fallback.

- **[should-fix]** `.github/workflows/ci.yml:902` — The new `sdk-build` job runs only on non-draft PRs when SDK/core/proto files change. Ensure draft PRs marked ready for review reliably re-trigger the full `sdk-build`/`sdk-test` chain; the `ci-gate` now depends on it.

- **[should-fix]** `sdk/Makefile:312` — `test-grpc-single` starts and stops the grpc-server inside each matrix job. If a test crashes, verify the `_cleanup-grpc` / kill path always runs so subsequent matrix jobs are not blocked by stale server processes on a shared runner.

- **[should-fix]** `.github/workflows/ci.yml:1161` — `ci-gate` now depends on both `sdk-build` and `sdk-test`. Confirm the status-evaluation script treats `skipped` `sdk-build`/`sdk-test` correctly for connector-only PRs, otherwise unrelated PRs could be blocked.

*(Note: this agent can only post PR-level comments; line-level inline comments are not available in its current toolset.)*

shuklatushar226 (approved):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #1423, opened 2026-05-26
feat(adyen): support AliPayHk wallet

10X-GRACE:
## Summary

Implement **Wallet** flow for **Adyen** AliPayHk payment method type.

This PR registers `(Wallet, AliPayHk)` in `ADYEN_SUPPORTED_PAYMENT_METHODS` (gap discovered during re-validation; transformer wiring `WalletData::AliPayHkRedirect -> AdyenPaymentMethod::AliPayHk` was already present at `crates/integrations/connector-integration/src/connectors/adyen/transformers.rs:1554`).

## Changes

- Added `(PaymentMethod::Wallet, PaymentMethodType::AliPayHk)` row to `ADYEN_SUPPORTED_PAYMENT_METHODS` with mandates=NotSupported, refunds=Supported, mirroring the recent MbWay/WeChatPay registrations.

## Files Modified

- `crates/integrations/connector-integration/src/connectors/adyen.rs`

## gRPC Test Results

**Status: PASS** — Adyen sandbox returned `AUTHENTICATION_PENDING` (HTTP 200) with a `https://checkoutshopper-test.adyen.com/checkoutshopper/alipayHk?redirectData=...` redirect URL for HKD/HK — the documented AliPay HK redirect customer-action flow.

<details>
<summary>grpcurl Authorize call (credentials redacted)</summary>

```
grpcurl -plaintext \
  -H 'x-connector: adyen' \
  -H 'x-auth: body-key' \
  -H "x-api-key: <REDACTED>" \
  -H "x-key1: <REDACTED>" \
  -H 'x-merchant-id: test_merchant' \
  -H 'x-tenant-id: default' \
  -H "x-request-id: test_alipayhk_<ts>" \
  -H "x-connector-request-reference-id: alipayhk_<ts>" \
  -d '{
    "merchant_transaction_id": "alipayhk_<ts>",
    "amount": {"minor_amount": 1000, "currency": "HKD"},
    "payment_method": {"ali_pay_hk_redirect": {}},
    "customer": {"email": {"value": "test@juspay.in"}, "id": "cust_<ts>"},
    "address": {
      "billing_address": {"country_alpha2_code": "HK", "first_name": {"value": "Test"}, "last_name": {"value": "Buyer"}, "email": {"value": "test@juspay.in"}}
    },
    "auth_type": "NO_THREE_DS",
    "return_url": "https://hyperswitch.io/return",
    "capture_method": "AUTOMATIC",
    "browser_info": {
       "language": "en-US",
[...]

PR #2240, opened 2026-09-05
fix(adyen): support Wallet payment methods in SetupMandate (Apple Pay / Google Pay)

10X-GRACE:
## Closes

- Fixes juspay/hyperswitch-cloud#23167 -- https://github.com/juspay/hyperswitch-cloud/issues/23167 (adyen / setup_mandate / apple_pay)
- Fixes juspay/hyperswitch-cloud#23168 -- https://github.com/juspay/hyperswitch-cloud/issues/23168 (adyen / setup_mandate / google_pay)
- Fixes juspay/hyperswitch-cloud#23169 -- https://github.com/juspay/hyperswitch-cloud/issues/23169 (adyen / setup_mandate / google_pay)
- Fixes juspay/hyperswitch-cloud#23170 -- https://github.com/juspay/hyperswitch-cloud/issues/23170 (adyen / setup_mandate / apple_pay)

## Summary

- Adyen's `SetupMandate` transformer only handled `PaymentMethodData::Card` -- every other variant (including decrypted Apple Pay / Google Pay wallets) fell through to a catch-all `IntegrationError::NotImplemented("payment method")` (gRPC code 12).
- This mirrors how HS's own Adyen connector handles it: HS reuses its Authorize builder for SetupMandate rather than duplicating payment-method logic, since Adyen's API doesn't distinguish the two flows beyond `recurringProcessingModel`.
- Added a `SetupMandate x Wallet` builder for `SetupMandateRequest<T>`, reusing the existing `*_for_setup_mandate` helper functions (already used by the Card builder) plus the `payment_method`/`mpi_data` (3DS liability-shift) construction already used by the Authorize `x Wallet` builder.
- To share the wallet-to-`AdyenPaymentMethod` conversion between Authorize and SetupMandate, its signature was narrowed from a flow-specific `&RouterDataV2<Authorize, ...>` to the flow-agnostic `&PaymentFlowData` -- the only field it ever read off that router data (billing phone number, for MbWay) lives on `PaymentFlowData` already, so no flow-specific data was actually needed.

Confirmed via ClickHouse as `grpc_code=12 Unimplemented` / `NOT_IMPLEMENTED: payment method` for `adyen/setup_mandate` with `apple_pay` and `google_pay`.

### Self-review fix: shopper_reference consistency

An initial version of the Wallet arm took `shopper_reference` from `get_recurring_processing_model_for_setup_mandate`'s return value, which always derives `"{merchant_id}_{customer_id}"` and never looks at `connector_customer`. The sibling Card arm ignores that same value and instead prefers the already-known `connector_customer`, falling back to the derived format only when it's absent. Left uncorrected, a Wallet and a Card SetupMandate for the same customer would have sent Adyen two different `shopperReference` values. Fixed to match the Card arm exactly.

## Investigated but not changed: #23231, #23234 (ali_pay_hk / authorize)

`WalletData::AliPayHkRedirect` support for Adyen's `Authorize` flow was already added in PR #1423 (Jun 2026): it's mapped to `AdyenPaymentMethod::AliPayHk`, the capability is declared supported in `adyen.rs`, and the Authorize wallet builder already calls this successfully with no other gate in between. The reported occurrences (Aug 19 - Sep 4) are *after* that fix shipped, and unlike #23169/#23170 these two issues were only "likely" classified in the original triage, with no confirmed `grpc_code`/`error_code` log row backing them.

No implementation gap was found for this path in the current codebase. Recommend re-checking the live UCS error payload for #23231/#23234 (e.g. via ClickHouse/Loki) before assuming a missing-payment-method cause -- it may be a different, unrelated failure now that the underlying wallet support already exists.

## Test plan

- [x] `cargo build --package connector-integration` -- clean
- [x] `cargo clippy` -- clean
- [x] `cargo fmt --check` -- clean
- [x] Existing `adyen::test::tests::authorize::*` tests still pass (confirms the `AdyenPaymentMethod` signature refactor didn't change Authorize behavior)
- [x] **Verified live against Adyen's real API**: built `grpc-server` locally and sent a `SetupRecurring` (SetupMandate) request with a Google Pay wallet via `grpcurl` (deliberately invalid API key/merchant account) -- now returns a genuine Adyen 401 (`"HTTP Status Response - Unauthorized"`) instead of `NOT_IMPLEMENTED`, confirming the request reaches Adyen correctly.


XyneSpaces:
[should-fix] This adds Wallet handling for Adyen `SetupMandate`, but the PR only updates transformer code, examples, and generated docs; it does not add or update a connector spec/global-suite test for the new Apple Pay / Google Pay setup-mandate path. Add coverage under `crates/internal/ucs-connector-tests/` so the new request mapping and MPI-data branches are exercised.

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2241, opened 2026-09-05
fix(stripe): support reused PaymentMethod token in SetupMandate

10X-GRACE:
## Closes

- Fixes juspay/hyperswitch-cloud#23265 -- https://github.com/juspay/hyperswitch-cloud/issues/23265 (stripe / setup_mandate / card credit)
- Fixes juspay/hyperswitch-cloud#23268 -- https://github.com/juspay/hyperswitch-cloud/issues/23268 (stripe / setup_mandate / card debit)
- Fixes juspay/hyperswitch-cloud#23273 -- https://github.com/juspay/hyperswitch-cloud/issues/23273 (stripe / setup_mandate / card debit)

## Summary

Root-caused with production ClickHouse evidence (`hyperswitch_prod.prism_connector_events`):

1. HS calls `PaymentMethodToken` first, tokenizing the card via Stripe's `/v1/payment_methods` API and getting back a `pm_xxx` id.
2. HS then calls `SetupMandate` with `payment_method_data` as `PaymentMethodData::PaymentMethodToken { token: pm_xxx }`.
3. connector-service's `SetupMandateRequest` builder only had a match arm for `PaymentMethodData::Card` -- `PaymentMethodToken` fell through to the same catch-all as every other non-card variant, returning `grpc_code=12 Unimplemented` / `"This feature is not implemented: Selected payment method through stripe"` **before ever attempting a Stripe API call**.

Stripe's API has two distinct ways to reference a payment method: create one inline via `payment_method_data[type]=...` (what the existing Card arm does), or reuse an existing PaymentMethod by id via a top-level `payment_method=pm_xxx` field. connector-service's Authorize path (`PaymentIntentRequest`) already supports the latter for its own `PaymentMethodToken` handling; `SetupMandateRequest` had no equivalent field.

- Added `payment_method: Option<Secret<String>>` to `SetupMandateRequest` (mirroring `PaymentIntentRequest`).
- Made `payment_data: Option<StripePaymentMethodData<T>>` so it can be omitted for the reused-token case.
- Branch on `PaymentMethodData::PaymentMethodToken` before the existing card-only `StripePaymentMethodData::try_from` call, routing the token through the new field with `payment_method_types` left unset (matching how Authorize skips a data-specific `payment_method_types` for a reused PaymentMethod).

### Also included: optional `object`/`metadata` on Stripe response structs

Made `PaymentIntentResponse`/`SetupMandateResponse`'s `object` and `metadata` fields `Option<...>` instead of required, matching HS's own equivalents. This is a real, independently-confirmed parity gap (HS has always treated these as optional) -- but ClickHouse showed it is **not** what was actually failing for these three issues, so it's kept as a genuine hardening fix, with the `PaymentMethodToken` support above being the actual root-cause fix.

### Self-review fix: Apple Pay / Google Pay-typed tokens

An initial version of the `PaymentMethodToken` arm sent *every* reused token through the bare top-level `payment_method` field, regardless of what was actually tokenized. Authorize's own handling of the identical domain type branches on `token_payment_method_type`: an Apple Pay / Google Pay decrypted-token payload was exchanged for a Stripe *token* via the Tokens API and must go nested as `payment_method_data[card][token]` (`StripePaymentMethodData::CardTokenPayment`); only a plain saved-PaymentMethod-id token (`token_payment_method_type: None`, the case in #23265/#23268/#23273) uses the bare field. SetupMandate now mirrors that same split.

## Test plan

- [x] `cargo build --package connector-integration` -- clean
- [x] `cargo clippy` -- clean
- [x] `cargo fmt --check` -- clean
- [x] **Verified live against Stripe's real API**: built `grpc-server` locally and sent a `SetupMandate` request with a `PaymentMethodToken` payment method via `grpcurl` (deliberately invalid API key) -- now returns a genuine Stripe 401 (`"Invalid API Key provided"`) instead of `NOT_IMPLEMENTED`, confirming the request reaches Stripe correctly. `SetupMandateResponse` deserialization (with `object`/`metadata` omitted and with both present) verified at the unit level.


XyneSpaces:
[blocking] The new `match token_data.token_payment_method_type` in `StripePaymentMethodData` only handles `Some(ApplePay | GooglePay)` and `None`; please add an explicit branch for the remaining `Some(TokenPaymentMethod::...)` variants, returning `NotImplemented` if they are unsupported for setup mandates. As written, a connector-token setup-mandate request with any other token payment method cannot compile because the match is non-exhaustive.

JeevaRamu0104 (approved):

AmitsinghTanwar007 (approved):

PR #2244, opened 2026-09-05
feat(proto): PaymentMethodType enum gaps + sender_payment_instrument_id on authorize response

gopikrishna000:
Two proto additions surfaced by the shadow migration-diff comparison. Both are additive (new fields/values, no renames or renumbers).

---

### Root cause

HS's `common_enums::PaymentMethodType` has several variants with no equivalent in `payments_grpc::PaymentMethodType`. HS's bridge conversion (`ForeignTryFrom<common_enums::PaymentMethodType> for payments_grpc::PaymentMethodType`, `crates/router/src/core/unified_connector_service/transformers.rs`) has an exhaustive match with a catch-all `_ => Err(...)` arm - any payment method type without a proto equivalent hits that arm and hard-errors with "Payment Method Type not yet supported" before a UCS request is even attempted (shadow execution fails silently, primary/direct call is unaffected).

Confirmed via real production log evidence (stripe klarna/pay_later PSync, request encoding failed at exactly this match, error originating from HS's own request-encoding code before any gRPC call reached UCS).

### Fix

Adds 19 missing `PaymentMethodType` enum values: `KLARNA`, `BHN_CARD_NETWORK`, `BLUECODE`, `BREADPAY`, `EFT_DEBIT_ORDER`, `FLEXITI`, `INDONESIAN_BANK_TRANSFER`, `MIFINITY`, `PAYJUSTNOW`, `PAYSERA`, `PAYSHAP`, `PAYSHAP_PROXY`, `PIX_AUTOMATICO_PUSH`, `PIX_AUTOMATICO_QR`, `PIX_EMV`, `PIX_KEY`, `PIX_QR`, `QRIS`, `SEPA_GUARENTEED_DEBIT`.

Two other HS variants missing from the bridge match (`InstantBankTransferFinland`, `InstantBankTransferPoland`) already have proto equivalents and are not touched here - only HS's own bridge match needs updating for those two.

---

### Root cause

The field exists on `PaymentServiceGetResponse` (PSync, proto field 30) but not on `PaymentServiceAuthorizeResponse`. A connector transformer that computes a payer instrument id on authorize (e.g. PayPal `payer_id`) writes it into `PaymentFlowData.sender_payment_instrument_id`, but the RouterDataV2 -> `PaymentServiceAuthorizeResponse` builder had no field to serialize it into, so it was silently dropped. hyperswitch's direct connector path sets `router_data.sender_payment_instrument_id` on authorize; the UCS path could not, producing a RouterData diff (`sender_payment_instrument_id` string on HS-direct, null on UCS) on paypal authorize.

### Fix

- Adds `optional string sender_payment_instrument_id = 29;` to `PaymentServiceAuthorizeResponse`.
- Wires it from `resource_common_data.sender_payment_instrument_id` in the authorize gRPC-response builder (`domain_types/src/types.rs`), success branch; `None` on the error branch.

---

Companion HS PR: [juspay/hyperswitch#14076](https://github.com/juspay/hyperswitch/pull/14076) (bridge match arms + authorize-response read-back).


XyneSpaces:
[blocking] This PR is coupled to juspay/hyperswitch#14076 for both the `sender_payment_instrument_id` contract and the new `PaymentMethodType` variants. Please merge/deploy the Prism and gateway changes together and validate generated proto/domain mappings end-to-end; if this lands alone, one side can expose values the other side does not understand.

JeevaRamu0104 (approved):

JeevaRamu0104 (dismissed):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2246, opened 2026-09-07
feat(common_utils): introduce Money Framework — opaque MinorUnit

AmitsinghTanwar007:
## Problem

Connectors could directly manipulate payment amounts because `MinorUnit` was fully open:

```rust
pub struct MinorUnit(pub i64);           // anyone could read/write the raw value
#[derive(Serialize, Deserialize)]        // anyone could put it in wire structs
pub fn new(i64) -> Self                  // anyone could construct arbitrary amounts
pub fn get_amount_as_i64() -> i64        // anyone could extract the raw value
impl Add, Sub, Display                   // anyone could do arithmetic, print raw values
```

This meant a connector could:
- **Bypass `AmountConvertor`** — use `MinorUnit` directly in wire structs, skipping currency-specific decimal conversion (JPY has 0 decimals, KWD has 3)
- **Construct fake amounts** — `MinorUnit::new(9999)`
- **Extract and tamper** — `amount.get_amount_as_i64() + 100`
- **Use deprecated i64 fields** — `request.refund_amount` (i64) instead of `request.minor_refund_amount` (MinorUnit)

### 1. Make `MinorUnit` opaque

| What | Before | After |
|---|---|---|
| Inner field | `pub i64` | **private** `i64` |
| `new(i64)` | Public method | Behind `MinorUnitProtoAccess` **trait** (requires import) |
| `get_amount_as_i64()` | Public method | Behind `MinorUnitProtoAccess` **trait** (requires import) |
| `Display` | Implemented (leaks raw value via `.to_string()`) | **Removed** |
| `Add`, `Sub`, `Mul`, `Sum` | Public trait impls | **Removed** (kept as `pub(crate)` for domain math) |
| `Serialize` / `Deserialize` | Always available | Stays — needed by domain event/logging structs |

### 2. Introduce `ConnectorMinorUnit`

A new type for connector wire structs. Connectors use this instead of `MinorUnit`.

- Has `Serialize`, `Deserialize`, `Display` — everything connectors need
- **No public constructor** — only obtainable via `AmountConvertor::convert()`
- `MinorUnitForConnector` now outputs `ConnectorMinorUnit` instead of `MinorUnit`

### 3. `proto_boundary` trait module

`MinorUnit::new()` and `get_amount_as_i64()` moved to `MinorUnitProtoAccess` trait.
`Money::new()`, `money.amount()`, `money.into_parts()` moved to `MoneyProtoAccess` trait.

Without importing the trait, the compiler rejects the call:
```
error[E0599]: no method named `get_amount_as_i64` found for struct `MinorUnit`
  help: trait `MinorUnitProtoAccess` provides `get_amount_as_i64`, consider importing it
```

### 4. Two layers of enforcement

**Layer 1 — Compiler**: Trait must be imported to call proto methods. Compiler blocks without import.

**Layer 2 — CI lint**: Grep blocks the import in connector code:
```bash
grep -rn "proto_boundary" crates/integrations/ --include="*.rs" | grep -v test.rs && exit 1
```

### 5. The safe path for connectors

```
MinorUnit (domain, opaque)
  → AmountConvertor::convert()
    → ConnectorMinorUnit (has Serialize + Display, no public constructor)
      → serde / Display
        → JSON / XML / String (wire format)
```

## What this PR includes

- Opaque `MinorUnit` with private field, removed `Display`/`Add`/`Sub`/`Mul`/`Sum`
- `ConnectorMinorUnit` type with `Serialize`/`Deserialize`/`Display`
- `proto_boundary` module with `MinorUnitProtoAccess` and `MoneyProtoAccess` traits
- `Money` fields made private with trait-gated accessors
- `Money::convert()` wrapper for safe connector-side amount conversion
- `compute_capturable_amount()` domain utility for amount arithmetic
- Full migration of `imerchantsolutions` connector to `ConnectorMinorUnit`
- HMAC signature fixes for adyen and novalnet
- Design document at `docs/money-framework-design.html`

### PR 2: Migrate connectors batch 1 (stacked — already raised as #2248)
- 15 connectors: affirm, billwerk, celero, datatrans, dlocal, fiuu, givepayments, grabpay, helcim, multisafepay, placetopay, shift4, silverflow, volt, worldpayvantiv
- Change wire struct fields from `MinorUnit` → `ConnectorMinorUnit`
- Use `AmountConvertor` for request construction, `convert_back` for response handling

### PR 3: Migrate connectors batch 2
- ~15 connectors: adyen, braintree, checkout, flywire, novalnet, nuvei, paypal, paysafe, stax, tamara, travelhub, truelayer, trustpay, worldpay, and others
- Same pattern as batch 1

### PR 4: Migrate connectors batch 3
- Remaining connectors: finix, getnet, nexixpay, payme, peachpayments, powertranz, ppro, revolut, sanlam_common, saferpay, tsys_transit, twoc_twop_paco, etc.
- Same pattern as batch 1

### PR 5: Change `AmountConvertor` trait to use `Money`
- Change `convert(&self, amount: MinorUnit, currency: Currency)` → `convert(&self, money: &Money)`
- Change `convert_back(...)` → return `Money` instead of `MinorUnit`
- Update 6 trait impls, 2 macro definitions (auto-covers 36 connectors), 4 utils helpers
- Update ~17 connectors that call `convert()` directly
- Update ~13 connectors that call `convert_back()` directly
- Rename to `Money::amount_convert()` and `Money::amount_convert_back()` as the public API

### PR 6: Remove deprecated `i64` amount fields from domain request structs
- Remove `amount: i64`, `refund_amount: i64`, `amount_to_capture: i64` from `PaymentsAuthorizeData`, `RefundsData`, `PaymentsCaptureData`, etc.
- Keep only `minor_amount: MinorUnit` (or `Money` if PR 5 is done)
- Update all connectors that still reference the `i64` fields

### PR 7: Remove `get_amount_as_i64()` from `MinorUnitProtoAccess` trait
- After PRs 5-6, no caller should need raw `i64` extraction
- Remove the method from the trait
- Simplify `proto_boundary` module

### PR 8 (future): Unify request structs to use `Money` instead of separate `minor_amount` + `currency`
- Replace `minor_amount: MinorUnit` + `currency: Currency` → `amount: Money` in domain request structs
- Address challenges: multiple amounts per currency, optional amount with required currency, response path currency threading
- This is the largest change — 672+ `.request.currency` and 241+ `.request.minor_amount` usages across 108 connector files

### PR 9 (future): Add CI lint for `proto_boundary` and `AmountConvertor` direct usage
- Add grep-based CI check blocking `proto_boundary` imports in connector code
- Add grep-based CI check blocking direct `AmountConvertor::convert()` calls in connector code (must go through `Money::amount_convert()`)

## What cannot be solved at the type system level

- **`MinorUnit` serde stays available** — domain structs (`CaptureSyncResponse`, surcharge types, mandate types, payment address types) derive `Serialize`/`Deserialize` and contain `MinorUnit` fields for event logging/Kafka. Cannot remove without redesigning event serialization.
- **`Debug` leaks the value as `MinorUnit(5000)`** — needed for error messages and logging. Not cleanly parseable, but technically visible.
- **`ConnectorMinorUnit::to_string()` gives the raw number** — by design, for wire format serialization. Extraction + tampering is possible but obvious in code review (same pattern as `Secret::expose()`).
- **Grep-based CI lint can be bypassed by CI config changes** — requires repo admin access, visible in git history. A compiler-level unbypassable lint would require a separate crate for `proto_boundary`.

## Test plan

- [x] `cargo check` — 0 errors
- [x] `cargo test -p ucs_common_utils` — 27 tests pass
- [x] `cargo test --test test_amount_conversion` — 3 tests pass
- [x] CI lint passes: 0 `proto_boundary` imports in non-test connector code
- [ ] Verify adyen webhook HMAC verification
- [ ] Verify novalnet webhook HMAC verification
- [ ] Verify imerchantsolutions partial capture flow

### DIFF
--- a/crates/common/common_utils/src/types.rs
+++ b/crates/common/common_utils/src/types.rs
@@ -1,11 +1,6 @@
 //! Types that can be used in other crates
 
-use std::{
-    fmt::Display,
-    iter::Sum,
-    ops::{Add, Mul, Sub},
-    str::FromStr,
-};
+use std::{fmt::Display, str::FromStr};
 
 use common_enums::enums;
 use error_stack::ResultExt;
@@ -130,64 +125,107 @@ impl AmountConvertor for FloatMajorUnitForConnector {
     }
 }
 
-/// Connector required amount type
+/// Connector required amount type – outputs ConnectorMinorUnit so connectors
+/// never touch domain MinorUnit directly.
 #[derive(Default, Debug, serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq)]
 pub struct MinorUnitForConnector;
 
 impl AmountConvertor for MinorUnitForConnector {
-    type Output = MinorUnit;
+    type Output = ConnectorMinorUnit;
     fn convert(
         &self,
         amount: MinorUnit,
         _currency: enums::Currency,
     ) -> Result<Self::Output, error_stack::Report<ParsingError>> {
-        Ok(amount)
+        Ok(ConnectorMinorUnit(amount))
     }
     fn convert_back(
         &self,
-        amount: MinorUnit,
+        amount: ConnectorMinorUnit,
         _currency: enums::Currency,
     ) -> Result<MinorUnit, error_stack::Report<ParsingError>> {
-        Ok(amount)
+        Ok(amount.0)
     }
 }
 
-/// This Unit struct represents MinorUnit in which core amount works
-#[derive(
-    Default,
-    Debug,
-    serde::Deserialize,
-    serde::Serialize,
-    Clone,
-    Copy,
-    PartialEq,
-    Eq,
-    Hash,
-    ToSchema,
-    PartialOrd,
-)]
+/// Connector payload amount that serializes as a minor-unit number.
+///
+/// This keeps connector request/response structs from using domain `MinorUnit`
+/// directly while preserving connector payloads that expect raw numeric minor
+/// units. Connectors obtain this type **only** via `AmountConvertor::convert`.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, ToSchema, PartialOrd)]
+pub struct ConnectorMinorUnit(MinorUnit);
 
-pub struct MinorUnit(pub i64);
+impl Serialize for ConnectorMinorUnit {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_i64(self.0 .0)
+    }
+}
+
+impl<'de> Deserialize<'de> for ConnectorMinorUnit {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value = <i64 as Deserialize>::deserialize(deserializer)?;
+        Ok(ConnectorMinorUnit(MinorUnit(value)))
+    }
+}
+
+impl Display for ConnectorMinorUnit {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0 .0)
+    }
+}
+
+/// This Unit struct represents MinorUnit in which core amount works.
+///
+/// The inner field is **private**. Construction and extraction are gated behind
+/// the `proto-conversion` cargo feature so that only the proto/domain boundary
+/// crates can create or inspect raw values. Connector code receives `MinorUnit`
+/// in domain structs but can only pass it through `AmountConvertor`.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash, ToSchema, PartialOrd)]
+pub struct MinorUnit(i64);
+
+impl Serialize for MinorUnit {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_i64(self.0)
+    }
+}
+
+impl<'de> Deserialize<'de> for MinorUnit {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value = <i64 as Deserialize>::deserialize(deserializer)?;
+        Ok(MinorUnit(value))
+    }
+}
 
 impl MinorUnit {
-    /// gets amount as i64 value will be removed in future
-    pub fn get_amount_as_i64(self) -> i64 {
+    // ── internal (crate-only) constructors/extractors ──────────────────
+    // These are used by AmountConvertor impls, convert_back, Sum, etc.
+    // within common_utils itself.
+
+    /// Crate-internal constructor.
+    pub(crate) fn from_i64(value: i64) -> Self {
+        Self(value)
+    }
+
+    /// Crate-internal extractor.
+    pub(crate) fn as_i64(self) -> i64 {
         self.0
     }
 
-    /// forms a new minor default unit i.e zero
-    pub fn zero() -> Self {
-        Self(0)
-    }
-
-    /// forms a new minor unit from amount
-    pub fn new(value: i64) -> Self {
+    /// Construct from a raw i64 in tests.
+    #[cfg(test)]
+    pub fn test_new(value: i64) -> Self {
         Self(value)
     }
 
     /// checks if the amount is greater than the given value
     pub fn is_greater_than(&self, value: i64) -> bool {
-        self.get_amount_as_i64() > value
+        self.0 > value
+    }
+
+    /// Returns true if the amount is positive (> 0)
+    pub fn is_positive(&self) -> bool {
+        self.0 > 0
     }
 
     /// Convert the amount to its major denomination based on Currency and return String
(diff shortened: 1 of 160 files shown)


crates/common/common_utils/src/types.rs
```diff
@@ -130,64 +125,111 @@ impl AmountConvertor for FloatMajorUnitForConnector {
...
 impl MinorUnit {
-    /// gets amount as i64 value will be removed in future
-    pub fn get_amount_as_i64(self) -> i64 {
-        self.0
+    // ── internal (crate-only) constructors/extractors ──────────────────
+    // These are used by AmountConvertor impls, convert_back, Sum, etc.
+    // within common_utils itself.
+
+    /// Crate-internal constructor.
+    pub(crate) fn from_i64(value: i64) -> Self {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -130,64 +125,111 @@ impl AmountConvertor for FloatMajorUnitForConnector {
...
+        Self(value)
     }
 
-    /// forms a new minor default unit i.e zero
-    pub fn zero() -> Self {
-        Self(0)
+    /// Crate-internal extractor.
+    pub(crate) fn as_i64(self) -> i64 {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -130,64 +125,111 @@ impl AmountConvertor for FloatMajorUnitForConnector {
...
+        self.0
     }
 
-    /// forms a new minor unit from amount
-    pub fn new(value: i64) -> Self {
+    /// Construct from a raw i64 in tests.
+    #[cfg(test)]
+    pub fn test_new(value: i64) -> Self {
```
jarnura: Why a test_new is required?

crates/common/common_utils/src/types.rs
```diff
@@ -130,64 +125,111 @@ impl AmountConvertor for FloatMajorUnitForConnector {
...
     }
 
     /// checks if the amount is greater than the given value
     pub fn is_greater_than(&self, value: i64) -> bool {
-        self.get_amount_as_i64() > value
+        self.0 > value
+    }
+
+    /// Returns true if the amount is positive (> 0)
+    pub fn is_positive(&self) -> bool {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -130,64 +125,111 @@ impl AmountConvertor for FloatMajorUnitForConnector {
...
+        self.0 > 0
+    }
+
+    /// Returns true when the amount is zero.
+    pub fn is_zero(&self) -> bool {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -254,37 +296,22 @@ impl MinorUnit {
...
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.0)
-    }
-}
-
-impl Add for MinorUnit {
-    type Output = Self;
-    fn add(self, a2: Self) -> Self {
-        Self(self.0 + a2.0)
+#[allow(dead_code)]
```
jarnura: Why this ia Dead Code, if dead remove the code do not allow

crates/common/common_utils/src/types.rs
```diff
@@ -254,37 +296,22 @@ impl MinorUnit {
...
+impl MinorUnit {
+    pub(crate) fn add(self, other: Self) -> Self {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -254,37 +296,22 @@ impl MinorUnit {
...
+        Self(self.0 + other.0)
     }
-}
 
-impl Sub for MinorUnit {
-    type Output = Self;
-    fn sub(self, a2: Self) -> Self {
-        Self(self.0 - a2.0)
+    pub(crate) fn sub(self, other: Self) -> Self {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -254,37 +296,22 @@ impl MinorUnit {
...
+        Self(self.0 - other.0)
     }
-}
-
-impl Mul<u16> for MinorUnit {
-    type Output = Self;
 
-    fn mul(self, a2: u16) -> Self::Output {
-        Self(self.0 * i64::from(a2))
+    pub(crate) fn mul_u16(self, factor: u16) -> Self {
```
jarnura: These functions should be private and not public

crates/common/common_utils/src/types.rs
```diff
@@ -254,37 +296,22 @@ impl MinorUnit {
...
+        Self(self.0 * i64::from(factor))
     }
-}
 
-impl Sum for MinorUnit {
-    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
-        iter.fold(Self(0), |a, b| a + b)
+    pub(crate) fn sum_iter(iter: impl Iterator<Item = Self>) -> Self {
```
jarnura: These functions should be private and not public

jarnura (changes requested):

crates/integrations/connector-integration/src/connectors/adyen/transformers.rs
```diff
@@ -4409,7 +4393,7 @@ where
             ),
             resource_common_data: PaymentFlowData {
                 status: adyen_payments_response_data.status,
-                amount_captured: minor_amount_captured.map(|amount| amount.get_amount_as_i64()),
+                amount_captured: None,
```
jarnura: Why these values are made into null?

crates/integrations/connector-integration/src/connectors/adyen/transformers.rs
```diff
@@ -4532,7 +4516,7 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
```
jarnura: Why these values are made into null?

crates/integrations/connector-integration/src/connectors/adyen/transformers.rs
```diff
@@ -6907,7 +6891,7 @@ impl<F, T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Se
...
```
jarnura: Why these values are made into null?

jarnura (changes requested):

crates/integrations/connector-integration/src/connectors/adyen/transformers.rs
AmitsinghTanwar007: we will be deprecating this field but in connector code it requires get_amt_as_i64() which i am avoiding so i am populating this value from minor amt capture in domain which has access to as_i64()

crates/common/common_utils/src/types.rs
AmitsinghTanwar007: i was thinking of having these(proto_boundary functions) inside one file in proto_boundary.rs who's side effect is this change
but i may try to move them inside types.rs which may allow me to keep things private.

AmitsinghTanwar007: and i made this decision previously because i was trying to use grep lint to avoid import of these function in connector layer (due to different file the path becomes unique)
but i can use clippy disallowed_methods instead and can have them in tyeps.rs
but the trade off is whenever new function is added here then it needs to be banned in the clippy as well
but i need to check what blast radius does pub(crate) holds

crates/common/common_utils/src/types.rs
AmitsinghTanwar007: these are used by connector for verification of the value and this functions dont manipulate the real amount

crates/common/common_utils/src/types.rs
AmitsinghTanwar007: this also dont manipulate the amount and used by connector

PR #2257, opened 2026-09-08
fix(external-services): retry once on connection closed before message completed

AmitsinghTanwar007:
## Summary

- Ports the connection-closed retry logic from Hyperswitch's `external_services` crate to UCS
- When hyper's connection pool selects an idle TCP connection at the exact moment the remote server sends a FIN, the request fails with `is_incomplete_message()`. Since the full HTTP request was never transmitted, it is safe to retry exactly once on a fresh connection
- Adds `http_client` module with `send_request_with_retry` — clones the request before sending, detects the hyper incomplete-message error, and retries once with the clone
- Adds OTel counter `ucs_auto_retry_connection_closed` (gated behind `otel` feature) for observability into retry frequency per connector host
- Replaces inline send logic in `call_connector_api` with the new retry-aware function

## Context

Intermittent HTTP 500 / HE_00 errors were observed against Fiserv (`connect.fiservapis.com`) on:
- 2026-08-14 20:30:28 UTC
- 2026-08-28 21:20:45 UTC (payment `PRSJ_96434409`)
- 2026-09-01 23:43:26 UTC

Root cause: TCP connection-pool race condition — hyper picks a dying idle connection at the exact moment the remote server sends a TCP FIN. This is a known issue already solved in Hyperswitch (`http_client.rs` lines 172-277).

Reference: Hyperswitch implementation in `crates/external_services/src/http_client.rs`

## Files Changed

| File | Change |
|------|--------|
| `http_client.rs` (new) | `send_request_with_retry`, `send_request`, `is_connection_closed_before_message_could_complete` |
| `lib.rs` | Register `pub mod http_client` |
| `otel_metrics.rs` | New OTel counter `ucs_auto_retry_connection_closed` |
| `service.rs` | Replace inline send with retry-aware function, emit metric + structured log on retry |

## Test plan

- [ ] `cargo check -p external-services` passes
- [ ] Verify OTel metric `ucs_auto_retry_connection_closed` appears in Grafana after deployment
- [ ] Monitor Fiserv `HE_00` error rate — should drop to near zero for connection-closed cases
- [ ] Confirm no regression in existing connector flows

### DIFF
--- a/crates/common/external-services/src/http_client.rs
+++ b/crates/common/external-services/src/http_client.rs
@@ -31,8 +31,7 @@ pub async fn send_request_with_retry(
 
     match response {
         Err(ref error)
-            if error.current_context()
-                == &ApiClientError::ConnectionClosedIncompleteMessage =>
+            if error.current_context() == &ApiClientError::ConnectionClosedIncompleteMessage =>
         {
             match cloned_request {
                 Some(cloned) => {
--- a/crates/common/external-services/src/http_client.rs
+++ b/crates/common/external-services/src/http_client.rs
@@ -1,0 +1,90 @@
+use std::error::Error as StdError;
+
+use common_enums::ApiClientError;
+use error_stack::{report, Report};
+
+/// Sends an HTTP request via reqwest, automatically retrying once if the
+/// connection was closed by the remote side before the message completed.
+///
+/// This handles a well-known race condition in HTTP connection pooling: hyper
+/// maintains a pool of idle keep-alive connections and occasionally selects one
+/// at the exact moment the remote server sends a TCP FIN to close it. Since
+/// hyper has already begun writing the request bytes, it cannot silently retry
+/// — the server may have acted on partial data. However, the
+/// `is_incomplete_message()` error tells us the full HTTP request was never
+/// transmitted, making it safe to retry exactly once on a fresh connection.
+///
+/// The request is cloned via `try_clone()` before the first attempt. If the
+/// body is a stream that cannot be cloned, the retry is skipped and the
+/// original error is returned.
+///
+/// Returns `(response, retried)` where `retried` is `true` if the first attempt
+/// hit a connection-closed error and a retry was attempted.
+pub async fn send_request_with_retry(
+    request: reqwest::RequestBuilder,
+) -> (Result<reqwest::Response, Report<ApiClientError>>, bool) {
+    // Clone the request before sending so we have a backup for retry.
+    // `try_clone()` returns None for streaming bodies that cannot be cloned.
+    let cloned_request = request.try_clone();
+
+    let response = send_request(request).await;
+
+    match response {
+        Err(ref error)
+            if error.current_context()
+                == &ApiClientError::ConnectionClosedIncompleteMessage =>
+        {
+            match cloned_request {
+                Some(cloned) => {
+                    tracing::info!(
+                        "Retrying request due to connection closed before message completed"
+                    );
+                    (send_request(cloned).await, true)
+                }
+                None => {
+                    tracing::info!(
+                        "Cannot retry connection-closed error: request body is not cloneable"
+                    );
+                    (response, false)
+                }
+            }
+        }
+        response => (response, false),
+    }
+}
+
+/// Sends a single HTTP request and maps reqwest errors to `ApiClientError`.
+async fn send_request(
+    request: reqwest::RequestBuilder,
+) -> Result<reqwest::Response, Report<ApiClientError>> {
+    request.send().await.map_err(|error| {
+        let api_error = match error {
+            error if error.is_timeout() => ApiClientError::RequestTimeoutReceived,
+            error if is_connection_closed_before_message_could_complete(&error) => {
+                ApiClientError::ConnectionClosedIncompleteMessage
+            }
+            _ => ApiClientError::RequestNotSent(error.to_string()),
+        };
+        report!(api_error)
+    })
+}
+
+/// Checks whether a `reqwest::Error` was caused by hyper's
+/// "connection closed before message completed" condition.
+///
+/// This walks the error source chain to find the underlying `hyper::Error`
+/// and checks `is_incomplete_message()`. This error occurs when hyper's
+/// connection pool selects an idle connection at the same moment the remote
+/// server sends a TCP FIN to close it.
+fn is_connection_closed_before_message_could_complete(error: &reqwest::Error) -> bool {
+    let mut source = error.source();
+    while let Some(err) = source {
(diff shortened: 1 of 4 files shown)


crates/common/external-services/src/http_client.rs
```diff
@@ -0,0 +1,89 @@
...
+/// `is_incomplete_message()` error tells us the full HTTP request was never
+/// transmitted, making it safe to retry exactly once on a fresh connection.
+///
+/// The request is cloned via `try_clone()` before the first attempt. If the
+/// body is a stream that cannot be cloned, the retry is skipped and the
+/// original error is returned.
+///
+/// Returns `(response, retried)` where `retried` is `true` if the first attempt
+/// hit a connection-closed error and a retry was attempted.
+pub async fn send_request_with_retry(
```
JeevaRamu0104, line 28: By default the retry count is 1. can we have configurable retry count.

AmitsinghTanwar007: bro as per me, no configurability is needed because when trying again it creates a new connection of tcp so that there is no chances of getting FIN for that so always for this situation the retry will be 1.

crates/common/external-services/src/http_client.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// This walks the error source chain to find the underlying `hyper::Error`
+/// and checks `is_incomplete_message()`. This error occurs when hyper's
+/// connection pool selects an idle connection at the same moment the remote
+/// server sends a TCP FIN to close it.
+fn is_connection_closed_before_message_could_complete(error: &reqwest::Error) -> bool {
+    let mut source = error.source();
+    while let Some(err) = source {
+        if let Some(hyper_err) = err.downcast_ref::<hyper::Error>() {
+            if hyper_err.is_incomplete_message() {
```
gopikrishna000, line 88: how are we sure here, that connector havent received request?
and retry should be with idempotency key?

AmitsinghTanwar007: hey bro so the thing is hyper::Error::is_incomplete_message() specifically means hyper was still writing request bytes to the socket when connection died.
like the HTTP framing itself broke mid-write. server cant process a request it never fully received right HTTP needs complete headers + complete body
to even parse it. incomplete frame just gets discarded by server's HTTP parser.

this is different from timeout where server might have got full request and is processing it. here the pipe broke while we were still pouring data into it so connector never saw a valid request.

about idempotency key :
not needed here because this is not application level retry its transport level recovery. from connector's perspective nothing happened, no request was received so there is nothing to be idempotent against. its like the request was never sent in first place.

this is same approach hyperswitch follows in production at scale (crates/external_services/src/http_client.rs lines 215-248)

gopikrishna000: Checked hyper's source (1.8.1, the one reqwest 0.12 pulls in here) and I don't think that characterisation holds.

`is_incomplete_message()` is raised on the READ side, after the request is already in flight, when hyper hits EOF while waiting for / parsing the response:

- `proto/h1/io.rs:219` "parse eof": read returned 0 bytes while parsing response headers
- `proto/h1/conn.rs:472` "found unexpected EOF on busy connection" (`should_error_on_eof` = only when NOT idle, i.e. a request has been written)
- `proto/h1/conn.rs:503` same, mid-message

A failure while still writing request bytes comes back as an io error (broken pipe / connection reset), not `IncompleteMessage`. Hyper's own client test for this variant (`client/tests.rs`) is explicit about the scenario: the mock server reads the second request and only then disconnects. Its comment: "we write on an assumedly good connection, and only reset the close AFTER we wrote bytes."

Hyper-util also deliberately does NOT auto-retry this. Its `retry_canceled_requests` only covers "requests that get disrupted before ever starting to write" on a reused connection, with the note "a fresh connection means we definitely can't retry". `IncompleteMessage` is excluded precisely because the client cannot know whether the server processed the request.

So this error means "the response could not be read", not "the request was never received". In the stale keep-alive race (server already closed, kernel RSTs the bytes) the connector indeed never saw it and a retry is safe. But a server / LB that read the full request and then crashed, hit an upstream timeout and sent FIN, or was shutting down produces the identical error, and a retry there double-submits.

Hyperswitch running this for a year tells us the race case dominates in practice; it doesn't make the claim provably true, and the cost of being wrong is a duplicate Authorize / Capture / Refund.

Suggestions:
1. Fix the doc comment: it says hyper "has already begun writing the request bytes" and then that "the full HTTP request was never transmitted". The first half is right, the second isn't.
2. Gate the retry on idempotency: GET, sync flows (PSync / RSync), or requests carrying a connector idempotency key. For non-idempotent POSTs either skip the retry or make it opt-in per connector.
3. Keep the metric, it's the right thing to have. It will show per-connector how often this actually fires before widening the scope.

AmitsinghTanwar007: hey bro you're right on this, i dug into hyper source as well and the analysis checks out. all 3 raise sites for is_incomplete_message are on the read
  path - parse(), require_empty_read(), mid_message_detect_eof(). none of them touch write buffer. the "message" that didnt complete is the response not the
   request.

  write failures actually produce different error kind - is_io() style errors like broken pipe / connection reset. so if request write itself had failed our
   retry wouldnt even trigger since we only match on is_incomplete_message.

  i already updated the doc comment in the latest commit to reflect this accurately - it now says hyper writes the request, transitions to Busy, then gets
  EOF on read side. and that request bytes may have reached server since FIN is a half-close.

  the retry is still safe in practice because idle connection cleanup uses close() not shutdown(SHUT_WR) which closes both directions and RSTs incoming
  data. also this has exact parity with hyperswitch main which has been running this same blind retry in production for years without double-charge issues
  (ref: https://github.com/juspay/hyperswitch/pull/3426#discussion_r1472535371)

  about the idempotency gating suggestion - valid improvement but should go as a separate enhancement on both HS and UCS together. for now the goal is to
  stop the Fiserv HE_00 errors with the same proven approach.

crates/common/external-services/src/service.rs
```diff
@@ -1457,21 +1458,23 @@ pub async fn call_connector_api(
...
-            );
-            report!(api_error)
-        })
-    };
+    let (response, retried) = crate::http_client::send_request_with_retry(request).await;
+
+    if retried {
+        #[cfg(feature = "otel")]
+        crate::otel_metrics::record_auto_retry_connection_closed(&connector_host);
+        tracing::info!(
```
gopikrishna000, line 1466: should it be warn

gopikrishna000 (approved):

JeevaRamu0104 (approved):

PR #2266, opened 2026-09-08
feat(Connector): Kount Refer the sessionId from metadata for DDC.

itzcharanteja:
## Description
- The Kount connector integration now consumes the merchant-provided DDC session id sent in the request metadata (`{"sessionId": "..."}` — the Euler `UcsFrmMetadata` shape), instead of always deriving one internally.
- Added helpers in `connectors/kount/transformers.rs`:
  - `session_id_from_metadata_value` — reads `sessionId` from metadata as a parsed JSON value; `None` on absent metadata, missing key, or empty value.
  - `session_id_from_metadata_str` — same, for metadata carried as a JSON-encoded string (the `PreRiskCheckRequest` metadata field); non-JSON strings are tolerated and yield `None`.
  - `resolve_session_id` — uses the provided `sessionId` when present, else falls back to the existing `hash_session_id` (SHA-256 of the merchant transaction id, first 32 chars).
- Wired into both DDC touchpoints:
  - `handle_pre_authenticate_response` (PreAuthenticate / DDC HTML response)
  - Evaluate Order (pre_risk_check) request builder

## Motivation and Context
The prism previously derived the Kount DDC session id unconditionally as `sha256(merchant_transaction_id)[:32]`. That works when the shopper uses the connector-served DDC script, but merchant-managed DDC flows collect device data against their **own** session reference, which Euler threads through the composite (pre_authenticate / pre_risk_check) request metadata. Ignoring it meant the device data collected on the merchant page never correlated with the inquiry sent to Kount — the inquiry was evaluated without the collected device signals.

With this change the provided `sessionId` wins at both the DDC response and the Evaluate Order inquiry, keeping the collection and evaluation on the same session id; the hash remains as the fallback so existing flows are unaffected.

EULER-80735

### Additional Changes

Neither applies — the metadata field already existed on the request types; only its consumption changed.

## How did you test it?
- Unit tests (`session_id_tests` in `transformers.rs`): sessionId resolved from parsed metadata value; missing/empty `sessionId` and non-JSON metadata ignored; string-metadata parsing; hash fallback (32-char SHA-256); provided id wins over hash.
- E2E against the Euler stack (euler-api-txns → prism grpc-server → Kount):
  - Google Pay DPAN flow with the merchant's DDC-collected reference threaded via FRM request metadata — the Kount Evaluate Order inquiry carried the same `deviceSessionId` (verified in connector request logs).
  - Flow without a provided reference — session id falls back to `sha256(order_id)[:32]` in both the DDC HTML and the inquiry, matching pre-change behavior.
Note: your working tree also has uncommitted local tweaks (config/development.toml, grpc-server/main.rs, flake.*) — the commit itself contains only transformers.rs, so the PR diff matches this description.


crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -738,6 +740,34 @@ pub fn hash_session_id(raw: &str) -> String {
...
 }
 
+/// Read the merchant-provided DDC session id from request metadata parsed as
+/// JSON (`{"sessionId": "..."}` — the Euler `UcsFrmMetadata` shape). `None`
+/// when the metadata is absent, carries no `sessionId`, or the value is empty.
+pub fn session_id_from_metadata_value(metadata: Option<&serde_json::Value>) -> Option<String> {
+    metadata
+        .and_then(|metadata| metadata.get("sessionId"))
+        .and_then(|session_id| session_id.as_str())
+        .filter(|session_id| !session_id.is_empty())
```
kanikac199: use `trim().is_empty()` instead of `is_empty()` to reject whitespace-only ids

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -738,6 +740,34 @@ pub fn hash_session_id(raw: &str) -> String {
...
+        .and_then(|metadata| serde_json::from_str::<serde_json::Value>(metadata.peek()).ok());
+    session_id_from_metadata_value(parsed.as_ref())
+}
+
+/// Resolve the Kount DDC session id. A merchant-provided `sessionId` in the
+/// request metadata wins — merchant-managed DDC scripts correlate on their own
+/// session id, and Euler threads the collected reference through both the
+/// pre_authenticate and pre_risk_check composite calls. Otherwise derive it
+/// from the fallback reference (the merchant transaction id) as before.
+pub fn resolve_session_id(provided_session_id: Option<String>, fallback_ref: &str) -> String {
```
kanikac199, line 761: `resolve_session_id` now returns the ID as is. Before it always went through `hash_session_id(...)`, so we always got a fixed 32-char hex value.

the DDC script `js_string_escape` only handles XSS, and `to_session_id` truncation doesn’t run on this path.

can you confirm if Kount accepts any format for deviceSessionId?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1626,7 +1664,13 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
 
         Ok(Self {
             order_id: order_id.clone(),
-            session_id: hash_session_id(&order_id),
+            // Merchant-provided `sessionId` from the request metadata (the
+            // DDC-collected reference) wins; otherwise hash the merchant
+            // transaction id — the same value the DDC script derives.
+            session_id: resolve_session_id(
```
kanikac199, line 1661: +1

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -738,6 +740,34 @@ pub fn hash_session_id(raw: &str) -> String {
         .unwrap_or_else(|_| to_session_id(raw))
 }
 
+/// Read the merchant-provided DDC session id from request metadata parsed as
+/// JSON (`{"sessionId": "..."}` — the Euler `UcsFrmMetadata` shape). `None`
```
kanikac199: can we remove any "euler' references from comments

XyneSpaces:
[should-fix] Add focused tests for the new metadata-based Kount session-ID resolution, covering a valid `sessionId`, malformed metadata, and the transaction-ID fallback. These inputs now choose different DDC identifiers, so a parsing regression would silently break Kount session matching.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
itzcharanteja: @kanikac199 , we have introduced this sessionId for the cases where merchant has integrated the FRM provider's device data collection script at their payment page side. In these cases merchant will send us the sessionID that they've used and we should relay the same to Kount. Merchant should adhere to the sessionID requirements defined by Kount.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
itzcharanteja: https://github.com/juspay/hyperswitch-prism/pull/2266/changes#r3978291971

kanikac199 (approved):

Aishwariyaa-Anand (approved):

PR #2267, opened 2026-09-08
feat(proto): typed 3DS authenticate contract for external authentication (Netcetera)

gopikrishna000:
## Summary

First cut of moving the Netcetera external-3DS request fields off the opaque `metadata` / `connector_feature_data` JSON blobs onto typed proto fields. Every added field is an EMVCo AReq element consumed by all four 3DS-server integrations (netcetera, gpayments, threedsecureio, unified_authentication_service), so it is orchestrator input, not connector-private data.

Review doc this implements: "UCS proto changes for external auth". Verdict per field is below.

## Proto changes

| Proposed in doc | Decision | Where it landed |
|---|---|---|
| force_3ds_challenge (bool) | Rejected as a bool, accepted as EMVCo enum | `three_ds_requestor_challenge_indicator` (01-09) |
| three_ds_requestor_authentication_ind | Accepted | `three_ds_requestor_authentication_indicator` (01-06) |
| device_channel | Accepted, plus 3RI | `device_channel` enum |
| sdk_information | Accepted, JWK typed | `ThreeDsSdkInformation` |
| acquirer_bin / merchant_id / country | Accepted | `AcquirerDetails` |
| merchant_name / MCC / country | Accepted by extending the existing message | `MerchantDetails` (was FRM-only) |
| notification_url | Rejected, duplicate | use `return_url` |
| results_response_notification_url | Rejected, duplicate | use `webhook_url` |
| merchant_configuration_id, three_ds_requestor_id/name, endpoint_prefix | Rejected from request, account config | `NetceteraConfig` (same pattern as `AdyenConfig.endpoint_prefix`) |
| pull_mechanism_for_external_3ds_enabled | Rejected, router-only flag | nowhere |
| not in doc | Added, were hardcoded in connector | `message_category`, `threeds_completion_indicator` |

`NetceteraConfig` also makes the mTLS certificate pair optional (doc note 2: VGS may terminate TLS).

## Compatibility: two contracts, never mixed

Sandbox merchants exercise the legacy transport today and the router deployment cannot be coordinated, so both contracts produce exactly the same AReq as before:

- **Legacy contract** (request carries none of fields 17-24, connector config is `NoKey`): byte-for-byte the pre-typed behaviour. device_channel / sdk_information from the `metadata` JSON, acquirer / merchant / URLs / force_3ds_challenge from the `connector_feature_data` blob, blob notification_url over return_url. Nothing changes for current callers.
- **Typed contract** (request carries any of fields 17-24, or `NetceteraConfig` is sent): only typed fields, `return_url`, `webhook_url` and `NetceteraConfig` are read. The blobs are never consulted; a missing typed value is absent, not a fallback.

Selection is one flag, `uses_typed_three_ds_contract`, evaluated once. The connector builds the AReq merchant side through one of two pure functions (`build_merchant_side_from_typed` / `build_merchant_side_from_legacy_blob`). The legacy branch logs `netcetera_legacy_3ds_transport=true`.

Wire: additive only, `buf lint` and `buf breaking` pass.

## Deprecation plan

1. This PR deploys first. No router change needed for it to be safe.
2. Router PR sends fields 17-24, `webhook_url`, and `NetceteraConfig`. Deploy whenever convenient. Note for the router side: `acquirer_country_code` and `merchant_country_code` are `CountryAlpha2` on the typed contract and UCS emits the ISO numeric code, so the AReq is identical only if the legacy blob also carried numeric codes.
3. After 2026-10-23 and once the Loki count of `netcetera_legacy_3ds_transport=true` is zero: one PR deleting the legacy branch, `AuthenticateSdkMetadata`, the transaction-level fields on `NetceteraMeta`, and the deprecation notes in the proto comments. No proto field is removed.

## Netcetera behaviour changes

- `threeDSCompInd`, `messageCategory`, `threeDSRequestorAuthenticationInd` were hardcoded (U / 01 / 01). Now from the request, same defaults when absent.
- Legacy `force_3ds_challenge = true` still maps to challenge indicator 04 through the fallback.
- Empty merchant object is omitted so the 3DS Server keeps using its stored merchant configuration.

## Deliberately out of scope (to discuss)

- Response side: DDC `form_fields` magic keys vs a typed `RedirectForm` variant.
- Response side: home for `acs_signed_content` / `acs_reference_number` (currently in response `connector_feature_data`).
- `ds_trans_id` being populated from the pre-auth card range `directory_server_id` (router reads it back as directory_server_id, so it is a coordinated hack; needs a `directory_server_id` field on the response side).
- Router PR to send the typed fields and populate `webhook_url`.

### Additional Changes
- [x] This PR modifies the API contract (`crates/types-traits/grpc-api-types/proto/payment.proto`, additive only)

## How did you test it?

`buf lint` locally. Rust compile and clippy left to CI (repo practice, no local builds). No Netcetera sandbox run yet; the typed path is not exercised by the current router, and the fallback path is byte-for-byte the previous behaviour except the `return_url` precedence noted above.

## Not built locally

Per repo practice, verification is left to CI.
---

## Review follow-ups (second pass)

Proto:
- `ThreeDsRequestorAuthenticationIndicator`: added `BILLING_AGREEMENT = 7`. The connector's own enum already had `07`; the proto stopped at `06`, so the typed path could not express a value the connector models.
- `AuthenticationData` +4 response fields: `acs_signed_content = 18`, `acs_reference_number = 19`, `directory_server_id = 20`, `scheme_id = 21`. The first two move the ACS challenge values off response `connector_feature_data`. The last two fix a contract defect: PreAuthenticate writes the card-range *directory server id* into `ds_trans_id`, a per-transaction field, and drops `scheme_id` entirely (doc note 1). Both are now populated correctly.
  The `ds_trans_id` write is **kept** so existing readers are unaffected; it carries a condition, not a date. Dropping it needs a reader-side migration and is safe to never do.
- `NetceteraConfig`: renumbered so credentials/identity are `1..5` and URL config is `50..51`. The previous layout skipped `51` by copying Adyen's `dispute_base_url` slot, which Netcetera has no use for. Oneof tag moved `157 -> 163` after rebase (main took 157-162 for new connectors).
- `MerchantDetails.merchant_id` doc restated as a definition rather than a prohibition.
- `ThreeDsSdkInformation` comment now states when to send it. Members stay required inside the optional message: the SDK either ran and produced all of them, or it did not run.

Domain:
- Single `MerchantDetails` domain struct. The previous revision added a second struct of the same name in `connector_types` while FRM kept its own two-field version, so FRM silently dropped `merchant_name` / `merchant_country_code`. FRM's conversion is now fallible and both paths share one type.
- Enum conversions no longer `unwrap_or_default()` an unknown discriminant into `UNSPECIFIED` and then report `InvalidDataFormat`. Absent field -> absent (never an error); present-but-unknown -> an error naming the field.
- Dropped the `MissingRequiredField` on `sdk_ephem_pub_key`; the `Option` there is a nested-message artifact, not semantics.

Legacy isolation:
- Legacy blob handling moved to `netcetera/legacy_blob_transport.rs`, prefixed `legacy_`, with the branch inverted so the surviving path is the fallthrough arm. Removal is: delete the file, delete its `mod` line, collapse one `match`. No renames, no judgement calls.
- **Fixed:** the legacy `warn!` fired unconditionally, including for callers that send neither typed fields nor a blob. Since the deprecation gate is "zero `netcetera_legacy_3ds_transport=true` lines", that guaranteed the gate could never go green. It now fires only when a blob is actually read.
- Deprecation date `2026-09-23 -> 2026-10-23` across all sites, gated on the Loki count rather than the date alone.

Every change here is additive or behaviour-preserving. If the cleanup PR is never raised, nothing that works today breaks.


crates/integrations/connector-integration/src/connectors/netcetera.rs
```diff
@@ -63,13 +65,34 @@ const MERCHANT_ENDPOINT_PREFIX_TEMPLATE: &str = "{{merchant_endpoint_prefix}}";
...
+            .as_deref()
+            .map(|prefix| base_url.replace(MERCHANT_ENDPOINT_PREFIX_TEMPLATE, prefix))
+            .ok_or_else(|| {
+                error_stack::report!(IntegrationError::InvalidConnectorConfig {
+                    config: "netcetera.endpoint_prefix",
+                    context: Default::default(),
+                })
+            });
+    }
+
```
sahkal: Netcetera has removed endpoint prefix please re check this

gopikrishna000: Good catch, confirmed and fixed in e752444.

The prefix mechanism was retired in hyperswitch#13145 (2026-07-20), which moved Netcetera from per-merchant subdomains to a flat host and deleted the substitution logic. UCS copied the old config on 2026-07-09, eleven days before that change, and never picked it up — so the base URLs here have been pointing at the retired `*.netcetera-cloud-payment.ch` hosts since.

Changes:
- Dropped `endpoint_prefix` from `NetceteraConfig`, the domain `ConnectorSpecificConfig::Netcetera`, and `NetceteraMeta`.
- Deleted `resolve_netcetera_base_url` and its three call sites; `base_url()` now returns the configured host directly, as the hyperswitch connector does.
- Updated the base URLs to match hyperswitch exactly: `https://api.prev.euc1.acquiring.cloud.netcetera.com/3ds-server` for dev/sandbox and `https://api.prod.euc1.acquiring.cloud.netcetera.com/3ds-server` for production.

Worth flagging since it widens the blast radius beyond the typed-contract work: this changes the host for both the typed and the legacy path, so the legacy path is no longer byte-for-byte identical to main. It is a fix rather than a regression — the previous hosts are the ones hyperswitch abandoned — but it is a runtime change and I have called it out in the PR description rather than leaving it under the byte-for-byte claim.

gopikrishna000: Correcting one thing I said above: I claimed the old hosts were retired and that the config was therefore broken. That is not right, and I had not checked before saying it.

Both still resolve. In production they are the same machine:

    flowbird.3ds-server.prod.netcetera-cloud-payment.ch  ->  3.74.255.88
    api.prod.euc1.acquiring.cloud.netcetera.com          ->  3.74.255.88

On prev they differ (35.158.11.132 vs 18.156.49.155) but both answer. So Netcetera kept the per-merchant subdomains working as aliases, and requests were reaching the right place as long as `endpoint_prefix` was set in the merchant's MCA metadata. If it was not set, the old code hard-errored before dispatch, which is a config failure rather than a wrong-host one.

The change still stands, but on the narrower argument: `endpoint_prefix` should not go into a new proto message for a mechanism hyperswitch has already dropped, because additive-only means we could not remove it cleanly later. Aligning the base URL with hyperswitch is then the consistent follow-through, and it is a no-op at the IP level in production.

Flagging the residual risk honestly: on prev this does move traffic from a merchant-specific host to the shared one. Hyperswitch has been on the shared host since July without issue, but that is hyperswitch's evidence rather than ours. If you would rather keep the base URL untouched here and take only the proto and plumbing removal, that is a reasonable call and I am happy to split it.

crates/types-traits/domain_types/src/router_data.rs
```diff
@@ -1557,6 +1569,7 @@ impl ConnectorSpecificConfig {
             Imerchantsolutions { api_key },
             Interpayments { api_key },
             Paydotcom { api_key },
+            Netcetera { base_url },
```
sahkal: why is base_url taken as connectorConfig where as for others its api_key?

sahkal (dismissed):

sahkal (dismissed):

sahkal (approved):

PR #2269, opened 2026-09-09
feat(connector): add ApplePay/GooglePay pre decryption, SetupMandate, and RepeatPayment flows for paydotcom

Shubhodip900:
## Description

What's added

Payment methods
- Apple Pay pre-decrypted (DPAN) — source_data.type = "network_token", token_type = "applepay". ECI and cryptogram are required; missing either is a hard error.
- Google Pay pre-decrypted (DPAN) — same path with token_type = "googlepay". Requires CRYPTOGRAM_3DS auth method; PAN_ONLY tokens are rejected.

Encrypted blob forwarding (direct Apple Pay / Google Pay SDK) is explicitly unsupported with an actionable error pointing to the DPAN path.

Flows
- SetupMandate — CIT via POST /v1/charges with source_data.setup_future_usage = "off_session". Response extracts underlying_network_id and returns it as mandate_reference.connector_mandate_id. Supports cards and pre-decrypted wallets.
- RepeatPayment — MIT via POST /v1/charges with off_session = true. Selects the richest available variant: Variant C (pm_id + underlying_network_id) → Variant B (pm_id only) → Variant A (customer_reference_id only).

Structural changes

- PaydotcomSourceData extended with network_token and setup_future_usage optional fields; card made optional.
- PaydotcomChargeResponse / PaydotcomHoldResponse — added underlying_network_id: Option<String>.
- PaydotcomMitSourceRef.id and PaydotcomMitSourceData.underlying_network_id typed as Secret<String>.
- specs.json updated with PaymentService/SetupRecurring and RecurringPaymentService/Charge.

### Additional Changes

## How did you test it?

No creds available. Use `cargo build` and `cargo run --bin grpc-server` .


crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -676,12 +728,172 @@ fn build_create_resource_request<T: PaymentMethodDataTypes>(
...
+                };
+                (!addr.is_empty()).then_some(addr)
+            },
+        });
+
+    match wallet_data {
+        WalletData::ApplePay(apple_pay) => match &apple_pay.payment_data {
+            ApplePayPaymentData::Decrypted(decrypted) => {
+                let cryptogram_data = &decrypted.payment_data;
+                let eci = cryptogram_data.eci_indicator.clone().ok_or_else(|| {
```
JeevaRamu0104: `eci_indicator` is optional on the decrypted payload and Amex/Discover Apple Pay tokens routinely omit it — this rejects them client-side. Every other connector passes it through as an `Option` (`adyen/transformers.rs:2497`, `tesouro/transformers.rs:347`). Does Pay.com actually reject `network_token` without an ECI?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +955,196 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+            >,
+            T,
+        >,
+    ) -> Result<Self, Self::Error> {
+        let item = &value.router_data;
+
+        // SetupMandate always creates a new Charge/Hold; there is no Confirm leg.
+        is_manual_capture(item.request.capture_method)?;
+
+        let minor_amount = item.request.minor_amount.ok_or_else(|| {
```
JeevaRamu0104: `minor_amount` is `Option` because zero-amount card-on-file setup is the normal shape for this flow — erroring here blocks it before any wire call. Default to `MinorUnit::zero()`?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +955,196 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
                             Some(build_authentication_context(
                                 item.request.browser_info.as_ref(),
                             )?),
-                            // A 3DS Authorize that was not opened by PreAuthenticate can only
-                            // finish here if the authentication turns out frictionless, so ask
-                            // for `automatic` rather than forcing a challenge nobody can answer.
-                            // The challenge journey runs PreAuthenticate first (see paydotcom.rs).
-                            PaydotcomThreeDsRequest::Automatic,
-                        ),
+                            PaydotcomThreeDsRequest::Challenge,
```
JeevaRamu0104: This forces a challenge regardless of `auth_type`, and there is no Confirm leg for SetupMandate (line 1098) — when Pay.com returns an `AuthenticationSession` the mandate arm yields `None` at line 1233 and the mandate is never stored. A merchant on `NoThreeDs` also gets a hard `MissingRequiredField: browser_info`. Can this route through `three_ds_mode()` like Authorize does?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1157,112 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                        .request
+                        .customer_id
+                        .as_ref()
+                        .map(|id| id.get_string_repr().to_string()),
+                    payment_method_options: None,
+                    source_data,
+                })))
             }
+
+            _ => Err(only_cards_error()),
```
JeevaRamu0104: nit: `only_cards_error()` says "Only card payments are supported" but the arm above handles wallets — the Authorize fallthrough uses the updated message. It also leaks into `data/field_probe/paydotcom.json` under `token_setup_recurring`.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1157,112 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+    ) -> Result<Self, Self::Error> {
+        let status = item.response.attempt_status();
+
+        // Extract `underlying_network_id` as the mandate reference.
+        let mandate_reference = match &item.response {
+            PaydotcomPaymentsResponse::Charge(charge) => {
+                charge.underlying_network_id.clone().map(|id| {
+                    Box::new(MandateReference {
+                        connector_mandate_id: Some(id),
+                        payment_method_id: None,
```
JeevaRamu0104, line 1363: `payment_method_id: None` means Pay.com's `pm_…` id is never stored, so RepeatPayment's Variant B/C arms are unreachable and every MIT falls to the empty Variant A. Hyperswitch overwrites `payment_method_id` with its own id anyway — round-trip the connector token through `mandate_metadata` instead (`airwallex/transformers.rs:1544-1548`). Does the charge response carry a `source` field?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2368,177 @@ pub fn hold_cancel_path(
...
+                .connector_request_reference_id
+                .clone(),
+        );
+
+        // `connector_mandate_id` carries the `underlying_network_id` that Pay.com returned
+        // during the CIT (SetupMandate). `payment_method_id` carries the `pm_card_…` id.
+        let (connector_mandate_id, payment_method_id) = match &item.request.mandate_reference {
+            MandateReferenceId::ConnectorMandateId(mandate_data) => (
+                mandate_data.get_connector_mandate_id(),
+                mandate_data.get_payment_method_id().cloned(),
```
JeevaRamu0104: `get_payment_method_id()` returns Hyperswitch's own payment-method id, not Pay.com's `pm_…` — Hyperswitch overwrites it after SetupMandate. Airwallex hits the same thing and reads the connector token back from `mandate_metadata` (`airwallex/transformers.rs:2664`). As written, Variants B and C send Pay.com an id it never issued.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2368,177 @@ pub fn hold_cancel_path(
...
+        //
+        // Variant C — PM id + network mandate id both present: most explicit; preferred
+        //             when multiple recurring sequences may exist for the same customer.
+        // Variant B — PM id only: let Pay.com choose the correct sequence internally.
+        // Variant A — neither: Pay.com auto-selects the latest eligible PM for the customer.
+        let (source, source_data) = match (payment_method_id, connector_mandate_id) {
+            (Some(pm_id), Some(network_id)) => (
+                None,
+                Some(PaydotcomMitSourceData {
+                    source_type: "source".to_string(),
```
JeevaRamu0104: nit: bare `String` where the file's other source types are a serde enum (`PaydotcomSourceType`).

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2368,177 @@ pub fn hold_cancel_path(
...
+            _ => (None, None),
+        };
+
+        Ok(Self {
+            off_session: true,
+            currency: item.request.currency,
+            amount,
+            // `customer_id` on RepeatPaymentData corresponds to the customer_reference_id
+            // used in the original CIT — it is optional but Pay.com uses it for PM selection.
+            customer_reference_id: None, // RepeatPaymentData has no dedicated customer_id field
```
JeevaRamu0104: `PaymentFlowData.customer_id` is on `item.resource_common_data` — it is what the CIT sent as `customer_reference_id`. Hardcoding `None` means Variant A (the `_ => (None, None)` fallback above) posts `{off_session, currency, amount, reference}` with nothing identifying a customer or payment method.

```suggestion
            customer_reference_id: item
                .resource_common_data
                .customer_id
                .as_ref()
                .map(|id| id.get_string_repr().to_string()),
```

crates/integrations/connector-integration/src/connectors/paydotcom.rs
```diff
@@ -445,6 +460,78 @@ macros::macro_connector_implementation!(
...
+        }
+
+        fn get_url(
+            &self,
+            req: &RouterDataV2<RepeatPayment, PaymentFlowData, RepeatPaymentData<T>, PaymentsResponseData>,
+        ) -> CustomResult<String, IntegrationError> {
+            // MIT charges always go to `/v1/charges`; Pay.com has no hold (manual-capture)
+            // variant for off-session transactions because no cardholder is present to
+            // approve a later capture.
+            Ok(format!("{}{}", self.connector_base_url_payments(req), PATH_CHARGES))
```
JeevaRamu0104, line 568: `RepeatPaymentData.capture_method` is never checked, so a merchant asking for MANUAL on an MIT gets funds captured immediately with no error. If Pay.com really has no off-session hold, reject it the way `is_manual_capture` rejects `ManualMultiple` rather than silently ignoring the field.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -662,26 +763,242 @@ fn build_create_resource_request<T: PaymentMethodDataTypes>(
...
+                                .peek()
+                                .to_string(),
+                        ),
+                        token_type: PaydotcomNetworkTokenType::Applepay,
+                        expiry_month: decrypted.get_expiry_month(),
+                        expiry_year: decrypted.get_four_digit_expiry_year(),
+                        three_ds: PaydotcomNetworkTokenThreeDs { eci, cryptogram },
+                    }),
+                    billing_details,
+                    setup_future_usage: None,
```
Vani-1107: why is this None?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -662,26 +763,242 @@ fn build_create_resource_request<T: PaymentMethodDataTypes>(
...
+                                .peek()
+                                .to_string(),
+                        ),
+                        token_type: PaydotcomNetworkTokenType::Googlepay,
+                        expiry_month,
+                        expiry_year,
+                        three_ds: PaydotcomNetworkTokenThreeDs { eci, cryptogram },
+                    }),
+                    billing_details,
+                    setup_future_usage: None,
```
Vani-1107: why is this None ?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1295,112 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                charge.underlying_network_id.clone().map(|id| {
+                    Box::new(MandateReference {
+                        connector_mandate_id: Some(id),
+                        payment_method_id: None,
+                        connector_mandate_request_reference_id: None,
+                        mandate_metadata: None,
+                    })
+                })
+            }
+            PaydotcomPaymentsResponse::Hold(hold) => hold.underlying_network_id.clone().map(|id| {
```
Vani-1107, line 1373: Is same behavior expected in case of both Hold and Charge status?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1191,6 +1821,10 @@ pub struct PaydotcomHoldResponse {
     pub failure_code: Option<String>,
     #[serde(default)]
     pub failure_message: Option<String>,
+    /// The network transaction identifier returned by Pay.com after mandate setup.
+    /// Used as `connector_mandate_id` in the SetupMandate response.
+    #[serde(default)]
```
Vani-1107: is this needed?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2530,177 @@ pub fn hold_cancel_path(
...
+    /// Always `true` for MIT — signals that no cardholder interaction is available.
+    pub off_session: bool,
+    #[serde(serialize_with = "paydotcom_currency::serialize")]
+    pub currency: common_enums::Currency,
+    pub amount: MinorUnit,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub customer_reference_id: Option<String>,
+    /// Variant B: explicit payment method id (e.g. `pm_card_…`).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub source: Option<String>,
```
Vani-1107: Should this be Secret ?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
Shubhodip900: Fixed. source: Option<String> was added to PaydotcomChargeResponse/PaydotcomHoldResponse and is stored in mandate_metadata (lines 1345, 1358). RepeatPayment reads the pm_… id from mandate_metadata via get_mandate_metadata() (line 2638),

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
Shubhodip900: yes

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +1051,224 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                                .as_ref()
+                                .map(|customer_id| customer_id.get_string_repr().to_string()),
+                            three_ds,
+                            authentication_context,
+                            request_threed_secure,
+                        )?)))
+                    }
+
+                    PaymentMethodData::Wallet(wallet_data) => {
+                        let sfu = item.request.setup_future_usage.and_then(|u| {
```
JeevaRamu0104: The card arm never does this — `build_create_resource_request` hardcodes `setup_future_usage: None` (line 803). A card payment carrying `setup_future_usage = OffSession` through Authorize never tells Pay.com to store the method, and `transaction_response` returns `mandate_reference: None` (line 2008), so nothing comes back either. Why do wallets and cards differ here?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +1051,224 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
                             Some(build_authentication_context(
                                 item.request.browser_info.as_ref(),
                             )?),
-                            // A 3DS Authorize that was not opened by PreAuthenticate can only
-                            // finish here if the authentication turns out frictionless, so ask
-                            // for `automatic` rather than forcing a challenge nobody can answer.
-                            // The challenge journey runs PreAuthenticate first (see paydotcom.rs).
-                            PaydotcomThreeDsRequest::Automatic,
-                        ),
+                            PaydotcomThreeDsRequest::Challenge,
```
JeevaRamu0104: The NoThreeDs branch below fixes half of this, but the ThreeDs branch still asks for `Challenge` — and line 1211 still says there is no Confirm leg for SetupMandate, while line 1376 maps `AuthenticationSession` to no mandate reference. The shopper completes the challenge and no mandate is stored. Authorize asks for `Automatic` in the same position for exactly this reason.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1281,131 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+    ) -> Result<Self, Self::Error> {
+        let status = item.response.attempt_status();
+
+        // Extract `underlying_network_id` as the mandate reference.
+        // Round-trip Pay.com's pm_… id through `mandate_metadata` so RepeatPayment can
+        // use it for Variant B/C without conflating it with Hyperswitch's own
+        // payment_method_id (see RepeatPayment transformer).
+        let mandate_reference = match &item.response {
+            PaydotcomPaymentsResponse::Charge(charge) => {
+                charge.underlying_network_id.clone().map(|id| {
```
JeevaRamu0104, line 1360: `mandate_reference` is built inside `.map()` on `underlying_network_id`, so a response carrying `source` but no network id yields no mandate at all — the setup call succeeds and the caller has no signal that nothing was stored. Should a missing `underlying_network_id` be an error instead?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1148,49 +1785,54 @@ impl From<PaydotcomRefundStatus> for RefundStatus {
...
     pub failure_message: Option<String>,
+    /// The network transaction identifier returned by Pay.com after mandate setup.
+    /// Used as `connector_mandate_id` in the SetupMandate response.
+    pub underlying_network_id: Option<String>,
+    /// Pay.com payment-method id (`pm_card_…`) tied to this charge, when the charge was
+    /// created from a stored payment method. Deserialized defensively — omitted if Pay.com
+    /// does not include it in the response. When present, stored in `mandate_metadata` so
+    /// RepeatPayment can use it for Variant B/C without conflating it with Hyperswitch's own
+    /// payment_method_id (which overwrites the field after SetupMandate).
+    pub source: Option<String>,
```
JeevaRamu0104: `source`'s wire shape is unverified and this struct backs Authorize, PSync, Capture, Void and SetupMandate. Missing is safe; a type mismatch is not — if Pay.com returns an object rather than a bare id, all five flows fail to deserialize. `serde_json::Value` (or `{ id }` with `#[serde(default)]`) fails soft at no cost. Has this been seen on a live response?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2522,230 @@ pub fn hold_cancel_path(
...
+                Some(PaydotcomMitSourceData {
+                    source_type: PaydotcomSourceType::Source,
+                    source: PaydotcomMitSourceRef {
+                        id: Secret::new(pm_id),
+                    },
+                    underlying_network_id: Some(Secret::new(network_id)),
+                }),
+            ),
+            (Some(pm_id), None) => (Some(Secret::new(pm_id)), None),
+            _ => (None, None),
```
JeevaRamu0104, line 2734: A `ConnectorMandateId` holding neither a metadata pm id nor a connector mandate id, on a payment with no customer id, still posts `{off_session, currency, amount, reference}` with nothing identifying a payment source. Much narrower than before — an explicit `MissingRequiredField` here beats a Pay.com 400.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +1051,239 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                                    )?),
+                                    // A 3DS Authorize that was not opened by PreAuthenticate can
+                                    // only finish here if the authentication turns out frictionless,
+                                    // so ask for `automatic` rather than forcing a challenge nobody
+                                    // can answer.
+                                    PaydotcomThreeDsRequest::Automatic,
+                                ),
+                            };
+
+                        // Mirror the wallet arm: if the Authorize carries
```
JeevaRamu0104, line 1094: This only covers the Create leg. When the orchestrator drives PreAuthenticate to Authenticate to Authorize, the Charge/Hold is created by PreAuthenticate (line 1507), which calls `build_create_resource_request` directly — and that hardcodes `setup_future_usage: None` at line 803. The Authorize leg is then `Confirm` with an empty body, so the flag never travels.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -748,39 +1051,239 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                            &item.resource_common_data,
+                            item.request.email.clone(),
+                            item.request
+                                .customer_id
+                                .as_ref()
+                                .map(|customer_id| customer_id.get_string_repr().to_string()),
+                            three_ds,
+                            authentication_context,
+                            request_threed_secure,
+                        )?;
```
JeevaRamu0104, line 1120: Pay.com stores the method now, but `transaction_response` hardcodes `mandate_reference: None` (line 2041), so the `underlying_network_id` and `source` that come back are dropped. Nothing usable for a later MIT reaches the caller — request side wired, response side not. Can Authorize inject the mandate reference the way SetupMandate does at line 1367?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1296,151 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+        };
+
+        let connector_feature_data = item.response.pending_metadata().map(Secret::new);
+
+        let response = match status {
+            AttemptStatus::Failure => Err(item.response.in_band_error(item.http_code, status)),
+            // Charge/Hold terminal-success with no underlying_network_id means Pay.com did
+            // not store the mandate — the setup call succeeded but the result is unusable for
+            // MIT. Surface this as a failure rather than returning an empty mandate silently.
+            AttemptStatus::Charged | AttemptStatus::Authorized if mandate_reference.is_none() => {
```
JeevaRamu0104: On the `Charged` path Pay.com has already taken the money — `/v1/charges` is the auto-capture endpoint — and this returns `Failure` without voiding or refunding. `resource_common_data.status` is still `Charged` from line 1399, so the two disagree. `Authorized` is recoverable by voiding; `Charged` is not. Should the two cases split?

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -793,12 +1296,151 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                Err(ErrorResponse {
+                    status_code: item.http_code,
+                    code: NO_ERROR_CODE.to_string(),
+                    message: NO_ERROR_MESSAGE.to_string(),
+                    reason: None,
```
JeevaRamu0104: `NO_ERROR_CODE` / `NO_ERROR_MESSAGE` / `reason: None` gives the merchant nothing on a call that reached Pay.com and succeeded there. Set `reason` to say the response carried no `underlying_network_id`.

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
```diff
@@ -1896,6 +2555,257 @@ pub fn hold_cancel_path(
...
+        };
+
+        let customer_reference_id = item
+            .resource_common_data
+            .customer_id
+            .as_ref()
+            .map(|id| id.get_string_repr().to_string());
+
+        // Variant A (customer_reference_id only) requires at least the customer id.
+        // Without it AND without a pm_id / mandate id, the request carries nothing
```
JeevaRamu0104: This guard defeats the field prober, which now reports `recurring_charge: error` ("Stuck on field: mandate_reference or customer_id"). That dropped the `RecurringPaymentService.Charge` section from `docs-generated/connectors/paydotcom.md`, removed the flow from `llms.txt`, and deleted the recurring example from all four `examples/paydotcom/*` files — while `specs.json` still declares it. Requiring `customer_id` unconditionally would keep it probeable.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
Shubhodip900: Added NOTE: explaining it only applies to the direct Create leg; gateway-3DS (PreAuthenticate → Confirm) always has it hardcoded to None

crates/integrations/connector-integration/src/connectors/paydotcom/transformers.rs
Shubhodip900: Added the same mandate-reference extraction block in the Authorize response transformer that SetupMandate has (lines 1366-1391), then injects it into the TransactionResponse

JeevaRamu0104 (dismissed):

Vani-1107 (approved):

JeevaRamu0104 (approved):

PR #2271, opened 2026-09-09
fix(psync): run connector pre-flight validation before dispatching a PSync

gopikrishna000:
## Description

`validate_psync_reference_id` existed on `ConnectorValidation`, but that trait is not a supertrait of `ConnectorServiceTrait`, so `PaymentService::get` never called it and the overrides on Adyen, Calida, Razorpay and Zift were dead code. Hyperswitch's direct integration path runs this check before every PSync and skips the connector call when it fails (`psync_flow.rs` `build_flow_specific_connector_request`). UCS did not, and two production symptoms followed:

- **Adyen, PSync without `encoded_data`** (the SDK's post-redirect force sync, and every scheduler retry): `build_request_v2` returns `None`, no HTTP call is made, and the flow answers with the untouched default `ErrorResponse` (`HE_00`, "Something went wrong", `status_code: 500`) as a gRPC-OK reply. Hyperswitch persists that on the attempt and returns it to the SDK, which reads the error block as a failed payment. For merchant `gooutdoors` this left 13 authorized 3DS payments uncaptured in 7 days; the same payments on the direct path capture at 92%.
- **Cybersource, PSync with an empty `connector_transaction_id`** (redirect return racing the webhook-driven Authorize): the empty string was wrapped as `ResponseId::ConnectorTransactionId("")`, so the connector's own missing-id check did not fire and UCS called `tss/v2/transactions/` with an empty path segment, producing an HTML 404 that the kill switch counts as `connector_outcome`.

### Changes

1. `ValidationTrait::validate_psync_reference_id` (moved from `ConnectorValidation`, same default: a connector transaction id is required). The four existing connector overrides move to their `ValidationTrait` impls unchanged.
2. `PaymentService::get` calls it after building `PaymentFlowData` and before router data. On failure it returns a no-op reply without calling the connector: gRPC OK, `PAYMENT_STATUS_UNSPECIFIED`, no `error`, `connector_transaction_id` and `merchant_transaction_id` echoed. The rejection reason is logged at info level.
3. `PaymentsSyncData::foreign_try_from(PaymentServiceGetRequest)` maps an empty `connector_transaction_id` to `ResponseId::NoResponseId`.

### Why a no-op reply rather than a gRPC error

UCS is stateless and cannot return the caller's current status, so it cannot literally "skip and keep status". A gRPC error would surface to the merchant as a failed sync (hyperswitch's `psync_gateway` returns any `payment_get` error to the caller), which is the symptom being fixed. Hyperswitch already maps `PaymentStatus::Unspecified` to "keep the previous attempt status" and writes no error when the reply carries none, so the no-op reply reproduces the direct path's skip with no hyperswitch change and no connector knowledge outside UCS.

## Motivation and Context

Prod evidence, merchant `gooutdoors` on Adyen, 7 days: 13 payments authorized through UCS whose browser force sync received `HE_00`, 0 captured; 68 whose force sync stayed on the direct path, 62 captured. Cybersource `flowbird`: 5 PSync 404s from empty transaction ids on 2026-09-09, all counted as kill switch failures.

### Additional Changes

## How did you test it?

tested locally

image

got no error field
Only attempt status as unspecified, hyperswitch can use previous status

## Checklist

- [x] I formatted the code `cargo +nightly fmt --all`
- [ ] I addressed lints thrown by `cargo clippy`
- [x] I reviewed the submitted code

### DIFF
--- a/crates/grpc-server/grpc-server/src/server/payments.rs
+++ b/crates/grpc-server/grpc-server/src/server/payments.rs
@@ -1098,11 +1098,7 @@ impl PaymentService for Payments {
                     // status" and write no error for it.
                     if let Err(validation_error) = connector_data.connector.validate_psync_reference_id(
                         &payments_sync_data,
-                        matches!(
-                            payment_flow_data.auth_type,
-                            common_enums::AuthenticationType::ThreeDs
-                        ),
-                        payment_flow_data.status,
+                        &payment_flow_data,
                     ) {
                         // Log with the same fields `IntoGrpcStatus` would emit, while the
                         // report's frames (connector-supplied context) are still attached;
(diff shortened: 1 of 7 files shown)


JeevaRamu0104 (dismissed):

JeevaRamu0104 (dismissed):

crates/grpc-server/grpc-server/src/server/payments.rs
```diff
@@ -1085,6 +1085,44 @@ impl PaymentService for Payments {
...
+                    // "skip" here is a no-op reply: gRPC OK, `PAYMENT_STATUS_UNSPECIFIED`, no
+                    // error. Callers already treat an unspecified status as "keep the previous
+                    // status" and write no error for it.
+                    if let Err(validation_error) = connector_data.connector.validate_psync_reference_id(
+                        &payments_sync_data,
+                        matches!(
+                            payment_flow_data.auth_type,
+                            common_enums::AuthenticationType::ThreeDs
+                        ),
+                        payment_flow_data.status,
```
AmitsinghTanwar007: logic part should reside behind connector abstraction

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2276, opened 2026-09-09
feat(superposition): remote workspace with file fallback, déjà read boundary, sampler targeting key

iemyashasvi:
## Description

Builds on #2132 (Saurav's commit `274a6534a` is kept as-is; every change here is in the commits on top of it) to make Superposition a **remote, runtime-editable policy source** for Prism, with the baked `config/superposition.toml` as the fallback, and to make the déjà sampler use it end to end.

**Source selection at boot (`common_utils::superposition_config`)**

- New typed `[superposition]` table → `SuperpositionClientConfig { source, endpoint, token, org_id, workspace_id, polling_interval, request_timeout, backup_file_path }` (same shape as hyperswitch's `SuperpositionClientConfig`; `source` is the one addition). `source: SuperpositionSource` is a two-value enum, `file` | `remote`, and the same enum `SuperpositionConfig::source()` reports at boot.
- `source = "remote"` → `HttpDataSource` (POST `/config`, `/experiment-config`) as primary, `FileDataSource` (backup path or the baked file) as fallback, `RefreshStrategy::Polling`. `source = "file"` (the shipped default) → `FileDataSource` + `Watch`, i.e. exactly what #2132 does today.
- **Boot never fails because of Superposition.** An incomplete remote block is reported and the source set back to `file` (`validate()` is a `fp_utils::when(...).and_then(...)` chain, one exit path); an unreachable remote initialises from the fallback file; a bad fallback degrades to the baked file. `SuperpositionConfig::source()` / `experiments_supported()` say which source won, and boot logs it.
- `resolve_with(dimensions, targeting_key)` replaces the connector-only `resolve`: the targeting key is the experiment bucketing identifier (not a dimension), which is what native Superposition experiments need.

**déjà**

- Every Superposition read is a **déjà read boundary** (`kind = superposition`, `ReplayStrategy::Substitute`). On replay the recorded policy value is substituted, so a policy edit between record and replay cannot change the candidate's behaviour; a novel read misses gracefully into a typed `ResolutionError` and the caller falls back to static config. Mirrors hyperswitch's `deja_boundary::read` around its three Superposition reads.
- Sampler (`grpc-server/src/deja/sampler.rs`): resolves `{environment, rpc_method, rpc_service}` with the request id as targeting key, bounded by `[deja.sampler] timeout_ms` (25 ms, hyperswitch's value). Timeout / error → `fail_closed` policy, once-per-rpc warning. The decision memo is gone: it was unsound under a refreshing source.

**Observability**

- `SUPERPOSITION_RESOLVE_TOTAL{consumer, outcome}` (consumer ∈ `sampler`, `connector_urls`; outcome ∈ hit / miss / key_missing / timeout / error) and `SUPERPOSITION_SOURCE{kind}` gauge (`remote` / `file`).

**Ops**

- `scripts/seed_superposition.sh`: idempotently seeds a workspace from `config/superposition.toml` (dimensions → default-config → contexts, 409-tolerant) and self-checks with `POST /config/resolve`. Same shape as hyperswitch's seed script.

## Motivation and Context

Today the sampling and connector-URL policy is image-baked: changing which requests déjà records (or a connector base URL) needs a rebuild and a rollout. Hyperswitch already reads its policy from a Superposition workspace with the file as fallback and treats those reads as déjà boundaries; this brings Prism to parity so the same workspace/dashboard workflow (contexts, native experiments with ramps) drives Prism, and so replays stay deterministic regardless of what the policy says at replay time.

Design and change spec reviewed in the déjà-in-UCS RFC set (`docs/rfcs/deja-ucs-superposition-{design,changes}.md`, to land separately with the docs).

### Additional Changes

New `[superposition]` table (default `source = "file"`) and `[deja.sampler] timeout_ms` in `config/development.toml`, `config/sandbox.toml`, `config/production.toml`. No behaviour change for a deployment that does not set `superposition.source = "remote"`. Env override prefix as usual: `CS__SUPERPOSITION__{SOURCE,ENDPOINT,TOKEN,ORG_ID,WORKSPACE_ID,POLLING_INTERVAL}`.

## How did you test it?

Unit / integration tests (all green with the `deja` feature on and off, `clippy -D warnings` clean):

- `common_utils` — settings default, `validate()` rejects each missing remote field, dimension/targeting-key helpers.
- `ucs_env/tests/superposition_config_load.rs` — loads through the real config loader: `source = "remote"` with nothing else set boots with the source back on `file`; a complete block loads as written; the shipped development config is file-first.
- `grpc-server/tests/test_superposition_config.rs` — Saurav's two file-watch tests plus: `file` settings select the baked file; unreachable remote initialises from the fallback file and serves its policy; targeting key is inert without experiments and is not a dimension; unreachable remote with a bad fallback degrades to the baked file.
- `grpc-server/tests/deja_superposition_boundary.rs` — under a déjà replay hook with an empty lookup table, `resolve` misses gracefully into `ResolutionError` while `resolve_with` (outside the boundary) still resolves.
- `sampler.rs` — 11 tests incl. a policy edit on disk reaching the next decision and bounded per-rpc failure warnings.

```bash
cargo test -p common_utils -p ucs_env -p grpc-server --features deja
cargo test -p common_utils -p ucs_env -p grpc-server
cargo clippy --workspace --all-targets --features deja -- -D warnings
```

Manual: the file-mode path is the same code the `deja-sampler-sandbox-50` branch ran on the sandbox custom pod (50 % sampling verified 22/22 id-for-id). Remote mode is to be verified on the sandbox custom pod once the `prism` workspace exists and the values carry the endpoint/token (infra PR to follow).


crates/common/common_utils/src/superposition_config.rs
```diff
@@ -1,44 +1,220 @@
...
-    #[error("Failed to read superposition config file '{path}': {source}")]
-    FileReadError {
-        path: String,
-        source: std::io::Error,
-    },
-    /// Failed to parse the TOML configuration
-    #[error("Failed to parse superposition.toml: {0}")]
-    ParseError(String),
-    /// Failed to resolve configuration for given context
-    #[error("Failed to resolve configuration: {0}")]
```
AmitsinghTanwar007: why this errors are removed they might be required for the errors when you are using the fallback of file read

crates/common/common_utils/src/superposition_config.rs
```diff
@@ -1,44 +1,220 @@
...
+        if self.org_id.trim().is_empty() {
+            return invalid("superposition.org_id cannot be empty");
+        }
+        if self.workspace_id.trim().is_empty() {
+            return invalid("superposition.workspace_id cannot be empty");
+        }
+        Ok(())
+    }
+}
+
```
AmitsinghTanwar007, line 127: try to avoid early returns

config/development.toml
```diff
@@ -258,6 +258,23 @@ keys = ["x-request-id","x-merchant-id","x-lineage-ids","x-reference-id","x-conne
...
 [connector_response_masking.connector_keys]
+# --- Superposition: where policy comes from ---------------------------------
+# Off (default): the baked config/superposition.toml, watched for changes.
+# On: the same provider over a remote workspace, polled every polling_interval
+# seconds, with the baked file as the provider's init-time fallback. Boot never
+# fails because of Superposition: enabling without endpoint/token/org/workspace,
+# or an unreachable workspace, is reported and the baked file serves policy.
+# Env overrides: CS__SUPERPOSITION__*.
+[superposition]
+enabled = false
```
AmitsinghTanwar007: this should be superposition enum which has two values file /remote something like this

crates/common/common_utils/src/superposition_config.rs
```diff
@@ -1,44 +1,220 @@
...
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            endpoint: String::new(),
+            token: Secret::new(String::new()),
+            org_id: String::new(),
+            workspace_id: String::new(),
+            polling_interval: 15,
+            request_timeout: None,
+            backup_file_path: None,
```
iemyashasvi: the default should point to superposition.toml

crates/common/common_utils/src/superposition_config.rs
```diff
@@ -48,19 +224,46 @@ impl SuperpositionConfig {
...
-            std::fs::read_to_string(path).map_err(|e| SuperpositionConfigError::FileReadError {
-                path: path.to_string(),
-                source: e,
-            })?;
+    pub async fn from_file(path: &str) -> Result<Self, SuperpositionConfigError> {
+        let source = FileDataSource::new(PathBuf::from(path))
+            .map_err(SuperpositionConfigError::InitializationError)?;
+        Self::with_sources(
+            Box::new(source),
+            None,
```
iemyashasvi: why none?

gopikrishna000 (dismissed):

crates/common/external-services/src/shared_metrics.rs
```diff
@@ -58,6 +58,24 @@ lazy_static! {
...
+        "Superposition policy resolutions by consumer and outcome",
+        &["consumer", "outcome"]
+    )
+    .unwrap();
+    // Set once at boot: which source this process resolves policy from (remote | file).
+    // A pod that fell back to the file will not follow the workspace — worth a panel.
+    pub static ref SUPERPOSITION_SOURCE: IntGaugeVec = register_int_gauge_vec!(
+        "SUPERPOSITION_SOURCE",
+        "Superposition source this process resolves policy from (1 = active)",
+        &["kind"]
```
AmitsinghTanwar007, line 76: its correct but now we are using otel

AmitsinghTanwar007 (approved):

gopikrishna000 (approved):

PR #2286, opened 2026-09-10
feat: add generated connector sanity layer

10X-GRACE:
## Summary
- Add `OsBasedReturnUrl` and `optional os_based_return_url` to `AuthenticatorClientAuthenticationContext` for Plaid client-auth redirect configuration.
- Add descriptor-driven `auto_populate` codegen that finds request/message types from the protobuf `FileDescriptorSet` and generates typed populate traits plus service wrappers.
- Add a feature-gated `grpc-server` sanity layer for Plaid that reads Euler-only redirect keys from raw `x-connector-config`, populates `os_based_return_url.return_url_map`, and strips those extra keys before downstream typed config parsing.
- Keep normal UCS runtime path untouched unless `connector-sanity-layer` is enabled.

## How it works
At build time, `grpc-api-types` reads the compiled protobuf descriptor set and generates `auto_populate_generated.rs`. For each field declared in `FIELD_SPECS`, the generator emits a trait such as `PopulateOsBasedReturnUrl`. Request messages that directly contain the field get real setter implementations; request messages that do not support it get no-op implementations. Nested and oneof request shapes delegate to the active child context when the field is reachable there.

The generated `SanityLayer<S, Z>` wraps tonic service implementations. Before forwarding to the real handler, it splits the `tonic::Request` into metadata, extensions, and body, lets the sanitizer mutate metadata/body, then rebuilds the request. This lets the Plaid sanity layer both populate the typed request and clean raw metadata before the existing connector config parser runs.

For Plaid, the runtime sanitizer reads these raw config keys when present:
- `ios_redirect_uri`
- `android_package_name`
- `web_redirect_uri`

It builds `OsBasedReturnUrl { os_type: String::new(), return_url_map }` and calls `populate_os_based_return_url`. The generated setter only fills the map when the request already contains `os_based_return_url` with a non-empty `os_type`, preserving the request-provided OS selection.

## Enablement
The runtime wrapper is disabled by default. Enable it when running/building `grpc-server` with:

```bash
cargo run -p grpc-server --features connector-sanity-layer
```

`ServiceType::Grpc` vs `ServiceType::Http` is still selected through config:

```toml
[server]
type = "grpc" # or "http"
```

When the feature is disabled, `crate::sanity_layer::wrap(service)` returns the original service and the generated wrapper is not in the runtime call path. When enabled, it returns the generated `SanityLayer` and sanity runs before handlers for both gRPC and HTTP service calls.

## Validation
Ran:

```bash
cargo fmt -p grpc-api-types -p grpc-server
cargo check -p grpc-server
cargo check -p grpc-server --features connector-sanity-layer
cargo test -p grpc-api-types --test auto_populate_test
cargo build -p grpc-server --features connector-sanity-layer
```

Live gRPC smoke used `types.MerchantAuthenticationService/CreateClientAuthenticationToken` with `x-auth-connector: plaid`, request body containing `authenticator.osBasedReturnUrl.osType = "ios"`, and `x-connector-config` containing the three Euler-only Plaid redirect keys.

Observed expected downstream failure:

```text
InvalidArgument: Missing required field: customer. customer is required for Plaid Link token creation
```

Before that failure, request logs showed the sanity layer populated:

```json
"os_based_return_url": {
  "os_type": "ios",
  "return_url_map": {
    "ios_redirect_uri": "https://merchant.example/plaid-ios-return",
    "android_package_name": "com.example.android",
    "web_redirect_uri": "https://merchant.example/plaid-web-return"
  }
}
```

### DIFF
--- a/crates/grpc-server/grpc-server/Cargo.toml
+++ b/crates/grpc-server/grpc-server/Cargo.toml
@@ -101,6 +101,7 @@ serial_test = "3.2.0"
 default = []
 kafka = ["tracing-kafka", "common_utils/kafka"]
 connector-request-kafka = ["dep:connector_request_kafka", "external-services/connector-request-kafka"]
+connector-sanity-layer = []
 # Push gRPC metrics over OTLP to an OpenTelemetry Collector, in addition to the
 # Prometheus /metrics scrape endpoint.
 otel = ["ucs_env/otel", "external-services/otel"]
--- a/crates/grpc-server/grpc-server/src/app.rs
+++ b/crates/grpc-server/grpc-server/src/app.rs
@@ -29,6 +29,27 @@ use crate::{
     utils,
 };
 
+type SanityLayer<S> = crate::sanity_layer::SanityLayer<S>;
+type CompositePaymentsService = composite_service::payments::Payments<
+    crate::server::payments::Payments,
+    crate::server::payments::MerchantAuthentication,
+    crate::server::payments::Customer,
+    crate::server::refunds::Refunds,
+    crate::server::payments::PaymentMethodAuthentication,
+>;
+type CompositeEventService = composite_service::events::CompositeEvents<
+    crate::server::events::EventServiceImpl,
+    crate::server::payments::MerchantAuthentication,
+>;
+type CompositePaymentMethodService = composite_service::payment_methods::PaymentMethods<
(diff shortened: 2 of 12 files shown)


10X-GRACE:
Closing because this PR needs to be raised from the author's GitHub account.

PR #2287, opened 2026-09-10
feat: add generated connector sanity layer

AmitsinghTanwar007:
## Summary
- Add `OsBasedReturnUrl` and `optional os_based_return_url` to `AuthenticatorClientAuthenticationContext` for Plaid client-auth redirect configuration.
- Add descriptor-driven `auto_populate` codegen that reads the protobuf `FileDescriptorSet` and generates typed populate traits plus service wrappers.
- Refactor auto-populate generation to use `quote`/`proc_macro2` with `AutoPopulateField` + `AutoPopulateSpec`, instead of string-built Rust setter bodies.
- Add a feature-gated `grpc-server` sanity layer with connector-based dispatch. Plaid reads Euler-only redirect keys from raw `x-connector-config` and populates `os_based_return_url.return_url_map`.
- Keep normal UCS runtime path untouched unless `connector-sanity-layer` is enabled. Metadata is inspected but not rewritten.

## How it works
At build time, `grpc-api-types` reads the compiled protobuf descriptor set and generates `auto_populate_generated.rs`. For each field listed in `AUTO_POPULATE_FIELDS`, the generator uses that field's `AutoPopulateSpec` implementation to emit a trait such as `PopulateOsBasedReturnUrl`.

Request messages that directly contain the field get real setter implementations. Request messages that do not support it get no-op implementations. Nested and oneof request shapes delegate to the active child context when the field is reachable there.

The generated `SanityLayer<S, Z>` wraps tonic service implementations. Before forwarding to the real handler, it clones request metadata for read-only inspection, lets the sanitizer mutate the typed request body, and then forwards the original request to the inner service. No metadata/header rewrite is performed.

The runtime sanitizer is connector-oriented:

```rust
match sanity_connector(metadata) {
    Some(SanityConnector::Plaid) => PlaidSanity.apply(metadata, req),
    Some(SanityConnector::Other(connector)) => {
        tracing::debug!(connector = %connector, "no connector sanity registered");
    }
    None => {}
}
```

Plaid behavior is behind a `ConnectorSanity` trait implementation. That keeps connector-specific config extraction and request mutation outside the generated code, while making future connector sanity handlers explicit.

For Plaid, the runtime sanitizer reads these raw config keys when present:
- `ios_redirect_uri`
- `android_package_name`
- `web_redirect_uri`

It builds `OsBasedReturnUrl { os_type: String::new(), return_url_map }` and calls `populate_os_based_return_url`. The generated setter only fills the map when the request already contains `os_based_return_url` with a non-empty `os_type`, preserving the request-provided OS selection.

## How to add another auto-populated field
1. Add the destination field to the relevant `.proto` message or messages.
2. Add a new variant to `AutoPopulateField` in `crates/types-traits/grpc-api-types/codegen/auto_populate.rs`.
3. Add that variant to `AUTO_POPULATE_FIELDS`.
4. Create a small spec type, for example `NativeAppIdentifierSpec`.
5. Implement `AutoPopulateSpec` for that spec type:
   - `field_name`: exact proto field name.
   - `trait_name`: generated Rust trait name.
   - `method_name`: generated Rust method name.
   - `value_type`: Rust type accepted by the generated setter.
   - `setter_body`: shape-based setter logic using `quote!` for `FieldShape::Optional` and `FieldShape::Required`; reject unsupported shapes with `compile_error!`.
6. Add delegation for the new variant in `impl AutoPopulateSpec for AutoPopulateField`.
7. Add or update focused tests in `grpc-api-types/tests/auto_populate_test.rs`.
8. Build or test `grpc-api-types`; the generated file in `OUT_DIR` will include trait impls for every descriptor-discovered matching message.

No request/message list is needed. The protobuf schema remains the source of truth.

If runtime connector code needs to call the new generated populate method, update the sanitizer bounds in `grpc-server/src/sanity_layer.rs` so the generic request type includes the new populate trait.

## How to add another connector sanity handler
1. Make sure the connector can be resolved to a `ConnectorVariant` by the existing metadata/config parsing path.
2. Add a connector-specific sanity struct in `crates/grpc-server/grpc-server/src/sanity_layer.rs`, for example `FooPaySanity`.
3. Implement `ConnectorSanity` for that struct:
   - `raw_config` extracts only that connector's raw config from `x-connector-config`.
   - `apply` converts connector-specific config keys into UCS's generic request value and calls the generated populate method.
4. Add one match arm in `sanity_connector` for the connector enum variant.
5. Add one match arm in `ConnectorSanitizer::sanitize` to call that connector's sanity implementation.
6. Leave unsupported connectors on the `Other(connector)` path; they should log `no connector sanity registered` and perform no mutation.
7. Add a focused test or smoke command that sends raw `x-connector-config` for the connector and confirms the typed request is populated before the handler runs.

The connector handler owns the translation from merchant/Euler-specific config shape into UCS's generic request shape. The auto-populate generator only owns finding destination fields and emitting typed setters.

## Enablement
The runtime wrapper is disabled by default. Enable it when running/building `grpc-server` with:

```bash
cargo run -p grpc-server --features connector-sanity-layer
```

`ServiceType::Grpc` vs `ServiceType::Http` is still selected through config:

```toml
[server]
type = "grpc" # or "http"
```

When the feature is disabled, `crate::sanity_layer::wrap(service)` returns the original service and the generated wrapper is not in the runtime call path. When enabled, it returns the generated `SanityLayer` and sanity runs before handlers for both gRPC and HTTP service calls.

## Validation
Ran:

```bash
cargo fmt -p grpc-api-types
cargo test -p grpc-api-types --test auto_populate_test
cargo clippy -p grpc-api-types --all-targets -- -D warnings
cargo check -p grpc-server --features connector-sanity-layer
cargo check -p grpc-server
```

Earlier live gRPC smoke used `types.MerchantAuthenticationService/CreateClientAuthenticationToken` with `x-auth-connector: plaid`, request body containing `authenticator.osBasedReturnUrl.osType = "ios"`, and `x-connector-config` containing the three Euler-only Plaid redirect keys.

Observed expected downstream failure:

```text
InvalidArgument: Missing required field: customer. customer is required for Plaid Link token creation
```

Before that failure, request logs showed the sanity layer populated:

```json
"os_based_return_url": {
  "os_type": "ios",
  "return_url_map": {
    "ios_redirect_uri": "https://merchant.example/plaid-ios-return",
    "android_package_name": "com.example.android",
    "web_redirect_uri": "https://merchant.example/plaid-web-return"
  }
}
```

## Scope note / follow-up
This PR is primarily for the generated sanity-layer infrastructure and typed request mutation path. It proves that the runtime sanity layer can read Euler-only Plaid redirect keys from raw `x-connector-config` and populate `AuthenticatorClientAuthenticationContext.os_based_return_url.return_url_map` before the handler sees the request.

One remaining product-flow step is intentionally separate from the generic sanity-layer work: the new `os_based_return_url` value still needs to be propagated through the domain request model and consumed by the Plaid client-auth transformer. Today the existing domain conversion only carries `return_url` and `native_app_identifier`, so the map populated by this PR is not yet selected into Plaid's final `redirect_uri` / `android_package_name` request fields. That follow-up should also normalize map keys to the runtime OS contract, e.g. `ios`, `android`, and `web`.


crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -3195,6 +3193,11 @@ message PaymentClientAuthenticationContext {
   optional string return_url = 8;
 }
 
+message OsBasedReturnUrl {
+  string os_type = 1;
+  map<string, string> return_url_map = 2;
```
AmitsinghTanwar007: should os_type be a enum?
@JeevaRamu0104

Aishwariyaa-Anand: @AmitsinghTanwar007 at hyperswitch we have an `enum ClientPlatform` for the same

crates/grpc-server/grpc-server/src/sanity_layer.rs
```diff
@@ -0,0 +1,152 @@
...
+            None => {}
+        }
+    }
+}
+
+#[cfg(feature = "connector-sanity-layer")]
+trait ConnectorSanity {
+    fn raw_config(&self, metadata: &MetadataMap) -> Option<serde_json::Value>;
+
+    fn apply<T: PopulateOsBasedReturnUrl>(&self, metadata: &MetadataMap, req: &mut T);
```
jarnura: we might need to have multiple functions specific to the override `PopulateOsBasedReturnUrl`. Apply function can have a default implementation which call all the override in the default implementation itself and the user just call apply only.

crates/grpc-server/grpc-server/src/sanity_layer.rs
```diff
@@ -0,0 +1,152 @@
...
+        .ok()?;
+    let json: serde_json::Value = serde_json::from_str(header_value).ok()?;
+    let config = json.get("config")?;
+    connector_keys
+        .iter()
+        .find_map(|key| config.get(*key).cloned())
+}
+
+#[cfg(feature = "connector-sanity-layer")]
+fn plaid_os_based_return_url(plaid_config: serde_json::Value) -> Option<OsBasedReturnUrl> {
```
jarnura: move this function plaid specific folder in connector integration

crates/grpc-server/grpc-server/src/sanity_layer.rs
```diff
@@ -0,0 +1,152 @@
...
+#[cfg(not(feature = "connector-sanity-layer"))]
+pub fn wrap<S>(inner: S) -> SanityLayer<S> {
+    inner
+}
+
+#[cfg(feature = "connector-sanity-layer")]
+impl RequestSanitizer for ConnectorSanitizer {
+    fn sanitize<T: PopulateOsBasedReturnUrl>(&self, metadata: &MetadataMap, req: &mut T) {
+        match sanity_connector(metadata) {
+            Some(SanityConnector::Plaid) => PlaidSanity.apply(metadata, req),
```
jarnura: Instead of doing the apply here we can convert the enum variant into dyn trait(RequestSanitizer). So that Plaid owns everything related to plaid, framework layer only hold the overall logic which is generic to all connector.

jarnura (changes requested):

crates/grpc-server/grpc-server/src/sanity_layer.rs
```diff
@@ -0,0 +1,152 @@
...
+                );
+            }
+            None => {}
+        }
+    }
+}
+
+#[cfg(feature = "connector-sanity-layer")]
+trait ConnectorSanity {
+    fn raw_config(&self, metadata: &MetadataMap) -> Option<serde_json::Value>;
```
jarnura: Never use serde_json::Value directly, always use secret for that

crates/types-traits/grpc-api-types/codegen/auto_populate.rs
```diff
@@ -0,0 +1,548 @@
...
+                    compile_error!(#error);
+                }
+            }
+        }
+    }
+}
+
+impl AutoPopulateSpec for AutoPopulateField {
+    fn field_name(self) -> &'static str {
+        match self {
```
jarnura, line 55: The usage of trait is required if it is polymorphic, everything here is handled as normal match, a impl based function approach itself suffice for match based handling, don't use trait if the handling is by match

jarnura (changes requested):

AmitsinghTanwar007:
## Doc rewrite + expanded test coverage (commit 1281ca075)

**Docs:** `codegen/auto_populate.md` rewritten short and diagram-led — why the sanity layer exists, how the generator decides real-setter vs. delegate vs. no-op, one dummy example, and the actual generated shape. Also fixes stale content left over from this branch's own earlier commits (it still described the removed `NoopSanity`, and the legacy-header-first connector resolution that was fixed to be `x-connector-config`-first).

**Codegen simplification:** the generated trait now carries one shared default body instead of every no-path message repeating `let _ = value;` — those messages collapse to a bare `impl Trait for X {}`. Generated file for today's schema: 1588 → 1128 lines (~29% smaller). Verified the real setter and delegating bodies are untouched.

**Test coverage — 1 test → 8, each with a comment stating what it depicts:**

| Test | Scenario it proves |
|---|---|
| `direct_setter_fills_map_when_os_type_already_set` | Happy path: `os_type` already concrete → map gets filled, `os_type` untouched |
| `direct_setter_does_not_create_missing_os_based_return_url` | **Negative** — field absent entirely → setter must not invent it |
| `direct_setter_does_not_fill_map_when_os_type_unspecified` | **Negative** — the core safety guarantee: `os_type` unspecified → map must stay untouched. This was the one guarantee left unverified since the original PR |
| `direct_setter_replaces_existing_map_entirely_rather_than_merging` | Documents real behavior that's easy to assume wrong: the setter **replaces** `return_url_map`, it does not merge into it |
| `populates_nested_os_based_return_url_map_when_authenticator_context_exists` | End-to-end: exercises the full `oneof` delegation path from the actual top-level RPC request, not just the direct setter |
| `oneof_set_to_non_qualifying_variant_is_a_noop` | **Negative** — `domain_context` set to the `payment` branch (no path to the field) → must be left completely unmodified |
| `oneof_unset_is_a_noop` | **Negative** — `domain_context` never set → no panic, nothing populated |
| `unrelated_request_with_no_path_to_the_field_is_a_true_noop` | Proves the trait-default mechanism itself: a real RPC request type with zero path to the field gets the shared default and is provably unaffected |

**Verified before pushing:** 8/8 auto_populate tests, 177/177 existing connector-integration tests, 4/4 grpc-server tests (feature enabled) all pass; `clippy -D warnings` and `cargo fmt --check` clean on `grpc-api-types`; `connector-integration` and `grpc-server` (both feature states) build clean against the updated generated output.

jarnura (approved):

Aishwariyaa-Anand (approved):

PR #2291, opened 2026-09-13
fix(config): correct production base urls for connectors

### LINKED ISSUE #2234: [BUG] multiple connectors - all flows - sandbox/test base URLs in production config
## What is broken

- **Connector**: `checkout`, `cashfree`, `multisafepay`, `fiserv`, `powertranz`, `givepayments`
- **Payment method**: all (connector-level base URL, not payment-method specific)
- **Flow**: all flows
- **Environment**: production

`config/production.toml` has six connectors whose `base_url` points at the connector's **sandbox / test / staging / cert** host. Any deployment running the production config sends live traffic for these connectors to a non-production environment. The same six are also wrong or missing in `config/superposition.toml`.

### Affected entries — `config/production.toml`

| Line | Connector | Value in `production.toml` | Host type |
|---|---|---|---|
| 66 | `multisafepay` | `https://testapi.multisafepay.com/v1/json` | test |
| 94 | `fiserv` | `https://cert.api.fiservapps.com/` | cert (test) |
| 97 | `checkout` | `https://api.sandbox.checkout.com/` | sandbox |
| 100 | `cashfree` | `https://sandbox.cashfree.com/` | sandbox |
| 141 | `powertranz` | `https://staging.ptranz.com/api` | staging |
| 161 | `givepayments` | `https://api-sandbox.givepayments.com` | sandbox |

For all six, the production value is byte-identical to both `sandbox.toml` and `development.toml`.

### Also affected — `config/superposition.toml`

`fiserv` has an explicit production-context override at **line 202** that is still the cert host:

```toml
_context_ = { connector = "fiserv", environment = "production" }
connector_base_url = "https://cert.api.fiservapps.com/"
```

`checkout`, `cashfree`, `multisafepay` and `powertranz` have only a default (sandbox) `connector_base_url` block and **no production override at all**, so production falls back to the sandbox default. 72 other connectors do have a production override, so this is a gap rather than the intended design.

### Why this looks unintentional

**1. It's an outlier.** Of 115 connectors with a `base_url` in both files, 92 have a production value that differs from sandbox. Of the 23 that match, most are legitimately single-endpoint (Stripe, Razorpay, Mollie, NMI — one host, environment selected by API key). These six are the ones where the shared host is explicitly a test hostname.

**2. Sibling connectors in the same family get it right.** The Fiserv group shows the repo already treats `cert` as Fiserv's test environment:

| Connector | production.toml | sandbox.toml |
|---|---|---|
| `fiservemea` | `prod.emea.api.fiservapps.com/ipp/...` | `prod.emea.api.fiservapps.com/sandbox/ipp/...` |
| `authipay` | `prod.emea.api.fiservapps.com/ipp/...` | `prod.emea.api.fiservapps.com/sandbox/ipp/...` |
| `fiservcommercehub` | `connect.fiservapis.com/ch/` | `connect-cert.fiservapis.com/ch/` |
| `fiserv` | `cert.api.fiservapps.com/` :x: | `cert.api.fiservapps.com/` |

`fiservcommercehub` uses `-cert` as its *sandbox* host, while `fiserv` uses `cert.` as its *production* host.

**3. The file annotates uncertainty elsewhere, but not here.** `production.toml` carries explicit caveats where a production URL wasn't confirmed — the JPMorgan Orbital note at lines 73-75 and the Netcetera note at 157-158. None of these six carries any such comment.

**4. Git history shows the copy.** `production.toml` was created in `ab48178d6` ("feat: Introduce production/sandbox configs", #179) with sandbox values copied in verbatim, identical to `sandbox.toml` at that same commit, and untouched since. `multisafepay` came in later via #244 with the same pattern.

### Impact

Per connector, one of two outcomes, neither of which surfaces as a config error:

- Live payments are silently processed against a test environment, or
- Production credentials are rejected by the sandbox host and every payment for that connector fails.

### Suggested fix

1. Replace each `base_url` in `production.toml` with the connector's documented production host, confirmed against each connector's API docs during the fix.
2. Fix the `fiserv` production override in `superposition.toml:202`, and add missing production overrides for `checkout`, `cashfree`, `multisafepay` and `powertranz`.
3. Follow-up worth considering: a CI check that fails when `production.toml` (or a production-context block in `superposition.toml`) contains a host matching `sandbox|testapi|staging|cert\.`, with an allowlist for any that are genuinely correct.

### Note

`ilixium` (`https://prprocessing.ilixium.com/platform/ili`) is also identical across all three env files. I couldn't determine whether `prprocessing` is production or pre-production, so I've left it out of the table — worth confirming alongside these.

## Logs / error response

Reproduction — no server logs, this is reproducible statically from the config:

```
$ grep -nEi '= *"https?://[^"]*(sandbox|testapi|staging|cert\.)[^"]*"' config/production.toml
66:multisafepay.base_url = "https://testapi.multisafepay.com/v1/json"
94:fiserv.base_url = "https://cert.api.fiservapps.com/"
97:checkout.base_url = "https://api.sandbox.checkout.com/"
100:cashfree.base_url = "https://sandbox.cashfree.com/"
141:powertranz.base_url = "https://staging.ptranz.com/api"
161:givepayments.base_url = "https://api-sandbox.givepayments.com"

$ git show ab48178d6:config/production.toml | grep -E '^(checkout|cashfree)\.base_url'
checkout.base_url = "https://api.sandbox.checkout.com/"
cashfree.base_url = "https://sandbox.cashfree.com/"

$ git show ab48178d6:config/sandbox.toml | grep -E '^(checkout|cashfree)\.base_url'
checkout.base_url = "https://api.sandbox.checkout.com/"
cashfree.base_url = "https://sandbox.cashfree.com/"
```

piy-ushk:
This PR fixes [Issue #2234](https://github.com/juspay/hyperswitch-prism/issues/2234) where multiple connectors were incorrectly using sandbox/test/staging hostnames in the production configuration, causing live payments to silently route to test environments or fail authentication.

Changes Made
config/production.toml: Updated the base_url values for multisafepay, fiserv, checkout, cashfree, powertranz, and givepayments to their respective production API endpoints.
config/superposition.toml:
Corrected the existing production environment override for fiserv from cert.api.fiservapps.com to prod.api.fiservapps.com.
Added missing production context overrides for checkout, cashfree, multisafepay, and powertranz so they correctly fall back to production endpoints when running in the production environment.
Note: ilixium (https://prprocessing.ilixium.com/platform/ili) has been left unchanged as it requires further validation on whether prprocessing is a pre-production identifier or the actual production domain.

Impact
Resolves the issue where production credentials were automatically being routed to the sandbox environments for these six specific payment gateways. Live payments will now route to correctly provisioned production servers.


XyneSpaces:
[blocking] `givepayments.base_url` is switched to production, but `config/superposition.toml` has no matching `{ connector = "givepayments", environment = "production" }` override while every other connector changed here adds one. Add the GivePayments production override so Superposition URL patching cannot retain the sandbox/default endpoint.

```toml
# GivePayments Production
[[overrides]]
_context_ = { connector = "givepayments", environment = "production" }
connector_base_url = "https://api.givepayments.com"
```

piy-ushk:
Good catch @XyneSpaces I originally left it out since givepayments wasn't in superposition.toml to begin with, but it makes complete sense to add it to prevent fallback issues. I've pushed the override. Thanks for reviewing.

PR #2293, opened 2026-09-14
feat(Connector): Kount khash payment token

itzcharanteja:
# PR: Kount KHASH payment tokens

## Description

Adds Kount **KHASH** payment-token hashing to the Kount connector. This replaces the current `HMAC-SHA256` payment token with Kount's prescribed **KHASH** algorithm ("the payment token hashed with salted SHA"), so Kount can reliably link the same card/instrument across orders for fraud scoring.

The change:
- Introduces a Kount KHASH implementation (`connectors/kount/khash.rs`): Kount-flavoured Ascii85 config-key decoding, KHASH's non-UTF-8 (CESU-8 + CRLF-normalised) message encoding, SHA-1 digest, and the prefix-preserving base-36 suffix token format.
- Adds the Kount-issued config key (the KHASH salt) and the Apple Pay merchant token identifier to the gRPC contract, threaded through `ConnectorSpecificConfig` / `KountAuthType`.
- Rewires the Evaluate Order `payment` block so each instrument type is sent per Kount's guidance.
- Fixes an existing invalid payment-type value (`CARD`, which is not present in Kount's enum).

KHASH implementation reference: [https://github.com/bdrhoa/rust_khash/tree/master](https://github.com/bdrhoa/rust_khash/tree/master)

## Motivation and Context

Kount requires the merchant to submit a Kount-specific `paymentToken` so it can link a payment instrument across orders *without* receiving the raw PAN.

Our current implementation hashes the PAN with `HMAC-SHA256(key = api_key)`. Kount's backend cannot reproduce that hash — it expects **KHASH**, a distinct algorithm — so it cannot link instruments across orders, defeating the purpose of `paymentToken` for fraud scoring.

Kount's guidance (and the reference Rust example they shared) specifies:
- **`CREDIT_CARD` / `DEBIT_CARD`** → use **KHASH** of the card number.
- **All other payment types** → a salted hash (kept as the existing `HMAC-SHA256`).
- **`NONE`** type → `paymentToken` must be left empty.
- **`TOKEN`** → emitted for processor/network-token payment methods (`NetworkToken`,
  `PaymentMethodToken`); the token itself is the hashed `paymentToken`. This arm is
  unchanged from `main` and does fire in the pre-auth Evaluate Order call.

We are live in alpha with Kount on cards, Apple Pay and Google Pay wallets, and this is the prerequisite for sending Kount-compatible instrument tokens on those flows.

### Payment-token mapping

| Case | `paymentType` | `paymentToken` |
|---|---|---|
| Card (raw PAN, network-token PAN, or gateway token) | `CREDIT_CARD` / `DEBIT_CARD` | **KHASH(number)** |
| Card with empty number (device-tokenization placeholder) | `NONE` | empty |
| Network-token / payment-method-token PMD | `TOKEN` | `HMAC-SHA256(token)` (unchanged from `main`) |
| Google Pay **DPAN** (decrypted token carries a cryptogram) | `GOOG` | `HMAC-SHA256(DPAN)` |
| Google Pay **FPAN** (decrypted, no cryptogram — raw PAN) | `CREDIT_CARD` | **KHASH(FPAN)** |
| Apple Pay decrypted (DPAN / MPAN) | `APAY` | `HMAC-SHA256(merchant token id, else DPAN)` |
| Google/Apple Pay pass-through (encrypted payload) | `GOOG` / `APAY` | none (payload is single-use, not stable) |

When no `khash_config_key` is configured, card instruments fall back to the legacy HMAC-SHA256 token with a warning, so unseeded merchants keep a stable token and nothing breaks during rollout.

### Coding conventions

- `Khash` deliberately has no `Debug` impl — the salt is a secret.
- KHASH's non-UTF-8 encoding and the config key are never exposed in logs.
- Adheres to the repo's strict lints (no `as`-conversions, no `unwrap` in production code).

### Alternative considered

- **Send the raw encrypted wallet payload as `paymentToken`** — rejected. Apple Pay / Google Pay encrypted payloads are single-use (a fresh cryptogram per transaction), so they are not consistent for a given card x device and would break instrument linkage. Instead we rely on decrypted stable identifiers (FPAN / DPAN / merchant token id) where available, and omit the token otherwise.

## Additional Changes

- [x] **This PR modifies the API contract** — `KountConfig.khash_config_key` and `ApplePayDecryptedData.merchant_token_identifier` added to `crates/types-traits/grpc-api-types/proto/payment.proto` and `payment_methods.proto`.
- [x] **This PR modifies application configuration/environment variables** — no changes. The `khash_config_key` is delivered per-merchant on `accountDetailsJson`/`KountConfig`, not via an env var.

## How did you test it?

- **Unit tests** — `cargo test -p connector-integration`: `187 passed / 0 failed`
  - **KHASH algorithm**: Ascii85 decode (`z` / `y` expansions, whitespace strip, byte-valued salts), CESU-8 + CRLF message encoding, token format, determinism, salt-sensitivity, suffix cap.
  - Kount's six official reference vectors are added as an `#[ignore]` test, runnable once the real config key is delivered (via `KHASH_KEY`), so the shipped key can be validated before enabling.
  - **Instrument mapping**: CREDIT_CARD/DEBIT_CARD + KHASH for cards; `NONE` for empty numbers; unknown card type now falls back to `CREDIT_CARD` (was `CARD`).
- **Clippy**: `cargo clippy -p connector-integration --all-targets` — clean.
- **Formatting**: `cargo +nightly fmt --all -- --check` — clean.
- **Workspace build**: `cargo check --workspace` — succeeds.
- **Manual reasoning** across the payment-type mapping table above (GPay DPAN/FPAN branching on cryptogram presence, Apple Pay merchant-token-id preference, encrypted-payload omission).

### 1. Unit test suite (no key required)

Covers: Ascii85 config-key decoding (canonical vector, `z`/`y` expansions, whitespace stripping, degenerate keys), byte-valued salt characters, CRLF normalisation + UTF-16 code-unit message encoding, token format (6-char BIN prefix + 14 base-36 chars), determinism, salt sensitivity, suffix-length capping.

```bash
cd hyperswitch-prism
nix develop -c cargo test -p connector-integration kount
```

```
test result: ok. 10 passed; 0 failed; 2 ignored; 0 measured; 177 filtered out
```

### 2. Kount official reference vectors (requires the config key)

Kount's KHASH example specifies six `<PAN> → <expected token>` pairs. These are recorded in `kount_reference_vectors` (test stays `#[ignore]`d — the key is supplied via env var and never enters the repo; only the expected tokens are committed).

```bash
KHASH_KEY='<kount-issued-config-key>' \
  nix develop -c cargo test -p connector-integration kount_reference_vectors -- --ignored --nocapture
```

```
4111111111111111 -> 411111WMS5YA6FUZA1KC (matches Kount's expected token)
5454545454545454 -> 545454E58W8101GXHU4U (matches Kount's expected token)
374410712128163  -> 3744101P95YO2CXFZDN0 (matches Kount's expected token)
6011000000000012 -> 601100HFMWO1DQWO6C79 (matches Kount's expected token)
6011180161073659 -> 601118XQEYTYFYQ67HBB (matches Kount's expected token)
38000000000006   -> 3800007KGYCD1ZE74JVH (matches Kount's expected token)

test result: ok. 1 passed; 0 failed
```

| Card | PAN | Expected token (Kount) | Result |
|---|---|---|---|
| Visa | `4111111111111111` | `411111WMS5YA6FUZA1KC` | ✅ |
| Mastercard | `5454545454545454` | `545454E58W8101GXHU4U` | ✅ |
| American Express | `374410712128163` | `3744101P95YO2CXFZDN0` | ✅ |
| Discover | `6011000000000012` | `601100HFMWO1DQWO6C79` | ✅ |
| Discover (bad card) | `6011180161073659` | `601118XQEYTYFYQ67HBB` | ✅ |
| Diners Club | `38000000000006` | `3800007KGYCD1ZE74JVH` | ✅ |

### 3. Optional: print tokens for arbitrary PANs

```bash
KHASH_KEY='<kount-issued-config-key>' \
  nix develop -c cargo test -p connector-integration print_tokens_with_real_key -- --ignored --nocapture
```

Prints KHASH tokens for five standard test PANs (never prints the key itself).


XyneSpaces:
[blocking] The new KHASH implementation is not validated against Kount’s official outputs: the enabled tests only check shape/self-consistency, while `kount_reference_vectors` is ignored and its vector list is empty. Please add the six official input→token expectations before merge; a decoder or encoding error here still produces deterministic, well-formed tokens that these tests cannot distinguish from Kount-compatible tokens.

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1378,32 +1390,65 @@ fn card_payment_type<T: PaymentMethodDataTypes>(card: &Card<T>) -> KountPaymentT
     {
         Some("credit") => KountPaymentType::CreditCard,
         Some("debit") => KountPaymentType::DebitCard,
-        _ => KountPaymentType::Card,
```
Aishwariyaa-Anand, line 1381: we dont not need Card payment type?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1465,6 +1525,49 @@ fn kount_instrument<T: PaymentMethodDataTypes>(
...
+/// the DPAN (stable per card x device); either is HMAC-hashed under
+/// paymentType=APAY. Encrypted (pass-through) payloads are single-use, so no
+/// paymentToken is sent.
+fn apple_pay_instrument(apay: &ApplePayWalletData) -> KountInstrument {
+    match &apay.payment_data {
+        ApplePayPaymentData::Decrypted(data) => {
+            let token_source = data
+                .merchant_token_identifier
+                .clone()
+                .unwrap_or_else(|| data.application_primary_account_number.peek().to_string());
```
Aishwariyaa-Anand, line 1563: why do we have fallback of token_source to application_primary_account_number

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
itzcharanteja: No, Kount doesn't support the payment type Card at all

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
itzcharanteja: It will always be either one of them.
For an applepay decrypted payment flow we'll receive one of the following reference for a payment instrucment
1. MPAN - Merchant primary account number i.e merchant_token_identifier
2. DPAN - Device specific primary account number i.e application_primary_account_number

Aishwariyaa-Anand (dismissed):

kanikac199 (dismissed):

crates/integrations/connector-integration/src/connectors/kount/khash.rs
```diff
@@ -0,0 +1,168 @@
...
+const MAX_SUFFIX_LENGTH: usize = 17;
+
+/// Number of leading characters of the input preserved verbatim in the token.
+const PRESERVED_PREFIX_LENGTH: usize = 6;
+
+/// Base-36 alphabet used for the hash suffix.
+const BASE36_ALPHABET: &[u8; 36] = b"0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+/// A KHASH generator seeded with a merchant's Kount configuration key.
+pub(crate) struct Khash {
```
JeevaRamu0104, line 40: The head commit (`e3d809f`) deleted every test this PR had — the Ascii85/CESU-8 unit tests here, the `kount_instrument` mapping tests in `transformers.rs`, and the `#[ignore]` `kount_reference_vectors` test. That last one is the only thing that lets anyone re-prove the tokens match Kount once the key is available; without it a refactor of `decode_config_key` or `encode_message` silently breaks instrument linkage. Inline `#[cfg(test)]` modules already exist in this crate (`connectors/juspay/crypto.rs`, `connectors/paynearme/transformers.rs`), so nothing forces the removal. Please restore them (and update the PR description, which still cites them).

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1606,27 +1709,48 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+        // use the Kount KHASH algorithm, salted with the Kount-issued
+        // configuration key; every other type uses the salted HMAC-SHA256.
+        // Without a configured KHASH key the card path falls back to the HMAC
+        // token so unseeded merchants keep a stable token instead of losing
+        // instrument linkage.
+        let auth = KountAuthType::try_from(&item.router_data.connector_config)?;
+        let khash = auth
+            .khash_config_key
+            .as_ref()
+            .map(|key| super::khash::Khash::new(key.peek()));
```
JeevaRamu0104, line 1722: An empty `khash_config_key` becomes `Khash { salt: "" }` instead of falling back to HMAC, so a merchant configured with `""` gets card tokens that match neither Kount's nor their pre-rollout tokens.

```suggestion
            .filter(|key| !key.peek().trim().is_empty())
            .map(|key| super::khash::Khash::new(key.peek()));
```

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -883,7 +892,7 @@ pub enum KountFulfillmentType {
 /// enumeration from the Kount Orders API is modelled here; only the variants
 /// with a UCS payment-method equivalent are ever produced (see
 /// [`kount_payment_type`]).
-#[derive(Debug, Clone, Copy, Serialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize)]
```
JeevaRamu0104, line 895: `Card` (`"CARD"`) is still declared five lines down even though the PR removes it precisely because Kount's enum has no such value, and the doc above says this is the full Kount enumeration. Drop the variant so it can't be reintroduced by the next arm.

crates/integrations/connector-integration/Cargo.toml
```diff
@@ -44,6 +44,7 @@ qrcode = "0.14.1"
 image = { version = "0.25.6", default-features = false, features = ["png"] }
 thiserror = { workspace = true }
 sha2 = "0.10"
+sha1 = "0.10"
```
JeevaRamu0104: nit: `ring` is already a direct dependency here — `ring::digest::digest(&ring::digest::SHA1_FOR_LEGACY_USE_ONLY, &bytes)` gives the same digest without adding `sha1`.

crates/types-traits/domain_types/src/payment_method_data.rs
```diff
@@ -1358,6 +1358,9 @@ pub struct ApplePayDecryptedData {
     pub application_expiration_year: Secret<String>,
     /// The payment data, which contains the cryptogram and ECI indicator
     pub payment_data: ApplePayCryptogramData,
+    /// Apple Pay merchant token identifier — present for merchant-provisioned
+    /// tokens (MPAN) only; stable per card x device x merchant
+    pub merchant_token_identifier: Option<String>,
```
JeevaRamu0104: This is a stable per-card×device×merchant identifier and the preferred `paymentToken` source for Apple Pay, and as a plain `String` it lands verbatim in `Debug` output — see the regenerated `data/field_probe/authorizedotnet.json:73`. The sibling identifiers in this struct are `Secret<_>`; should this be `Option<Secret<String>>`?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1433,28 +1478,43 @@ fn pmt_to_kount_payment_type(pmt: PaymentMethodType) -> Option<KountPaymentType>
...
+            if pan.trim().is_empty() {
+                KountInstrument::typed(K::None)
+            } else {
+                KountInstrument::card(card_payment_type(card), pan.to_string())
+            }
+        }
+        PaymentMethodData::CardRedirect(_) => KountInstrument::typed(K::CreditCard),
+        // Processor / network tokens: the token itself is the (hashed)
+        // paymentToken under paymentType=TOKEN.
         PaymentMethodData::NetworkToken(token_data) => {
```
JeevaRamu0104, line 1505: The description says `TOKEN` is "kept out of the pre-auth path", but this function feeds Evaluate Order and still emits `paymentType=TOKEN` for `NetworkToken`/`PaymentMethodToken`. Behaviour is unchanged from `main`, so it's the description that needs fixing — or was the arm meant to go?

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
```diff
@@ -1606,27 +1709,48 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+                    (None, None)
+                };
+                let payment_token = match (instrument.token_source.as_deref(), instrument.hash_kind)
+                {
+                    (Some(source), TokenHashKind::Khash) => match &khash {
+                        Some(khash) => {
+                            Some(khash.hash(source, super::khash::CARD_TOKEN_SUFFIX_LENGTH))
+                        }
+                        None => {
+                            tracing::warn!(
```
JeevaRamu0104: This fires on every card evaluation from every merchant without a key, so during rollout it's one `warn` per payment. `debug`, or a once-per-process guard, fits better.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/kount/transformers.rs
itzcharanteja: it's the description that needs fixing

itzcharanteja:
> [blocking] The new KHASH implementation is not validated against Kount’s official outputs: the enabled tests only check shape/self-consistency, while `kount_reference_vectors` is ignored and its vector list is empty. Please add the six official input→token expectations before merge; a decoder or encoding error here still produces deterministic, well-formed tokens that these tests cannot distinguish from Kount-compatible tokens.

JeevaRamu0104 (approved):

Aishwariyaa-Anand (approved):

kanikac199 (approved):

PR #2294, opened 2026-09-14
feat(connector): [D24] add LocalBankTransfer support to Authorize

shuklatushar226:
## Summary

Add the **LocalBankTransfer** payment method (`PaymentMethodData::BankTransfer(BankTransferData::LocalBankTransfer)`) to the existing **Authorize** flow of the **D24** (Directa24) connector.

The existing WebPay path (`CardRedirect(Webpay)`, merged in #2212) is unchanged. Its request body stays byte-for-byte the same, and a regression call confirms it (see test 4 and the Hyperswitch e2e below).

This change was generated and validated by **GRACE** (automated connector integration pipeline).

> **Testing mode: local mock server.** No D24 sandbox credentials were available, so nothing here was tested against D24 staging. Every call below went to a **local mock D24 HTTP server** with **dummy credentials**. The mock recorded each outbound request, and a separate script recomputed the `Authorization: D24 <hex>` HMAC-SHA256 signature and compared it with the one sent (match). Responses are copied from the documented examples in the D24 tech spec, so `connectorTransactionId`, `merchantTransactionId` etc. come from the mock and not from the request.

## Changes

`crates/integrations/connector-integration/src/connectors/d24/transformers.rs`
- `D24PaymentMethod` now covers the D24 `BANK_TRANSFER` methods that need no `reported_info`:
  - **MX:** `SE` (SPEI), `COD` (CoDi), `BM` (Banamex), `STS` (Santander SuperNet), `AF` (Afirme), `BQL` (Banorte)
  - **BR:** `IX` (Pix), `I` (Itaú), `NU` (Nubank), `ME` (MercadoPago)

  Each method has one wire code, one country and one local currency. The wire code is declared once per variant (`#[strum(serialize = ..)]`), and both `code()` and `Serialize` read it, so WebPay still goes out as `"WP"`.
- How the method code is picked for `LocalBankTransfer`:
  - If `bank_code` is set, it is read as the D24 method code (case-insensitive) and checked against the allow-list for the billing country.
  - If `bank_code` is not set, the country default is used: `SE` for MX, `IX` for BR.
  - Any other code or country fails locally with `NotSupported` and no request is sent.
- The Guard 2 (billing country) and Guard 3 (currency: local currency or USD) checks now depend on the payment method instead of being hardcoded to Chile/CLP.
- Payer document mapping (`d24_payer_document`):
  - CL + `OTHER` → `RUT` (validated as before)
  - BR + `CPF`/`CNPJ` → separators stripped, check digits validated, `document_type` set to `CPF`/`CNPJ`
  - MX + `OTHER` → alphanumeric, 7–18 characters, sent **without** `document_type` (CURP/RFC/IFE/PASS cannot be told apart)
  - A rejected document says which rule failed in the error's `additional_context` (the number is never echoed).
- Itaú (`I`) requires the full payer address (line1, city, state, zip), checked with the shared address getters.
- `success_url` / `back_url` / `error_url` are now `Option`:
  - WebPay still requires them and fails locally if they are missing.
  - Bank transfers send them only when a return URL is supplied.
- `language` is sent as the ISO 639-1 primary subtag (`es`, `en`, `pt`, `ja`) and omitted otherwise; `description` is truncated to D24's 100 characters.
- The ONE_SHOT `201` response now parses `iframe` and `payment_info` (CLABE / Pix QR code / digitable line / exact amount / expiry). These are returned as `connector_metadata` (gRPC `connectorFeatureData`) so the caller can render the transfer instructions itself. Parsing is lenient, so the `deposit_id` of a deposit that was already created is never lost. `payment_info.metadata` (which carries the payer's CPF and name and the CLABE) is `Secret`, so it is masked in logs and in the typed connector response; `connector_metadata` still carries the plain JSON.
- D24 error bodies: the KYC rejection `reason` / `reason_code` are kept in the error reason, and `NotSupported` messages are noun phrases so they no longer read "... not supported ... is not supported by d24".
- Bank transfers return `AUTHENTICATION_PENDING` (redirect URL present) or `PENDING`. They never return a final status at Authorize; the final state comes from PSync.

`crates/integrations/connector-integration/src/connectors/d24.rs`
- Comment-only updates describing the new bank-transfer Authorize path.

`data/integration-source-links.json`
- Added the `D24` provenance entry (the D24 API docs used for this integration).

## Files Modified

- `crates/integrations/connector-integration/src/connectors/d24.rs`
- `crates/integrations/connector-integration/src/connectors/d24/transformers.rs`
- `data/field_probe/d24.json`, `docs-generated/connectors/d24.md`, `docs-generated/all_connector.md` (regenerated by the CI auto-fix commits `aeb2c54be2` and `d7d2380dea`)

## Commits

| Commit | Change |
|---|---|
| `832ddbf9a8` | feat: LocalBankTransfer support in Authorize |
| `aeb2c54be2` | CI auto-fix: formatting and generated docs / field-probe data |
| `4e06679687` | fix: send only the ISO 639-1 `language` code D24 accepts (found in the Hyperswitch e2e run: `es-MX` / `pt-BR` were sent) |
| `01113ac124` | fix: truncate `description` to 100 characters (D24 rejects longer ones with `201`) |
| `a089111d82` | fix: keep the KYC rejection `reason` / `reason_code` in the error response |
| `7ae587bb32` | fix: stop repeating "not supported" in D24 `NotSupported` messages |
| `d7d2380dea` | CI auto-fix: regenerated `data/field_probe/d24.json` (probe error text) |
| `38986611ab` | fix: mask payer PII in `payment_info` (`metadata`, `secondary_metadata`, `user_id` are `Secret`); `payment_info.currency` is `Currency` and `type` an enum |
| `e8779427e1` | fix: say why a payer document is rejected (`additional_context` + `suggested_action`); shared getters for the Itaú address |
| `7f0401ae4e` | docs: note that bank-transfer deposit refunds are unsupported (comment-only) |
| `e2ed0f788e` | refactor: derive the payment method wire code instead of a hand-written `impl Serialize`; exhaustive local-currency match |
| `49bb198d9` | fix: measure the Mexican payer document in characters, so the rejection message reports the unit it tested |
| `1bfd8d0e3` | fix: reject refunds of `LocalBankTransfer` deposits locally, when the caller sends `payment_method` |
| `b143b7b8d` | fix: report a partially settled deposit on PSync — `PartialCharged` + `captured_amount`, and surface `status_reason` |
| `ee169d9eb` | test: four Authorize scenarios in `connector_specs/d24/override.json` (MX SPEI default, BR Itaú/CNPJ, BR Pix/CPF, rejected `bank_code`) |

## Known not implemented / out of scope

- Refunds for bank-transfer deposits. These need `bank_account` (D24 error 804), which UCS does not send. Hyperswitch rejects them before calling UCS, and since `1bfd8d0e3` UCS refuses them too — but only when the caller populates the optional `PaymentServiceRefundRequest.payment_method`; omit it and the call still reaches D24 and comes back as `804`.
- The `BankTransferData::Pix` variant. Only `LocalBankTransfer { bank_code: "IX" }` / BR default reaches Pix.
- Methods that need `reported_info` / `bank_account` (BR `BB`, `B`) and all `BANK_DEPOSIT`-typed methods. These are rejected locally with `NotSupported`.
- Countries other than MX and BR for bank transfer.
- MX `document_type` (not sent), `expiration`, `notification_url` / webhooks, `X-Idempotency-Key`.
- Local min/max amount checks.

## Known gaps / follow-ups

- **Bank-transfer instructions are not surfaced to the merchant.** `payment_info` goes out as opaque `connector_feature_data`; Hyperswitch stores it in `connector_metadata` but its payment response only shows the redirect. Follow-up: a typed instructions shape (short term, HS-shaped JSON as prism Stripe emits; properly, a proto field).
- **No webhooks, and HS stops PSync after about 26 minutes.** `IncomingWebhook` is empty and `notification_url` is not sent, so a transfer paid after HS's default PSync schedule (start 60s, 5 x 300s) stays pending until force-synced. Follow-up: D24 deposit notifications through `EventService`, or a longer PSync schedule for D24 in Hyperswitch.
- **juspay/hyperswitch `ef56dae37a` to be reworked into a tolerant parse.** It skips the bank-transfer next-steps parse by connector name in shared core; the generic fix is to warn and return `None` on a shape mismatch.
- **UCS pre-call rejections leave Hyperswitch payments in `processing`.** Unsupported bank code / missing document / missing country are rejected before any D24 call, but HS keeps the intent `processing` with a PSync task (generic HS handling).
- ~~**Settled amount / `AMOUNT_MISMATCH` ignored.**~~ Fixed in `b143b7b8d`: PSync compares the settled amount with the requested one, reports `PartialCharged` when it is short, sets `amount_captured` / `minor_amount_captured`, and surfaces `status_reason`. Remaining limit: a deposit whose settled currency cannot be lined up with the requested one leaves the amount unknown and falls back to the connector's status.
- **Raw state name in the address.** `billing.state` is forwarded as sent; a full state name instead of the ISO 3166-2 code makes D24 reject even a SPEI/Pix deposit that does not need an address.
- **The new scenarios are not executed in CI yet.** D24 is in `alpha_connectors.json` (no Directa24 sandbox credentials), so `certify-connectors.sh` skips it. `connector_specs/d24/override.json` is gated today only by `all_override_entries_match_existing_scenarios_and_proto_schema`, which builds and proto-schema-checks the four requests; live branch coverage follows the credentials.
- **`connector_specific_scenarios.json` cannot be executed by the harness.** It is the intended home for connector-private scenarios, but only the enumeration path merges it — `execute_single_scenario_with_context` and `load_effective_scenario_for_connector` resolve bodies through `load_scenario`, which reads the global suite file alone, so a private scenario fails with `not found in suite`. That is why D24's scenarios specialise global ones through `override.json` instead, and why they carry global names. Not D24-specific; worth a separate harness fix.
- **Field probe marks LocalBankTransfer unsupported.** `probe-config.toml` overrides are keyed per flow and connector only, and D24's authorize override must pin CL/RUT for WebPay, so the generated `data/field_probe/d24.json` and `docs-generated/connectors/d24.md` show Local bank transfer as unsupported. Fixing it needs a per-payment-method override in the field-probe engine.

## gRPC Test Results

**Status: PASS (against local mock D24 server, not D24 staging)**

Server: `./target/debug/grpc-server` with `CS__CONNECTORS__D24__BASE_URL=http://127.0.0.1:18924` (mock), `CS__SERVER__PORT=50124`. These grpcurl runs predate the follow-up commits; the error text in test 5 is the old wording (see the Hyperswitch e2e section for the current text).

| # | Scenario | Outbound `POST /v3/deposits` body (checked on the mock) | Result |
|---|---|---|---|
| 1 | MX, no `bank_code` → SPEI | `payment_method "SE"`, `amount 1000.0`, `MXN`, no `document_type`, 3 URLs, `client_ip`, `description`; HMAC recomputed, match | `AUTHENTICATION_PENDING`, `statusCode 201`, redirect + `payment_info` in `connectorFeatureData` |
| 2 | BR, `bank_code "ix"` → Pix | `payment_method "IX"`, `amount 15.06`, `BRL`, CPF separators stripped, `document_type "CPF"`, no URLs; HMAC match | `AUTHENTICATION_PENDING`, `statusCode 201`, Pix QR code / digitable line in `connectorFeatureData` |
| 3 | MX, mock returns `400` | request sent, mock replies `{code 201, description, details}` | gRPC `InvalidArgument`, `CONNECTOR_ERROR_RESPONSE`, http 400, code `"201"` (expected negative case) |
| 4 | CL WebPay regression | `payment_method "WP"`, `document_type "RUT"`, all three URLs, same shape as before | `AUTHENTICATION_PENDING`, `statusCode 201` |
| 5 | BR, `bank_code "BB"` | **no outbound request** | gRPC `Unimplemented`, `NOT_SUPPORTED` locally (expected negative case) |

Notes:
- In tests 3 and 5, grpcurl shows `Code: Internal … grpc-status-details-bin mismatch`. This is grpcurl complaining about the server's `grpc-status-details-bin` trailer. The real `grpc-status` it reports is `InvalidArgument` (3) and `Unimplemented` (5), and both are the intended rejections.
- In test 4 the mock gave the WebPay call the same canned ONE_SHOT/SPEI response body as test 1. That is why `connectorFeatureData` shows SPEI `payment_info`. The regression check is on the outbound WebPay **request** body, which is unchanged.
- The mock returns `redirect_url: "https://URL"` in test 2, so the gRPC response shows `https://url/`.

<details>
<summary>1. MX SPEI (default code), grpcurl Authorize (credentials and PII redacted)</summary>

```
grpcurl -plaintext \
  -H 'x-connector-config: <REDACTED>' \
  -H 'x-merchant-id: <REDACTED>' -H 'x-tenant-id: public' -H 'x-request-id: 1_mx_spei_default' \
  -d '{
  "merchant_transaction_id": "d24_bt_mx_0001",
  "amount": {"minor_amount": 100000, "currency": "MXN"},
  "payment_method": {"local_bank_transfer": {}},
  "customer": {"id": "cust11", "email": {"value": "<REDACTED>"},
    "customer_document_details": {"document_type": "OTHER", "document_number": {"value": "<REDACTED>"}}},
  "address": {"billing_address": {"first_name": {"value": "Ricardo"}, "last_name": {"value": "Carlos"},
    "country_alpha2_code": "MX"}},
  "auth_type": "NO_THREE_DS",
  "return_url": "https://www.merchant.com/deposit_completed",
  "browser_info": {"ip_address": "123.123.123.123"},
  "description": "SPEI bank transfer test"
}' \
  127.0.0.1:50124 types.PaymentService/Authorize

Outbound request captured by mock (POST /v3/deposits; Authorization: D24 <HMAC, verified>):
{"country":"MX","amount":1000.0,"currency":"MXN","invoice_id":"d24_bt_mx_0001","payer":{"id":"cust11","document":"<REDACTED>","email":"<REDACTED>","first_name":"Ricardo","last_name":"Carlos"},"payment_method":"SE","success_url":"https://www.merchant.com/deposit_completed","back_url":"https://www.merchant.com/deposit_completed","error_url":"https://www.merchant.com/deposit_completed","client_ip":"123.123.123.123","description":"SPEI bank transfer test"}

Response:
{
  "merchantTransactionId": "postmanTest388400617",
  "connectorTransactionId": "301626485",
  "status": "AUTHENTICATION_PENDING",
  "statusCode": 201,
  "redirectionData": {
    "form": {
      "endpoint": "https://payment-stg.depositcheckout.com/v1/checkout/eyJhbGciOiJIUzM...",
      "method": "HTTP_METHOD_GET"
    }
  },
  "connectorFeatureData": {
    "value": "{\"checkout_type\":\"ONE_SHOT\",\"iframe\":true,\"payment_info\":{\"type\":\"VOUCHER\",\"payment_method\":\"SE\",\"payment_method_name\":\"Spei\",\"amount\":1000.0,\"currency\":\"MXN\",\"expiration_date\":\"2025-07-27 19:57:40\",\"created_at\":\"2025-07-24 19:57:40\",\"metadata\":{\"beneficiary_name\":\"D24 S.A.\",\"temporal_clabe\":false,\"beneficiary_overridden\":false,\"updated_clabe_flag\":false,\"offline_payments_enabled\":false,\"payer_name\":\"Ricardo Carlos\",\"instructions_alert\":true,\"clabe\":\"<REDACTED>\",\"legacy_view\":false},\"secondary_metadata\":null}}"
  },
  "connectorReferenceId": "postmanTest388400617",
  "typedConnectorResponse": {
    "value": "{\"deposit_id\":301626485,\"checkout_type\":\"ONE_SHOT\",\"redirect_url\":\"https://payment-stg.depositcheckout.com/v1/checkout/eyJhbGciOiJIUzM...\",\"iframe\":true,\"user_id\":\"11\",\"merchant_invoice_id\":\"postmanTest388400617\",\"payment_info\":{\"type\":\"VOUCHER\",\"payment_method\":\"SE\",\"payment_method_name\":\"Spei\",\"amount\":1000.0,\"currency\":\"MXN\",\"expiration_date\":\"2025-07-27 19:57:40\",\"created_at\":\"2025-07-24 19:57:40\",\"metadata\":{\"beneficiary_name\":\"D24 S.A.\",\"temporal_clabe\":false,\"beneficiary_overridden\":false,\"updated_clabe_flag\":false,\"offline_payments_enabled\":false,\"payer_name\":\"Ricardo Carlos\",\"instructions_alert\":true,\"clabe\":\"<REDACTED>\",\"legacy_view\":false},\"secondary_metadata\":null}}"
  }
}
grpcurl_exit=0
```

</details>

<details>
<summary>2. BR Pix via <code>bank_code "ix"</code>, grpcurl Authorize (credentials and PII redacted)</summary>

```
grpcurl -plaintext \
  -H 'x-connector-config: <REDACTED>' \
  -H 'x-merchant-id: <REDACTED>' -H 'x-tenant-id: public' -H 'x-request-id: 2_br_pix_bank_code' \
  -d '{
  "merchant_transaction_id": "d24_bt_br_0001",
  "amount": {"minor_amount": 1506, "currency": "BRL"},
  "payment_method": {"local_bank_transfer": {"bank_code": "ix"}},
  "customer": {"email": {"value": "<REDACTED>"},
    "customer_document_details": {"document_type": "CPF", "document_number": {"value": "<REDACTED formatted CPF: ddd.ddd.ddd-dd>"}}},
  "address": {"billing_address": {"first_name": {"value": "Ricardo"}, "last_name": {"value": "Carlos"},
    "country_alpha2_code": "BR"}},
  "auth_type": "NO_THREE_DS"
}' \
  127.0.0.1:50124 types.PaymentService/Authorize

Outbound request captured by mock (POST /v3/deposits; Authorization: D24 <HMAC, verified>):
{"country":"BR","amount":15.06,"currency":"BRL","invoice_id":"d24_bt_br_0001","payer":{"document":"<REDACTED 11 digits, separators stripped>","document_type":"CPF","email":"<REDACTED>","first_name":"Ricardo","last_name":"Carlos"},"payment_method":"IX"}

Response:
{
  "merchantTransactionId": "postmanTest943044826",
  "connectorTransactionId": "300000025",
  "status": "AUTHENTICATION_PENDING",
  "statusCode": 201,
  "redirectionData": {
    "form": {
      "endpoint": "https://url/",
      "method": "HTTP_METHOD_GET"
    }
  },
  "connectorFeatureData": {
    "value": "{\"checkout_type\":\"ONE_SHOT\",\"iframe\":null,\"payment_info\":{\"type\":\"BANK_TRANSFER\",\"payment_method\":\"IX\",\"payment_method_name\":\"Pix\",\"amount\":1506.0,\"currency\":\"BRL\",\"expiration_date\":\"2020-06-17T07:04:16Z\",\"created_at\":\"2020-06-16T19:04:16Z\",\"metadata\":{\"payer_document\":\"<REDACTED>\",\"reference\":1161706605,\"show_terms_conditions\":true,\"payer_document_type\":\"CPF\",\"qr_code\":\"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAOQAAADkCAYAAACIV4iN...\",\"digitable_line\":\"00020126820014br.gov.bcb.pix2560pix.treeal.com/qr/v3/at/c...\"},\"secondary_metadata\":null}}"
  },
  "connectorReferenceId": "postmanTest943044826",
  "typedConnectorResponse": {
    "value": "{\"deposit_id\":300000025,\"checkout_type\":\"ONE_SHOT\",\"redirect_url\":\"https://url/\",\"iframe\":null,\"user_id\":\"11\",\"merchant_invoice_id\":\"postmanTest943044826\",\"payment_info\":{\"type\":\"BANK_TRANSFER\",\"payment_method\":\"IX\",\"payment_method_name\":\"Pix\",\"amount\":1506.0,\"currency\":\"BRL\",\"expiration_date\":\"2020-06-17T07:04:16Z\",\"created_at\":\"2020-06-16T19:04:16Z\",\"metadata\":{\"payer_document\":\"<REDACTED>\",\"reference\":1161706605,\"show_terms_conditions\":true,\"payer_document_type\":\"CPF\",\"qr_code\":\"data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAOQAAADkCAYAAACIV4iN...\",\"digitable_line\":\"00020126820014br.gov.bcb.pix2560pix.treeal.com/qr/v3/at/c...\"},\"secondary_metadata\":null}}"
  }
}
grpcurl_exit=0
```

</details>

<details>
<summary>3. MX, connector 400 error mapping, grpcurl Authorize (expected negative case)</summary>

```
grpcurl -plaintext \
  -H 'x-connector-config: <REDACTED>' \
  -H 'x-merchant-id: <REDACTED>' -H 'x-tenant-id: public' -H 'x-request-id: 3_mx_error_400' \
  -d '{
  "merchant_transaction_id": "d24_bt_error_0001",
  "amount": {"minor_amount": 100000, "currency": "MXN"},
  "payment_method": {"local_bank_transfer": {"bank_code": "SE"}},
  "customer": {"email": {"value": "<REDACTED>"},
    "customer_document_details": {"document_type": "OTHER", "document_number": {"value": "<REDACTED>"}}},
  "address": {"billing_address": {"first_name": {"value": "Ricardo"}, "last_name": {"value": "Carlos"},
    "country_alpha2_code": "MX"}},
  "auth_type": "NO_THREE_DS"
}' \
  127.0.0.1:50124 types.PaymentService/Authorize

Mock response: HTTP 400
{"code":201,"description":"Field validation error. Check details","details":["payer.document: Invalid document type and/or document","payer.address.state: Invalid State for Country"]}

Response:
ERROR:
  Code: Internal
  Message: grpc-status-details-bin mismatch: grpc-status=InvalidArgument, grpc-message="Field validation error. Check details", grpc-status-details-bin=message:"CONNECTOR_ERROR_RESPONSE"  1:"Field validation error. Check details"  3:400  4:"\n\x93\x01\n\x03201\x12%Field validation error. Check details\x1aepayer.document: Invalid document type and/or document, payer.address.state: Invalid State for Country\x1a\x93\x01\n\x03201\x12%Field validation error. Check details\x1aepayer.document: Invalid document type and/or document, payer.address.state: Invalid State for Country"  7:"{\"code\":201,\"description\":\"Field validation error. Check details\",\"message\":null,\"details\":[\"payer.document: Invalid document type and/or document\",\"payer.address.state: Invalid State for Country\"],\"type\":null}"
grpcurl_exit=77
```

</details>

<details>
<summary>4. CL WebPay regression, grpcurl Authorize (credentials and PII redacted)</summary>

```
grpcurl -plaintext \
  -H 'x-connector-config: <REDACTED>' \
  -H 'x-merchant-id: <REDACTED>' -H 'x-tenant-id: public' -H 'x-request-id: 4_cl_webpay_regression' \
  -d '{
  "merchant_transaction_id": "d24_wp_cl_0001",
  "amount": {"minor_amount": 10000, "currency": "CLP"},
  "payment_method": {"card_redirect": {"type": "WEBPAY"}},
  "customer": {"email": {"value": "test@example.com"},
    "customer_document_details": {"document_type": "OTHER", "document_number": {"value": "<REDACTED RUT>"}}},
  "address": {"billing_address": {"first_name": {"value": "John"}, "last_name": {"value": "Doe"},
    "country_alpha2_code": "CL"}},
  "auth_type": "NO_THREE_DS",
  "return_url": "https://example.com/return"
}' \
  127.0.0.1:50124 types.PaymentService/Authorize

Outbound request captured by mock (POST /v3/deposits; Authorization: D24 <HMAC, verified>):
{"country":"CL","amount":10000.0,"currency":"CLP","invoice_id":"d24_wp_cl_0001","payer":{"document":"<REDACTED>","document_type":"RUT","email":"test@example.com","first_name":"John","last_name":"Doe"},"payment_method":"WP","success_url":"https://example.com/return","back_url":"https://example.com/return","error_url":"https://example.com/return"}

Response (mock reused the canned ONE_SHOT/SPEI 201 body, so payment_info below is SPEI):
{
  "merchantTransactionId": "postmanTest388400617",
  "connectorTransactionId": "301626485",
  "status": "AUTHENTICATION_PENDING",
  "statusCode": 201,
  "redirectionData": {
    "form": {
      "endpoint": "https://payment-stg.depositcheckout.com/v1/checkout/eyJhbGciOiJIUzM...",
      "method": "HTTP_METHOD_GET"
    }
  },
  "connectorFeatureData": {
    "value": "{\"checkout_type\":\"ONE_SHOT\",\"iframe\":true,\"payment_info\":{\"type\":\"VOUCHER\",\"payment_method\":\"SE\",\"payment_method_name\":\"Spei\",\"amount\":1000.0,\"currency\":\"MXN\",\"expiration_date\":\"2025-07-27 19:57:40\",\"created_at\":\"2025-07-24 19:57:40\",\"metadata\":{\"beneficiary_name\":\"D24 S.A.\",\"temporal_clabe\":false,\"beneficiary_overridden\":false,\"updated_clabe_flag\":false,\"offline_payments_enabled\":false,\"payer_name\":\"Ricardo Carlos\",\"instructions_alert\":true,\"clabe\":\"<REDACTED>\",\"legacy_view\":false},\"secondary_metadata\":null}}"
  },
  "connectorReferenceId": "postmanTest388400617",
  "typedConnectorResponse": {
    "value": "{\"deposit_id\":301626485,\"checkout_type\":\"ONE_SHOT\",\"redirect_url\":\"https://payment-stg.depositcheckout.com/v1/checkout/eyJhbGciOiJIUzM...\",\"iframe\":true,\"user_id\":\"11\",\"merchant_invoice_id\":\"postmanTest388400617\",\"payment_info\":{...same SPEI payment_info as test 1, clabe <REDACTED>...}}"
  }
}
grpcurl_exit=0
```

</details>

<details>
<summary>5. BR <code>bank_code "BB"</code> rejected locally, grpcurl Authorize (expected negative case)</summary>

```
grpcurl -plaintext \
  -H 'x-connector-config: <REDACTED>' \
  -H 'x-merchant-id: <REDACTED>' -H 'x-tenant-id: public' -H 'x-request-id: 5_br_bb_rejected' \
  -d '{
  "merchant_transaction_id": "d24_bt_br_0002",
  "amount": {"minor_amount": 100000, "currency": "BRL"},
  "payment_method": {"local_bank_transfer": {"bank_code": "BB"}},
  "customer": {"email": {"value": "<REDACTED>"},
    "customer_document_details": {"document_type": "CPF", "document_number": {"value": "<REDACTED>"}}},
  "address": {"billing_address": {"first_name": {"value": "Ricardo"}, "last_name": {"value": "Carlos"},
    "country_alpha2_code": "BR"}},
  "auth_type": "NO_THREE_DS"
}' \
  127.0.0.1:50124 types.PaymentService/Authorize

(no outbound request reached the mock)

Response:
ERROR:
  Code: Internal
  Message: grpc-status-details-bin mismatch: grpc-status=Unimplemented, grpc-message="LocalBankTransfer bank_code \"BB\" in billing country BR; supported Directa24 bank-transfer codes are MX: SE, COD, BM, STS, AF, BQL and BR: IX, I, NU, ME (BANK_DEPOSIT methods and methods that require reported_info, such as BB or B, are not supported) is not supported by d24", grpc-status-details-bin=message:"NOT_SUPPORTED"  1:"LocalBankTransfer bank_code \"BB\" in billing country BR; supported Directa24 bank-transfer codes are MX: SE, COD, BM, STS, AF, BQL and BR: IX, I, NU, ME (BANK_DEPOSIT methods and methods that require reported_info, such as BB or B, are not supported) is not supported by d24"  4:"https://docs.hyperswitch.io/prism/architecture/concepts/error-codes"
grpcurl_exit=77
```

</details>

## Hyperswitch -> UCS -> mock D24 e2e (juspay/hyperswitch#14209)

Hyperswitch `router` (juspay/hyperswitch#14209) → this UCS build → mock D24, on a local merchant with a D24 MCA. The mock checked every deposit request against the spec (31 checks: headers, HMAC, country, amount, currency, method code, document and type, names, email, phone, address, return URLs, client IP, language, description, invoice id).

| # | Scenario | Hyperswitch result | Outbound D24 request |
|---|---|---|---|
| 1 | MX SPEI, `bank_code: "SE"` | HTTP 200, `requires_customer_action`, redirect | 31/31 |
| 2 | MX, no `bank_code` (SPEI default) | HTTP 200, `requires_customer_action` | 31/31 |
| 3 | BR Pix `IX`, CPF | HTTP 200, `requires_customer_action` | 31/31 |
| 4 | BR, CNPJ | HTTP 200, `requires_customer_action` | 31/31 |
| 5 | MX, mock answers 400 (`code 201`) | payment `failed`, `error_code 201`, D24 details as the message | 31/31 |
| 6 | BR `bank_code: "BB"` | HTTP 501 `IR_00`, no D24 call | none (expected) |
| 7 | CL WebPay regression | HTTP 200, `requires_customer_action` | 31/31 |
| 8 | PSync, deposit `COMPLETED` | `succeeded` | 7/7 (signed `GET /v3/deposits/{id}`, read-only key) |
| 9 | Missing document (9a) / missing billing country (9b) | HTTP 400 `IR_04` | none (expected) |

All 9 scenarios pass. The first run found the `language` (`es-MX`, `pt-BR`) and unbounded `description` bugs fixed in `4e06679687` and `01113ac124`, and a Hyperswitch HTTP 500 on bank-transfer next steps fixed in juspay/hyperswitch `ef56dae37a`.

**Re-run after the review follow-ups (`38986611ab`..`e2ed0f788e`).** UCS rebuilt from `e2ed0f788e` (built before the rebase onto `d7d2380dea`, identical Rust tree); same Hyperswitch build, merchant and mock D24:

| Scenario | Covers | Result |
|---|---|---|
| 1 MX SPEI `SE` | `payment_info` masking (`38986611ab`), derived wire code (`e2ed0f788e`) | HTTP 200 `requires_customer_action`; outbound 31/31; stored `payment_attempt.connector_metadata` identical to the earlier run; UCS log shows `metadata: "*** serde_json::value::Value ***"` and no CLABE |
| 3 BR Pix `IX`, CPF | same | HTTP 200 `requires_customer_action`; 31/31; `connector_metadata` identical; the payer CPF no longer appears in the UCS log (it was printed in the clear before) |
| 7 CL WebPay | `"WP"` still sent after the serializer refactor | HTTP 200 `requires_customer_action`; 31/31 |
| 5 MX, mock answers 400 | error mapping unchanged | payment `failed`, `error_code 201`, same error message as before; 31/31 |

## Validation Checklist

- [x] `make validate-pre-push-fix` steps 1–5 clean on `832ddbf9a8`: `cargo +nightly fmt --all` (no changes), `cargo check --all-features --all-targets`, `cargo clippy --all-features --all-targets -- -D warnings`, `check_connector_specs`, `make generate`
- [x] Follow-up commits: `cargo +nightly fmt --all` clean, `cargo clippy -p connector-integration` clean
- [x] Review round (`49bb198d9`..`ee169d9eb`): `cargo +nightly fmt --all --check` clean, `cargo clippy -p connector-integration` clean, `cargo run --bin check_connector_specs` OK (118/118), `cargo test -p integration-tests all_supported_scenarios_match_proto_schema_for_all_connectors` and `all_override_entries_match_existing_scenarios_and_proto_schema` both pass
- [x] Review round behaviour re-verified against a mock Directa24 server (no sandbox credentials exist), driving UCS directly: PSync paid-in-full / underpaid / overpaid / pending / amount-absent, a USD request against an MXN deposit, and the refund guard with and without `payment_method`
- [x] `cargo run --bin check_connector_specs`: all checks passed (117/117 connectors)
- [ ] `make validate-pre-push-fix` step 6 (`make docs`) **failed locally, not because of this PR**. The docs were generated, but the Kotlin example compile check reported 23 errors, all in `examples/{etisalat,globalpay,globalpayments_heartland,merchante,paydotcom,payhere}` and **none in d24**. Cause: the check compiles against the Maven-local SDK jar (`~/.m2/.../prism-0.0.0-dev.jar`, built 2026-09-05, before those connectors merged). Its silent `gradlew clean publishToMavenLocal` refresh fails locally with a Kotlin *internal compiler error* in `sdk/java :compileKotlin`. This PR touches no proto, Java or Kotlin files.
- [x] `typos --config ./.typos.toml` clean
- [x] grpcurl Authorize returned success status (2xx): `AUTHENTICATION_PENDING`, `statusCode 201` (local mock server, not D24 staging)
- [x] Staged file set verified against an explicit manifest (`git diff --cached --name-only`) for every commit

https://claude.ai/code/session_01LfT18m1rR4WvAnamsrx9uf


XyneSpaces:
[blocking] Add connector-spec scenarios for the new D24 LocalBankTransfer authorize paths, covering at least one MX method (such as SPEI) and one BR method (such as Pix). The PR adds country-, currency-, and payer-document-specific request behavior, but without corresponding `connector_specs/d24/specs.json` coverage those branches can regress without the connector harness detecting it.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -621,7 +1098,7 @@ pub struct D24SyncResponse {
     pub invoice_id: Option<String>,
     pub user_id: Option<Secret<String>>,
     pub country: Option<String>,
-    pub currency: Option<common_enums::Currency>,
+    pub currency: Option<Currency>,
```
JeevaRamu0104, line 1111: `amount`, `local_amount`, and `usd_amount` are parsed here but never used — `Completed` always maps to `Charged` below regardless of what was paid. That was safe for WebPay's card settlement, but this PR routes manual bank transfers (SPEI, Pix) through the same path, where underpaying would still get recorded as fully charged. Should this check settled vs. requested amount first?

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -344,45 +680,89 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
+        //
+        // A local bank transfer carries no country and at most a `bank_code`,
+        // so its Directa24 code is resolved together with the billing country.
         let payment_method = match &request.payment_method_data {
             PaymentMethodData::CardRedirect(card_redirect_data) => {
                 D24PaymentMethod::try_from(card_redirect_data)?
             }
+            PaymentMethodData::BankTransfer(bank_transfer_data) => {
+                match bank_transfer_data.deref() {
+                    BankTransferData::LocalBankTransfer { bank_code } => {
```
JeevaRamu0104, line 696: No scenarios in `connector_specs/d24/specs.json` cover this — the new bank-code allow-list, per-country document validation, and Itaú address requirement have no regression coverage in the suite `make test-scenario` runs. Can you add a few (MX default, explicit `bank_code`, BR Pix, one rejected code)?

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -693,6 +1170,14 @@ impl TryFrom<ResponseRouterData<D24SyncResponse, Self>>
 ///   confirmed from the documentation; if it is wrong D24 answers HTTP 400
 ///   `804 MISSING_BANK_ACCOUNT`, which is cheap and unambiguous. Guessing a
 ///   bank account instead would send money to the wrong place.
+///
+///   A local bank-transfer deposit (SPEI, Pix, ...) *does* need a bank account
```
JeevaRamu0104, line 1238: Nothing here rejects a refund on a `LocalBankTransfer` deposit locally — it goes to D24 and comes back as a raw `804 MISSING_BANK_ACCOUNT`. Hyperswitch blocks this upstream today, but UCS has no independent guard if that changes. Worth a local `NotSupported` here for defense in depth?

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
```diff
@@ -272,18 +507,92 @@ pub struct D24PayerAddress {
...
+        )),
+        (DocumentKind::Cpf, CountryAlpha2::BR) => {
+            brazilian_payer_document(document, D24DocumentType::Cpf, "CPF")
+        }
+        (DocumentKind::Cnpj, CountryAlpha2::BR) => {
+            brazilian_payer_document(document, D24DocumentType::Cnpj, "CNPJ")
+        }
+        (DocumentKind::Other, CountryAlpha2::MX) => {
+            let normalized = document.document_number.peek().trim().to_ascii_uppercase();
+            if !(7..=18).contains(&normalized.len())
```
JeevaRamu0104: nit: `.len()` here is byte length but the error message a few lines down reports `.chars().count()` — harmless since non-ASCII already fails the alphanumeric check, but worth using the same unit.

JeevaRamu0104 (changes requested):

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Good catch — this was a real bug, not just dead parsing. Fixed in b143b7b8d.

UCS can express it: `AttemptStatus::PartialCharged` exists and is allowed on PSync (`domain_types/src/flow_status.rs`), and the settled figure goes on `PaymentFlowData::amount_captured` / `minor_amount_captured`, which the PSync response builder maps to `PaymentServiceGetResponse.captured_amount`. The requested amount is available as `PaymentsSyncData::amount` (non-optional on that flow), so the comparison needed no new plumbing.

What it now does:

- `COMPLETED` with settled < requested -> `PartialCharged`; `amount_captured` / `minor_amount_captured` set to the settled amount.
- `status_reason` (`AMOUNT_MISMATCH`, `THIRD_PARTY_PAYER`) is parsed and surfaced as `raw_connector_status.reason`; it was being dropped.

Three deliberate non-behaviours:

- The status is only ever downgraded. An overpaid deposit stays `Charged` with the settled figure reported — the surplus is a reconciliation matter, not a payment-status one.
- Currency. `amount` and `local_amount` are both documented as the amount in the deposit's local currency, `usd_amount` as the USD equivalent, and a D24 deposit may be created in either (Authorize guard 3 allows local or USD). The comparison only uses a figure it can pin to the requested currency; a currency it cannot line up leaves the amount unknown, because comparing MXN against a USD request would flag every such deposit as underpaid.
- An unknown or unconvertible amount falls back to the connector's own status rather than inventing a partial payment or failing the sync.

Verified against a mock Directa24 — there are no D24 sandbox credentials, so mock is the only option. Driving `PaymentService/Get` directly at UCS: paid in full -> `CHARGED` / captured 50000 (unchanged from the previous behaviour, now with the amount reported); underpaid 300.00 of 500.00 MXN -> `PARTIAL_CHARGED` / captured 30000 with `reason: AMOUNT_MISMATCH`; overpaid -> `CHARGED` / captured 60000; `PENDING` -> `PENDING` with no captured amount; amounts absent from the response -> `CHARGED` with no captured amount; and a USD request against an MXN-denominated deposit correctly reads `usd_amount` (2500 full, 1500 underpaid).

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Yes — worth it, and it turned out to be possible. Added in 1bfd8d0e3.

The refund request does carry enough, just not where the old comment claimed. `RefundsData` has no payment method *type*, but it does have `payment_method_data: Option<PaymentMethodData<DefaultPCIHolder>>`, populated from `PaymentServiceRefundRequest.payment_method` via `convert_to_domain_model_for_non_card_payment_methods`, and that conversion does handle `LocalBankTransfer` -> `BankTransferData::LocalBankTransfer { bank_code }`. So the guard reads it and refuses locally.

The limit is honest and documented in the code: `payment_method` is optional on the refund request. When the caller omits it the refund is attempted as before and D24 still answers `804 MISSING_BANK_ACCOUNT`. Refusing on an absent optional was not an option — it would block WebPay refunds, which are supported. So this is defence in depth for callers that send the field, not a complete gate.

Message style follows the fix that stopped these repeating the phrase: the `message` is a noun phrase, so `NotSupported` renders it as `refunding a LocalBankTransfer deposit (Directa24 returns it to a bank account and requires bank_account, which no UCS refund field carries) is not supported by d24`, with the out-of-band suggestion in `suggested_action`.

Verified against a mock Directa24 (no sandbox credentials exist), driving `PaymentService/Refund` directly at UCS: with `payment_method.local_bank_transfer` it returns `Unimplemented` with that message and the mock records no `/v3/refunds` call at all; with the field omitted the request still reaches the connector. The doc comment on `D24RefundRequest` was updated to match.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Fixed in 49bb198d9 — the length is computed once as `normalized.chars().count()` and the same value is interpolated into the message, so the number reported is the number that was tested.

crates/integrations/connector-integration/src/connectors/d24/transformers.rs
shuklatushar226: Added in ee169d9eb, as `connector_specs/d24/override.json`. Four Authorize scenarios, covering the cases you asked for plus the two other gaps you named:

| Scenario | Covers |
|---|---|
| `no3ds_auto_capture_sepa_bank_transfer` | MXN, no `bank_code` -> SPEI (`SE`) default; Mexican `OTHER` document (7-18 alphanumeric) |
| `no3ds_auto_capture_ach_bank_transfer` | BRL, explicit `bank_code: "I"` (Itau) -> the full-billing-address requirement; CNPJ check-digit validation |
| `no3ds_auto_capture_bacs_bank_transfer` | BRL Pix, explicit `bank_code: "IX"`; CPF check-digit validation |
| `no3ds_fail_payment` | BRL `bank_code: "BB"` -> rejected by the allow-list before any Directa24 call |

Two things worth flagging.

**Why `override.json` and not `connector_specific_scenarios.json`.** The latter is the right home for connector-private scenarios and I wrote it that way first, but the harness only *enumerates* that file. `run_suite_test_with_options` merges it into the scenario set, then `execute_single_scenario_with_context` and `load_effective_scenario_for_connector` both resolve the body through `load_scenario`, which reads the global suite file alone — so every private scenario fails to build with `scenario 'x' not found in suite 'PaymentService/Authorize'`, and `all_supported_scenarios_match_proto_schema_for_all_connectors` fails on exactly that. That looks like a harness gap rather than anything D24-specific (no connector uses the file today), and fixing it is out of scope for this PR, so these go through `override.json`, which *is* applied on the execution path. Happy to file it separately if that's useful.

The cost is that the scenario names are the global ones, so `make test-scenario connector=d24 suite=authorize scenario=no3ds_auto_capture_sepa_bank_transfer` runs a Mexican SPEI deposit. Not ideal; it's what the mechanism allows today.

**What actually gates in CI right now.** D24 is in `alpha_connectors.json` ("No Directa24 sandbox credentials are configured in CI"), so `certify-connectors.sh` skips it — these will not be executed live until credentials land. What does run and gate this PR is `all_override_entries_match_existing_scenarios_and_proto_schema`, which builds and proto-schema-checks every one of these four requests. So today the coverage is "these requests are well-formed and pinned"; the branch-level regression coverage arrives with the credentials.

I did not add `supported_payment_methods` to `specs.json`: the payment-method filter reads the scenario's method *before* the override is applied, so declaring `local_bank_transfer` would filter out the very scenarios being specialised.

Validation: `cargo run --bin check_connector_specs` OK (118/118, no missing suites), and both schema tests pass.

shuklatushar226:
@XyneSpaces — connector-spec scenarios added in `ee169d9eb`, as `crates/internal/integration-tests/src/connector_specs/d24/override.json`. One MX method and two BR methods, plus a rejection:

- **MX / SPEI** (`no3ds_auto_capture_sepa_bank_transfer`) — MXN, no `bank_code`, exercising the `SE` default and the Mexican `OTHER` document rule (7-18 alphanumeric, no `document_type` emitted).
- **BR / Itau** (`no3ds_auto_capture_ach_bank_transfer`) — BRL, explicit `bank_code: "I"`, exercising the explicit allow-list path, the Itau full-billing-address requirement, and CNPJ check-digit validation.
- **BR / Pix** (`no3ds_auto_capture_bacs_bank_transfer`) — BRL, explicit `bank_code: "IX"`, with CPF check-digit validation.
- **BR / rejected code** (`no3ds_fail_payment`) — BRL `bank_code: "BB"`, which the allow-list refuses before any Directa24 call.

So each of the three country-, currency- and document-specific branches this PR adds now has a pinned request.

Two caveats I'd rather state than leave implied:

1. These go through `override.json` rather than `connector_specific_scenarios.json`. The latter is the intended home for connector-private scenarios, but the harness only enumerates it — the execution path resolves scenario bodies through `load_scenario`, which reads the global suite file alone, so a private scenario fails to build. Details in the inline thread; it looks like a general harness gap rather than anything D24-specific, and I left it alone here. The side effect is that the scenarios keep their global names.
2. D24 is in `alpha_connectors.json` (no Directa24 sandbox credentials in CI), so `certify-connectors.sh` skips it and these are not executed live yet. What gates today is `all_override_entries_match_existing_scenarios_and_proto_schema`, which builds and proto-schema-checks all four requests. Live branch coverage follows the credentials.

The behaviour itself was verified against a mock Directa24 server, since no sandbox credentials exist.

Validation on this push: `cargo +nightly fmt --all --check` clean, `cargo clippy -p connector-integration` clean, `cargo run --bin check_connector_specs` OK, and both `all_supported_scenarios_match_proto_schema_for_all_connectors` and `all_override_entries_match_existing_scenarios_and_proto_schema` pass.

JeevaRamu0104 (approved):

deepanshu-iiitu (approved):

deepanshu-iiitu (approved):

PR #2297, opened 2026-09-15
fix(connector): [PAYHERE] make billing address/city/country optional

peeyushshukla-juspay:
## Description

1. **Added error context in Authorize response handling** — when checkout field assembly fails, the connector now propagates the inner cause as the outer message (e.g. `payhere: failed to build checkout request fields: Missing required field: payment_method_data.billing.address.first_name`) instead of returning a bare `ResponseHandlingFailed`, so callers get the actual field name in error responses.

2. **Relaxed mandatory billing fields** — the following are now optional and degrade to an empty string in the hosted-checkout form when absent:
   - `billing.address.line1`
   - `billing.address.city`
   - `billing.address.country`

  `first_name` / `last_name` remain mandatory; `email` / `phone` keep the billing → customer fallback chain.
### Additional Changes


crates/integrations/connector-integration/src/connectors/payhere/transformers.rs
```diff
@@ -314,12 +314,25 @@ impl PayherePaymentsRequest {
...
-        let address = router_data.resource_common_data.get_billing_line1()?;
-        let city = router_data.resource_common_data.get_billing_city()?;
+        // address/city/country are OPTIONAL: send them when present,
+        // otherwise send an empty string in the checkout form.
+        let address = router_data
+            .resource_common_data
+            .get_optional_billing()
+            .and_then(|a| a.address.as_ref())
+            .and_then(|ad| ad.line1.clone())
+            .unwrap_or_default();
```
Aishwariyaa-Anand, line 322: can we use this - get_optional_billing_line1

crates/integrations/connector-integration/src/connectors/payhere/transformers.rs
```diff
@@ -314,12 +314,25 @@ impl PayherePaymentsRequest {
...
+        let city = router_data
+            .resource_common_data
+            .get_optional_billing()
+            .and_then(|a| a.address.as_ref())
+            .and_then(|ad| ad.city.clone())
+            .unwrap_or_default();
```
Aishwariyaa-Anand, line 326: get_optional_billing_city

crates/integrations/connector-integration/src/connectors/payhere/transformers.rs
Aishwariyaa-Anand, line 335: get_optional_billing_country

kanikac199 (approved):

crates/integrations/connector-integration/src/connectors/payhere/transformers.rs
```diff
@@ -314,12 +314,21 @@ impl PayherePaymentsRequest {
...
                 "billing.phone / customer.phone_number",
             ))?;
-        let address = router_data.resource_common_data.get_billing_line1()?;
-        let city = router_data.resource_common_data.get_billing_city()?;
+        // address/city/country are OPTIONAL: send them when present,
+        // otherwise send an empty string in the checkout form.
+        let address = router_data
+            .resource_common_data
+            .get_optional_billing_line1()
+            .unwrap_or_default();
```
Aishwariyaa-Anand, line 322: why do you need unwrap_or_default, aren't they optional? Is this field mandatory for connector?

peeyushshukla-juspay: PayHere's hosted-checkout API marks address, city and country as mandatory form fields. We relaxed them on our side (don't hard-fail the request when the caller omits them), unwrap_or_default() gives the empty-string body value, which is what the gateway receives when the caller skipped them. Omitting the key entirely would violate their spec

Aishwariyaa-Anand (approved):

PR #2279, opened 2026-09-09
feat(connector): [Shift4] enrich Authorize with AVS/CVV checks, issuer decline codes, shipping and merchant reference

shuklatushar226:
## Summary

Enrich the **Authorize** flow of the **Shift4** connector.

Shift4's Authorize flow already existed — this PR is an **enrichment**, not a new integration. It closes the gaps against the technical specification, verifies live the parts that were already correct, and fixes three defects found in the Hyperswitch-parity audit.

Generated and validated by **GRACE** (automated connector integration pipeline), then exercised end-to-end against the `api.shift4.com` sandbox.

## Scope delivered

| Scope item | Status |
|---|---|
| CIT card payments | IMPLEMENTED + TESTED (`type: "customer_initiated"` echoed on the charge) |
| Auto vs manual capture | IMPLEMENTED + TESTED (AUTOMATIC → `CHARGED`/`captured:true`; MANUAL → `AUTHORIZED`/`captured:false`; follow-up Capture → `CHARGED`) |
| $0 zero-amount auth (Account Name Inquiry) | IMPLEMENTED + TESTED (`amount:0` → `AUTHORIZED`, `aniCheck.result: "full_match"`) |
| AVS check result | IMPLEMENTED + TESTED (`avs_result: "full_match"` with card `4242000000000315` + zip `00000`; `"unavailable"` otherwise) |
| CVV/CVC check response | IMPLEMENTED + TESTED (`card_validation_result` `"match"` / `"no_match"`; charge status deliberately **not** derived from `cvvCheck`) |
| Billing details | IMPLEMENTED + TESTED (`billing.{name,email,phone,address}`; phone joined as `+1 4155551234`) |
| Shipping details | IMPLEMENTED + TESTED (`shipping.{name,phone,address}`) |
| Issuer/network error mapping (smart retry / GSM) — **Gap #2** | IMPLEMENTED + TESTED (402 decline surfaces `issuerDeclineCode "N7"` → `network_decline_code`, `chargeId` → `connector_transaction_id`, plus `adviceCode` / `networkAdviceCode`) |
| MAC codes | IMPLEMENTED (parsed on both decline shapes); value untested — Shift4 publishes no test card returning `do_not_try_again` |
| Merchant reference ID | IMPLEMENTED + TESTED (`external.vendorReference`) |
| L2/L3 data | **NOT SUPPORTED BY SHIFT4** — recorded as an explicit code comment; no wire field invented |
[...]

PR #2303, opened 2026-09-15
feat(connector): [Shift4] recurring payments for card, Apple Pay and Google Pay

shuklatushar226:
## Summary

Targets `main`, alongside #2279 (Shift4 **Authorize enrichment** + **Void**).

**Merge #2279 first.** The recurring code in this PR is written on top of the enrichment and Void commits, so those 4 commits ride along in this branch and show up in this diff until #2279 lands. Once #2279 merges, this PR's file diff reduces to the recurring work on its own, with no rebase needed and no conflict expected, since the content is identical. They cannot be made fully independent: replaying the recurring commits onto `main` without the enrichment conflicts on the first commit, and the recurring transformers call helpers that the enrichment introduces.

This PR adds **recurring payments** for **card**, **Apple Pay** and **Google Pay**:

- **SetupMandate** (SetupRecurring)
- **RepeatPayment** (RecurringPaymentService/Charge)
- **off-session CIT Authorize**, which returns a mandate

It also carries the **review, logic-parity and E2E fixes** found while validating the above, including:

- a **security fix**: the Shift4 `Authorization` header was logged unmasked (pre-existing on `main`)
- a **partial-capture fix**: `POST /charges/{id}/capture` now sends `amount`, so Shift4 settles the captured amount instead of the full authorization
- **captured-amount reporting**, so Hyperswitch records `amount_received`

Generated with **GRACE**, then exercised against the `api.shift4.com` sandbox by grpcurl, and end-to-end as Hyperswitch -> UCS -> Shift4 for both card and wallets.

Companion Hyperswitch PR: https://github.com/juspay/hyperswitch/pull/14210

**Per-run evidence** was posted as comments on #2279 while this work still lived on that branch:

- [SetupMandate / card](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672056983)
- [RepeatPayment / card + CIT mandate](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672450722)
- [SetupMandate / wallet](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672973323)
- [RepeatPayment / wallet + wallet Authorize](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673276858)
- [Fix run](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673818577) and [follow-up](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673935305)
- [HS -> UCS -> Shift4 final regression](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5674066016)
- [Captured-amount fix](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5674202261) and [its HS E2E](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5674245943)

> The tables below describe the Shift4 connector as a whole. Authorize enrichment and Void come from #2279; everything recurring, plus the fixes listed above, is this PR.

## What the connector does, per flow and payment method

| Flow | Card | Apple Pay / Google Pay (encrypted token) |
|---|---|---|
| **Authorize** | Inline `card`. Typed `customer_initiated`, or `first_recurring` for an off-session CIT. The CIT returns mandate `card_...`. | Inline `paymentMethod{type, token, billing}`. An off-session CIT returns mandate `pm_...`; a one-off sale returns no mandate. |
| **SetupMandate** (SetupRecurring) | `POST /charges` typed `first_recurring` under a Shift4 customer. Returns `card_...` and the customer. | Same shape, as `paymentMethod`. Returns `pm_...` and the customer. |
| **RepeatPayment** (RecurringPaymentService/Charge) | `card: "card_..."` + `customerId`. A NetworkMandateId sends the raw card + `external.schemeTransactionId`. | `paymentMethod: "pm_..."` + `customerId`. No charge-level billing. |
| **Capture** | `POST /charges/{id}/capture {amount}`, which settles exactly that amount. | Same. |
| **Void** | `POST /refunds {chargeId}` on the uncaptured charge. PSync reads `captured:false, refunded:true` as `VOIDED`. | Same. |
| **PSync / Refund / RSync / IncrementalAuthorization / CustomerService.Create / ClientAuthenticationToken** | Implemented. | PSync, Refund and RSync are shared. |

**Common behaviour**

- **Customer:** `customerId` always comes from `connector_customer_id`, never from the merchant customer id.
- **Mandate discriminator:** the `pm_` prefix of the Shift4 `connector_mandate_id`, defined once.
  - Hyperswitch payment method ids also start with `pm_`. The prefix is therefore only ever tested on `connector_mandate_id`.
- **Mandates returned:**
  - A `pm_` is returned only while a customer owns it and its `status` is `chargeable`.
  - A `card_` is returned only while a customer owns it. Card and Shift4 card-token (`tok_...`) charges return `paymentMethod: null`, so they always take the `card_` branch (sandbox-verified for an off-session token CIT and a token SetupRecurring).
- **Decline mapping:**
  - A decline over HTTP 402 is `FAILURE` on Authorize, SetupMandate and RepeatPayment, with `issuerDeclineCode` / `adviceCode` / `networkAdviceCode` / `chargeId` preserved.
  - An HTTP 200 `status: failed` charge is returned as an error, with the same codes.
- **Not supported by Shift4, and deliberately not approximated:**
  - L2/L3 data
  - a dynamic statement descriptor
- **Zero-amount charges** are always sent `captured: false`. Shift4 rejects capturing a zero amount with "Zero amount charge cannot be captured".

## Behaviour changes to call out

1. **Off-session CIT without `connector_customer_id` is refused (U17).**
   - **Affects:** card, token, Apple Pay and Google Pay Authorize with `setup_future_usage=OFF_SESSION` + customer acceptance, and every SetupRecurring.
   - **Result:** refused locally with `MissingRequiredField customer.connector_customer_id` (gRPC `InvalidArgument`, Hyperswitch IR_04) before the shopper is charged.
   - **Why:** Shift4 charges a stored card or payment method only together with its owning customer. A customer-less `pm_` would be claimed by whichever customer charges it first.
   - **Companion:** the Hyperswitch PR creates the Shift4 customer for these payments.
2. **SetupRecurring capture (U4).**
   - **Zero amount:** always an uncaptured verification, reported `CHARGED`. This applies to card and wallet alike.
   - **Non-zero amount:** always **captured, and the funds settle**, reported `CHARGED`. `PaymentServiceSetupRecurringRequest` has no `capture_method` field, so there is no way to ask for an open authorization. The `cc7aadee7d` commit message says a manual non-zero setup stays `AUTHORIZED`; that branch was unreachable and was removed in the follow-up commit.
   - **Previously:** a non-zero card setup was reported `CHARGED` over an uncaptured hold that never settled, and a non-zero wallet setup was reported `AUTHORIZED`.
   - **Hyperswitch:** routes a non-zero `payment_type=setup_mandate` to SetupMandate, so this is visible there.
   - **Sandbox check:** Shift4 accepts `captured: true` with `type: first_recurring`, still stores the card / payment method under the customer, and an MIT on it succeeds.
3. **Cardholder name (U5).**
   - `card.cardholderName` is the card's own holder name only, with no fallback to the billing name or the customer name, on every flow.
   - It is no longer required on a zero-amount charge; the sandbox accepts one without it.
4. **Partial capture (U12).**
   - Capture sends `amount`, and Shift4 settles exactly that amount and releases the rest.
   - Before this, a partial capture settled the full authorization (reproduced from Hyperswitch: HS recorded 400 while Shift4 captured 1000).
   - `amount` is not in the API reference, but is sandbox-verified:
     - the 1000 charge reads back 400 after capturing 400
     - an over-capture is refused with "Invalid Capture data"
     - a second capture is refused with "Requested Charge is already captured"
   - Multiple-capture intents stay rejected.
   - The Capture response reports `CHARGED` even after a partial capture. `PartialCharged` cannot be derived: Shift4 rewrites the charge `amount` to the captured amount and returns no authorized amount, and the capture request carries none either. Hyperswitch derives `partially_captured` from its own amounts, so it is unaffected; a direct gRPC/SDK caller has to compare `amount_to_capture` with the authorized amount itself.
5. **`customerId` on Authorize.**
   - Sent only when the charge stores the credential, or charges a token.
   - A one-off card, wallet or bank-redirect sale no longer attaches the card or payment method to a known customer.
6. **Authorize 402 is `FAILURE` (U10),** consistent with SetupMandate and RepeatPayment. Other errors still carry no attempt status, so the caller decides; Hyperswitch marks non-5xx errors failed.
7. **Wallet connector response (W2).** Apple Pay / Google Pay charges report the auth code under `additionalPaymentMethodData.applePay` / `googlePay`, not `card`. AVS / CVV are not reported for wallets.
8. **Typed response (CW7).** `typedConnectorResponse` of a card charge now also contains `"paymentMethod":null`. This is additive.
9. **Empty values are omitted:**
   - `Idempotency-Key` is omitted when neither `merchant_request_id` nor the merchant reference is set.
   - `external.vendorReference` is omitted when the merchant reference is empty.
   - `card.cvc` is omitted when absent or empty.
   - The key is `merchant_request_id`, else the merchant reference (`merchant_transaction_id` / `merchant_recurring_payment_id` / `merchant_charge_id`), and is never generated per call.

## Known limitations

- **3DS (U13).** PreAuthenticate / Authenticate / PostAuthenticate are not implemented for Shift4 in UCS. A Hyperswitch 3DS card CIT routed through UCS therefore fails at pre-authentication, before Authorize. Only no-3DS card flows are supported through UCS.
- **Pre-decrypted wallets and third-party SDK tokens.** Pre-decrypted Apple Pay / Google Pay and third-party SDK wallet tokens return `NotSupported` (Hyperswitch IR_00). Shift4 has no request field for a DPAN, cryptogram or ECI. The MCA must not enable Apple Pay "simplified" / decrypt-at-Hyperswitch or `google_pay.support_predecrypted_token` for shift4.
- **Google Pay PAN_ONLY.** Only `CRYPTOGRAM_3DS` tokens work. A `PAN_ONLY` token needs browser 3DS, and Shift4 declines it with a final HTTP 402 `authentication_required`.
- **Apple Pay JSON token (CW8).** A token that decodes to a JSON `paymentData` object is sent to Shift4 as an object; the sandbox `TEST_TOKEN:<amount><CCY>` tokens are sent as a string.
  - With a structurally valid PKPaymentToken carrying fake cryptography, Shift4 answered HTTP 400 `ApplePay token is malformed`. Raw probes got the same verdict for a fake token sent as an object and as a string.
  - Acceptance of a real production token is not verified, because no real token exists in the sandbox.
- **Redirect wallets.** AliPay, WeChat Pay, Paysera and Skrill, which native Hyperswitch Shift4 supports, return `NotImplemented` through UCS. This predates this branch.
- **Network transaction id MIT.** Accepts a raw `card` only. `CardDetailsForNetworkTransactionId` returns `NotSupported`; shift4 is not in Hyperswitch's NTI connector list.
- **SetupRecurring capture method.** A non-zero setup cannot be kept as an open authorization, because the SetupRecurring proto has no `capture_method`. Adding one is a proto change and is left as a follow-up.
- **Idempotency-Key and retries.** The key is the caller's reference. If a caller reuses one reference across retry attempts of a payment within 24 h, Shift4 replays the first response. Hyperswitch sends attempt-scoped references by default.

## Hyperswitch companion and E2E

**https://github.com/juspay/hyperswitch/pull/14210:**
- creates the Shift4 customer for off-session card CITs and zero-amount setups
- adds shift4 to the mandate payment-method lists
- fixes the create-then-confirm path so `setup_future_usage` set at confirm is honoured

**E2E Hyperswitch -> UCS -> Shift4:**
- **Card:** passed on the pre-fix binary. It surfaced U12 (partial capture), U13 (3DS) and U17.
- **Wallet:** pending.
- **Regression:** a final Hyperswitch regression runs on the binary built from this head.

## Test coverage

**`crates/internal/integration-tests/src/connector_specs/shift4/specs.json`**
- It declares `supported_payment_methods` (`card`, `apple_pay_sdk`, `google_pay_sdk`, `ideal`, `eps`).
- It also marks the global Google Pay Authorize scenario unsupported. The harness swaps that scenario's token for a browser-generated one, which needs `GPAY_HOSTED_URL` and a `metadata.google_pay` block in the creds; Shift4's creds have neither.

**Apple Pay / Google Pay integration scenarios are pending two harness gaps.** Earlier comments on this PR said the harness does not load connector-specific scenario files (CW2). That is not quite right: the file is loaded, but the harness cannot run its scenarios because of these gaps (line numbers at `cc7aadee7d`):
1. **Connector-specific scenarios cannot be built.**
   - `crates/internal/integration-tests/src/harness/scenario_api.rs:65-70`: `load_effective_scenario_for_connector` only calls `load_scenario` (`scenario_loader.rs:107`), which reads the global suite file. `scenario_api.rs:4868` has the same `load_scenario` call.
   - The suite run and the schema check both merge `connector_specific_scenarios.json` in (`scenario_loader.rs:294`), but every per-scenario request is built through that function, so a private scenario fails with `scenario '<name>' not found in suite`.
2. **Apple Pay `payment_data` is not normalized.**
   - `scenario_api.rs:3325-3334`: `normalize_proto_oneof_shapes` adds the prost wrapper for Google Pay `tokenization_data`, but not for Apple Pay `payment_data`.
   - `validate_tonic_payload_shape` (`:5209`) therefore reports `apple_pay_sdk.payment_data.encrypted_data` as an `unknown/ignored request field`.

Four Apple Pay scenarios are ready: Authorize one-off and off-session CIT, and SetupRecurring 1500 and 0, on Shift4's amount-bound sandbox token. They pass the schema validator only once both gaps are fixed. They are deferred to a separate harness PR rather than changing shared test code here.

Until then, wallet coverage is the grpcurl evidence below:
- Apple Pay and Google Pay SetupRecurring, off-session CIT, MIT, one-off sale and partial capture
- negative cases

A wallet RecurringPaymentService/Charge scenario would also need that suite's SetupRecurring dependency to resolve to a wallet setup.

**Field probe.** `probe-config.toml` gives the Shift4 setup flows a probe `customer.connector_customer_id`. `setup_recurring` / `proxy_setup_recurring` / `token_setup_recurring` now report `supported` instead of `error`, and the `all_connector.md` Shift4 cells SetupRecurring, TokenSetupRecurring and ProxySetupRecurring change from `?` to `✓`.

## Validation

Run on head `6b823ff959` with `CARGO_INCREMENTAL=0`, and again on `0d9e294cb8` after the follow-up commit, with the same results:

- **`scripts/validation/pre-push.sh --fix`**
  - formatting: no changes needed
  - `cargo check --all-features --all-targets`: pass
  - `cargo clippy --all-features --all-targets -- -D warnings`: pass
  - `check_connector_specs`: `[OK] shift4 (11 flows mapped)`
  - `make generate`: pass
  - `make docs`: fails only on 7 Kotlin compile errors in the stale `examples/globalpay/globalpay.kt` and `examples/paydotcom/paydotcom.kt`, unrelated to shift4. The Shift4 docs and examples regenerated, and `all_connector.md` was regenerated with `--no-syntax-check`.
- **Schema validators:** `all_supported_scenarios_match_proto_schema_for_all_connectors` ok, `all_override_entries_match_existing_scenarios_and_proto_schema` ok.
- **Connector crate:** `cargo clippy -p connector-integration --all-targets -- -D warnings` clean; `cargo +nightly fmt --all` clean.
- **Spelling:** `typos --config ./.typos.toml` clean on the changed files.
- **Build and grpcurl:** debug `grpc-server` builds; 29 of 29 fix-run grpcurl legs and 13 of 13 follow-up legs pass against the Shift4 sandbox.
- **Credentials:** 0 occurrences of the api key (raw, `base64(key:)`, `base64(key)`) in the server log after the U18 fix.

## grpcurl evidence (fix run, redacted)

**Setup:** debug `grpc-server` built from this head, port 50061, Shift4 sandbox. The api key was sent in `x-connector-config` and redacted as `<REDACTED>` everywhere. After the U18 masking fix, the server log contained **0** occurrences of the key in raw form, as `base64(key:)` and as `base64(key)`.

| # | PM | Leg | Outcome | Result |
|---|---|---|---|---|
| 1 | Card | CustomerService/Create | cust_ returned | PASS |
| 2 | Card | Authorize AUTOMATIC (one-off) | CHARGED, no customerId | PASS |
| 3 | Card | Authorize MANUAL -> Capture 1000 | AUTHORIZED -> CHARGED | PASS |
| 4 | Card | Authorize MANUAL 1000 -> Capture 400 (U12) | CHARGED, PSync amount 400 | PASS |
| 5 | Card | Authorize MANUAL -> Void | VOIDED, PSync VOIDED | PASS |
| 6 | Card | PSync | CHARGED, PAR returned | PASS |
| 7 | Card | Refund 300 -> RSync | REFUND_SUCCESS | PASS |
| 8 | Card | SetupRecurring 0, no card holder name | CHARGED (captured:false), card_ | PASS |
| 9 | Card | SetupRecurring 1500 (U4) | CHARGED (captured:true), card_ | PASS |
| 10 | Card | Charge on SetupRecurring card_ | CHARGED | PASS |
| 11 | Card | Authorize off-session CIT -> Charge on its card_ | CHARGED | PASS |
| 12 | Card | Charge without merchant_charge_id (U1/U2) | CHARGED, no Idempotency-Key, no external | PASS |
| 13 | Card | NetworkMandateId MIT, raw card (U3) | CHARGED, external.schemeTransactionId sent | PASS |
| 14 | Card | Authorize decline 4916018475814056 (U10) | 402 card_declined, status FAILURE | PASS |
| 15 | Card | Authorize card_cvc "" (U14) | CHARGED, no cvc sent | PASS |
| 16 | Card | SetupRecurring without connector_customer_id | InvalidArgument, no request | PASS |
| 17 | Card | NetworkMandateId MIT without card | Unimplemented, no request | PASS |
| 18 | Apple Pay | SetupRecurring 1500 -> Charge | CHARGED -> CHARGED | PASS |
| 19 | Apple Pay | SetupRecurring 0 -> Charge | CHARGED -> CHARGED | PASS |
| 20 | Google Pay | SetupRecurring CRYPTOGRAM_3DS -> Charge | CHARGED -> CHARGED | PASS |
| 21 | Apple Pay | Authorize off-session CIT -> Charge | CHARGED pm_ -> CHARGED | PASS |
| 22 | Google Pay | Authorize off-session CIT -> Charge | CHARGED pm_ -> CHARGED | PASS |
| 23 | Apple Pay | Authorize one-off | CHARGED, no customerId, no mandate | PASS |
| 24 | Apple Pay | Authorize MANUAL 2000 -> Capture 500 (W-P5) | CHARGED, PSync amount 500 | PASS |
| 25 | Apple Pay | Authorize JSON PKPaymentToken (CW8) | 400 ApplePay token is malformed (expected, fake crypto) | PASS |
| 26 | Apple Pay | Authorize CIT without connector_customer_id | InvalidArgument, no request | PASS |
| 27 | Apple Pay | Charge without connector_customer_id | InvalidArgument, no request | PASS |
| 28 | Apple Pay | SetupRecurring pre-decrypted | Unimplemented (NotSupported) | PASS |
| 29 | Google Pay | Authorize pre-decrypted | Unimplemented (NotSupported) | PASS |

<details>
<summary>Partial capture (U12): card 1000 -> capture 400; Apple Pay 2000 -> capture 500</summary>

Outgoing capture bodies (server log, Authorization stripped):

```
POST /charges/char_6WvtE28LBExzXuU23V3ZiKHm/capture  {"amount":400}   -> CHARGED
GET  /charges/char_6WvtE28LBExzXuU23V3ZiKHm           -> status CHARGED, amount 400, captured true
POST /charges/char_alxJQrOz0Dmrk0mJ91DvmnjH/capture  {"amount":500}   -> CHARGED
GET  /charges/char_alxJQrOz0Dmrk0mJ91DvmnjH           -> status CHARGED, amount 500, captured true
```

</details>

<details>
<summary>Name rule (U5), one-off customerId, empty values (U1/U2/U14)</summary>

- **Card Authorize, one-off** (`card_holder_name` "Card Holder", billing name "Billing Person"):
  - The outgoing body has no `customerId`.
  - Shift4 `GET /charges/char_gLf2TahtvFlrWLVdX38SRUFn` returns `type: customer_initiated`, `customerId: null`, `card.cardholderName: "Card Holder"`, `billing.name: "Billing Person"`.
- **Card off-session CIT** (`char_QdVAuoCujG5c4EN4Qmsb8Ifd`): `customerId` and `card.customerId` set, mandate `card_GUyLTSz0f1dOL2WodqklMMyL`.
- **Apple Pay one-off** (`char_81nOvsVNmTXsxMyZg4WaAKM6`): no `customerId`, `paymentMethod.customerId: null`, no mandate.
- **Zero-amount card SetupRecurring without a card holder name:** the body has no `cardholderName`, status `CHARGED`, mandate `card_A8oiiU8i5lGmFY3YZQxx3PQD`.
- **RecurringPaymentService/Charge without `merchant_charge_id`:** headers have no `Idempotency-Key`, the body has no `external`, status `CHARGED`.
- **Authorize with `card_cvc: ""`:** the card object has keys `number, expMonth, expYear, cardholderName`, no `cvc`, status `CHARGED`.

</details>

<details>
<summary>NetworkMandateId MIT (U3) and decline (U10)</summary>

- **NTI MIT:** the outgoing body has the raw card, no `customerId`, `type: merchant_initiated`, and `external: {vendorReference, schemeTransactionId}` carrying the network transaction id of the earlier CIT. Result `CHARGED` (`char_BAdIibalS7UuXQLha7JclAmp`).
- **NetworkMandateId with no card:** `Unimplemented` "NetworkMandateId without raw card details is not supported by Shift4 ...". No request is sent.
- **Decline card `4916018475814056` on Authorize:** HTTP 402 `card_declined`, issuer code `05`, `chargeId` preserved. The error details now carry payment status `FAILURE` (21), which is absent on the pre-fix Authorize decline.

</details>

<details>
<summary>Wallet legs</summary>

- **Apple Pay SetupRecurring 1500:** `captured: true`, `CHARGED`, `pm_C3Ow1S3VUj5dD177VtUCV9mL` -> Charge 800 `CHARGED`.
- **Apple Pay SetupRecurring 0:** `captured: false`, `CHARGED`, `pm_CRzH0Nd0PXogrTSJHsHsTAhy` -> Charge 500 `CHARGED`.
- **Google Pay CRYPTOGRAM_3DS SetupRecurring 1500:** `CHARGED`, `pm_cHdzGzdpDGfuXO2KZj8z4uue` -> Charge 900 `CHARGED`.
- **Off-session Apple Pay Authorize 2000:** `CHARGED`, `pm_s9a4VhGOZoHHMNHmQwQAkWmS` -> Charge 1200 `CHARGED`.
- **Off-session Google Pay Authorize 2500:** `CHARGED`, `pm_5IMLrLDfrUKVLruHd9lpGxHh` -> Charge 1300 `CHARGED`.
- **Connector response:** wallet responses carry `connectorResponse.additionalPaymentMethodData.applePay|googlePay.authCode`.
- **Refused locally, with no request sent:**
  - Apple Pay CIT without `connector_customer_id` -> `InvalidArgument`
  - wallet Charge without `connector_customer_id` -> `InvalidArgument`
  - pre-decrypted Apple Pay SetupRecurring -> `Unimplemented`
  - pre-decrypted Google Pay Authorize -> `Unimplemented`
- **Apple Pay JSON token (CW8):** see Known limitations.

</details>

Earlier grpcurl evidence, not repeated here:
- [Authorize enrichment](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5609488006)
- [SetupMandate / card](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672056983)
- [RepeatPayment / card](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672450722)
- [SetupMandate / wallet](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5672973323)
- [RepeatPayment / wallet](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673276858)
- [This fix run](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673818577)
- [Follow-up: setup capture, partial-capture status, token mandates](https://github.com/juspay/hyperswitch-prism/pull/2279#issuecomment-5673935305)

## Files modified

- `crates/integrations/connector-integration/src/connectors/shift4.rs`
- `crates/integrations/connector-integration/src/connectors/shift4/transformers.rs`
- `crates/internal/integration-tests/src/connector_specs/shift4/specs.json`
- `crates/internal/field-probe/probe-config.toml`
- generated: `data/field_probe/shift4.json`, `docs-generated/`, `examples/shift4/`

### DIFF
--- a/crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
+++ b/crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
@@ -16,15 +16,19 @@ use domain_types::{
         Shift4ClientAuthenticationResponse as Shift4ClientAuthenticationResponseDomain,
     },
     merchant_authentication_flow_data::MerchantAuthenticationFlowData,
+    payment_address,
     payment_method_data::{
         BankRedirectData, PaymentMethodData, PaymentMethodDataTypes, RawCardNumber,
     },
-    router_data::{ConnectorSpecificConfig, FlowStatus},
+    router_data::{
+        AdditionalPaymentMethodConnectorResponse, ConnectorResponseData, ConnectorSpecificConfig,
+        FlowStatus,
+    },
     router_data_v2::RouterDataV2,
     router_response_types::RedirectForm,
 };
 use error_stack::ResultExt;
-use hyperswitch_masking::{ExposeOptionInterface, Secret};
+use hyperswitch_masking::{ExposeOptionInterface, PeekInterface, Secret};
 use serde::{Deserialize, Serialize};
 use url::Url;
 
@@ -59,10 +63,51 @@ pub struct Shift4ErrorResponse {
     pub error: ApiErrorResponse,
 }
 
+/// Class of the Shift4 error envelope (`error.type`).
+///
+/// Shift4 documents exactly four values. `Unknown` keeps an unrecognised value
+/// from failing the whole error-body parse — an unparsable error body would
+/// otherwise surface as a deserialization failure instead of the decline the
+/// merchant needs to see.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum Shift4ErrorType {
+    InvalidRequest,
+    CardError,
+    GatewayError,
+    RateLimitError,
+    #[serde(other)]
+    Unknown,
+}
+
+/// Shift4's declined/failed error envelope (`HTTP 402` for card declines, `4xx`
+/// otherwise): `{ "error": { type, code, message, issuerDeclineCode, adviceCode,
+/// networkAdviceCode, chargeId, ... } }`.
+///
+/// **The wire encoding is camelCase**, so `rename_all = "camelCase"` is load
+/// bearing: the reference Hyperswitch struct carries `#[serde(rename = "camelCase")]`,
+/// which is a no-op (`rename`, not `rename_all`) and silently deserializes every
+/// network field as `None`. Do not "simplify" this attribute back.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
 pub struct ApiErrorResponse {
+    #[serde(rename = "type")]
+    pub error_type: Option<Shift4ErrorType>,
     pub code: Option<String>,
     pub message: String,
+    /// Decline code supplied by the card issuer. Declined charges only.
+    /// Maps to `ErrorResponse::network_decline_code`.
+    pub issuer_decline_code: Option<String>,
+    /// Merchant advice code (MAC). Shift4 documents exactly two states:
+    /// absent/`null`, or `do_not_try_again` (a hard stop for smart retry).
+    /// Maps to `ErrorResponse::network_error_message`.
+    pub advice_code: Option<String>,
+    /// Raw advice code from the issuer / card network. Shift4 publishes no
+    /// value table for it — it is pass-through.
+    /// Maps to `ErrorResponse::network_advice_code`.
+    pub network_advice_code: Option<String>,
+    /// Charge the decline belongs to. Maps to `ErrorResponse::connector_transaction_id`.
+    pub charge_id: Option<String>,
 }
 
 // ===== CREATE CUSTOMER FLOW STRUCTURES =====
@@ -153,6 +198,26 @@ pub struct Shift4PaymentsRequest<T: PaymentMethodDataTypes> {
     /// calls. Shift4 requires BOTH `captured=false` AND `options.authorizationType=pre`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub options: Option<Shift4ChargeOptions>,
+    /// Transaction initiation class. Shift4 documents no default, so an untyped
+    /// charge is classified by the scheme rather than by us. A one-off sale is
+    /// `customer_initiated`; a charge that is also storing the card on file for
+    /// later merchant-initiated use is `first_recurring`.
+    #[serde(rename = "type", skip_serializing_if = "Option::is_none")]
+    pub transaction_type: Option<Shift4TransactionType>,
+    /// Charge-level billed-party details. Distinct from the flat `card.address*`
+    /// fields (which take precedence for AVS): this is what Shift4 echoes on the
+    /// charge object and shows in the dashboard. Note `billing.address.country`
+    /// is **two-letter** ISO while `card.addressCountry` is three-letter.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub billing: Option<Shift4Billing>,
+    /// Recipient and delivery address.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub shipping: Option<Shift4Shipping>,
+    /// Carries the merchant reference. `external.vendorReference` is the only
+    /// merchant-controlled reference identifier on `POST /charges` — Shift4 has
+    /// no `orderId` / `invoice` / `merchantReference` field.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub external: Option<Shift4External>,
     #[serde(flatten)]
     pub payment_method: Shift4PaymentMethod<T>,
 }
@@ -197,7 +262,58 @@ pub struct Shift4CardData<T: PaymentMethodDataTypes> {
     pub number: RawCardNumber<T>,
     pub exp_month: Secret<String>,
     pub exp_year: Secret<String>,
-    pub cardholder_name: Secret<String>,
+    /// Card security code. Shift4 only runs the CVV check when this is present —
+    /// omitting it forces `cvvCheck.result` to `not_provided` on every sale.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cvc: Option<Secret<String>>,
+    /// Optional for a normal charge; **required** for the zero-amount Account
+    /// Name Inquiry (ANI) path, which is enforced in the request builder.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cardholder_name: Option<Secret<String>>,
+    /// Flat card-level address. This is the highest-precedence AVS source in
+    /// Shift4's documented lookup order (card address -> charge `billing`
+    /// address -> customer address -> payment-method billing address).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_line1: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_line2: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_city: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_state: Option<Secret<String>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_zip: Option<Secret<String>>,
+    /// **Three-letter** ISO country code — deliberately different from
+    /// `billing.address.country`, which Shift4 documents as two-letter.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address_country: Option<common_enums::CountryAlpha3>,
+}
+
+impl<T: PaymentMethodDataTypes> Shift4CardData<T> {
+    /// Build the inline `card` object from UCS card data plus the billing
+    /// address that drives AVS. `cardholder_name` is passed in because its
+    /// requiredness is amount-dependent (see the ANI rules on `amount == 0`).
+    fn new(
+        card_data: &domain_types::payment_method_data::Card<T>,
+        cardholder_name: Option<Secret<String>>,
+        billing_address: Option<&payment_address::AddressDetails>,
+    ) -> Self {
+        Self {
+            number: card_data.card_number.clone(),
+            exp_month: card_data.card_exp_month.clone(),
+            exp_year: card_data.card_exp_year.clone(),
+            cvc: Some(card_data.card_cvc.clone()),
+            cardholder_name,
+            address_line1: billing_address.and_then(|addr| addr.line1.clone()),
+            address_line2: billing_address.and_then(|addr| addr.line2.clone()),
+            address_city: billing_address.and_then(|addr| addr.city.clone()),
+            address_state: billing_address.and_then(|addr| addr.state.clone()),
+            address_zip: billing_address.and_then(|addr| addr.zip.clone()),
+            address_country: billing_address
+                .and_then(|addr| addr.country)
+                .map(common_enums::CountryAlpha2::from_alpha2_to_alpha3),
+        }
+    }
 }
 
 // BankRedirect Payment Structures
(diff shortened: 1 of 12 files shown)


shuklatushar226:
### Targets `main`; merge #2279 first

This PR now targets `main` rather than `feat/shift4_grace`.

The Shift4 recurring work is written on top of the Authorize enrichment and Void commits under review in #2279, so those 4 commits are part of this branch and appear in this diff for now. **Once #2279 merges, this diff reduces to the recurring work on its own** — no rebase, and no conflict, because the content is identical.

Making the two PRs fully independent isn't possible without rewriting the feature: replaying the 13 recurring commits onto `main` without the enrichment conflicts on the first commit, and the recurring transformers use helpers the enrichment adds.

**Review scope for this PR:** the 13 commits from `dc46f86149` onward, which is everything except the enrichment and Void commits.

crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
```diff
@@ -1519,37 +2825,44 @@ impl<T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Seria
...
-                    })?;
+        // Require the caller to specify an amount. Shift4 accepts 0 for
+        // card-on-file verification, but we don't silently default to it —
+        // the caller must pass 0 explicitly if that's what they mean, so a
+        // missing amount is always a client error rather than an implicit
+        // zero-dollar auth.
+        let amount = item.request.minor_amount.ok_or_else(|| {
+            error_stack::report!(IntegrationError::MissingRequiredField {
+                field_name: "amount",
+                context: Default::default(),
```
JeevaRamu0104: nit: this error context is still `Default::default()`, unlike the missing-customer error a bit further down in the same function. Worth giving it the same treatment, e.g. "Shift4 requires an explicit amount; pass 0 for a verification."

crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
```diff
@@ -185,6 +261,301 @@ pub struct Shift4TokenPayment {
...
+            .filter(|_| {
+                payment_method.customer_id.is_some()
+                    && payment_method.status == Some(Shift4PaymentMethodStatus::Chargeable)
+            })
+            .map(shift4_mandate_reference);
+    }
+    response
+        .card
+        .as_ref()
+        .filter(|card| card.customer_id.is_some() || response.customer_id.is_some())
```
JeevaRamu0104: The payment-method branch above requires both `payment_method.customer_id` and `status == Chargeable`, but this card branch accepts ownership via `card.customer_id.is_some() || response.customer_id.is_some()` — an OR across two different objects, no status check. Is `response.customer_id` guaranteed to name the same owner as `card.customer_id`, or could a charge return a `card_` mandate to a customer that isn't actually the card's owner?

crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
shuklatushar226: Filled in `10bef0b4b`.

`transformers.rs:2851` now carries `additional_context` ("Shift4 requires an explicit amount; `0` is a distinct card-on-file verification instruction, so a missing amount is a client error rather than an implicit zero-dollar auth"), a `suggested_action`, and `doc_url` → `#create-a-new-charge`. I filled the `NotSupported` at `:2904` in the same builder while I was there. The remaining `Default::default()` sites in this file (auth-type and bank-redirect paths) blame to commits that are ancestors of `main`, so I left them outside this PR's scope.

One caveat worth flagging, since it makes the new context less useful than it looks: **over gRPC this error is currently shadowed.** `domain_types/src/types.rs:4966` validates `amount` during the `SetupRecurringRequest` → `SetupMandateRequestData` conversion and raises its own `MissingRequiredField { field_name: "amount" }` with a default context first; that conversion also sets `minor_amount: Some(..)` unconditionally once `amount` is present, so the Shift4 branch cannot fire from a gRPC call at all. A test leg confirmed it: the error came back with the generic `docs.hyperswitch.io/.../error-codes` URL, not the new one.

The plumbing itself is fine — I exercised the other filled context (iDEAL on SetupRecurring) and it surfaced verbatim in the gRPC error. So the Shift4-level context is documentation plus defence for non-gRPC callers. Making it actually reachable means filling the framework-level context in `domain_types`, which affects every connector; happy to raise that separately if you'd like it fixed.

crates/integrations/connector-integration/src/connectors/shift4/transformers.rs
shuklatushar226: Good catch — tightened in `10bef0b4b`. The card branch is now:

```rust
.filter(|card| card.customer_id.is_some())
```

The `|| response.customer_id.is_some()` fallback is gone, so a card mandate is returned only when Shift4 itself says a customer owns that card.

**On whether the two could disagree:** I checked every charge recorded across the sandbox runs on this branch. `card.customerId` and the charge-level `customerId` are always both present or both absent. Shift4 does return unowned cards (`"card":{"id":"card_...","customerId":null}`) — the one-off no-customer leg is exactly that — but never paired with a non-null charge-level `customerId`, so the OR never independently decided a case. It was still looser than the invariant the function needs, which is what your question exposed.

**No mismatch check added**, deliberately: the card path no longer reads `response.customer_id` at all, so whether the two agree is now irrelevant to what the mandate says. The MIT's `customerId` comes from the caller's `connector_customer` (`Shift4RepeatPaymentRequest`, same file), never from the mandate, and Shift4 rejects a stored-card charge whose `customerId` is not the card's owner. A local equality check would be dead code guarding a condition the processor already enforces.

**On the missing status check:** Shift4's card object has no `status` field — 24 attributes, per [#card-object](https://dev.shift4.com/docs/api#card-object). `status` (`chargeable` / `pending` / `failed` / `used`) exists only on the payment-method object, which is why the wallet branch checks it and this one cannot. The doc comment now states that with both links, so the asymmetry reads as deliberate rather than as an oversight.

**Regression legs** (sandbox, after the change): card SetupRecurring 0 and 1500, off-session card Authorize CIT → Charge, `tok_` off-session CIT, `tok_` SetupRecurring, Apple Pay SetupRecurring → Charge — all still return their mandate and charge successfully. The one-off no-customer leg returns `card.customerId = null` and no mandate, which is the only case where a mandate is now withheld.

JeevaRamu0104 (approved):

deepanshu-iiitu (approved):

deepanshu-iiitu (approved):

PR #2313, opened 2026-09-16
fix(grpc-server): enable http2 keepalive so half-open client connections are dropped, not held

gopikrishna000:
Closes juspay/hyperswitch-cloud#23600

## Problem

Hyperswitch router connections to UCS are never pinged by the server. A connection that dies without a GOAWAY (pod rollout racing endpoint removal, or connection state expiring on the node path) stays half-open on both ends until the next RPC writes into it and fails with `transport error`. Production shows 7 such failures in 7 days on the router side, each the first call on a channel that had been idle for hours; evidence is in the router PR juspay/hyperswitch#14273.

## Change

`Server::builder()` in `grpc-server/src/app.rs` now enables HTTP/2 keepalive:

- `http2_keepalive_interval(30s)`: the server PINGs idle client connections, so a half-open connection is dropped here rather than held open until a client writes a payment into it.
- `http2_keepalive_timeout(20s)`: matches the gRPC-Go default. A tighter value would drop healthy connections during ordinary load spikes, which is failure amplification rather than protection.

Constants live in `common_utils::consts`. Both are provided tonic builder options; nothing is hand-rolled.

## Deliberately not included

`max_connection_age`. tonic retires connections on a plain timer with no jitter (`server/mod.rs`, `sleep_or_pending` then `graceful_shutdown`), unlike gRPC-Go which jitters. Every client connection is born within seconds of the others after a rollout, so all of them would GOAWAY on the same boundary, repeatedly. It would also add roughly 288 GOAWAY events per connection per day where there are none today, each one a chance to race an in-flight request, to buy pod-level load spreading that is not a problem at the current replica counts. Keepalive traffic already prevents a connection from being idle on the wire, which is what the setting was originally meant to address.

## Follow-up outside this repo

Helm chart: a `preStop` sleep of about 5s on the UCS deployment so endpoint removal completes before SIGTERM, and the GOAWAY from graceful shutdown is sent on a network path the router can still receive it on.

## Verification

Not built locally by policy; CI is the build. Graceful GOAWAY does not fail in-flight or subsequent requests.


XyneSpaces:
[should-fix] `.max_connection_age(Duration::from_secs(consts::GRPC_MAX_CONNECTION_AGE_SECS))` at `crates/grpc-server/grpc-server/src/app.rs:351` retires connections via GOAWAY without setting a grace period.

In-flight streams may be closed before they finish because the default `max_connection_age_grace` is zero. Add `.max_connection_age_grace(Duration::from_secs(...))` and a matching `GRPC_MAX_CONNECTION_AGE_GRACE_SECS` constant so open RPCs can complete before the connection shuts down.

XyneSpaces:
[should-fix] `crates/grpc-server/grpc-server/src/app.rs` sets `.max_connection_age(Duration::from_secs(consts::GRPC_MAX_CONNECTION_AGE_SECS))` without configuring `.max_connection_age_grace(...)`. The PR description states that in-flight streams should complete before the connection is retired, but the grace window that enforces that drain is missing; active streams may be force-closed at the 5-minute boundary.

Add an explicit `.max_connection_age_grace(...)` value large enough for the longest expected in-flight RPC, backed by a named constant and documented in the comment.

JeevaRamu0104 (dismissed):

AmitsinghTanwar007 (approved):

JeevaRamu0104 (approved):

PR #2316, opened 2026-09-17
fix(connector): [Adyen] substitute endpoint prefix when test_mode is absent

shuklatushar226:
## What was broken

- **Connector**: adyen
- **Payment method**: card (any; affects URL construction, not payment method handling)
- **Flow**: SetupMandate (`PaymentService/SetupRecurring`). Any Adyen flow that receives a request without `test_mode` is exposed.
- **Fixes**: N/A

## Description

`build_env_specific_endpoint` treated a missing `test_mode` as test mode (`test_mode.unwrap_or(true)`) and returned `base_url` unchanged. In production, Adyen's `base_url` is the live template `https://{{merchant_endpoint_prefix}}-checkout-live.adyenpayments.com/checkout/`, so the raw placeholder was dispatched and the call failed before leaving UCS:

```
Failed to send request to connector builder error for url
(https://{{merchant_endpoint_prefix}}-checkout-live.adyenpayments.com/checkout/v68/payments): Parsed Url is not a valid Uri
```

Hyperswitch router builds before juspay/hyperswitch#14049 do not set `test_mode` on `PaymentServiceSetupRecurringRequest` (Authorize/Execute do set it). As a result, every Adyen SetupMandate shadow call from those routers fails with gRPC `Internal`, while the same flows via the direct path succeed.

**Change:** when `test_mode` is absent, infer it from the configured URL. Only live configs carry the `{{merchant_endpoint_prefix}}` template (`sandbox.toml` / `development.toml` point at `checkout-test.adyen.com`), which is the same rule `netcetera.rs` already uses. Explicit `Some(true)` / `Some(false)` behave exactly as before.

| `test_mode` | base_url | Before | After |
|---|---|---|---|
| `Some(false)` | live template | prefix substituted | unchanged |
| `Some(true)` | any | raw base_url | unchanged |
| `None` | test URL (no template) | raw base_url | unchanged |
| `None` | live template | **raw template → invalid URI** | **prefix substituted** |

## Checklist

- [x] Regression
- [ ] Integrity

No TOML, feature matrix, URL whitelist, GSM, wasm or dashboard changes needed.

## How did you test it?

Direct `grpcurl` against a locally built `grpc-server`, comparing `origin/main` with this branch on identical inputs:
- `types.PaymentService/SetupRecurring` with an Adyen card.
- `x-connector-config` overrides `base_url` with the template in the host (`http://127.0.0.{{merchant_endpoint_prefix}}:9911/checkout/`) and sets `endpoint_prefix: "1"`, so all traffic went to a local mock Adyen server.

| Case | `origin/main` | This PR |
|---|---|---|
| `test_mode` absent, templated URL | `Internal: … Parsed Url is not a valid Uri` (same error as production) | `CHARGED`; mock received `POST /checkout/v68/payments` |
| `test_mode: false`, templated URL | `CHARGED` | `CHARGED` |
| `test_mode: true`, templated URL | invalid URI (template kept) | invalid URI (template kept), unchanged |
| `test_mode` absent, non-templated URL, no `endpoint_prefix` | `CHARGED` | `CHARGED` |


gopikrishna000 (approved):

shuklatushar226:
Closing: the root cause is on the router side. Hyperswitch builds before juspay/hyperswitch#14049 don't forward `test_mode` on `PaymentServiceSetupRecurringRequest` / `PaymentServiceIncrementalAuthorizationRequest`, so UCS receives `None`. That change is in hyperswitch release `2026.09.09.0` and later. Once the router is deployed with it, UCS gets `test_mode: false` and Adyen's existing `build_env_specific_endpoint` substitutes the endpoint prefix, matching the direct connector path. Keeping prism's behaviour identical to hyperswitch.

PR #2320, opened 2026-09-17
ci: require the deja-art label on every pull request

iemyashasvi:
## What

Adds a convention check, **Verify PR carries the deja-art label**, to `pr-convention-checks.yml`. A pull request without the `deja-art` label fails the check with a message saying what the label is for and how to re-run.

## Why

The `deja-art` label is what starts a Deja ART replay of the pull request against recorded sandbox traffic. Making the label mandatory means every change gets replayed. The replay's verdict is reported on the pull request but does **not** gate the merge; only the presence of the label does.

## How

- Same workflow and trigger as the conventional-commit title check (`pull_request_target`), so the check definition comes from `main` and a pull request cannot modify it away. The job reads only the event payload and checks nothing out.
- A pull request labelled after the check ran re-runs it on the next push or title edit.
- Automated pull requests by `github-actions[bot]` and `hyperswitch-bot[bot]` (release and hotfix flows) are exempt, since they cannot add a label.
- In the merge queue the job passes, because the pull request had to pass it to get there.

## Rollout

1. Merge this. The check appears on new pull requests but blocks nothing yet.
2. Add **Verify PR carries the deja-art label** to the required status checks of the Default Branch ruleset (repo admin). From then on a pull request without the label cannot merge.

## Checklist

- [x] Workflow YAML validated
- [x] Shell logic dry-run for labelled, unlabelled and bot-authored cases


.github/workflows/pr-convention-checks.yml
```diff
@@ -1,5 +1,11 @@
 name: Pull Request Convention Checks
 
+# Neither job needs the token for anything: both read the event payload and
+# check nothing out. `pull_request_target` would otherwise hand them a
+# write-capable token, which is the whole hazard of that trigger.
```
pixincreate: we can consider removing these jargon comments.

.github/workflows/pr-convention-checks.yml
```diff
@@ -55,3 +61,45 @@ jobs:
...
+      # The `deja-art` label is what starts a Deja ART replay of this pull
+      # request against recorded sandbox traffic. Every pull request must
+      # carry it, so every change gets replayed. The replay's VERDICT does not
+      # gate the merge; the label does. A pull request labelled after this
+      # check ran re-runs it on the next push or title edit.
+      #
+      # Automated pull requests (release and hotfix flows) are exempt: their
+      # authors cannot add a label, and a replay of a version bump adds nothing.
+      - name: Verify PR carries the deja-art label
+        if: ${{ github.event_name == 'pull_request_target' }}
```
pixincreate: don't.

use pull_request instead.

.github/workflows/pr-convention-checks.yml
pixincreate: replace this with `pull_request`

.github/workflows/pr-convention-checks.yml
```diff
@@ -55,3 +61,45 @@ jobs:
...
+        shell: bash
+        env:
+          LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
+          AUTHOR: ${{ github.event.pull_request.user.login }}
+        run: |
+          case "$AUTHOR" in
+            github-actions\[bot\]|hyperswitch-bot\[bot\])
+              echo "Automated pull request by $AUTHOR; the deja-art label is not required."
+              exit 0
+              ;;
```
pixincreate: this is not needed. gh actions will directly write to main. hs bot commits to the pr itself.

.github/workflows/pr-convention-checks.yml
```diff
@@ -55,3 +61,45 @@ jobs:
...
+          esac
+          if [[ ",$LABELS," == *",deja-art,"* ]]; then
+            echo "Label deja-art present; a Deja ART replay will run for this pull request."
+            exit 0
+          fi
+          echo "::error::This pull request must carry the 'deja-art' label."
+          echo "The label starts a Deja ART replay of this change against recorded sandbox traffic."
+          echo "Add the label, then push or edit the title to re-run this check."
```
pixincreate: why edit title?

.github/workflows/pr-convention-checks.yml
iemyashasvi: Removed in 9e7e428a8.

.github/workflows/pr-convention-checks.yml
iemyashasvi: Switched the workflow to `pull_request` in 9e7e428a8, and updated both jobs' event guards to match.

.github/workflows/pr-convention-checks.yml
iemyashasvi: Done, `pull_request` now (9e7e428a8).

.github/workflows/pr-convention-checks.yml
iemyashasvi: Removed the exemption in 9e7e428a8.

.github/workflows/pr-convention-checks.yml
iemyashasvi: No longer needed. The check now reads the labels from the API instead of the event payload, so adding the label and re-running the check is enough (9e7e428a8).

pixincreate (approved):

gopikrishna000 (approved):

PR #2325, opened 2026-09-18
chore(grace): derive the 3DS leg count from connector docs, and gate it

shuklatushar226:
## Why

A GRACE run implemented 3DS for Nuvei and produced a **latent double charge**. That is not a Nuvei mistake — it is the workflow's output.

Nuvei REST 1.0 has two 3DS endpoints, not three: `/initPayment.do`, then `/payment.do` **with** a `threeD` block, then `/payment.do` **without** it. GRACE's only authoritative statement about 3DS shape said *"ThreeDS = all three markers, one pass"*. Given three documented calls and a mandate for three markers, the mapping is forced: the settling call lands in `PostAuthenticate` — the one composite-loop arm with **no break** — and `next_authentication_step` then routes to `Authorize`, which charges again. Two full-amount charges, the second with no CAVV/ECI, so no liability shift.

**Three legs is the rare exception, not the rule:**

| | zero legs | one | two | three |
|---|---|---|---|---|
| Hyperswitch (~154 connectors) | ~62 | 6 | 4 | **2** — cybersource + barclaycard, a fork of it |
| UCS (20 with any marker) | 3 | 8 | 6 | **1** — getnet |

HS's three predicates all default to `false`; only 12 of ~154 connectors opt in. GRACE's map was the exact inverse.

## The invariant

It is **not** "authentication legs must not charge". From `composite-service/src/payments.rs` and `utils.rs:217-237`:

| arm | breaks when | may it charge? |
|---|---|---|
| PreAuthenticate | `redirection_data.is_some() \|\| is_failure_status` | only if it always redirects or fails |
| Authenticate | `redirection_data.is_some() \|\| is_terminal_status` (incl. `Charged`/`Authorized`) | **yes** — `redsys` relies on this |
| **PostAuthenticate** | **never** | **no, ever** |
| Authorize | unconditional | yes — this is the charge |

**`PostAuthenticate` must never charge, because it is the only step with no exit.**

Hyperswitch corroborates the target shape: it runs the same Nuvei connector without double-charging because it labels the settling `/payment.do` **`CompleteAuthorize`**. **UCS's `PostAuthenticate` is not HS's `CompleteAuthorize`; HS's `CompleteAuthorize` is UCS's `Authorize`.**

## What changes

- **`.gracerules_add_flow`** — the FLOW-GROUP MAP's fixed marker set becomes a **LEG-COUNT PROCEDURE** (find the charging call → it is always `Authorize` → classify what precedes it), with the invariant and a worked counter-example. The `Pre→Auth→Post` dependency chain is removed from **both** copies (markdown + JSON): it made the legitimate `Pre`+`Post` shape that nexixpay, moneris and worldpay actually ship formally unresolvable.
- **`pattern_authentication_dispatch.md` is added to the FLOW-MARKER → PATTERN MAP.** It already said *"PostAuthenticate has no exit"* and *"prove termination"* — and appeared **zero times** in the authoritative map, so agents following it never read the one file with the answer. Highest-value edit here.
- **`2.3a_plan.md`** — new §4a records the call map in `plan.json .three_ds`, enforced by 10 jq rules. A `charges: true` call mapped to `PostAuthenticate` now **fails the plan**, where the fix is cheap. Adds INV-23…INV-26.
- **`grace/rulesbook/codegen/tools/threeds_gate.py`** (new, stdlib-only) — TDS-01 PostAuthenticate charges · TDS-02 declared marker unreachable from the dispatch hook · TDS-03 non-termination. Wired into `2.3b` per-unit + finalize gates and a new `2.7` review lens at **S0**, with `stage_hint: PLANNER`.
- **`2.3_codegen.md:331`** no longer fails a correct two-leg implementation for not exercising a third marker — that gate made inventing a leg the cheapest escape.
- **`feedback.md` UCS-003 superseded.** Its auto-fix rule fired on every connector and pushed toward *more* legs; worldpay, the connector it was derived from, has dead legs today.
- **`1_orchestrator.md`** keeps only the alias row and points at the canonical procedure, so there is no second copy of the procedure to keep in sync. A de-duplication grep in the finalize gate fails if one reappears.

## Verification

| check | result |
|---|---|
| Gate fails Nuvei's 3DS implementation | ✅ TDS-01 on both signals (`get_from_capture_method`; `PostAuthenticate get_url == Authorize get_url`) |
| Gate passes cybersource (the genuine 3-leg shape) | ✅ exit 0 |
| Parser robustness across all 21 connectors | ✅ **0 unparsed** |
| Plan validator rejects the Nuvei mapping / accepts the corrected one | ✅ / ✅ |
| Techspec `Charges:` check flags the current Nuvei spec | ✅ 16 steps, 0 annotated |
| Anti-drift grep | ✅ green, and catches a re-introduced mandate |

Signal discrimination was checked against **all 9** connectors with a `PostAuthenticate` builder: only Nuvei fires. An amount-only check would have misfired on barclaycard and cybersource, which legitimately send an amount on a non-charging `authentication-results` call — so amount is a warning, never a failure.

## Audit — live defects found, reported not fixed

Found while deriving the taxonomy. **No connector code is changed in this PR.**

- **`moneris.rs:107-135` — infinite `PostAuthenticate` loop.** `(RedirectWithParams | RedirectWithoutParams, _)` swallows every `completed_step`, so the `(_, Some(PostAuthenticate)) => Authorize` arm below it can never fire. TDS-03 reproduces it from both redirect states.
- **Markers implemented but unreachable on the composite path** — `worldpay`, `saferpay`, `nexixpay`, `globalpay`, `ilixium`, `nmi`, `paydotcom` never override `next_authentication_step`, so the trait default returns `Authorize` and those legs are dead code. TDS-02 reproduces it. (`globalpay` is a special case: its `PostAuthenticate` is a *post*-Authorize confirmation.)
- **Nuvei's `Authenticate` leg is also charge-shaped** — safe only because that arm breaks on a terminal status and the challenge path always returns `redirection_data`. Worth knowing if it is re-touched.

## Deferred

A declarative "this leg charges" bit on the three auth request types would be the real architectural fix — they all carry `amount: MinorUnit` (non-`Option`) and `capture_method`, which is what invites the bug. It is a Rust change across ~33 connectors; HS lives without it and pays with hardcoded `Connector::Redsys` match arms.


shuklatushar226:
Superseded by #2326, which now targets `main` and contains this commit (`28dcf4f0`) unchanged, plus the test-oracle replacement that builds on it. Collapsing to one PR.

PR #2328, opened 2026-09-18
fix(stripe): stop dropping billing, shipping, browser info and L2/L3 on wallet tokens

shuklatushar226:
## Description

UCS built a materially different Stripe `POST /v1/payment_intents` body from hyperswitch for the same payment. Three independent causes, all on the Apple Pay / Google Pay token path:

**1. Tokenized payments lost their shipping and browser blocks.**

`shipping[*]` and `payment_method_data[ip]` / `payment_method_data[user_agent]` were gated on `payment_method_token.is_some()`, which is true for *every* tokenized payment. Hyperswitch gates the same blocks on `is_payment_method_tokenize_flow_required`, which additionally requires a **Stripe split payment**. So on an ordinary (non-split) Apple Pay / Google Pay token payment, hyperswitch sends shipping and browser info and UCS sent neither.

The gate is now `payment_method_token.is_some() && <stripe split payment>`, which leaves split-payment behaviour untouched and restores the blocks everywhere else.

**2. Stripe Level 2 / Level 3 purchase data was never sent.**

`l2_l3_data` arrives on `PaymentFlowData` but the Stripe connector never read it, so `payment_details[customer_reference]`, `payment_details[order_reference]` and `amount_details[shipping][to_postal_code]` were missing.

Ported hyperswitch's `StripeLineItemsData` / `create_stripe_line_items_data` in full — including `amount_details[line_items][*]` and the rule that Stripe takes tax at the order level *or* the line-item level but never both — rather than only the three fields this particular merchant's traffic happened to surface. Applied to both `PaymentIntentRequest` builders (Authorize and RepeatPayment), since hyperswitch sends it on the same intent for mandate payments too.

**3. Apple Pay lost its entire billing block, and both wallets sent the wrong `payment_method_types[0]`.**

An earlier revision of this PR left billing alone, on the reading that the binding is shadowed by the payment-method branch below it and that no billing field appeared in the report. Shadow traffic since then shows that was wrong: the shadowing branch is exactly the problem.

`billing_address` kept the old `payment_method_token.is_some()` gate, and the wallet-token branch then discarded it outright with `StripeBillingAddress::default()`. Hyperswitch reaches its equivalent branch only for a Stripe split payment; for every ordinary merchant it goes through `create_stripe_payment_method`, whose wallet arm returns the real `payment_request_details.billing_address`. So UCS dropped all nine `payment_method_data[billing_details][*]` fields on every Apple Pay payment — losing the billing postal code Stripe uses for AVS, on top of the IP and user-agent from cause 1 that Radar uses.

The same branch also hardcoded `payment_method_types[0]` to `card`. Stripe reads the wallet off the token, so hyperswitch's `get_stripe_payment_method_type_from_wallet_data` sends **nothing** for Apple Pay and `card` for Google Pay. The wallet kind is now carried alongside the token and mirrored, instead of sending `card` for both.

## Evidence

`merchant_1782222808739` (prod-use, Stripe, shadow), last 24h — 66/66 Apple Pay request comparisons show exactly these fields and cards show none of the wallet-specific ones:

| field group | direction | Apple Pay | cards |
|---|---|---|---|
| `payment_method_data[billing_details][*]` (8) | hyperswitch only | ✗ | — |
| `shipping_address[*]` (6) + `shipping[name]` | hyperswitch only | ✗ | — |
| `payment_method_data[ip]`, `[user_agent]` | hyperswitch only | ✗ | — |
| `payment_method_types[0]` | **UCS only** | ✗ | — |
| `ip`, `user_agent`, `amount_details[shipping][to_postal_code]`, `payment_details[customer_reference]`, `payment_details[order_reference]` | hyperswitch only | ✗ | ✗ |

Google Pay has no traffic on this merchant, but it shares the `ApplePay | GooglePay` match arm and so shares every defect above.

Note the response-side 500s on this merchant (`UCS error: Internal - h2 protocol error`) are a separate, unrelated problem — a gRPC error trailer exceeding the 16 KiB h2 frame cap on Stripe declines. Nothing in this PR addresses it.

## Not addressed here

UCS sending top-level `payment_method` where hyperswitch sends `payment_method_data[card][token]` is a router-side input rather than a connector bug. This crate already branches on `token_payment_method_type` (`stripe/transformers.rs`), and hyperswitch populates it for Apple Pay / Google Pay tokens as of juspay/hyperswitch#13827; the diff clears once that router build is live.

The `RepeatPayment` builder gates its shipping block the same way Authorize did. I have left it as-is: the report covers `authorize` only, and I would rather confirm the MIT shape against shadow traffic before changing it.

## Also in this PR: enable `return_raw_and_typed_connector_data`

Flipped `false` -> `true` in `config/development.toml`, `config/sandbox.toml` and `config/production.toml`, so raw and typed connector request/response payloads are included in client responses. The flag and its gating come from #2333, which this branch picks up by merging `main`.

Reviewers should note the warning carried in the config next to the flag:

```toml
# WARNING: Enabling this can produce large gRPC error trailers. Client applications
# and intermediate proxies must support a sufficiently large HTTP/2 maximum
# header-list size.
```

That is not hypothetical on Stripe. `merchant_1782222808739` already returns `UCS error: Internal - h2 protocol error: http2 error` on 100% of Apple Pay declines (267 occurrences in 7 days) from a gRPC error trailer exceeding the 16 KiB h2 cap — with the flag still off. Worth confirming the h2 max-header-list-size on the UCS server and any intervening proxy before this reaches production. Sandbox and development carry no such risk.

### Additional Changes

## How did you test it?

No automated tests in this PR.

Verified against prod shadow traffic instead — `merchant_1782222808739` (prod-use, Stripe), 24h of validation-service `RP_05` connector-request comparisons. 66/66 Apple Pay authorize comparisons show exactly the field set in the table above, and card comparisons show none of the wallet-specific ones, which is what pins the defect to the wallet-token branch rather than to the shared address handling.

- `cargo clippy -p connector-integration --all-targets` — clean
- `cargo fmt -p connector-integration`

Closes juspay/hyperswitch-cloud#23614


JeevaRamu0104 (dismissed):

AmitsinghTanwar007 (dismissed):

XyneSpaces:
[blocking] This PR changes Stripe request building for tokenized payments and adds L2/L3 data, but it adds no automated tests. Regressions in amount calculations or tokenized-payment gates are likely.

Add snapshot/form-encoded tests covering non-split tokenized payments (shipping/browser present), split-tokenized payments (shipping/browser omitted), L2/L3 with/without tax, and the RepeatPayment path.

[should-fix] L2/L3 amount semantics need verification:
- `amount_details[line_items][*][unit_cost]` at `transformers.rs:369` assumes `order_item.amount` is per-unit; if it is the line total, Stripe receives an inflated value.
- `discount_amount` at line 397 uses `unit_discount_amount` without multiplying by quantity, potentially understating the total discount.
- Order-level tax is suppressed whenever any line item has `total_tax_amount = Some(_)`, including `Some(0)`.

Confirm the semantics against Stripe docs and add tests that fail if any of these assumptions change.

XyneSpaces:
[should-fix] The new L2/L3 amount fields (`shipping_amount`, `order_tax_amount`, `discount_amount`, line-item `unit_cost`, tax, and discounts) wired into the Stripe intent request are not covered by unit tests. Add serialization tests for the line-item expansion and for the order-tax vs line-item-tax exclusivity rule so amount/currency semantics cannot drift.

```suggestion
// Add a test in crates/integrations/connector-integration/src/connectors/stripe/transformers.rs
// that constructs L2L3Data and asserts the serialized StripeLineItemsData shape.
```

XyneSpaces:
[should-fix] `crates/integrations/connector-integration/src/connectors/stripe/transformers.rs:2264` — `is_tokenize_flow` is now `payment_method_token.is_some() && <Stripe split payment>`. Upstream `is_payment_method_tokenize_flow_required` additionally restricts to card/NTID tokenize flows, so a non-card split token payment will drop shipping/browser here while upstream would keep it.

Align the gate with the upstream tokenize-flow predicate to avoid losing shipping/browser data for valid tokenized flows.

XyneSpaces:
[should-fix] `StripeLineItemsData.line_items` uses `HashMap<String, String>` with `#[serde(flatten)]`. `HashMap` iterates in a nondeterministic order, so the URL-encoded Stripe request body can vary between requests. This makes tests flaky and breaks any downstream body signature or cache-key computation that expects a stable encoding. Use `BTreeMap<String, String>` or an ordered map instead.

```suggestion
pub line_items: Option<BTreeMap<String, String>>,
```

XyneSpaces:
[should-fix]

`crates/integrations/connector-integration/src/connectors/stripe/transformers.rs` — New L2/L3 purchase-data serialization and the split-payment tokenize-flow address handling are not covered by tests in this diff.

Add unit or integration tests for:
- `create_stripe_line_items_data` with order-level tax vs line-item-level tax exclusion.
- The `is_tokenize_flow` branch suppressing `shipping_address` and `browser_info` for Stripe split payments.

This prevents silent regression in the tokenized-payment path and in L2/L3 field encoding.

AmitsinghTanwar007 (dismissed):

gopikrishna000 (dismissed):

gopikrishna000 (approved):

AmitsinghTanwar007 (approved):

PR #2329, opened 2026-09-18
fix(proto): make deprecated fields optional

pixincreate:
## Description
Five deprecated fields in `payment.proto` were still required. This marks them optional:

- `MandateAmountData.amount` and `MandateAmountData.currency`
- `PaymentServiceCaptureResponse.mandate_reference`
- `PaymentServiceSetupRecurringResponse.mandate_reference`
- `PaymentMethodServiceEligibilityResponse.eligibility`

Consumers can omit these fields without the payment flow breaking. Rust call sites that assumed the fields are always present are updated.

## Motivation and Context
A consumer removed these fields after they were marked deprecated. Since the fields were required, the removal broke their flow. Keeping deprecated fields optional avoids a hard break for HTTP consumers until the fields are removed in a major release.

Hyperswitch dependency update for this change: https://github.com/juspay/hyperswitch/pull/14312

### Additional Changes

## How did you test it?
`cargo check --workspace --all-targets` passes on this branch. The Hyperswitch side compiles against this change with the `unified-connector-service-client` dependency pointed at this branch.


Shubhodip900 (approved):

XyneSpaces:
[should-fix] `MandateAmountData` conversion at `crates/types-traits/domain_types/src/types.rs:13151` uses `.or(amount_data.amount).unwrap_or_default()`, so when both the new optional `amount_money` and the now-optional deprecated `amount` are absent the mandate amount silently becomes 0. A missing amount should fail explicitly rather than default to a monetary value.

Return an error or log a metric when neither amount field is present.

XyneSpaces:
[should-fix] `crates/types-traits/domain_types/src/types.rs` falls back to `unwrap_or_default()` when both `amount_money` and the deprecated `amount` are missing, which can produce a zero-amount mandate instead of failing validation. Return an explicit validation error or require `amount_money` when the deprecated field is absent.

```suggestion
    amount_money
        .map(|amount_money| amount_money.into())
        .or_else(|| amount.map(|amount| amount.into()))
        .ok_or_else(|| /* validation error: mandate amount is required */)?
```

XyneSpaces:
[blocking] `crates/types-traits/domain_types/src/types.rs:13151` silently defaults the mandate amount to `0` when both `amount_money` and the deprecated `amount` are absent.

The conversion computes `amount_data.amount_money.map(|amount_money| amount_money.minor_amount).or(amount_data.amount).unwrap_or_default()`. A missing amount on a mandate setup authorizes or sets up a zero-value mandate, which is a payment-correctness risk.

```suggestion
let amount = amount_data
    .amount_money
    .map(|amount_money| amount_money.minor_amount)
    .or(amount_data.amount)
    .ok_or(IntegrationError::MissingRequiredField {
        field_name: "amount".to_string(),
    })?;
```

JeevaRamu0104 (approved):

XyneSpaces:
[should-fix] `crates/types-traits/domain_types/src/types.rs:13144` — `MandateAmountData` conversion silently defaults `amount` to `0` when both `amount_money` and the deprecated `amount` are absent. A missing amount should fail fast rather than create a zero-value mandate.

Return an explicit error when no amount source is present.

[should-fix] `sdk/javascript/src/payments/_generated_grpc_client.ts`, `sdk/rust/src/_generated_grpc_client.rs`, `sdk/python/src/payments/_generated_*.py`, `sdk/java/src/main/kotlin/GeneratedFlows.kt`, `docs-generated/api-reference/domain-schema/README.md` — generated SDK/proto-doc files still encode the old required-field shapes for `MandateAmountData` / `PaymentMethodServiceEligibilityResponse` but are not regenerated in this PR. The public SDK surface will drift from the proto change.

Regenerate the SDK clients and docs from the updated proto before merging.

XyneSpaces:
:warning: `crates/types-traits/domain_types/src/types.rs` silently defaults the mandate amount to `0` when both `amount_money` and the deprecated `amount` field are absent:

```rust
amount: Some(
    amount_data
        .amount_money
        .map(|amount_money| amount_money.minor_amount)
        .or(amount_data.amount)
        .unwrap_or_default(),
),
```

With the deprecated field now `optional`, a missing value flows into `unwrap_or_default()` and produces a `0` amount instead of failing. This weakens payment-amount correctness. Return an explicit error when neither amount source is present.

XyneSpaces:
[should-fix]

`crates/types-traits/domain_types/src/types.rs:13150` — Defaulting missing mandate amount to 0 silently masks missing-data errors.

When both `amount_money` and the deprecated `amount` are absent, the new `.or(amount_data.amount).unwrap_or_default()` chain falls back to 0. For mandate amount conversion, propagate the missing value as an error or require `amount_money` to be present.

```suggestion
                    amount_data
                        .amount_money
                        .map(|amount_money| amount_money.minor_amount)
                        .or(amount_data.amount)
                        .ok_or_else(|| ConnectorError::MissingRequiredField { field_name: "amount".to_owned() })?
```

(Icon mapping: :warning: should-fix)

Aishwariyaa-Anand (approved):

PR #2333, opened 2026-09-21
fix: gate raw and typed connector data

AmitsinghTanwar007:
## Summary
- rename `return_raw_connector_data` to `return_raw_and_typed_connector_data`
- gate raw and typed connector request/response fields consistently
- default the option to `false` and document the HTTP/2 trailer-size implications
- add configuration deserialization tests

## Context
For request `01a0b355-555a-7dd0-a870-d1c52fe4d577`, UCS completed Stripe Authorize with a structured `402 card_declined`, while HS observed a generic H2 protocol error receiving the gRPC result. Large `grpc-status-details-bin` trailers are the leading hypothesis because connector errors currently carry raw and typed payloads, but production logs do not explicitly report a header-list overflow. This change bounds client-facing diagnostic data by default without claiming that the incident root cause is conclusively proven.

## Breaking configuration change
Replace:
```text
CS__COMMON__RETURN_RAW_CONNECTOR_DATA
```
with:
```text
CS__COMMON__RETURN_RAW_AND_TYPED_CONNECTOR_DATA
```
Deployment configuration will be updated in coordination with this release.

## Validation
- `cargo test -p ucs_env configs::tests`
- `cargo check -p external-services -p grpc-server`
- `cargo fmt --all -- --check`
- `git diff --check`


docs/superpowers/specs/2026-09-21-raw-typed-connector-data-flag-design.md
JeevaRamu0104, line 1: why we need this file

JeevaRamu0104 (approved):

shuklatushar226 (approved):

XyneSpaces:
[should-fix] `crates/common/external-services/src/service.rs` clears `typed_connector_response` when `return_connector_data` is false, but never assigns it from `resource_common_data.get_typed_connector_response()` when the flag is true (around lines 621–631). As a result, `typed_connector_response` is always `None` on connector errors regardless of the flag. Add the missing assignment inside the `if return_connector_data` block so it behaves consistently with the other raw/typed fields.

PR #2334, opened 2026-09-21
feat(reddot): add Red Dot Payment connector

MansiKhatri-37:
## Description

Adds a new connector integration for Red Dot Payment (RDP) — a Southeast Asia payment gateway — supporting card payments in SOP (merchant-collected card) redirect mode.
Flows implemented (per connector_specs/reddot/specs.json):
- Authorize — RDP Redirect API, POST /service/payment-api (card authorize, auto-capture sale, 3DS redirect support)
- PSync (PaymentService/Get) — Redirection enquiry, POST /service/Merchant_processor/query_redirection
- Refund — RDP Merchant API, POST /instanpanel/api/payment (form-urlencoded)
- RSync (RefundService/Get) — Merchant enquiry, POST /instanpanel/api/enquiry
- Incoming webhooks — parse-only handling of notify_url callbacks. RDP callbacks are unverifiable (no documented signature scheme), so they're treated as unverified triggers requiring a follow-up PSync for authoritative status (mirrors the maya/glomopay pattern).

## Motivation and Context

New connector implementation.

## How did you test it?
https://docs.google.com/spreadsheets/d/1gIqevjC16CG41dyB99SXeQ-S1MBTLw_d2XEyLpjAVB0/edit?gid=0#gid=0


XyneSpaces:
[should-fix] A few issues in the Reddot connector PR:

1. `config/development.toml` carries local-environment overrides (`return_raw_connector_data = true`, `port = 8011`, `type = "http"`, `metrics.port = 8090`) that do not belong in committed defaults. Please revert these to the repo-standard development values; the Reddot base URLs can stay.

2. `crates/types-traits/grpc-api-types/proto/payment.proto` adds the `REDDOT` enum value and `ReddotConfig` message, but the diff does not include the generated Rust under `crates/types-traits/grpc-api-types/src/`. If those generated files are checked in, regenerate them so the proto and Rust stay in sync.

3. The webhook event-type mapping in the new Reddot connector falls through to `EventType::PaymentIntentFailure` for any unknown code (`_ => EventType::PaymentIntentFailure`). That will misclassify future or refund-related events as payment failures. Introduce an `Unknown`/`Unsupported` variant, or ignore unrecognized codes instead of assuming failure.

config/development.toml
```diff
@@ -18,12 +18,12 @@ buffer_limit = 100000
...
 host = "0.0.0.0"
-port = 8000
-type = "grpc"
+port = 8011
+type = "http"
 
 [metrics]
 host = "0.0.0.0"
-port = 8080
+port = 8090
```
Aishwariyaa-Anand: revert this

config/superposition.toml
```diff
@@ -34,7 +34,7 @@ server_type = { value = "grpc", schema = { type = "string" } }
 
 # Metrics configuration
 metrics_host = { value = "0.0.0.0", schema = { type = "string" } }
-metrics_port = { value = 8080, schema = { type = "integer" } }
+metrics_port = { value = 8090, schema = { type = "integer" } }
```
Aishwariyaa-Anand: revert

crates/grpc-server/grpc-server/src/main.rs
```diff
@@ -5,8 +5,17 @@ use grpc_server::{self, app};
 use ucs_env::{configs, logger};
```
Aishwariyaa-Anand, line 5: revert changes in this file

crates/internal/integration-tests/src/connector_specs/reddot/specs.json
```diff
@@ -0,0 +1,9 @@
+{
+  "connector": "reddot",
+  "supported_suites": [
```
Aishwariyaa-Anand, line 3: add supported payment methods

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,569 @@
...
+    ) -> CustomResult<ErrorResponse, errors::ConnectorError> {
+        // RDP's edge (Imperva Incapsula WAF) serves `403 + HTML` to blocked
+        // clients — a strict JSON parse of that page would surface a
+        // misleading RESPONSE_DESERIALIZATION_FAILED, so fall back to a
+        // synthesized error that keeps the HTTP status and the body snippet.
+        let response: reddot::ReddotErrorResponse = if res.response.is_empty() {
+            reddot::ReddotErrorResponse {
+                response_status: "ERROR".to_string(),
+                response_code: res.status_code.to_string(),
+                response_msg: "Red Dot returned an empty error body".to_string(),
```
Aishwariyaa-Anand: use NO_ERROR_CODE and NO_ERROR_MESSAGE const

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,569 @@
...
+        Ok(Some(WebhookResourceReference::Payment(
+            PaymentWebhookReference {
+                // RDP's gateway-side id for this payment.
+                connector_transaction_id: body.transaction_id,
+                // `merchant_reference` = full Juspay txnId echoed by RDP —
+                // this is the integrity-key euler's verifyTxnId compares
+                // against txnDetail.txnId (TrackerVerification.hs).
+                // Fallback to `order_id` (our random 16-hex id) never matches
+                // a Juspay txnId, so it degrades to a loud integrity failure
+                // when the acquirer didn't round-trip merchant_reference.
```
Aishwariyaa-Anand: this comment is not required

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,569 @@
...
+                merchant_transaction_id: body.merchant_reference.or(body.order_id),
```
Aishwariyaa-Anand: do we pass merchant_transaction_id in request for order_id

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+/// authorize time (ucs-side generation; euler never sees it directly).
+/// Juspay txnUuids (19-21 chars) overflow RDP's 20-char `order_id` ceiling,
+/// so instead of truncating a Juspay id we mint our own 64-bit-random hex
+/// (collision space ~2^64/payment rate) and ask euler to persist it: it is
+/// returned inside `connector_metadata` on the Authorize response, which
+/// euler merges into `second_factor.gateway_auth_req_params`, and the refund
+/// request hands that blob back via `connector_feature_data` (channel 5 of
+/// the UCS state channels — the only one that round-trips connector-private
+/// data). PSync keys off `transaction_id`/`merchant_reference` instead, so
+/// only authorize + refund touch this key.
```
Aishwariyaa-Anand: remove comments related to Euler flows

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+            .clone();
+
+        match &router_data.request.payment_method_data {
+            PaymentMethodData::Card(card_data) => {
+                // Mandatory `payer_name` fallback chain (mirrors the email chain):
+                //   1. card_holder_name as collected (skip Juspay's sentinel
+                //      "name" that euler's TempCard backfills when missing)
+                //   2. Customer record's name (euler composes first+last into
+                //      UCS CustomerDetails.name)
+                //   3. literal "name" — RDP requires the field non-empty
```
Aishwariyaa-Anand, line 426: do we need this fallback, will the request go through for default "name"?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+                    merchant_reference: Some(merchant_reference),
+                    signature: Secret::new(signature),
+                    bill_to_forename: billing_details.and_then(|ad| ad.first_name.clone()),
+                    bill_to_surname: billing_details.and_then(|ad| ad.last_name.clone()),
+                    bill_to_address_line1: billing_details.and_then(|ad| ad.line1.clone()),
+                    bill_to_address_city: billing_details.and_then(|ad| ad.city.clone()),
+                    bill_to_address_country: billing_details
+                        .and_then(|ad| ad.country.map(|country| country.to_string())),
+                    bill_to_address_postal_code: billing_details.and_then(|ad| ad.zip.clone()),
+                    bill_to_phone: billing_phone.and_then(|phone| phone.number.clone()),
```
Aishwariyaa-Anand: use util functions

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+pub struct ReddotRefundRequest {
+    pub response_type: String,
+    pub action_type: String,
+    pub mid: Secret<String>,
+    /// Original payment order number — the order_id we sent at authorize.
+    pub order_number: String,
+    /// Original RDP transaction id of the payment being refunded.
+    pub transaction_id: String,
+    pub amount: StringMajorUnit,
+    pub currency: String,
```
Aishwariyaa-Anand:
```suggestion
    pub currency: Currency,
```

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+/// RDP Merchant API refund response body. `result_status` values per docs:
+/// accepted | failed | pending | capture pending | pending third party.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct ReddotRefundResponse {
+    pub result_status: Option<String>,
+    pub reason_code: Option<String>,
+    pub refund_id: Option<String>,
+    pub order_number: Option<String>,
+    pub amount: Option<StringMajorUnit>,
+    pub currency: Option<String>,
```
Aishwariyaa-Anand:
```suggestion
    pub currency: Option<Currency>,
```

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1458 @@
...
+/// Mint a random order_id: 16 lowercase hex chars (= 8 random bytes);
+/// RDP's enforced ceiling is 20 (it rejects 22 with `-1014`), so 16 sits
+/// comfortably underneath.
+fn gen_reddot_order_id() -> String {
+    let oid = hex::encode(rand::random::<[u8; 8]>());
+    // 8 bytes always hex-encode to 16 chars; keep the assertion so the
+    // constant stays load-bearing at the generator, not only in tests.
+    debug_assert_eq!(oid.len(), RDP_ORDER_ID_GENERATED_LEN);
+    oid
+}
```
Aishwariyaa-Anand, line 54: use merchant_transaction_id, if length is greater than required then generate

MansiKhatri-37: we are sending the merchant_transaction_id in merchant_reference ,should we still add this change?

XyneSpaces:
[should-fix] Please revert the unrelated `config/development.toml` defaults changed in this connector PR: enabling `return_raw_connector_data` and changing the development server/metrics ports and server type alters every developer/CI environment and can expose raw connector payloads by default. Keep only the Reddot-specific configuration required by the connector.

XyneSpaces:
[blocking] Unrelated local-dev config changes in `config/development.toml`

The PR changes `return_raw_connector_data = true`, `server.port = 8011`, `server.type = "http"`, and `metrics.port = 8090`. These are local debugging overrides and not part of the Reddot connector; revert them.

[blocking] Invalid duplicate keys in `config/superposition.toml`

Lines around 103 and 107 add a second `rpc_method`/`rpc_service` block while the keys already exist earlier in the file. TOML parsers will reject or overwrite these, breaking shared defaults. Deduplicate the block.

[blocking] Custom Tokio runtime added to `grpc-server/src/main.rs`

`crates/grpc-server/grpc-server/src/main.rs` replaces `#[tokio::main]` with a hand-built runtime setting `thread_stack_size(128 * 1024 * 1024)`. This affects all production deployments and is unrelated to the connector; revert or move to a separate PR.

[blocking] `amount_captured` assigned minor-unit value

`crates/integrations/connector-integration/src/connectors/reddot/transformers.rs` (around line 1081) sets `amount_captured` from `amount.get_amount_as_i64()`, which is in minor units. `PaymentFlowData::amount_captured` expects major units, so this will report 100x the captured amount for most currencies. Convert to major units first.

[blocking] Placeholder `payer_name` defaults to `"name"`

In the same transformers file (around line 432), `payer_name` falls back to the literal string `"name"` when cardholder/customer name is unavailable. Send a real value or fail fast; a placeholder will cause declines or 3DS failures.

[should-fix] Inconsistent reference-id mapping between Authorize and PSync

Authorize uses `response.order_id` as `connector_response_reference_id`, while PSync prefers `response.merchant_reference`. Use `merchant_reference` consistently (with fallback to `order_id`) so reconciliation and tracker integrity do not depend on which response path ran.

crates/integrations/connector-integration/src/connectors/reddot.rs
MansiKhatri-37: no, we generate a random id which we send in order_id
the merchant_transaction_id is passed in the merchant_reference_key.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
MansiKhatri-37: yes

config/superposition.toml
```diff
@@ -92,16 +92,8 @@ api_tags_tags_psync = { value = "", schema = { type = "string" } }
 # Environment dimension - determines which environment configuration to use
 environment = { position = 1, schema = { type = "string", enum = ["sandbox", "development", "production"] } }
 
-# Full rpc path as observed at deja ingress, e.g. "/types.PaymentService/Authorize".
-rpc_method = { position = 3, schema = { type = "string" } }
-
-# Service class derived from the rpc path ("/types.PaymentService/Authorize" ->
-# "payment") - per-class sampling without a maintained path map, the file-side
-# analogue of hyperswitch's deja_dimension cohort.
-rpc_service = { position = 4, schema = { type = "string" } }
```
Aishwariyaa-Anand, line 101: why is this removed

crates/integrations/connector-integration/src/connectors/reddot.rs
Aishwariyaa-Anand: do we need fallback to order_id then?
why is connector_txn_id mapping removed?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1474 @@
...
+#[derive(Debug, Serialize)]
+pub struct ReddotAuthorizeRequest<
+    T: PaymentMethodDataTypes + std::fmt::Debug + Sync + Send + 'static + Serialize,
+> {
+    pub mid: Secret<String>,
+    pub api_mode: String,
+    pub payment_type: String,
+    pub order_id: String,
+    pub amount: StringMajorUnit,
+    pub ccy: String,
```
Aishwariyaa-Anand:
```suggestion
    pub ccy: Currency,
```

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1474 @@
...
+                    exp_date: exp_date.peek(),
+                    cvv2: card_data.card_cvc.peek(),
+                    secret_key: auth.secret.peek(),
+                })?;
+
+                // bill_to_* only for CyberSource-acquired MIDs (MGA account
+                // detail flag); for those accounts billing is mandatory, so a
+                // missing billing block is a hard error with a clear message.
+                let bill_to = if auth.is_cybersource_acquired() {
+                    Some(get_bill_to(&router_data.resource_common_data)?)
```
Aishwariyaa-Anand: if this is optional do we need to propagate error through `?`

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1474 @@
...
+                        resource_id: ResponseId::ConnectorTransactionId(transaction_id),
+                        redirection_data: None,
+                        mandate_reference: None,
+                        connector_metadata: None,
+                        network_txn_id: None,
+                        network_txn_link_id: None,
+                        connector_response_reference_id: response
+                            .merchant_reference
+                            .clone()
+                            .or(response.order_id.clone()),
```
Aishwariyaa-Anand: both order_id and merchant_reference gives back different values, which of these should be mapped in connector_response_reference_id. Why do we need a fallback here?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
MansiKhatri-37, line 463: yes, these params are necessary if acquirer is cybersource

config/superposition.toml
kanikac199, line 101: revert this

config/superposition.toml
```diff
@@ -92,16 +92,8 @@ api_tags_tags_psync = { value = "", schema = { type = "string" } }
 # Environment dimension - determines which environment configuration to use
 environment = { position = 1, schema = { type = "string", enum = ["sandbox", "development", "production"] } }
 
-# Full rpc path as observed at deja ingress, e.g. "/types.PaymentService/Authorize".
-rpc_method = { position = 3, schema = { type = "string" } }
```
kanikac199, line 96: revert this

flake.lock
```diff
@@ -48,11 +48,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1771470520,
-        "narHash": "sha256-PvytHcaYN5cPUll7FB70mXv1rRsIBRmu47fFfq3haxA=",
+        "lastModified": 1788077403,
```
kanikac199: revert changes in this file

Aishwariyaa-Anand (dismissed):

kanikac199 (dismissed):

Aishwariyaa-Anand (approved):

Aishwariyaa-Anand (approved):

kanikac199 (approved):

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+                        status_code: http_code,
+                    }),
+                )
+            }
+            // Terminal failure; response_msg carries the reason.
+            // Status mapping mirrors maya: `-7995`/`-7997` (3DS-auth
+            // failures) → AUTHENTICATION_FAILED; everything else
+            // (`-1` bank/acquirer decline, `-1003` signature/validation,
+            // ...) → AUTHORIZATION_FAILED.
+            error_code => {
```
JeevaRamu0104: `-1009` ("not been paid or expired") lands here and becomes `AuthorizationFailed`. Per `specs.json` that's what an unpaid session returns, so a PSync while the customer is still on the 3DS page terminally fails the payment — and if they then pay, we've taken money on a failed attempt. `-1003` (our bad signature) and bare error bodies hit the same arm. Only explicit declines should fail; everything else should stay `Pending`.

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,571 @@
...
+        let (status, error_code, error_message, error_reason) = match body.response_code.as_str() {
+            "0" => (AttemptStatus::Charged, None, None, None),
+            "-01" => (AttemptStatus::Pending, None, None, None),
+            "-7995" | "-7997" => (
+                AttemptStatus::AuthenticationFailed,
+                Some(body.response_code.clone()),
+                Some(body.response_msg.clone().unwrap_or_default()),
+                body.response_msg.clone(),
+            ),
+            error_code => (
```
JeevaRamu0104: Same catch-all as PSync: any unrecognised `response_code` becomes `AuthorizationFailed`. Can this list the decline codes explicitly and leave unknown codes `Pending`?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+    type Error = error_stack::Report<ConnectorError>;
+
+    fn try_from(
+        item: ResponseRouterData<ReddotRefundSyncResponse, Self>,
+    ) -> Result<Self, Self::Error> {
+        let raw_connector_response = serde_json::to_string(&item.response).ok().map(Secret::new);
+
+        // Only `refunded` is success; `refund pending` is non-final;
+        // EVERYTHING else is terminal failure.
+        let refund_status = match item.response.transaction_status.as_deref() {
```
JeevaRamu0104: Any `transaction_status` other than `refunded`/`refund pending` — including an error body with no status, a WAF block, or the payment's own status before the refund registers — marks the refund `Failure` terminally. Unknown should stay `Pending`.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+
+    fn try_from(
+        item: ReddotRouterData<
+            RouterDataV2<RSync, RefundFlowData, RefundSyncData, RefundsResponseData>,
+            T,
+        >,
+    ) -> Result<Self, Self::Error> {
+        let router_data = &item.router_data;
+        let auth = ReddotAuthType::try_from(&router_data.connector_config)?;
+        let transaction_id = router_data.request.connector_transaction_id.clone();
```
JeevaRamu0104: RSync looks up by the payment's `transaction_id` because Refund never sends `refund_id`, so with two partial refunds on one payment it can't tell which refund it's reporting. Send `refund_id` on Refund and key RSync on it.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+const RDP_ORDER_ID_GENERATED_LEN: usize = 16;
+
+/// Key inside the `connector_feature_data` JSON blob under which the
+/// persisted order_id lives between authorize and refund.
+pub(crate) const RDP_ORDER_ID_BLOB_KEY: &str = "reddot_order_id";
+
+/// Mint a random order_id: 16 lowercase hex chars (= 8 random bytes);
+/// RDP's enforced ceiling is 20 (it rejects 22 with `-1014`), so 16 sits
+/// comfortably underneath.
+fn gen_reddot_order_id() -> String {
```
JeevaRamu0104: This mints a random `order_id` inside the transformer, so every retry of Authorize is a new order at RDP. Deriving it deterministically from `connector_request_reference_id` (e.g. a truncated hash, ≤20 chars) keeps retries stable.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+                        // The generated order_id goes back to euler inside
+                        // connector_metadata; euler merges this connector-
+                        // feature-data blob into second_factor.gateway_auth_
+                        // req_params, and the refund request hands it right
+                        // back. Refund's `order_number` needs the identical
+                        // value (see reddot_order_id_from_blob).
+                        connector_metadata: response
+                            .order_id
+                            .clone()
+                            .map(|oid| serde_json::json!({ RDP_ORDER_ID_BLOB_KEY: oid })),
```
JeevaRamu0104: The persisted blob uses `response.order_id`, not the id we generated and sent. If RDP omits it, the payment can never be refunded. Store the request-side `order_id` instead.

Also: fixture `transaction_id`s look like `{order_id}_{acquirer_id}` — if that's guaranteed, Refund could derive `order_number` from `connector_transaction_id` and drop the blob entirely.

crates/internal/field-probe/src/normalizer.rs
```diff
@@ -296,6 +296,10 @@ fn replace_json_dynamic_fields(s: &str) -> String {
         r = replace_json_alphanum_value(&r, key, placeholder);
     }
 
+    // Reddot generates a fresh 16-character hex order_id for every probe.
```
JeevaRamu0104: This only exists to hide the random `order_id` from probe snapshots. Once the id is deterministic, this can go.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -5991,6 +5992,16 @@ message PayhereConfig {
   optional string base_url = 50;
 }
 
+message ReddotConfig {
+  SecretString mid = 1;
+  // RDP merchant secret key, used to build request signatures.
+  SecretString secret = 2;
+  // MGA account detail, string-typed as stored in account_details:
+  // when "true" (CyberSource-acquired MID), bill_to_* fields are sent to RDP.
+  string acquirer_cybersource = 3;
```
JeevaRamu0104: Why a `"true"`/`"false"` string? This should be a `bool`; string parsing silently treats typos as false and skips mandatory billing for CyberSource MIDs.

crates/types-traits/domain_types/src/router_data.rs
```diff
@@ -4268,6 +4290,15 @@ impl ForeignTryFrom<(&ConnectorAuthType, &connector_types::ConnectorVariant)>
                     }),
                     _ => Err(err().into()),
                 },
+                ConnectorEnum::Reddot => match auth {
+                    ConnectorAuthType::BodyKey { api_key, key1 } => Ok(Self::Reddot {
+                        mid: api_key.clone(),
+                        secret: key1.clone(),
+                        acquirer_cybersource: "false".to_string(),
```
JeevaRamu0104: The legacy auth path hardcodes `"false"`, so CyberSource-acquired MIDs coming through it silently drop the mandatory `bill_to_*` block.

crates/types-traits/domain_types/src/connector_types.rs
```diff
@@ -3630,6 +3632,9 @@ pub struct RefundsData {
     pub split_settlement_refund: Option<Box<SplitSettlementRefund>>,
     /// Connector-side identifier for the original payment that this refund targets.
     pub connector_order_id: Option<String>,
+    /// Connector-configured metadata (`PaymentServiceRefundRequest.metadata`,
+    /// proto field 9) passed through verbatim for connector-specific needs.
+    pub metadata: Option<SecretSerdeValue>,
```
JeevaRamu0104: `RefundsData.metadata` isn't read anywhere in the reddot connector. Can this core change come out of this PR?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+    pub acquirer_authorized_amount: Option<String>,
+    pub acquirer_authorized_ccy: Option<String>,
+    pub acquirer_response_code: Option<String>,
+    pub acquirer_response_msg: Option<String>,
+    pub acquirer_authorization_code: Option<String>,
+    pub acquirer_mpi_eci: Option<String>,
+    pub created_timestamp: Option<String>,
+    pub acquirer_created_timestamp: Option<String>,
+    pub request_timestamp: Option<String>,
+    pub first_6: Option<String>,
```
JeevaRamu0104: `payer_name`, `first_6`, `last_4` (and `exp_date` on the webhook body at L1013) are plain `String` on `Debug` structs and get re-serialized into `raw_connector_response`. Wrap them in `Secret`.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+                    .or_else(|| {
+                        router_data
+                            .request
+                            .customer_name
+                            .clone()
+                            .map(|customer_name| customer_name.trim().to_string())
+                            .filter(|customer_name| !customer_name.is_empty())
+                    })
+                    .map(Secret::new)
+                    .unwrap_or_else(|| Secret::new("name".to_string()));
```
JeevaRamu0104: Sending the literal `"name"` as `payer_name` puts placeholder data on a real payment. Should this return a missing-field error instead?

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+                        // req_params, and the refund request hands it right
+                        // back. Refund's `order_number` needs the identical
+                        // value (see reddot_order_id_from_blob).
+                        connector_metadata: response
+                            .order_id
+                            .clone()
+                            .map(|oid| serde_json::json!({ RDP_ORDER_ID_BLOB_KEY: oid })),
+                        network_txn_id: None,
+                        network_txn_link_id: None,
+                        connector_response_reference_id: response.order_id.clone(),
```
JeevaRamu0104: Authorize sets `connector_response_reference_id` to `order_id`, while PSync (L883) and the webhook use `merchant_reference`. Which one is it meant to be? It should be the same across flows.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+            Some("accepted") => common_enums::RefundStatus::Success,
+            Some("failed") => common_enums::RefundStatus::Failure,
+            _ => common_enums::RefundStatus::Pending,
+        };
+
+        let connector_refund_id = item
+            .response
+            .refund_id
+            .clone()
+            .unwrap_or_else(|| item.router_data.request.refund_id.clone());
```
JeevaRamu0104: Falling back to the merchant's `refund_id` puts a merchant-owned id in `connector_refund_id` (RSync does the same at L1354). If RDP returns none, leave it to the caller rather than aliasing.

crates/internal/integration-tests/src/connector_specs/reddot/override.json
```diff
@@ -0,0 +1,85 @@
+{
+  "PaymentService/Authorize": {
+    "no3ds_manual_capture_credit_card": {
```
JeevaRamu0104: Overriding the manual-capture scenarios to `AUTOMATIC` makes them pass without testing manual capture. They belong in `unsupported_scenarios`, like incremental auth.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+            ..item.router_data
+        })
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
+#[allow(clippy::expect_used)]
+#[allow(clippy::panic)]
+mod tests {
```
JeevaRamu0104: No tests cover the response transformers — PSync/RSync status mapping or `response_code` number-vs-string parsing. Those are where the status bugs above live; please add fixture-based cases.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+        .as_ref()
+        .and_then(|cfd| cfd.peek().get(RDP_ORDER_ID_BLOB_KEY))
+        .and_then(|v| v.as_str())
+        .map(str::to_owned)
+        .ok_or_else(|| {
+            IntegrationError::MissingRequiredField {
+                field_name: "connector_feature_data.reddot_order_id",
+                context: errors::IntegrationErrorContext {
+                    additional_context: Some(
+                        "reddot_order_id missing from second_factor.gateway_auth_req_params."
```
JeevaRamu0104: This merchant-facing error names `second_factor.gateway_auth_req_params`, an Euler internal. Same for the Euler/Juspay references in comments throughout. Keep the connector caller-agnostic.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+// WEBHOOK TYPES (notify_url callback)
+// =============================================================================
+
+/// RDP notify_url callback body. Same field shapes as the query_redirection
+/// data response (`response_code` serialized the same string-or-number way),
+/// so the status mapping intentionally mirrors the PSync transform.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ReddotWebhookBody {
+    pub mid: Option<String>,
+    /// Merchant reference we sent at authorize (Juspay txnUuid for REDDOT),
```
JeevaRamu0104: nit: this doc says `order_id` is the merchant reference / Juspay txnUuid — it's the random 16-hex id we generate.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
```diff
@@ -0,0 +1,1466 @@
...
+/// append `secret_key`, SHA-512 hex.
+///
+/// This is a DIFFERENT algorithm from [`compute_payment_signature`] — do not
+/// conflate them. The input must NOT contain a `signature` entry.
+///
+/// `pub` so the future webhook/response signature verification (implemented on
+/// the `Reddot` connector struct in the parent module) can reuse it —
+/// [`compute_payment_signature`] stays private as it is only needed to build
+/// the Authorize request body.
+pub fn compute_generic_signature(mut params: Vec<(&str, &str)>, secret_key: &str) -> String {
```
JeevaRamu0104: nit: nothing outside this module uses these signature helpers yet; `pub(crate)` is enough (also `compute_merchant_api_signature` at L1033).

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,571 @@
...
+            Ok(format!(
+                "{}/instanpanel/api/payment",
+                instanpanel_base_url(&req.resource_common_data.connectors)?
+            ))
+        }
+    }
+);
+
+// =============================================================================
+// REFUND SYNC FLOW (RDP Enquiry API — same endpoint/signature scheme as psync)
```
JeevaRamu0104: nit: the comment says same endpoint as psync, but this hits `/instanpanel/api/enquiry` on the secondary host.

crates/integrations/connector-integration/src/connectors/reddot.rs
```diff
@@ -0,0 +1,571 @@
...
+    generic_type: T,
+    [PaymentMethodDataTypes + Debug + Sync + Send + 'static + Serialize]
+);
+
+// ===== FLOW STATUS IMPLEMENTATIONS =====
+// Emits marker-trait impls AND stub ConnectorIntegrationV2 impls for every
+// flow listed. Each stub's get_url returns
+// IntegrationError::connector_flow_not_implemented(...).
+//
+// To implement a flow:
```
JeevaRamu0104: nit: leftover template boilerplate ("To implement a flow…") — can be dropped.

crates/types-traits/grpc-api-types/proto/payment.proto
```diff
@@ -5991,6 +5992,16 @@ message PayhereConfig {
   optional string base_url = 50;
 }
 
+message ReddotConfig {
+  SecretString mid = 1;
```
JeevaRamu0104: nit: `mid` needs a comment saying what it holds (RDP merchant id).

crates/types-traits/domain_types/src/types.rs
```diff
@@ -11462,11 +11466,15 @@ impl ForeignTryFrom<grpc_api_types::payments::PaymentServiceRefundRequest> for R
...
+            metadata: value
+                .metadata
+                .map(|m| ForeignTryFrom::foreign_try_from((m, "metadata")))
+                .transpose()?,
             payment_method_data: value.payment_method.and_then(|pm| {
                 payment_method_data::PaymentMethodData::<
                     payment_method_data::DefaultPCIHolder,
                 >::convert_to_domain_model_for_non_card_payment_methods(pm)
-                .ok()
+                    .ok()
```
JeevaRamu0104: nit: unrelated indentation change.

JeevaRamu0104 (changes requested):

crates/types-traits/grpc-api-types/proto/payment.proto
MansiKhatri-37: In Euler , this is the standard, we don't store boolean values in account details.

crates/integrations/connector-integration/src/connectors/reddot/transformers.rs
MansiKhatri-37: since txnuuid can be greater than 16 charx, we generate random order id

crates/types-traits/grpc-api-types/proto/payment.proto
MansiKhatri-37: connector expects mid.

PR #2335, opened 2026-09-21
fix(connector): [Adyen] scope Cancelled to Failure only for FRAUD-CANCELLED

shuklatushar226:
## Description

Adyen answers Authorize with `resultCode: "Cancelled"` / `refusalReason: "FRAUD-CANCELLED"` when RevenueProtect reverses an authorization the issuer had already approved — the response carries `additionalData.refusalReasonRaw: "00 : Approved or completed successfully"`.

`get_adyen_payment_status` mapped `Cancelled -> AttemptStatus::Voided` unconditionally, so UCS returned `PaymentStatus::VOIDED`, the router's UCS boundary copied it into `ErrorResponse.attempt_status`, and `payment_attempts.status` was written as `voided`. The direct hyperswitch connector writes `failure` for the identical response.

This scopes the failure mapping to the refusal reason that actually identifies it:

```rust
AdyenStatus::Cancelled => match (refusal_reason_code, refusal_reason) {
    (Some(ADYEN_FRAUD_CANCELLED_CODE), _) | (_, Some(ADYEN_FRAUD_CANCELLED_REASON)) => {
        AttemptStatus::Failure
    }
    _ => AttemptStatus::Voided,
},
```

Both the code (`22`) and the reason text (`FRAUD-CANCELLED`) are matched, since Adyen populates `refusalReasonCode` and `refusalReason` independently across API versions.

## Motivation and Context

**The same Adyen response produced two different attempt statuses depending on routing.** The direct connector writes `failure` here, because it puts `attempt_status: None` in its `ErrorResponse` and the router falls through to `Failure` (`crates/router/src/core/payments/operations/payment_response.rs:2508-2533`). UCS preserves the status it is given, so it wrote `voided`. Merchants saw a payment's terminal state change based on a rollout percentage they cannot observe.

`voided` is also the wrong word for the merchant: it means "an authorized payment we deliberately reversed", not "a payment that did not go through". And because success rate is computed as `charged / (charged + failure)`, `voided` attempts silently leave the denominator — SR appears to improve as UCS traffic ramps, with no change in approved volume.

**Observed in production** on one Adyen Apple Pay merchant, 2026-09-19 → 2026-09-21:

| Path | `payment_attempts.status` | Count |
|---|---|---|
| direct | `failure` | 34 |
| UCS | `voided` | 5 |

Zero crossover in either direction. The tightest pair is 69 seconds apart, same merchant, same payment method, same `$44.99`, same Adyen refusal:

```
17:25:48  pay_oUAH83KlH4a0GbG1mf0c  direct  ->  failure
17:26:57  pay_zRnCXnUWxfUSIQds6MCB  UCS     ->  voided
```

The 5 UCS attempts: `pay_XFD3axHt3eGVbxSiCfk6`, `pay_hSdjmDs1oGLQBH5XVlTj`, `pay_6JE4OOWwbO7fabcWXtvD`, `pay_bXshfbNkjCxDUT5AMosY`, `pay_zRnCXnUWxfUSIQds6MCB`.

### Why the mapping is scoped to the refusal, not applied to all `Cancelled`

`resultCode: Cancelled` is not reliably a failure. Adyen `voided` attempts across all merchants over the last four weeks:

| `error_message` | Count | What it is |
|---|---|---|
| *(no error)* | **500** | genuine merchant-initiated voids |
| `Something went wrong` | 17 | `HE_00`, hyperswitch-internal, not an Adyen refusal |
| `FRAUD-CANCELLED` | **5** | the UCS attempts above |

Remapping `Cancelled -> Failure` outright, or deriving `Failure` from the mere presence of an `ErrorResponse`, would have mislabelled those 500 legitimate voids as failed payments. Only the FRAUD-CANCELLED refusal identifies the risk-engine reversal, so only it maps to `Failure`.

### Deliberate difference from the direct connector

| Adyen response | hyperswitch | this change |
|---|---|---|
| `Cancelled`, no refusal | `voided` | `voided` |
| `Cancelled` + FRAUD-CANCELLED | `failure` | `failure` |
| `Cancelled` + some other refusal | `failure` | **`voided`** |

hyperswitch marks the third case a failure because it discards its computed status whenever *any* refusal is present. No such response appears in four weeks of production traffic, so the two agree on every observed case; this version is the stricter reading and is intentional, not an oversight.

### Genuine voids are unaffected

`get_adyen_payment_status` is module-private and reachable only from the five Authorize-family response builders. A merchant-initiated void never reaches it: the Cancel flow maps `CancelStatus` separately, and the terminal `Voided` comes from `AdyenWebhookStatus::Cancelled` and `WebhookEventCode::Cancellation` on success. Neither is touched.

### Additional Changes

## How did you test it?

- `cargo clippy -p connector-integration` and `cargo +nightly fmt --all` pass.
- Verified against production `connector_events` that the rest of the chain is already correct: prism maps `AttemptStatus -> PaymentStatus` 1:1, and in the same window on the same merchant, Adyen codes 2, 10, 12, 20 and 31 all returned UCS `status: FAILURE` and were written as `failure`. Code 22 was the only one returning `VOIDED`, so correcting the mapping here is sufficient — no router-side change is needed.
- No companion hyperswitch change: its persisted behaviour is already correct for every case that occurs in production.

## Follow-ups (not in this PR)

- The 5 rows already written as `voided` will not self-correct and still skew that merchant's SR denominator until backfilled. The intents are `cancelled` and need correcting too, and the ClickHouse analytics copies are fed by the application's Kafka producer rather than DB CDC, so they do not follow a Postgres update.
- GSM maps Adyen codes 2/20/22/31 to `UE_3000 "Technical issue with PSP"`, which is wrong for pure declines (117 of 151 failures in that window). Shared by both paths, so it is a separate fix.


gopikrishna000 (dismissed):

JeevaRamu0104 (approved):

XyneSpaces:
The `FRAUD-CANCELLED` → `Failure` mapping is applied to the synchronous/PSync/redirect response paths. Adyen also reports status via webhooks (mapped through `AdyenWebhookStatus`). If a webhook arrives with `Cancelled` + fraud reason, it would still map to `Voided` unless the same guard is added to the webhook transformation path. Please confirm whether the webhook path is already handled or needs the same treatment for consistency.

AmitsinghTanwar007 (approved):

PR #2338, opened 2026-09-21
feat(connector): [Revolv3] MITs for Applepay and Googlepay using Decrypted payload + NTID

Vani-1107:
### Additional Changes


XyneSpaces:
[should-fix] Both the Apple Pay path (`revolv3/transformers.rs:421`) and the Google Pay path (`revolv3/transformers.rs:450`) silently drop the `get_card_issuer()` parse error with `.ok()`. Invalid wallet-token metadata is hidden instead of failing the MIT.

```suggestion
card_brand: token.get_card_issuer().map(Revolv3CardBrand::from_card_issuer)?,
```

AkshayaFoiger (approved):

JeevaRamu0104 (approved):

